PR #13, opened 2018-08-05
Add support for automated builds using Travis CI and AppVeyor

k0ekk0ek:
The changes in this pull request enable automated builds with Travis CI (Debug and Release with GCC 8 and Clang 6.0 on Linux and Debug and Release with Xcode 9.4 on macOS) and AppVeyor (Debug and Release for Microsoft Visual Studio 2015 and Microsoft Visual Studio 2017 on Windows 32-bit and 64-bit).

Build dependencies (specifically CUnit and Criterion) are managed with Conan so that the build process is the same across platforms. The Conan recipes are kept in git repositories maintained by ADLINK ATO and can be found on [GitHub](https://github.com/atolab). The resulting binaries are published to a [Bintray](https://bintray.com/atolab/public-conan) repository, also maintained by ADLINK ATO. Note that users can also benefit from the dependencies being managed with Conan as it saves them from having to install all of them by hand.

Some tests still fail on certain platforms, no coverage or static analysis builds are available yet and GitHub integration may not be perfect, but I think this will serve as an excellent base for qualifying future work done on cyclonedds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,74 @@
+language: c
+
+# Platform descriptions
+# NOTE: These can be used in creating the build matrix by making use of the
+#       anchor/alias YAML features.
+linux_gcc8: &linux_gcc8
+  os: linux
+  compiler: gcc
+  addons:
+    apt:
+      update: true
+      sources: [ ubuntu-toolchain-r-test ]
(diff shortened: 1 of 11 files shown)


eboasson:
😁

PR #18, opened 2018-09-10
Performance improvements (and a few odds and ends)

eboasson:
This pull request primarily changes the way network input is handled:

- It adds support for having multiple receive threads, such that receive threads handling a single socket avoid input multiplexing. This has a shockingly large effect on macOS, and a small effect on Linux.
- On macOS it now does input multiplexing using a kqueue instead of `select` because a kqueue is much faster.

These improvements show up as lower latencies in ping-pong tests and as less system time spent for handling incoming packets.

There are some attendant changes, of course ...

- Latency measurement tool now reports one-way latencies rather than roundtrip times and adds the 99% percentile latency; and no longer tries to delete non-existent entities when terminating.
- It turns out the code for generating stack traces when some threads seem to not be making progress can't handle the case where a thread stops just when it a stack trace of it is requested (it hangs).
- Disabling multicast when the selected interface doesn't support it (mostly because it means it becomes impossible to stop single-socket receive threads on a multicast socket).
- Eliminating some superfluous small memsets save a surprising amount of time.
- It no longer closes stdout/stderr if those are used for tracing.
- It no longer keeps rescheduling sending a participant discovery message if there's no writer to use.
- It no longer uses the multicast receive socket when sending application data.
- It prints a final line in the DDSI trace ("Finis.") when everything has ended, making it trivial to check the stack shut down cleanly.
- It no longer crashes when checking whether an invalid address is a multicast address.

and finally ...

- builds now default to RelWithDebInfo if no CMAKE_BUILD_TYPE is specified. The default behaviour of cmake makes no sense whatsoever, and RelWithDebInfo is the most reasonable compromise.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -11,6 +11,17 @@
 #
 cmake_minimum_required(VERSION 3.6)
 
+# Set a default build type if none was specified
+set(default_build_type "RelWithDebInfo")
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
+  set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
+      STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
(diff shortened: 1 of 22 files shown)


kydos:
@eboasson this looks good for me. 

+1

Issue #23, opened 2018-10-11
Three major issues found for Cyclone.


raiars:
Hello,
     I have found three major issues for Cyclone till now. 
     1. Domain Id is restricted only to DDS_DOMAIN_DEFAULT. Different domain Ids (1~230) cannot be passed in which is different from the API description.
     2. Get Instance handle is not supported for publisher, subscriber and topic.
     3. StatusCondition seems not supported by Cyclone right now. So how can I wait for a certain status to be triggered using WaitSet?
Thanks!

eboasson:
Hi,

Thanks for trying and especially for asking: the situation is not as bad as it may seem at first (though there are certainly a number of issues that need fixing):

#### 1. Domain id

Whenever you specify ``DDS_DOMAIN_DEFAULT``, the domain id you get is whatever is specified as the value for the CycloneDDS/Domain/Id element in the configuration XML file. Without a configuration file, it defaults to 0. The ``CYCLONEDDS_URI`` environment variables specifies the configuration to use. It is a URI, but I'm pretty sure at the moment it only supports either plain file names or the ``file`` protocol — in other words, something like ``file:///home/baz/cyclonedds/etc/cyclonedds.xml``. There is an example in ``etc/cyclonedds.xml``.

It will also accept the domain id that is given by that configuration file. Without specifying a configuration file, it defaults the domain id to 0 and only accepts ``DDS_DOMAIN_DEFAULT`` and 0. If, for example, you have a configuration file specifying domain id 1, then ``DDS_DOMAIN_DEFAULT`` and 1 are allowed values for the domain id in ``dds_create_participant``. 

And yes, this is a bit bizarre and undesirable behaviour. (I know how it came to be this way, but I'll spare you the history …) Fortunately, it is pretty easy to fix. I'll see what I can do, but in the meantime, please try using a configuration file. Also note that the configuration file can use environment variable expansion, in case you want to be able to change it with relative ease.

#### 2. Publisher, subscriber and topic Instance handles

You're right, these calls are not yet implemented. In my view, the bigger issue here is that built-in topics are not properly implemented yet (only DCPSParticipant, but even that's a bit problematic). The usefulness of the instance handles for these three entities is rather limited without having built-in topics. (I'm actually working on getting built-in topics fully supported — but that I am doing so is rather invisible at the moment …)

[...]

Issue #22, opened 2018-09-24
ThroughputPublisher Termination


kydos:
On MacOS (not verified if the same issue appears on Linux) a running instance of ThroughputPublisher cannot be terminated by Ctrl-C. It seams that the only way to stop it is to send a kill -9 signal.

eboasson:
@kydos, I have been able to reproduce this behaviour if I first stop the subscriber. Would that be consistent with your observations?

In that case I get the following deadlock:
```
    2731 Thread_144215   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2731 start  (in libdyld.dylib) + 1  [0x7fff6fc78085]
    +   2731 main  (in ThroughputPublisher) + 381  [0x104c0500d]  publisher.c:94
    +     2731 start_writing  (in ThroughputPublisher) + 154  [0x104c0552a]  publisher.c:244
    +       2731 dds_write  (in libddsc.0.dylib) + 103  [0x104c9aef7]  dds_write.c:44
    +         2731 dds_write_impl  (in libddsc.0.dylib) + 653  [0x104c9b28d]  dds_write.c:224
    +           2731 write_sample_gc  (in libddsc.0.dylib) + 79  [0x104c68aaf]  q_transmit.c:1140
    +             2731 write_sample_eot  (in libddsc.0.dylib) + 750  [0x104c68dae]  q_transmit.c:1028
    +               2731 os_mutexLock  (in libddsc.0.dylib) + 85  [0x104cb4085]  os_platform_sync.c:123
    +                 2731 _pthread_mutex_firstfit_lock_slow  (in libsystem_pthread.dylib) + 226  [0x7fff6fe684c7]
    +                   2731 _pthread_mutex_firstfit_lock_wait  (in libsystem_pthread.dylib) + 96  [0x7fff6fe6ada0]
    +                     2731 __psynch_mutexwait  (in libsystem_kernel.dylib) + 10  [0x7fff6fdb3872]
```
and
```
    + 2731 thread_start  (in libsystem_pthread.dylib) + 13  [0x7fff6fe69425]
    +   2731 _pthread_start  (in libsystem_pthread.dylib) + 70  [0x7fff6fe6d2a7]
    +     2731 _pthread_body  (in libsystem_pthread.dylib) + 126  [0x7fff6fe6a33d]
    +       2731 os_startRoutineWrapper  (in libddsc.0.dylib) + 170  [0x104cb552a]  os_posix_thread.c:176
[...]

PR #24, opened 2018-10-14
More sensible domain id handling; stopping throughput publisher

eboasson:
This pull request partially addresses two open issues:
* One commit addresses the first issue raised in #23, by allowing any domain id to be used in ``dds_create_participant`` as long as the configuration file doesn't explicitly set a domain id. To accomplish this, the domain id specification in the configuration file now allows the keyword ``any`` (also the new default). This doesn't affect any cases that used to work already.
* The other commit addresses the observation in #22 that ^C doesn't work when there is no subscriber. It ignores the termination flag set by the ^C handler while waiting for the subscriber to show up. This fix addresses it by delaying the setting of the handler, until the termination flag is respected.
The other points raised in these two issues are not addressed.

### DIFF
--- a/src/core/ddsc/src/dds__init.h
+++ b/src/core/ddsc/src/dds__init.h
@@ -32,7 +32,7 @@ dds__check_domain(
  *-# Returns 0 on success or a non-zero error status
  **/
 dds_return_t
-dds_init(void);
+dds_init(dds_domainid_t domain);
 
 /* Finalization function, called from main */
 
--- a/src/core/ddsc/src/dds_builtin.c
+++ b/src/core/ddsc/src/dds_builtin.c
@@ -137,8 +137,8 @@ dds__create_builtin_participant(
     }
 
     pp->m_entity.m_guid = guid;
-    pp->m_entity.m_domain = dds_domain_create (config.domainId);
-    pp->m_entity.m_domainid = config.domainId;
+    pp->m_entity.m_domain = dds_domain_create (config.domainId.value);
+    pp->m_entity.m_domainid = config.domainId.value;
     pp->m_entity.m_deriver.delete = dds__delete_builtin_participant;
 
 fail:
--- a/src/core/ddsc/src/dds_init.c
+++ b/src/core/ddsc/src/dds_init.c
@@ -41,7 +41,7 @@ dds_globals dds_global = { .m_default_domain = DDS_DOMAIN_DEFAULT };
 static struct cfgst * dds_cfgst = NULL;
 
 dds_return_t
-dds_init(void)
+dds_init(dds_domainid_t domain)
 {
   dds_return_t ret = DDS_RETCODE_OK;
   const char * uri;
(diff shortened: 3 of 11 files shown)


eboasson:
@k0ekk0ek, I am pretty sure the one test failure is Criterion hanging in termination. Could you please have a look and let me know if you are happy with this change and, if so, think it reasonable to merge it despite this failure?

k0ekk0ek:
@eboasson, I agree with the changes. And yes, please merge the changes. We know Criterion sometimes misbehaves (it'll be removed at some point not too far in the future) and I don't see how any of your changes should lead to difficulties during termination.

joxoby:
> This pull request partially addresses two open issues:
> 
> * One commit addresses the first issue raised in #23, by allowing any domain id to be used in `dds_create_participant` as long as the configuration file doesn't explicitly set a domain id. To accomplish this, the domain id specification in the configuration file now allows the keyword `any` (also the new default). This doesn't affect any cases that used to work already.

What's the reason for having the option of setting the domain id in the config file?

eboasson:
Hi Juan,

There are various reasons behind it:
It is partially historical, inherited from OpenSplice’s DDSI service;
It is also in part because in my experience it is often much nicer to set the domain id from the environment instead of programmatically, and as the configuration processing performs environment variable expansion, you can use that;
It is in part because the day (hopefully soon) that multiple simultaneous domains are supported, it will allow you to have multiple configurations and pick the one that matches the requested domain id.
For the latter bit, chances are that it will change from an id or “any” to a list of ids or “any”, but that’s a mere detail. Also haven’t yet given thought to whether there would be multiple configurations in a single file, or whether there’d be multiple files, &c.

First order of business in this is to get the old global variables inherited from OpenSplice’s service out of the way.

Best regards,
Erik

joxoby:
Hi Erik,

Thanks for your answer. 

What bugs me about it is that the domain-id is not a parameter of a DDS instance, since a single DDS instance should be able to support simultaneously every domain-id. Having a config option for a default domain-id makes sense, but this value should not have precedence over a domain-id explicitly supplied by the user programmatically when creating a new participant.
Also, multiple configuration sets for different domains in a single DDS instance might create a whole new set of complications without gaining much from it.
Again, I think that the domain-id should be interpreted as a "variable" of an instance instead of as a parameter.

Thank you.

eboasson:
Hi Juan,

I agree with you … and your earlier comment actually did the trick to get me sufficiently annoyed by the current state of affairs to sit down for the unpleasant work of finally eliminating the global variables that today prevent supporting multiple domains simultaneously. So thank you for that! ☺️

With that, where today you effectively set the domain for the entire process when creating the first participant, soon you’ll be allowed to freely specify a domain id when creating a (additional) participant. That’ll incidentally solve the RMW layer’s problem today, where it usually needs to create something before a node has been created, and thus needs to create a participant without knowing the intended domain id, and thereforce locks the entire process to that domain id.

For now, the configuration will be chosen based on that domain id — and in the first iteration, that will probably always be the same configuration, parameterised by the domain id. Whether multiple configurations for different domain ids in the end will be worth it is something we’ll have to see. I know of some cases where it would matter given what you can specify today in the configuration, but that’s not to say there do not exist more elegant ways. One step at the time.

Regards,
Erik

PR #52, opened 2018-11-21
Change QoS and listener operations naming to bring them in line with the rest of the API

eboasson:
This pull request changes the names of the QoS and listener operations to bring them in line with the rest of the API (that is named like operation _ type, e.g., dds_create_qos). Functions with the old names are still available as deprecated exports, causing compilation warnings from gcc, clang and, I presume, Microsoft's compiler.


TheFixer:
@eboasson I noticed that the changes in the naming scheme you made did not make it to the documentation. For example, the [user manual](http://cdds.io/docs/pdf/cdds.pdf) that is prominently advertised on github (see [here](https://github.com/eclipse-cyclonedds/cyclonedds)  stills talks about dds_qos_create() iso dds_create_qos(). I would have expected that the manual is updated too.

PR #56, opened 2018-11-22
Rename dds_waitset_set_trigger to dds_trigger_waitset

FirasSahliADLinktech:
Change the function dds_waitset_set_trigger to dds_trigger_waitset in file where it appears


eboasson:
@FirasSahliADLinktech, thank you for this, but I'm not inclined to accept the change (quite apart from the IP validation issue and not keeping the old name as a deprecated interface). The reason is that the original name is actually correct: this function doesn't trigger a waitset, it rather sets the value of the trigger flag on a waitset.

eboasson:
Apparently no-one disagrees with my reasons for not accepting this change, so closing the pull request.

PR #60, opened 2018-11-23
Some missing unit tests and correct signature

FirasSahliADLinktech:
Add required unit tests for dds_time, dds_qos_copy and dds_instance_get_key. And correct the signature of the function dds_instance_get_key

### DIFF
--- a/src/core/ddsc/tests/instance_get_key.c
+++ b/src/core/ddsc/tests/instance_get_key.c
@@ -1,0 +1,96 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <criterion/criterion.h>
+#include <criterion/logging.h>
+
+#include "os/os.h"
+#include "ddsc/dds.h"
+#include "RoundTrip.h"
+
+static dds_entity_t participant = DDS_ENTITY_NIL;
+static dds_entity_t topic = DDS_ENTITY_NIL;
+static dds_entity_t publisher = DDS_ENTITY_NIL;
+static dds_entity_t writer = DDS_ENTITY_NIL;
+
+static dds_instance_handle_t handle = DDS_HANDLE_NIL;
+
+static RoundTripModule_Address data;
+
+/* Fixture to create prerequisite entity */
+static void setup(void)
+{
+    participant = dds_create_participant(DDS_DOMAIN_DEFAULT, NULL, NULL);
+    cr_assert_gt(participant, 0);
+    topic = dds_create_topic(participant, &RoundTripModule_Address_desc, "ddsc_instance_get_key", NULL, NULL);
+    cr_assert_gt(topic, 0);
+
+    publisher = dds_create_publisher(participant, NULL, NULL);
+    cr_assert_gt(publisher, 0);
+
+    writer = dds_create_writer(publisher, topic, NULL, NULL);
+    cr_assert_gt(writer, 0);
+
+    memset(&data, 0, sizeof(data));
+    data.ip = os_strdup("some data");
+    cr_assert_not_null(data.ip);
+    data.port = 1;
+}
+
+/* Fixture to delete prerequisite entity */
+static void teardown(void)
+{
+    RoundTripModule_Address_free(&data, DDS_FREE_CONTENTS);
+
+    dds_delete(writer);
+    dds_delete(publisher);
+    dds_delete(topic);
+    dds_delete(participant);
+}
+
+Test(ddsc_instance_get_key, bad_entity, .init=setup, .fini=teardown)
+{
+    dds_return_t ret;
+
+    ret = dds_instance_get_key(participant, handle, &data);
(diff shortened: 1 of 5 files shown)


src/core/ddsc/tests/instance_get_key.c
```diff
@@ -0,0 +1,91 @@
...
+
+Test(ddsc_instance_get_key, null_handle, .init=setup, .fini=teardown)
+{
+    dds_return_t ret;
+
+    ret = dds_register_instance(writer, DDS_HANDLE_NIL, NULL);
+    cr_assert_eq(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER, "Argument data is NULL");
+
+    ret = dds_instance_get_key(writer, DDS_HANDLE_NIL, &data);
+    cr_assert_eq(dds_err_nr(ret), DDS_RETCODE_PRECONDITION_NOT_MET);
```
eboasson: ``PRECONDITION_NOT_MET`` is not listed among the possible return values in the header file (nor in the DDS specification). If this test passes, there is a discrepancy that needs addressing. Can you look into that please?

src/core/ddsc/tests/instance_get_key.c
```diff
@@ -0,0 +1,91 @@
...
+}
+
+Test(ddsc_instance_get_key, registered_instance, .init=setup, .fini=teardown)
+{
+    dds_return_t ret;
+
+    ret = dds_register_instance(writer, &handle, &data);
+    cr_assert_eq(ret, DDS_RETCODE_OK, "dds_register_instance succeeded (ret: %d)", dds_err_nr(ret));
+    ret = dds_instance_get_key(writer, handle, &data);
+    cr_assert_eq(dds_err_nr(ret), DDS_RETCODE_OK);
```
eboasson: I'd rather have get_key return the result in another variable, and then also verify that the returned key value is actually the same as the one that was put in.

src/core/ddsc/tests/qos.c
```diff
@@ -240,6 +239,44 @@ qos_fini(void)
...
+Test(ddsc_qos, copy_with_partition, .init=qos_init, .fini=qos_fini)
+{
+	dds_return_t result;
+	dds_qos_t *qos;
+	struct pol_partition p = { 0, NULL };
+
+	qos = dds_qos_create();
+	cr_assert_not_null(qos);
+
+	dds_qset_partition(g_qos, g_pol_partition.n, c_partitions);
```
eboasson:
```suggestion
	dds_qset_partition(g_qos, g_pol_partition.n, g_pol_partition.ps);
```

src/core/ddsc/tests/qos.c
```diff
@@ -240,6 +239,44 @@ qos_fini(void)
...
+	result = dds_qos_copy(qos, g_qos);
+
+	cr_assert_eq(result, DDS_RETCODE_OK);
+	dds_qget_partition(qos, &p.n, &p.ps);
+	cr_assert_eq(p.n, 2);
```
eboasson: Asserting twice than p.n is equal to the same value makes no sense; checking against g_pol_partition.n is the better one, so this one should go.

src/core/ddsc/tests/qos.c
```diff
@@ -240,6 +239,44 @@ qos_fini(void)
...
+	cr_assert_eq(p.n, g_pol_partition.n);
+	cr_assert_str_eq(p.ps[0], g_pol_partition.ps[0]);
```
eboasson: Can you turn this into a loop, please? That makes the test independent of the number of partitions actually used.

src/core/ddsc/tests/time.c
```diff
@@ -0,0 +1,17 @@
...
+{
+    dds_time_t now, then;
+    dds_duration_t pause = 1 * DDS_NSECS_IN_MSEC;
+
+    now = dds_time();
+    cr_assert_gt(now, 0);
+    /* Sleep for 1 millisecond, every platform should (hopefully) support that */
+    dds_sleepfor(pause);
+    then = dds_time();
+    cr_assert_gt(then, now);
```
eboasson: The test could be strengthed: if it is required that ``then > now``, then ``then >= now + pause`` should hold as well.

And what about platforms that can't sleep for a 1ms — do we guarantee that sleepfor will then sleep for longer, or does it merely yield the CPU? Sleeping a bit longer would avoid that trouble, so why not increase ``pause`` to something that really "should" be supported?

eboasson:
@FirasSahliADLinktech — and also please sort out the IP validation.

src/core/ddsc/tests/instance_get_key.c
FirasSahliADLinktech: DDS_RETCODE_PRECONDITION_NOT_MET is the returned value for the function dds_instance_get_key when the returned value of the function dds_tkmap_get_key is false (The test passes. )
The specification of this function in the header file does not include this value ( one of them should be updated).

Issue #57, opened 2018-11-22
CMake exits with an error message if a file does not contain any test signatures


k0ekk0ek:
When a source file passed to `add_cunit_executable` does not contain any test signatures, CMake exits with the following error:

```
CMake Error at cmake/modules/CUnit.cmake:177 (list):
  list sub-command REMOVE_DUPLICATES requires list to be present.
Call Stack (most recent call first):
  cmake/modules/CUnit.cmake:219 (process_cunit_source_file)
  os/tests/CMakeLists.txt:30 (add_cunit_executable)
```

PR #68, opened 2018-12-05
Replace Criterion by CUnit

k0ekk0ek:
This pull request completely removes the dependency on the Criterion test framework from the codebase. This has a couple of advantages:

1. All tests now use CUnit, which is supported virtually everywhere. This means we can run all tests on all platforms. Even the embedded ones.
2. Criterion no longer needs to be available to run tests.
2.1. We do not have to maintain that many dependency recipes. The Conan recipe for CUnit got included by Bincrafters, who add newer platforms when available. e.g. xcode 10 was recently added by someone else than me. For Criterion this wasn't the case.
2.2. This makes it easier to get Cyclone included in distributions, because virtually all ship CUnit, but none ship Criterion.
2.3. Developers can install the CUnit dev package that ships with their distribution. So, no dependency on Conan if someone really has a problem with it.
3. Easier debugging. I personally had trouble with Criterion not being able to setup the debug server. With CUnit I just launch it in GDB and wait till it breaks.
4. Test runs and ci builds are quicker, less dependencies and tests generally run quicker (at least on Linux).


k0ekk0ek:
Btw, this also fixes issue #57.

eboasson:
Thank you very much for this effort, @k0ekk0ek! Allow me to nitpick a little bit, though, because I did some checks and some of the changes aren't entirely correct, or at least exhibit slightly different behaviour. E.g., some cr_asserts turned into non-fatal CU_ASSERTs, some string equality checks changed to pointer equality checks and the special return code assertions lost the extraction of the error number.

It goes without saying that I did not find these by manually inspecting everything, I used the below perl hack on the diff with the intent of verifying the changes to the most common checks, and the details I mentioned were discovered by examing its output. First column ">" means it recognised an assert in the old version, "<" that it recognised a matching one in the new version. A "?" in the first column means something failed to match.

Would you please have another look at the changes? Just once more, so the obvious mistakes can be fixed. After that, a green light on Travis and Appveyor will be accepted as sufficient evidence :)

```perl
%t=("gt"=>">","lt"=>"<","eq"=>"==","neq"=>"!=","geq"=>">=","leq"=>"<=");

while (<>) {
  print $_;
  if (/cr_(expect|assert)_(gt|lt|eq|neq|geq|leq)\s*\(\s*([^,]+),\s*(?:([^,]+),\s*(?:[A-Z_0-9]{3,}\s*)*"[^"]*".*\)\s*;|(.+)\)\s*;)/) {
    my $fat = $1 eq "expect" ? "" : "_FATAL"; my $op = $t{$2}; my $a = $3; my $b = defined $4 ? $4 : $5;
    $a =~ s/\s+$//;
    $b =~ s/\s+$//;
    push @checks, [ "CU_ASSERT$fat", $a, $op, $b ];
    print ">@{$checks[-1]}\n";
  } elsif (/(CU_ASSERT(?:_FATAL)?)\s*\(\s*(.+?)(<|<=|==|!=|>=|>)\s*(.+?)\)\s*;/) {
    my $macro = $1; my $a = $2; my $op = $3; my $b = $4;
    $a =~ s/\s+$//;
    $b =~ s/\s+$//;
    match ($macro, $a, $op, $b);
  } elsif (/(CU_ASSERT_((?:NOT_)?)EQUAL(?:_FATAL)?)\s*\(\s*([^,]+),\s*(.+?)\)\s*;/) {
    my $macro = $1; my $op = $2 eq "" ? "==" : "!="; my $a = $3; my $b = defined $4 ? $4 : $5;
    $macro =~ s/(_NOT)?_EQUAL//;
    $a =~ s/\s+$//;
    $b =~ s/\s+$//;
    match ($macro, $a, $op, $b);
  }
}

sub match {
  my ($macro, $a, $op, $b) = @_;
  my $i = 0;
  for ($i = 0; $i < 10 && $i < @checks; $i++) {
    my @x = @{$checks[$i]};
    last if $macro eq $x[0] && $a eq $x[1] && $op eq $x[2] && $b eq $x[3];
  }
  if ($i < 10 && $i < @checks) {
    my @x = @{$checks[$i]};
    print "<$i @x\n";
    splice @checks, 0, $i+1;
  } else {
    my @x = @{$checks[0]};
    printf "?%d\n", scalar @checks;
    print "?MACRO|$macro|$x[0]|\n" unless $macro eq $x[0];
    print "?A|$a|$x[1]|\n" unless $a eq $x[1];
    print "?OP|$op|$x[2]|\n" unless $op eq $x[2];
    print "?B|$b|$x[3]|\n" unless $b eq $x[3];
  }
}
```

k0ekk0ek:
Used the script to go over the diff. Most of the "inconsistent" changes had to do with assume being replaced by a normal assert. The assert is stricter as it actually fails the test whereas assume skips the current iteration of the theory but doesn't fail the test. Apart from that, I did remove the need to add the atolab conan repository as well.

PR #70, opened 2018-12-06
Do not require table to contain address if interface is down (Fixes #55)

### LINKED ISSUE #55: Tests abort on windows when one the network interfaces is down
While trying to run the test cases on my laptop all tests that somehow use a network interface abort. The easiest way to reproduce is to simply run CUnit_os_getifaddrs_ipv4. The output I get is the following:

```
ctest -T test -C DEBUG -R CUnit_os_getifaddrs_ipv4 -V
UpdateCTestConfiguration  from :C:/cyclonedds/build.1/DartConfiguration.tcl
Parse Config file:C:/cyclonedds/build.1/DartConfiguration.tcl
   Site: DESKTOP-S7FK1CA
   Build name: Win32-MSBuild
UpdateCTestConfiguration  from :C:/cyclonedds/build.1/DartConfiguration.tcl
Parse Config file:C:/cyclonedds/build.1/DartConfiguration.tcl
Test project C:/cyclonedds/build.1
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 18
    Start 18: CUnit_os_getifaddrs_ipv4

18: Test command: C:\cyclonedds\build.1\bin\abstraction.exe "-a" "-r" "os_getifaddrs-ipv4" "-s" "os_getifaddrs" "-t" "ipv4"
18: Environment variables:
...
18: Test timeout computed to be: 10
18: Assertion failed: eq != 0, file c:\cyclonedds\src\os\src\windows\os_platform_ifaddrs.c, line 190
1/2 Test #18: CUnit_os_getifaddrs_ipv4 ..........***Failed    1.12 sec
test 22
    Start 22: CUnit_os_getifaddrs_ipv4_n_ipv6

22: Test command: C:\cyclonedds\build.1\bin\abstraction.exe "-a" "-r" "os_getifaddrs-ipv4_n_ipv6" "-s" "os_getifaddrs" "-t" "ipv4_n_ipv6"
22: Environment variables:
...
22: Test timeout computed to be: 10
22: Assertion failed: eq != 0, file c:\cyclonedds\src\os\src\windows\os_platform_ifaddrs.c, line 190
2/2 Test #22: CUnit_os_getifaddrs_ipv4_n_ipv6 ...***Failed    1.30 sec

0% tests passed, 2 tests failed out of 2

Total Test time (real) =   3.14 sec

The following tests FAILED:
         18 - CUnit_os_getifaddrs_ipv4 (Failed)
         22 - CUnit_os_getifaddrs_ipv4_n_ipv6 (Failed)
Errors while running CTest
```

k0ekk0ek:
This pull request fixes issue #55 by not requiring the address to be in the address table if the interface is down. Apart from that it also fixes build warnings by passing a message to CU_PASS. The warnings were introduced by my previous pull request.

### DIFF
--- a/src/core/ddsc/tests/read_instance.c
+++ b/src/core/ddsc/tests/read_instance.c
@@ -298,7 +298,7 @@ CU_Theory((dds_entity_t *ent, void **buf, dds_sample_info_t *si, size_t bufsz, u
         ret = dds_read_instance(*ent, buf, si, bufsz, maxs, g_hdl_valid);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -319,7 +319,7 @@ CU_Theory((dds_entity_t *ent, void **buf, dds_sample_info_t *si, uint32_t maxs),
         ret = dds_read_instance_wl(*ent, buf, si, maxs, g_hdl_valid);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -349,7 +349,7 @@ CU_Theory((dds_entity_t *ent, void **buf, dds_sample_info_t *si, size_t bufsz, u
         ret = dds_read_instance_mask(*ent, buf, si, bufsz, maxs, g_hdl_valid, mask);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
--- a/src/core/ddsc/tests/reader.c
+++ b/src/core/ddsc/tests/reader.c
@@ -306,7 +306,7 @@ CU_Theory((void **buf, dds_sample_info_t *si, size_t bufsz, uint32_t maxs), ddsc
         ret = dds_read(g_reader, buf, si, bufsz, maxs);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -540,7 +540,7 @@ CU_Theory((void **buf, dds_sample_info_t *si, size_t bufsz, uint32_t maxs), ddsc
         ret = dds_read_mask(g_reader, buf, si, bufsz, maxs, mask);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -1676,7 +1676,7 @@ CU_Theory((void **buf, dds_sample_info_t *si, size_t bufsz, uint32_t maxs), ddsc
         ret = dds_take(g_reader, buf, si, bufsz, maxs);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -1911,7 +1911,7 @@ CU_Theory((void **buf, dds_sample_info_t *si, size_t bufsz, uint32_t maxs), ddsc
         ret = dds_take_mask(g_reader, buf, si, bufsz, maxs, mask);
         CU_ASSERT_EQUAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
--- a/src/core/ddsc/tests/reader_iterator.c
+++ b/src/core/ddsc/tests/reader_iterator.c
@@ -405,7 +405,7 @@ CU_Theory((void **buf, dds_sample_info_t *si), ddsc_read_next, invalid_buffers, 
         ret = dds_read_next(g_reader, buf, si);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
@@ -521,7 +521,7 @@ CU_Theory((void **buf, dds_sample_info_t *si), ddsc_read_next_wl, invalid_buffer
         ret = dds_read_next_wl(g_reader, buf, si);
         CU_ASSERT_EQUAL_FATAL(dds_err_nr(ret), DDS_RETCODE_BAD_PARAMETER);
     } else {
-        CU_PASS();
+        CU_PASS("Skipped");
     }
 }
 /*************************************************************************************************/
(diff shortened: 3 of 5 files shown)


eboasson:
@k0ekk0ek, thank you very much, and it is definitely an improvement over the older version. However, before I merge it, I'd like to first try once again to understand a few things ...

Why complicate something very simple by adding unnecessary state? I've seen tons of code that does it and this particular bit of code is in no way unique in this regard; and I am really only using this change as a specimen. I just cannot help but wonder what prompted you to initialize ``eq`` to -1 to avoid doing anything, when it really means none of the statements in the IPv4 branch have any effect when the interface is down.

Secondly, ``eq`` really serves not much of a purpose; it would have, had ``i`` been declared within the ``for`` statement as a local iterator, but as it is, it is superfluous. To me a simple ``for`` loop with a break when the matching element has been found, followed by a check on the value of the loop index makes for a more sensible idiom. Less state, easier to track.

Thirdly, though the likelihood of it happening is negligible, it can leak memory (it could before the fix you did, it is just that the preceding made me wonder and have a good look). If any of the ``addr``, ``netmask`` or ``broadaddr`` allocations fail, only ``ifa`` gets freed and ``name`` through whichever was the last successful allocation will leak.

I am aware that the use of deeply nested if statements — which, I note, you only inherited from elsewhere — combined with variables to track state is entirely in line with many gurus' statements on what constitutes good programming style, but that doesn't mean those gurus are right. I suppose it may well be an excellent way of training one's memory to have a near-infinite stack for tracking allocations and conditions in nested if-statements, and perhaps that is the ultimate goal those gurus have in mind ... but my poor brain dislikes that and prefers simple control flow. In C, for exception handling, that favours ``goto``s and early ``return``s. I'm convinced the hate for ``goto`` comes from those same gurus. (And not from E. W. Dijkstra, contrary to common belief — but not many have actually read his famous letter to the editor or taken the time to understand the context.)

Finally, the network address the function copies presumably must be IPv4 or IPv6 address, or else the ``index`` assignment makes no sense in the non-IPv4 case, a detail that would be much more obvious by a small rewrite. I furthermore do wonder whether it wouldn't be more clear if the check whether the interface is up were done earlier, but that is a minor detail.

Perhaps I should try to show what I mean ... completely untested (not even syntax checked!):
```C
static int
copyaddr(
    os_ifaddrs_t **ifap,
    const PIP_ADAPTER_ADDRESSES iface,
    const PMIB_IPADDRTABLE addrtable,
    const PIP_ADAPTER_UNICAST_ADDRESS addr)
{
  os_ifaddrs_t *ifa;
  if ((ifa = os_calloc(1, sizeof(*ifa))) == NULL)
    goto err_alloc_ifa;
  ifa->flags = getflags(iface);
  (void)os_asprintf(&ifa->name, "%wS", iface->FriendlyName);
  if (ifa->name == NULL)
    goto err_alloc_ifa_name;
  if ((ifa->addr = os_memdup(addr->Address.lpSockaddr, addr->Address.iSockaddrLength)) == NULL)
    goto err_alloc_ifa_addr;
  switch (ifa->addr->sa_family)
  {
    case AF_INET:
      if (ifa->flags & IFF_UP)
      {
        const struct sockaddr_in * const sa = (struct sockaddr_in *)addr->Address.lpSockaddr;
        DWORD i;
        for (i = 0; i < addrtable->dwNumEntries; i++)
          if (sa->sin_addr.s_addr == addrtable->table[i].dwAddr)
            break;
        if (i < addrtable->dwNumEtries)
        {
          if ((ifa->netmask = os_calloc(1, sizeof(netmask))) == NULL)
            goto err_alloc_ifa_netmask;
          if ((ifa->broadaddr = os_calloc(1, sizeof(broadaddr))) == NULL)
            goto err_alloc_ifa_broadaddr;
          ifa->index = addrtable->table[i].dwIndex;
          ifa->netmask.sin_family = AF_INET;
          ifa->netmask.sin_addr.s_addr = addrtable->table[i].dwMask;
          ifa->broadaddr.sin_family = AF_INET;
          ifa->broadaddr.sin_addr.s_addr = sa->sin_addr.s_addr | ~(netmask.sin_addr.s_addr);
        }
      }
      break;
    case AF_INET6:
      ifa->index = iface->Ipv6IfIndex;
      break;
    default:
      assert(0);
  }
  *ifap = ifa;
  return 0;
err_alloc_ifa_broadaddr:
  os_free(ifa->netmask);
err_alloc_ifa_netmask:
  os_free(ifa->addr);
err_alloc_ifa_addr:
  os_free(ifa->name);
err_alloc_ifa_name:
  os_free(ifa);
err_alloc_ifa:
  return ENOMEM;
}
```

k0ekk0ek:
I agree that using `goto` for exception handling is good practice, as are early returns. However, to me, that just creates extra branches. Which I avoid out of personal preference as opposed to clinging to gotos are bad. I tend to use early returns as long as nothing is allocated and prefer to keep one path if anything is. All if statements etc check the error variable (the same that's returned by the function), then at the end of the function I check one last time if an error occurred, and if it did, clean up. In this case `os_free` should have been `os_freeifaddrs` as in the posix specific code. To me that's easier to follow (even if this isn't the best example).

I do agree that introducing an extra state wasn't the best idea. It was the quickest.

The `eq` variable did serve a purpose though. I wanted to make sure that the address could be found in the table if I had an "interface". The reason for that is that I need two calls to get the information os_getifaddrs needs to return on Windows. One to get the interfaces and one to get the addresses that go with the interfaces. In my opinion the operating system should offer that in one call so that I at least get a consistent view of a given point in time. In that sense, the assert actually served it's purpose and I know that in theory I can get a list of interfaces in which it's considered up but have a list of addresses that was compiled with the interface being down.

I don't know if it's worth the effort of implementing going through the process of fetching the interfaces and address again if the situation occurs?

I do agree that checking the index vs. the number entries would have been a better choice.

eboasson:
I have no interest in blocking this pull request, nor in fighting over style, but at least the memory leaks need addressing. They haven't been introduced by this and I probably failed to spot them while reviewing an earlier version of this code, so I am willing to fix them myself. I do feel the urge to comment on a small detail, though:

> I agree that using goto for exception handling is good practice, as are early returns. However, to me, that just creates extra branches. 

This is an intriguing point of view. A branch, by definition, occurs when you have to choose whether to do A or rather B (either which could be empty). The only way to avoid the branches is by not checking, I'd say. Especially

> All if statements etc check the error variable (the same that's returned by the function), then at the end of the function I check one last time if an error occurred, and if it did, clean up.

means that you add state that needs to be tracked when reasoning about the behaviour without actually reducing the number of branches. It works kind of ok when all returns types are the same, but in my experience dealing with (and debugging) code written in that style, I have invariably found it harder to deal with (because of the extra state) and very impractical that there are no good places to set unconditional breakpoints ...

To each his own.

> I do agree that introducing an extra state wasn't the best idea. It was the quickest.

Nah, not really. Just adding it as a conjunct in the IPv4 check would've been even quicker. But whatever.

> One to get the interfaces and one to get the addresses that go with the interfaces. In my opinion the operating system should offer that in one call so that I at least get a consistent view of a given point in time. In that sense, the assert actually served it's purpose and I know that in theory I can get a list of interfaces in which it's considered up but have a list of addresses that was compiled with the interface being down.

The assertion as written is independent of the use of a variable ``eq`` or not, but more important is that it isn't the right way to deal with such behaviour from the underlying operations. As you yourself suggest, if you fail to find a match, retry from the beginning is the more sensible approach. (That is, if there is no information available that allows you to determine whether the configuration has changed since the last time you checked.)

But no matter how closely you manage to approximate extracting a consistent view from the operating system, by the time of use, it may very well be stale already ... That the code currently has no provision for handling such changes is obviously a shortcoming.

> I don't know if it's worth the effort of implementing going through the process of fetching the interfaces and address again if the situation occurs?

Not worth it right now, the code is currently in a state that is simple enough to describe: no support whatsoever for network configurations that change. My guess is that handling those changes properly will mean the effort on making this "less incorrect" will turn out to be wasted.

> In this case ``os_free`` should have been ``os_freeifaddrs`` as in the posix specific code.

I had forgotten about that one, thanks.

k0ekk0ek:
Fixed the pull request. Made the code a little simpler. Tests run locally, let's hope everything passes on Travis and AppVeyor as well.

PR #71, opened 2018-12-10
IDL compiler: New front-end supporting modules and structs

FransFaase:
In order to support X-Types the current IDL compiler needs to be extended and the exchange of type information between nodes need to be extended. It would be nice if the IDL compiler and the module for exchanging type information could be based on the same code, with (at least) two front-ends and a number of back-ends. The fact that the current IDL compiler is implemented in Java, makes it difficult to reuse the code.

It is important that the design of the data structures for representing the type definitions fits with the various front and back-ends that need to be implemented. It seems a good idea to guide the design with the implementation of a front-end of limited part of the IDL 4.2 grammar and the implementation of a back-end.

This pull request contains the data structures and the front-end for a limited part of the IDL 4.2 grammer restricted to modules, structs, and the necessary base types. Of the literals only the integer literal is included, because it is needed for some of the base types, such as the bounded sequence.


src/types/src/dds_tt.h
```diff
@@ -0,0 +1,214 @@
...
+  union {
+    bool bln;
+    char chr;
+    unsigned long wchr;
+    char *str;
+    unsigned long long ullng;
+    signed long long llng;
+    long double ldbl;
+  };
+} dds_ts_literal_t;
```
k0ekk0ek: This type was already defined somewhere else. I get the feeling that the same goes for multiple types, but haven't checked. Is there a special reason this is defined twice? It seems wrong...

FransFaase: I have removed the type.h file, an older version of dds_tt.h, that got included in the commit.

src/tools/idlc/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+# v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+
+# Find Bison and Flex
```
k0ekk0ek: Please remove the commented out CMake code.

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+set(LEXER_PATH "${GENERATED_SOURCE_DIR}/${LEXER_BASENAME}")
+
+bison_target( parser "src/idl.y" "${PARSER_PATH}")
+flex_target(
+  lexer "src/idl.l" "${LEXER_PATH}"
+  COMPILE_FLAGS
+    "--header-file=${GENERATED_SOURCE_DIR}/idl.lexer.h")
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+set(CMAKE_C_FLAGS -fms-extensions)
```
k0ekk0ek: Please indent in if statements.

src/types/src/type.h
```diff
@@ -0,0 +1,116 @@
+/*
```
k0ekk0ek: I thought the type.h was to be removed from the commit?

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+set_source_files_properties("${PARSER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion")
+set_source_files_properties("${LEXER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion -Wno-unused-parameter -Wno-unused-function")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+set(CMAKE_C_FLAGS "-fms-extensions -Wno-microsoft-anon-tag")
+set_source_files_properties("${PARSER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion")
+set_source_files_properties("${LEXER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion -Wno-unused-parameter -Wno-unused-function")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+set(CMAKE_C_FLAGS "/wd4005")
```
k0ekk0ek: This is suppressing redefinition warnings I think? I specifically don't like the fact that you add it to the CMAKE_C_FLAGS here. Which is probably propagated to all other targets? Have you checked?

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+    "--header-file=${GENERATED_SOURCE_DIR}/idl.lexer.h")
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+set(CMAKE_C_FLAGS -fms-extensions)
+set_source_files_properties("${PARSER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion")
+set_source_files_properties("${LEXER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion -Wno-unused-parameter -Wno-unused-function")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+set(CMAKE_C_FLAGS "-fms-extensions -Wno-microsoft-anon-tag")
+set_source_files_properties("${PARSER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion")
```
k0ekk0ek: Why is the sign conversion warning suppressed? Is this something that we cannot work around in another fashion?

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+set_source_files_properties("${LEXER_PATH}" PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-sign-conversion -Wno-unused-parameter -Wno-unused-function")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+set(CMAKE_C_FLAGS "/wd4005")
+endif()
+
+add_library(types "src/parser.c" "src/tt_create.c" "src/type_walker.c" "src/stringify.c" "${BISON_parser_OUTPUTS}" "${FLEX_lexer_OUTPUTS}")
+target_include_directories(types PRIVATE "${CMAKE_SOURCE_DIR}")
+target_include_directories(types PRIVATE "${CMAKE_SOURCE_DIR}/os/include")
```
k0ekk0ek: This shouldn't be here.

src/types/CMakeLists.txt
FransFaase: These warnings are in the code generated by bison and flex. I have added one define ("#define YYTYPE_INT16 int") to yy_decl.h to remove one of the warnings in idl.parser.c. The other warning, and also most of the warnings in idl.lexer.c need explicit casting to be added to the generated code, for which there is no work-around as far as I could find.

eboasson:
@k0ekk0ek: it seems to me that most of your review comments have been addressed, but that @FransFaase asked for some clarification that you haven't responded to

src/types/src/idl.y
```diff
@@ -0,0 +1,477 @@
...
+  if (dds_ts_context_get_ignore_yyerror(context)) {
+    return 0;
+  }
+
+  fprintf(stderr, "ERROR: %d.%d %s\n", yylloc->first_line, yylloc->first_column, text);
+  return 0;
+}
+
+int
+illegale_identifier(const char *token)
```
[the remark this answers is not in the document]
FransFaase: This function implements the checks as described in Section 7.2.4 (of the IDL 4.2 specification docment) and the escaping mechanism as mentioned in Section 7.2.3.2. Section 7.2.4 mentions that identifiers like Boolean and BOOLEAN are illegal, because the collide with the keyword boolean.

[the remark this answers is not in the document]
k0ekk0ek: I think that the lexer shouldn't return that it's an identifier if it's not(?). Unless it complicates the parser logic, or you need it to generate a specialized error message. Both are not the case(?). Apart from that, I think you may have intended to write illegal as opposed to illegale(?).

src/types/src/parser.c
```diff
@@ -0,0 +1,140 @@
...
+dds_ts_parse_file(const char *file, void (*error_func)(int line, int column, const char *msg))
+{
+  int err = 0;
+  FILE *fh;
+
+  assert(file != NULL);
+
+  if ((fh = fopen(file, "rb")) == NULL) {
+    err = errno;
+    if (error_func != NULL)
```
eboasson: My beloved subject for playing the pedant! Thank you 😁

``NULL`` must be treated as a pointer-to-object because it is simply defined as a "null pointer constant", which in turn is defined as:

> An integer constant expression with the value 0, or such an expression cast to type void *, is called a null pointer constant.66) If a null pointer constant is converted to a pointer type, the resulting pointer, called a null pointer, is guaranteed to compare unequal to a pointer to any object or function.

and the possible definition ``(void *)0`` is a pointer-to-object.

The lovely little thing about C is that conversion between pointer-to-object and pointer-to-function is not guaranteed to work. (It is guaranteed in POSIX and on Windows, but certainly not on all platforms.) So technically, this _may_ result in an invalid conversion.

So technically, the ``NULL`` should be replaced by ``0``. The chances of this ever making any difference are smallish.

src/types/include/typetree.h
```diff
@@ -0,0 +1,216 @@
...
+};
+
+void dds_ts_free_node(dds_ts_node_t *node);
+
+
+/* Type specifications */
+
+/* Type specification (type_spec) */
+typedef struct {
+  dds_ts_node_t;
```
eboasson: This is a C11-ism … as far as I am concerned there is no reason to limit ourselves to C99, what matters is that features are well-supported across compilers.

Gcc and clang support for this has been around for a long time; Visual Stdio appears to have no trouble with, as it builds fine on AppVeyor. From a bit of searching most do, nowadays.

However, my beloved Illumos build with SunPro 12.1 gets broken by it ... so, unless there's a good reason to allow this C11-ism, I'd prefer to see its use removed.

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+add_subdirectory(tests)
+
+# Find Bison and Flex
+find_package(BISON REQUIRED)
+find_package(FLEX REQUIRED)
+find_package(Abstraction REQUIRED)
```
eboasson: Just a note: this is no longer needed with @k0ekk0ek's latest cmake modifications

src/types/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+# v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+add_subdirectory(tests)
```
eboasson: This should be conditional on ``BUILD_TESTING``

src/types/src/idl.y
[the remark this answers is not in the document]
FransFaase: An illegal identifier could be viewed as a lexical error or a semantic error. With respect to parsing, there is no problem to regard it as a semantic error, and just proceed parsing the remainder of the input. The question is also whether you want parsing to stop at the first error or report as much errors as possible. Just like an identifier can 'collide' with a keyword, it can also 'collide' with another identifier. ('collide' means, according to the standard, that they are not the same when taking into account case, but are when compared case insensitive.) When we stop parsing on an illegal identifier, maybe we should also stop parsing when we find an colliding identifier.

src/types/src/tt_create.c
```diff
@@ -0,0 +1,402 @@
...
+  assert(context->cur_node->parent != NULL);
+  context->cur_node = context->cur_node->parent;
+}
+
+void dds_ts_add_struct_forward(dds_ts_context_t *context, dds_ts_identifier_t name)
+{
+  assert(cur_scope_is_definition_type(context, DDS_TS_MODULE));
+  dds_ts_forward_declaration_t *forward_dcl = (dds_ts_forward_declaration_t*)os_malloc(sizeof(dds_ts_forward_declaration_t));
+  if (forward_dcl == NULL) {
+    abort();
```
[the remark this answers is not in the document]
FransFaase: I have modified it. I have changed function into bool functions, that return false when this happens. Also a Boolean flag in context is set. In idl.y YYABORT is called when each of the functions return false.

src/types/src/dds_tt.h
```diff
@@ -0,0 +1,214 @@
...
+  dds_ts_type_spec_t;
+} dds_ts_base_type_t;
+
+/* Sequence type (sequence_type) */
+typedef struct {
+  dds_ts_type_spec_t;
+  dds_ts_type_spec_t* element_type;
+  bool bounded;
+  unsigned long long max;
+} dds_ts_sequence_type_t;
```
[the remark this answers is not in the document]
FransFaase: I have this renamed this into dds_ts_sequence_t. Also did the same with for: dds_ts_string_type_t, dds_ts_fixed_pt_type_t and dds_ts_map_type_t.

src/types/src/dds_tt.h
```diff
@@ -0,0 +1,214 @@
+/*
```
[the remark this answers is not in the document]
FransFaase: I have moved/renamed it to src/types/include/typetree.h.

src/types/src/idl.l
```diff
@@ -0,0 +1,107 @@
...
+"//"                      { BEGIN(COMMENT); }
+<COMMENT>"\n"             { BEGIN(INITIAL); }
+<COMMENT>.                { }
+
+
+{integer_literal}         {
+                            yylval->literal.flags = DDS_TS_UNSIGNED_LONG_LONG_TYPE;
+                            /* strtoll recognizes if the value is dec, oct or hex if base is zero */
+                            yylval->literal.value.ullng = os_strtoull(yytext, NULL, 0); 
+                            if (   os_getErrno() == ERANGE
```
k0ekk0ek: @FransFaase, I don't think this is correct. errno is only set to ERANGE if the number exceeds the maximum. The way to check here is to set errno to 0, then call os_strtoull and see if errno was set. If no error occurred, set it back to the original value.

src/types/src/parser.c
```diff
@@ -0,0 +1,152 @@
...
+    dds_ts_context_set_error_func(context, error_func);
+    yyscan_t scanner;
+    dds_ts_parser_lex_init(&scanner);
+    dds_ts_parser_set_in(fh, scanner);
+    err = dds_ts_parser_parse(scanner, context);
+    if (err == 0) {
+      char buffer[1000];
+      dds_ts_stringify(dds_ts_context_get_root_node(context), buffer, 1000);
+OS_WARNING_MSVC_OFF(4996);
+      printf("Result: '%s'\n", buffer);
```
k0ekk0ek: @FransFaase, I take it this is for debugging purposes and will be removed at some point? Please note it with a FIXME and initialize the first element in the buffer to zero. dds_ts_stringify probably always terminates, but since there's no return value, there's no way to check. Also, is it really necessary to turn warnings off for printf? I find that a bit hard to believe.

src/types/include/typetree.h
```diff
@@ -0,0 +1,216 @@
...
+/* Base type specification (base_type_spec) */
+typedef struct {
+  dds_ts_type_spec_t type_spec;
+} dds_ts_base_type_t;
+
+/* Sequence type (sequence_type) */
+typedef struct {
+  dds_ts_type_spec_t type_spec;
+  dds_ts_type_spec_t* element_type;
+  bool bounded;
```
k0ekk0ek: @FransFaase, I originally called the flags member of the node *flags* because it did more than just denote the type. I called it that because it could contain a bit for indicating that the type was bounded (useful for strings, sequences, maps, ...) and in case of e.g. sequences the type specifier would be the type of element in a sequence and a bit for the sequence. In case of a struct member a bit to denote it's a member (although that could be checked by checking the parent node). I'm not rejecting this pull request because of it, but I'd like to know why that idea was ditched as it seemed useful to me?

src/types/src/parser.c
```diff
@@ -0,0 +1,152 @@
...
+      if (len > 20)
+        os_strlcpy(buffer, "OUT_OF_MEMORY", len);
+      return 2;
+    }
+    yyscan_t scanner;
+    dds_ts_parser_lex_init(&scanner);
+    dds_ts_parser__scan_string(str, scanner);
+    err = dds_ts_parser_parse(scanner, context);
+    if (err != 0) {
+      if (dds_ts_context_get_out_of_memory_error(context))
```
k0ekk0ek: @FransFaase, please use curly brackets with if statements.

src/types/src/parser.c
```diff
@@ -0,0 +1,152 @@
...
+
+  assert(str != NULL);
+  assert(buffer != NULL);
+
+  if (str == NULL || buffer == NULL) {
+    err = -1;
+  } else {
+    dds_ts_context_t *context = dds_ts_create_context();
+    if (context == NULL) {
+      if (len > 20)
```
k0ekk0ek: @FransFaase, this is a rather funny statement in my opinion. strlcpy always nul-terminates and the `len > 20` is chosen arbitrarily unless I'm missing something. Please assert len is larger than zero (better to check if it's large enough to hold the error message) before doing anything and remove this statement. Also, not a fan of the magic values being used to indicate errors. I'm ok with it for now because it's merely for testing and I'm working on exporting error codes from the abstraction layer.

k0ekk0ek:
@eboasson, I think one of the commits without a *signed-off* is causing ip-validation to fail? If possible, can you update the destination branch? Of course, we could also accept that it isn't passing because of a commit that is already in master?

src/types/src/parser.c
FransFaase: Indeed, that statement should have been removed.
With respect to the error codes returned, I will wait for your changes. I expect that some of the parsing functions will move to the tool in the future.

src/types/src/parser.c
FransFaase: I have added curly brackets, here and elsewhere.

src/types/src/parser.c
FransFaase: Yes, it is only temporarily there to show some 'result' of the parsing. It is going to be replaced by a call to (one of) the back-end.

src/types/src/typetree.c
```diff
@@ -0,0 +1,153 @@
...
+  dds_ts_free_node(&struct_member->member_type->node);
+  free_node(&struct_member->node);
+}
+
+static void free_array_size(dds_ts_array_size_t *array_size)
+{
+  free_node(&array_size->node);
+}
+
+extern void dds_ts_free_node(dds_ts_node_t *node)
```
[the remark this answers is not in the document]
FransFaase: I already considered the solution of a free callback function. I have implemented it and also moved code from tt_create.c to typetree.c. In tt_create.c now contains the code that is specific to how the type tree is constructed from calls within idl.y.

eboasson:
@FransFaase and @k0ekk0ek: I am trying to figure out what exactly happened here. Did you rewrite the original two commits and rebase them on top of master?

image

it seems weird that it says committed on 28 November, but then it also happened the other day. The IP validation check details are not available right now, but I think I saw earlier that it was the "other" address that it still complains about, but that one I can't find in these commits ... So consider me a little bit confused. Perhaps the easiest way to deal with it would be essentially throw away the old xtypes branch (it seems unlikely that anybody would be relying on it already) and restart the process.

Furthermore it seems sensible me to decide whether or not rewriting history on the eclipse-cyclonedds main repository is a good idea or not. My initial sentiment is that it isn't.

FransFaase:
Hello,
yes that what I/we did. I fear that once you have accepted a commit that failed the IP validation that everything else based on it will also fail. So, restarting the xtypes branch might be the only way to resolve the issue.
Have a nice weekend,
Frans
----Origineel Bericht----
Van : notifications@github.com
Datum : 18/01/2019 15:59
Aan : cyclonedds@noreply.github.com
Cc : faase009@planet.nl, mention@noreply.github.com
Onderwerp : Re: [eclipse-cyclonedds/cyclonedds] IDL compiler: New front-end supporting modules and structs (#71)
@FransFaase and @k0ekk0ek: I am trying to figure out what exactly happened here. Did you rewrite the original two commits and rebase them on top of master?
it seems weird that it says committed on 28 November, but then it also happened the other day. The IP validation check details are not available right now, but I think I saw earlier that it was the "other" address that it still complains about, but that one I can't find in these commits ... So consider me a little bit confused. Perhaps the easiest way to deal with it would be essentially throw away the old xtypes branch (it seems unlikely that anybody would be relying on it already) and restart the process.
Furthermore it seems sensible me to decide whether or not rewriting history on the eclipse-cyclonedds main repository is a good idea or not. My initial sentiment is that it isn't.

eboasson:
At this point I don't see a more practical way for sorting this out than simply deleting the old xtypes branch, and pushing these two commits one a newly create "xtypes" branch. In the future ... please try to avoid having to resort to such drastic measures.

I'm closing this PR now, that new branch should be there in a little bit. If not 🤔 you'll have to resubmit the PR. Hope this is ok with everyone.

PR #103, opened 2019-01-24
Xtypes

FransFaase:
IDL compiler: generating C99 header file and development of the back-end API.

This pull request worked on generating the C99 header file. It also involves work on the back-end API. To make the API suitable for more types of back-ends, also the generation of meta data in XML (which is included in the C99 source file) was taken into account.

The idea of the API design is to have all tree traverse logic encapsulated and have call-back functions for only generating code. The way the tree is traversed is controlled by flags, currently only one. The idea is that ways of traversal will be common among different back-ends. This design also leaves open the possibility of implementing call-back functions that read from a template file.

At this moment, all the code is included in a single file. This is only for the practical reason that this makes developing the code, at this stage of the development process, easier. Once the interfaces have stabilized (or when a second back-end is being developed), these should be split into a number of files.


FransFaase:
This pull request requires some more study to see how the walker pattern, which was introduced for the stringify function, could be applied.

PR #105, opened 2019-01-28
Implement dds contains

FirasSahliADLinktech:
Implement dds_contains will check if an entity is contained or observed by an other entity


eboasson:
Closing because it is superseded by #106

PR #106, opened 2019-01-28
Implement dds_contains and some minor renames

k0ekk0ek:
This pull request implements dds_contains and a bunch of test cases to verify correct operation.

Apart from that it adds a macro called DDS_IS_ENTITY_KIND to use in dds functions. The macro first tests if the value is greater than zero and, if true, tests if the value binary and DDS_ENTITY_KIND_MASK is greater than zero. This intention of this macro is to quickly test if a value looks like it is a valid entity handle.

I've also opted to rename DDS_KIND_DONTCARE to DDS_KIND_ANY to better represent the intention and because it's shorter and to rename dds_valid_hdl to dds_is_entity to better represent the intent.

This pull request builds on work initiated by @FirasSahliADLinktech, but I found that his implementation tested too much and, at the same time, not enough. Apart from that there was a bug in his implementation that caused already deleted parent entities to slip through. Of course it's not really that big of a deal, but still. Oh and the check in the switch when the parent was a reader wasn't correct. I took credit for the addition of DDS_IS_ENTITY_KIND only because I actually implemented it.

### DIFF
--- a/src/core/ddsc/include/ddsc/dds.h
+++ b/src/core/ddsc/include/ddsc/dds.h
@@ -3326,7 +3326,9 @@ dds_get_topic(
  * @retval DDS_RETCODE_BAD_PARAMETER
  *             Entity or parent parameter is not a valid parameter.
  * @retval DDS_RETCODE_ILLEGAL_OPERATION
- *             The combination of entity and parent is not legal
+ *             The combination of entity and parent is not legal.
+ * @retval DDS_RETCODE_ALREADY_DELETED
+ *             Either entity or parent is already deleted.
  */
 DDS_EXPORT dds_entity_t
 dds_contains (
--- a/src/core/ddsc/include/ddsc/dds.h
+++ b/src/core/ddsc/include/ddsc/dds.h
@@ -2083,7 +2083,7 @@ dds_waitset_get_entities(
  * @param[in]  waitset  The waitset to attach the given entity to.
  * @param[in]  entity   The entity to attach.
  * @param[in]  x        Blob that will be supplied when the waitset wait is
- *                      triggerd by the given entity.
+ *                      triggered by the given entity.
  *
  * @returns A dds_return_t indicating success or failure.
  *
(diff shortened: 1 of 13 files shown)


k0ekk0ek:
I just remember not updating the documentation to reflect it's possible for the function to return already deleted. One (hopefully last) change coming up...

eboasson:
Thanks for the PR, but allow me to raise two objections:

1. I think the decision to include the entities observed by a waitset as "contained" entities confuses two concepts. In my view, it makes more sense to define "A contains B" as that B is included in the set of descendants as determined by recursively calling "get_children", or that B cannot exist in the absence of A.
2. I interpret a function "is_entity" as a test whether the passed-in object is indeed a valid instance of an entity (so akin to "is instance of" in OO languages, where "entity" would be the superclass of all entities). The "valid handle" function that is proposed to be renamed to "is_entity" tests whether the handle is valid, and optionally, whether that entity is of a certain type. The expected error is that the handle is invalid, rather than the referenced object not being an entity. Opinions can differ on the meaning, but I think it was named correctly.

The rename of ``DDS_KIND_DONTCARE`` to ``DDS_KIND_WRITER`` is something I don't care — pun intended — about.

I struggle to see the use cases for this function in the first place, other than as a formality for providing compatibility with the full DCPS API as specified by the OMG. That usage doesn't require the ambiguity between descendants and observers, so I really would like prefer to see it simplified to only deal with descendants.

The "valid_hdl"/"is_entity" check ... well ... those few uses that exist suffer from time-of-check-to-time-of-use issues, so the sensible thing to do is to rewrite the few cases where it gets used and eliminate it altogether. A most elegant way to address my objection to the name change, I'd say 😁

eboasson:
It has been quite a while since I raised my concerns about this PR and as of yet there has been neither a comment that expresses a different view, nor an update of the code. I will close this PR soonish unless there's a response to my concerns.

eboasson:
Another two weeks passed — closing it.

PR #107, opened 2019-01-30
fix and enable SSL support when OpenSSL is available

eboasson:
This PR fixes and enables the SSL/TLS support if OpenSSL is available at build time. The "fixes" bit should not be read as "it was broken and is now correct", but rather as "it was broken enough not to build and now has come to life". I'm sure we have all kept an eye on the interesting ways in which one can use openssl to lock the front door while replacing the back door by a revolving one for easier access, and so I'm rather unwilling to claim that it protects against all foreseeable forms of attack.

But at least some initial tests do suggest it is not willing to accept any random certificate as a valid peer and that it does really require TLS1.2. So, consider it a first step towards proper support for SSL/TLS.

If anyone reading this is knowledgeable about SSL/TLS or OpenSLL … please comment!


k0ekk0ek:
@eboasson would you mind adding a build option to enforce building the with SSL support? That way possible breakage of SSL support will be noticed. Adding `OpenSSL/1.1.1a@conan/stable` to `conanfile.txt` allows you to ensure the dependency is available on all platforms at build time and that users can easily install the necessary dependencies. Then add REQUIRED to the **find_package** line if the option is enabled.

PR #110, opened 2019-02-08
fixup following rename SYSDEPS to OS_MSGHDR_FLAGS

eboasson:
The name change from SYSDEPS_MSGHDR_FLAGS to OS_MSGHDR_FLAGS missed the uses of the macro, with the result that datagram truncation on reception does not result in warning (but in the default configuration, truncation cannot occur); and that the message flags are undefined on sending datagrams, but judging by the man page, the likelihood of this causing problems is also small in practice.


eboasson:
Test CI failure is a connectivity problem on AppVeyor after successfully building 6 out of 8 platforms but I can't restart it other than by pushing changes to the PR ... The source branch has it built successfully, so merging.

PR #122, opened 2019-02-26
Xtypes

FransFaase:
In order to develop the back-end API, work was started on generating the header file of the C99 back-end. This implementation improves on the existing implementation with respect to recursive types. Because the generation of the header file is rather limited with respect to different ways to traverse the type tree, also code was developed to generate the meta data for the source file of the C99 back-end. This was used as input for developping an API based on the walker mechanism that covers both the generation of the header file and the meta data.

The implementation in gen_c99.c makes use of three walkers. The first walker is used to generate the header file. From the walker several functions are called for generating the actual header file. The 'largest' function, for generating the declartion of a member of a struct, contains some logic that could also have been implemented in the walker, but would have resulted in a lot of smaller functions to be defined. Finding the right balance between what to define in a walker and what in a function, is a point of discussion.
A second walker is used for generating the meta data for a struct. This walker, besided making calls to functions, also makes directly generates part of the output.
This second walker is called from a function that is called from the third walker, which is used to generate (part of) the source file. This shows how a walker can be used within a walker.

The walker, which was also used in the initial commit as part of the tests, has been adapted for the new use cases needed for generating the header and source files.

Other changes in this commit:

The grammar was modified to include embedded structs within structs. Support for the #pragma keylist has not been implemented. Code has been added to determine if a struct is used within another struct. (This is used in the source file generation to generate only code for structs that are not part of another struct.)

Code was added to link forward struct declarations to struct definitions. This also required matching reopening of modules by adding a previous pointer to the dds_ts_module_t type.

A bug with processing array declarations in the grammar was fixed.

Tests where added to check the generation of the header file. There are no tests with respect to the meta data that is generated. (The code probably does not produce valid XML in case multiple modules and/or recursive types are used.)

This pull request also includes a merge commit from master (with a small fix for the new Windows builds for Travis).

Added document gen_C99_op.md describing the details about generating the operation codes.


k0ekk0ek:
Hi @FransFaase. Thanks for the pull request. It looks pretty good at first glance. However, doing an in depth review with the current state of the branch isn't very practical. I'd like to propose that @eboasson updates the xtypes branch with the current master and that you update your pull request afterwards. Also, I think it would be better if you don't merge master after that and instead request one of the committers to merge master into this branch every once in a while until the work is ready to be pulled into master itself.

eboasson:
done — I had to resolve a conflict in .travis.yml, I hope I did that correctly.

Issue #91, opened 2019-01-14
lookup_thread_state_real bug found


raiars:
Hi, I found a bug for thread management for user application threads. 
Imagine the usecase: the user is creating multiple threads in a for loop.(5 threads created, once they done, 5 new threads created again). In this case, thread id for a new thread may be same with that of an old thread(which is finished by the user application).  
However, in Cyclone thread management, the old application thread information will be recorded in the global thread pool.
Function lookup_thread_state:
```C
    if ((ts1 = tsd_thread_state) == NULL) {
        if ((ts1 = lookup_thread_state_real()) == NULL) {
            /* this situation only arises for threads that were not created
               using create_thread, aka application threads. since registering

               thread state should be fully automatic the name will simply be
               the identifier */
            tid = os_threadIdSelf();
            (void)snprintf(
                tname, sizeof(tname), "0x%"PRIxMAX, os_threadIdToInteger(tid));
            os_mutexLock(&thread_states.lock);
            ts1 = init_thread_state(tname);
            if (ts1 != NULL) {
                os_osInit();
                ts1->extTid = tid;
                ts1->tid = tid;   //The old thread information is recorded here!
                DDS_LOG(DDS_LC_INFO, "started application thread %s\n", tname);
                os_threadCleanupPush(&cleanup_thread_state, NULL);
            }
            assert(ts1->state == THREAD_STATE_ALIVE);
            os_mutexUnlock(&thread_states.lock);
        }
        tsd_thread_state = ts1;
    }
```
When a new thread with the same thread id comes, it will step into lookup_thread_state_real function. It will satisfy os_threadEqual (thread_states.ts[i].tid, tid) condition and return the information about an old thread. Then it will cause two real threads to both acquire one cyclone thread in the pool. This may cause unexpected problems. Can you have a look at it?
```C
struct thread_state1 *lookup_thread_state_real (void)
{
  if (thread_states.ts) {
    os_threadId tid = os_threadIdSelf ();
    unsigned i;
    for (i = 0; i < thread_states.nthreads; i++) {
      if (os_threadEqual (thread_states.ts[i].tid, tid) ) {   //The new thread with the same thread id will satisfy this condition
        return &thread_states.ts[i];  //An old finished thread information can be returned here!
      }
    }
  }
  return NULL;
}
```

raiars:
My Solution is just not to store a user application thread id into the global pool and it seems work fine. ^_^

eboasson:
[...]

PR #124, opened 2019-03-01
Some minor changes

k0ekk0ek:
First of all, sorry for doing this as a **big bang** commit. While there was no escape for doing everything at once for the rearrangement of the abstraction layer, some changes would have been easier to review if they were done as a separate commit.

This pull request rearranges and fixes up the operating system abstraction layer. The biggest change is that it now operate feature based as opposed to target based, which should make it a lot easier to mix-and-match targets. e.g. use lwIP on Linux and Windows. The intention is to also allow features to be missing if a given target simply does not support it, like was already the case with IPv6. The latter should make it a lot easier to port to other platforms.

All of this is mainly done to make my port to FreeRTOS+lwIP cleaner as I didn't want to mimic and/or create all kinds of stub implementations for things that aren't offered by embedded targets.

- Offer resource usage numbers on all currently supported platforms (Linux, macOS and Windows).
  This makes finding hotspots and problem analysis a little easier.

- Replace os_result by dds_retcode_t and move DDS return code defines down.
  Eliminates the need to convert between different return code types.

- Move dds_time_t down and remove os_time.
  Eliminates the need to convert between different time representations and
  reduces code duplication.

- Remove use of Microsoft source-code annotation language (SAL).
  SAL annotations are Microsoft specific and not very well documented. This
  makes it very difficult for contributers to write.

- Replace the os prefix by ddsrt to avoid name collisions.

- Remove various portions of unused and unwanted code.

- Export thread names on all supported platforms.

- Remove system abstraction for errno. The os_getErrno and os_setErrno were
  incorrect. Functions that might fail now simply return a DDS return code
  instead.

- Remove thread-specific memory abstraction. os_threadMemGet and accompanying
  functions were a mess and their use has been eliminated by other changes in
  this commit.

- Replace attribute (re)defines by ddsrt_ prefixed equivalents to avoid name
  collisions and problems with faulty __nonnull__ attributes.

This rework touches thread creation and destruction, but doesn't fix issue #91.


eboasson:
Thanks @k0ekk0ek! While it will be a pain to review all the details I'm happy to do it. In general, I am in agreement with what you have done, so while there may be a few minor things to come out of a detailed review of all the changes, I don't expect anything major.

I have noticed two things from just trying it out:
- the timestamps in the trace are no longer simply wall-clock times and different threads have different ideas of the time though they do seem to agree that time progresses — perhaps you could have a look at that already?
- on Fedora 29 (which I happen to use), the ``ddsrt`` and ``util`` tests no longer build because of a missing -lpthread on the linker command-line:
```
cmake -GNinja -DBUILD_TESTING=ON -DUSE_SANITIZER=address -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=$PWD/install ../src && ninja
/usr/bin/ld: ddsrt/tests/CMakeFiles/cunit_ddsrt.dir/__/src/sync/posix/sync.c.o: undefined reference to symbol 'pthread_rwlock_tryrdlock@@GLIBC_2.2.5'
/usr/bin/ld: //usr/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
[41/261] Building JAR file /home/erik/.../target/idlc-jar-with-dependencies.jar
ninja: build stopped: subcommand failed.
```
They do build & run just fine if I hack it in, but I know there must exists a better way than what I did, so if you could have a look at that as well, I'd be grateful?

k0ekk0ek:
Definitely has to do with the way the USE_SANITIZER option sets the flags. Builds work fine on both GCC and Clang if I don't include it. @eboasson: Are you seeing the issues with the USE_SANITIZER option disabled too?

k0ekk0ek:
[sanitizer.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/2927034/sanitizer.patch.txt)
@eboasson: Can you try the attached patch, I'm not having the `-lpthread` problem you describe when I use that. Seems the sanitizer option doesn't work as we'd expect. The option *CMAKE_LINKER_FLAGS* doesn't exist, so setting it has no effect, at least on my machine.

eboasson:
@k0ekk0ek, it doesn't make any difference for me ... I wonder what is different on my Fedora installation ...

eboasson:
@k0ekk0ek, it seems I got walked into one of cmake's many traps. It does never *quite* do what it claims to do, especially not when it re-runs automatically. So, false alarm, me idiot. I'm sorry to have wasted your time.

eboasson:
@k0ekk0ek I found two more small things:

* on macOS, it looks like ``pthread_getname_np`` (and therefore ``ddsrt_thread_getname``) returns an empty string on macOS unless ``pthread_setname_np`` has been called. For most processes that means the names for user threads will end up as "(anon)" in the log. I'd prefer to have the thread id instead (like it used to be).
* on Fedora 29 + VMWare Fusion + NAT mode, hostname lookup for unknown hostnames takes forever because of a timeout, and this is visible as a long timeout in starting a process when an network interface name is specified. Perhaps the most sensible approach would be to first check whether the name matches that of an interface, and only then to try doing a hostname lookup.

Callstack for the second one (in case it is useful):
```
#0  0x00007ffff6aed3e8 in __GI___poll (fds=fds@entry=0x7fffffffa498, nfds=nfds@entry=1, 
    timeout=5000, 
    timeout@entry=<error reading variable: That operation is not available on integers of more than 8 bytes.>) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00007ffff41b571b in send_dg (ansp2_malloced=<optimized out>, resplen2=<optimized out>, 
    anssizp2=<optimized out>, ansp2=<optimized out>, anscp=<optimized out>, 
    gotsomewhere=<synthetic pointer>, v_circuit=<synthetic pointer>, ns=<optimized out>, 
    terrno=0x7fffffffa458, anssizp=0x7fffffffa590, ansp=0x7fffffffa448, 
    buflen2=<optimized out>, buf2=<optimized out>, buflen=<optimized out>, 
    buf=<optimized out>, statp=<optimized out>) at res_send.c:1103
#2  __res_context_send (ctx=ctx@entry=0x6040000002d0, 
    buf=buf@entry=0x7fffffffa5c0 "t\033\001", buflen=buflen@entry=33, buf2=buf2@entry=0x0, 
    buflen2=buflen2@entry=0, ans=<optimized out>, ans@entry=0x7fffffffb140 "", 
    anssiz=<optimized out>, ansp=<optimized out>, ansp2=<optimized out>, 
    nansp2=<optimized out>, resplen2=<optimized out>, ansp2_malloced=<optimized out>)
    at res_send.c:510
#3  0x00007ffff41b30ea in __GI___res_context_query (ctx=ctx@entry=0x6040000002d0, 
    name=name@entry=0x7fffffffa7f0 "lo0.localdomain", class=class@entry=1, type=type@entry=1, 
    answer=answer@entry=0x7fffffffb140 "", anslen=anslen@entry=1024, answerp=0x7fffffffb5a8, 
    answerp2=0x0, nanswerp2=0x0, resplen2=0x0, answerp2_malloced=0x0) at res_query.c:208
#4  0x00007ffff41b3756 in __res_context_querydomain (ctx=ctx@entry=0x6040000002d0, 
    name=name@entry=0x602000000230 "lo0", domain=0x60c000000168 "localdomain", 
    class=class@entry=1, type=type@entry=1, answer=answer@entry=0x7fffffffb140 "", 
    anslen=1024, answerp=0x7fffffffb5a8, answerp2=0x0, nanswerp2=0x0, resplen2=0x0, 
    answerp2_malloced=0x0) at res_query.c:593
#5  0x00007ffff41b3f9e in __GI___res_context_search (ctx=ctx@entry=0x6040000002d0, 
    name=name@entry=0x602000000230 "lo0", class=class@entry=1, type=type@entry=1, 
    answer=answer@entry=0x7fffffffb140 "", anslen=anslen@entry=1024, answerp=<optimized out>, 
    answerp2=<optimized out>, nanswerp2=<optimized out>, resplen2=<optimized out>, 
    answerp2_malloced=<optimized out>) at res_query.c:416
#6  0x00007ffff41c8520 in gethostbyname3_context (ctx=ctx@entry=0x6040000002d0, 
    name=name@entry=0x602000000230 "lo0", af=af@entry=2, result=result@entry=0x7fffffffbc90, 
    buffer=buffer@entry=0x7fffffffbf00 "\254\020]\001", buflen=buflen@entry=1024, 
    errnop=0x7ffff668b6c8, h_errnop=0x7ffff668b72c, ttlp=0x0, canonp=0x0)
    at nss_dns/dns-host.c:218
#7  0x00007ffff41c8f3c in _nss_dns_gethostbyname3_r (name=0x602000000230 "lo0", af=2, 
    result=0x7fffffffbc90, buffer=0x7fffffffbf00 "\254\020]\001", buflen=1024, 
    errnop=0x7ffff668b6c8, h_errnop=0x7ffff668b72c, ttlp=0x0, canonp=0x0)
    at nss_dns/dns-host.c:164
#8  0x00007ffff41c8f95 in _nss_dns_gethostbyname2_r (name=<optimized out>, 
    af=<optimized out>, result=<optimized out>, buffer=<optimized out>, 
    buflen=<optimized out>, errnop=<optimized out>, h_errnop=0x7ffff668b72c)
    at nss_dns/dns-host.c:282
#9  0x00007ffff6b0b1bd in __gethostbyname2_r (name=name@entry=0x602000000230 "lo0", 
    af=af@entry=2, resbuf=resbuf@entry=0x7fffffffbc90, buffer=<optimized out>, buflen=1024, 
    result=result@entry=0x7fffffffbc88, h_errnop=<optimized out>) at ../nss/getXXbyYY_r.c:315
#10 0x00007ffff6ae1c53 in gaih_inet (name=name@entry=0x602000000230 "lo0", 
    service=<optimized out>, req=req@entry=0x7fffffffcca0, pai=pai@entry=0x7fffffffbe88, 
--Type <RET> for more, q to quit, c to continue without paging--
    naddrs=naddrs@entry=0x7fffffffbe84, tmpbuf=tmpbuf@entry=0x7fffffffbef0)
    at ../sysdeps/posix/getaddrinfo.c:575
#11 0x00007ffff6ae2808 in __GI_getaddrinfo (name=<optimized out>, service=<optimized out>, 
    hints=0x7fffffffcca0, pai=0x7fffffffcc60) at ../sysdeps/posix/getaddrinfo.c:2265
#12 0x00007ffff72469f4 in __interceptor_getaddrinfo (node=0x602000000230 "lo0", service=0x0, 
    hints=0x7fffffffcca0, out=0x7fffffffcc60)
    at ../../../../libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:2353
#13 0x00007ffff6e45f44 in ddsrt_gethostbyname (name=0x602000000230 "lo0", af=2, 
    hentp=0x7fffffffcd80) at /home/erik/cyclonedds/src/ddsrt/src/sockets.c:275
#14 0x00007ffff6cf9099 in ddsi_ipaddr_from_string (tran=0x7ffff71d69e0 <ddsi_udp_factory_g>, 
    loc=0x7fffffffd0a0, str=0x602000000230 "lo0", kind=1)
    at /home/erik/cyclonedds/src/core/ddsi/src/ddsi_ipaddr.c:102
#15 0x00007ffff6cf5675 in ddsi_udp_address_from_string (
    tran=0x7ffff71d69e0 <ddsi_udp_factory_g>, loc=0x7fffffffd0a0, str=0x602000000230 "lo0")
    at /home/erik/cyclonedds/src/core/ddsi/src/ddsi_udp.c:467
#16 0x00007ffff6cf2494 in ddsi_locator_from_string (loc=0x7fffffffd0a0, 
    str=0x602000000230 "lo0") at /home/erik/cyclonedds/src/core/ddsi/src/ddsi_tran.c:280
#17 0x00007ffff6d6a819 in find_own_ip (requested_address=0x602000000230 "lo0")
    at /home/erik/cyclonedds/src/core/ddsi/src/q_nwif.c:601
#18 0x00007ffff6d5aaaa in rtps_init () at /home/erik/cyclonedds/src/core/ddsi/src/q_init.c:938
#19 0x00007ffff6df2336 in dds_init (domain=-1)
    at /home/erik/cyclonedds/src/core/ddsc/src/dds_init.c:132
#20 0x00007ffff6de85e4 in dds_create_participant (domain=-1, qos=0x0, listener=0x0)
    at /home/erik/cyclonedds/src/core/ddsc/src/dds_participant.c:162
#21 0x000000000041a38e in common_init (
    argv0=0x7fffffffe5bd "/home/erik/cyclonedds/build/bin/pubsub")
    at /home/erik/cyclonedds/src/tools/pubsub/common.c:235
#22 0x0000000000416068 in main (argc=2, argv=0x7fffffffe348)
    at /home/erik/cyclonedds/src/tools/pubsub/pubsub.c:2535
```

eboasson:
Some more things:

1. Moving the include files from ``ddsc`` to ``dds`` kinda makes sense to me (there's always a risk of name clashes, but those can always be avoided by installing the entirety of Cyclone DDS in some special directory), but what I don't like is making such a change without a decent heads-up to anyone using it. Indeed, some people may object to the change.
Adding a ``ddsc/dds.h`` that forwards to ``dds/dds.h`` should suffice to retain compatibility, So I would like to see that added, and additionally to have this change announced as a separate GitHub issue. Later on we can then add ``#warning`` and eventually remove it.

2. ``dds_sleepfor`` on Windows should round the sleep duration up, so should use ``(n + DDS_NSECS_IN_MSEC - 1) / DDS_NSECS_IN_MSEC``, like ``ddsrt_cond_waituntil`` does. There is the possible overflow problem as well, but as long sleeps should never occur, that can be addressed by asserting (or aborting).

3. On Windows, it makes more sense to define ``ddsrt_cond_waituntil`` in terms of ``ddsrt_cond_waitfor``, rather than the other way around as it is now. Also, the ``DDS_NEVER`` case doesn't really get used (although it can realistically happen when in the year 2260 or so), and so I think it makes more sense to let ``nsecs_to_msecs_ceil`` deal with it — a spurious wakeup might result, but that's a non-issue. Finally, ``nsecs_to_msecs_ceil`` could be rewritten to a form I find easier to read (and that I'm guessing the compilers will turn it into), but I don't think that's a big deal.

k0ekk0ek:
@eboasson, I've changed the behavior of `ddsrt_thread_getname` to return a string containing the native (system) thread identifier if pthread_getname_np returns an empty string. Also, time calculation in `dds_log` should be fixed.

eboasson:
Some more comments:
* ``ddsi_tcp_conn_read_plain`` and ``ddsi_tcp_conn_read_ssl`` have a terribly broken interface, with an error code as an out parameter and a length as a return parameter (this was so before), but it turns out their caller (``ddsi_tcp_conn_read``) first checks the return value, and only looks at the error code if the return value is < 0.
The implementation of (at least) ``ddsi_tcp_conn_read_plain`` assumes that the received length is not modified by the ``ddsrt_recv`` in case of an error, but this seems like a frightfully fragile interface. The current implementation does behave that way, but I would nonetheless prefer seeing the implementation of ``ddsi_tcp_conn_read_plain`` changed to avoid this assumption.
Longer-term, the interface of those functions needs to be changed, but I appreciate the effort you've put in here and I don't want to insist on fixing ever more broken interfaces as part of this PR. Hence, the request for merely changing the code to avoid that assumption.
(Note: I haven't checked the write functions in detail, but it certainly looks like it suffers from the same.)
* there are cases where an ``ddsi_ssl_accept`` is reported only as a NULL return, not because the socket has been set to "invalid socket". The pragmatic solution is to change the check in ``ddsi_tcp_accept``. (This was probably wrong before …)
* in ``ddsi_is_ssm_mcaddr``, I don't understand why the ``#ifdef DDSI_INCLUDE_SSM`` is needed
* in ``sockets/windows.h`` it sets ``DDSRT_HAVE_SSM`` to 0, but MSDN suggests it is supported. Could you please double-check?

martinbremmer:
The ddsrt_recvmsg() does not work on Windows. The rcvd argument is _not_ set when recvfrom succeeds.

k0ekk0ek:
@martinbremmer. Thanks for the comment. I guess I focussed to much on translation of error codes that this got left out. As you can guess, Windows isn't my primary platform :). Sloppy, but good thing that you noticed.

martinbremmer:
@k0ekk0ek, I've created a test that reproduces the recvmsg issue in the ctest run. Please see https://github.com/martinbremmer/cyclonedds/commit/ecfd1e7e8f9ff1fd78f6a4f4ce4ee45be40a683e

k0ekk0ek:
Review comments so far are processed. I cannot fix the conflict from my end unless I rebase as a merge makes the ip check fail. I propose that we resolve the conflicts using the "Resolve conflicts" button that should be enabled for committers. Let me know if there's anything else in this pull request you want to see changed.

eboasson:
@k0ekk0ek: GitHub tells me the conflicts are too complex to use the web-editor!

I have two problems trying to build on RH6.5:

With CMake 3.6.1 — I know that cmake version is officially too old, but whatever — the check for the presence of clock_gettime fails because the second check never does anything serious, if I change ``ddsrt/CMakeLists.txt`` as follows:
```diff
diff --git a/src/ddsrt/CMakeLists.txt b/src/ddsrt/CMakeLists.txt
index 30627b1..9a5ff1f 100644
--- a/src/ddsrt/CMakeLists.txt
+++ b/src/ddsrt/CMakeLists.txt
@@ -170,8 +170,9 @@ elseif(UNIX)
   check_library_exists(c clock_gettime "" HAVE_CLOCK_GETTIME)
   if(NOT HAVE_CLOCK_GETTIME)
     # Before glibc 2.17, clock_gettime was in librt.
-    check_library_exists(rt clock_gettime "time.h" HAVE_CLOCK_GETTIME)
-    if(HAVE_CLOCK_GETTIME)
+    check_library_exists(rt clock_gettime "time.h" HAVE_CLOCK_GETTIME_1)
+    if(HAVE_CLOCK_GETTIME_1)
+      set(HAVE_CLOCK_GETTIME TRUE)
       target_link_libraries(ddsrt INTERFACE rt)
     endif()
   endif()
```
it works. Maybe we should just do it this way?

Secondly, there is a call to ``ddsrt_thread_toid`` in OpenSSL 1.0 compatibility code, but ``ddsrt_thread_toid`` doesn't exist. I'd prefer a fix for that over dropping the compatibility code and requiring that OpenSSL should be ≥ 1.1.0. What would you suggest?

k0ekk0ek:
`ddsrt_thread_id` has been replaced with a platform native `ddsrt_gettid`. `ddsrt_thread_id` returned a `pthread_t` casted to `uintptr_t` for POSIX platforms. `ddsrt_gettid` returns the native thread identifier. Given that the call to `ddsrt_thread_id (ddsrt_thread_self ())` more-or-less equals `ddsrt_gettid ()` (they both return a unique thread identifier), replacing the code with a call to `ddsrt_gettid` is the easiest/best way to solve the issue.

I'm fine with the CMake change.

One commit coming up...

eboasson:
@k0ekk0ek getting there ... three more, none of them incorrect code, just warnings from clang on macOS with all sorts of warnings enabled, and one of them from clang's static analyzer:

* ``dds_alloc``: null pointer argument in call to ``memset`` — easily solved by making the memset conditional on the allocation succeeding; more elegant would be to add a dds_log_fatal function that always sets DDS_LC_FATAL and is marked as "noreturn";
* ``ddsrt_duration_to_timeval_ceil``: possible misuse of comma operator because of the ``assert`` in the conditional expression; I'd convert the assertion to a compile-time one;
* ``ddsrt_getrusage``: declaration of "ret" shadows a local variable in the macOS specific code.

If you could address those too ... I'll be happy to accept it.

That then leaves the problem of the conflict and the various small fix-ups. I'd be happy with you doing a rebase (combining commits as you deem appropriate) and a force push. It will be easy enough for me to check my local copy against the result of the force push.

k0ekk0ek:
@eboasson. I processed all the comments and squashed all the commits as they were just fixups to the initial commit. At least now we have a single working commit. Awaiting the merge...

PR #128, opened 2019-03-15
Fix possible indefinite blocking in writer throttling

eboasson:
This PR fixes two separate problems: one that is theoretical (the one that changes the use of the ``throttling`` flag to counting concurrent calls), the other I just happened to run into. It surprises me it evaded detection for so long ...

The second one may be the cause of #125 but I haven't yet got my hands on a reliable Windows 10 machine and so haven't been able to dig into that problem properly.


eboasson:
Closing because I think the committer agreement issue is now resolved but maybe requires reopening the pull request.

PR #130, opened 2019-03-18
Update and extend type support design document

k0ekk0ek:
After having reviewed and tested the code submitted as part of the initial design, I had some remarks. In my opinion the tree was to hard to navigate, specifically to many indirections, and the design document didn't make it clear why certain decisions were made or explained the "architecture" in a way that is easy to grasp. Please consider pulling my as part of the work needed to change the inner workings and api. Please also shine a light on my proposal if you can find the time.


docs/dev/typesupport.md
```diff
@@ -0,0 +1,185 @@
...
+  type_t *target;
+} forward_decl_t;
+
+/* Constructed types introduce a complexity when the TypeTree is destructed
+   because they can be both embedded and referenced. A constructed type must
+   only be destructed from the scope in which it is defined, never when merely
+   referenced. */
+typedef struct {
+  typespec_t type;
+  type_t *members;
```
FransFaase: It is important to note that the members are heterogeneous, in the sense that the elements may have different types, just like a module can contain different type of members, such struct and union definitions.

There is a certain overlap between what can occur in a module and a struct. If one would define a walk function that walks over different types in the type tree, it would be handy if somehow the members are shared, so that in the implementation you do not need to differentiate between them.

k0ekk0ek:
@eboasson. There were still some warnings in the Flex generated code. Those have been fixed. When the Travis build finishes, perhaps you can check my AppVeyor builds to avoid having to wait yet another day?

PR #135, opened 2019-03-25
Possible assert when ddsi socket write fails.

martinbremmer:
The pull request #124 was a major overhaul of the operating system abstraction layer.
One small change within that pull request was a syntax change of the socket send functions.

Previously, the 'nr of written bytes' variable contained an error value when the send failed. Now, however, that value is not touched when the ddsrt socket send fails. But the DDSI write still expected it to contain an error value when the send failed.

Now, the DDSI write(s) dependent on the ddsrt socket return value when the send failed, instead of the 'nr of written bytes'.

NB. The socket recv usages don't have this issue.


martinbremmer:
I've closed this pull request, because I didn't have a Eclipse account yet and my commits were not signed.

martinbremmer:
Just for future reference, the new PR that has an eclipse signature is #136.

PR #142, opened 2019-04-01
Fix format strings and signatures for fixed size integers

k0ekk0ek:
This pull request contains some signature and format string fixes fixed size integers (mostly) that I ran into while getting everything to compile with the Xilinx SDK. It contains one or two small other changes too, but no real functional changes.

### DIFF
--- a/src/core/ddsc/src/dds_write.c
+++ b/src/core/ddsc/src/dds_write.c
@@ -185,10 +185,10 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
     if (!config.whc_batch)
       nn_xpack_send (wr->m_xp, false);
     ret = DDS_RETCODE_OK;
-  } else if (w_rc == ERR_TIMEOUT) {
+  } else if (w_rc == Q_ERR_TIMEOUT) {
     DDS_ERROR ("The writer could not deliver data on time, probably due to a reader resources being full\n");
     ret = DDS_ERRNO (DDS_RETCODE_TIMEOUT);
-  } else if (w_rc == ERR_INVALID_DATA) {
+  } else if (w_rc == Q_ERR_INVALID_DATA) {
     DDS_ERROR ("Invalid data provided\n");
     ret = DDS_ERRNO (DDS_RETCODE_ERROR);
   } else {
@@ -224,10 +224,10 @@ dds_return_t dds_writecdr_impl_lowlevel (struct writer *ddsi_wr, struct nn_xpack
     if (!config.whc_batch && xp != NULL)
       nn_xpack_send (xp, false);
     ret = DDS_RETCODE_OK;
-  } else if (w_rc == ERR_TIMEOUT) {
+  } else if (w_rc == Q_ERR_TIMEOUT) {
     DDS_ERROR ("The writer could not deliver data on time, probably due to a reader resources being full\n");
     ret = DDS_ERRNO(DDS_RETCODE_TIMEOUT);
-  } else if (w_rc == ERR_INVALID_DATA) {
+  } else if (w_rc == Q_ERR_INVALID_DATA) {
     DDS_ERROR ("Invalid data provided\n");
     ret = DDS_ERRNO (DDS_RETCODE_ERROR);
   } else {
--- a/src/core/ddsi/include/dds/ddsi/ddsi_tran.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_tran.h
@@ -26,6 +26,12 @@
 
 #define DDSI_TRAN_ON_CONNECT 0x0001
 
+#if DDSRT_HAVE_IPV6 == 1
+# define DDSI_LOCATORSTRLEN INET6_ADDRSTRLEN_EXTENDED
+#else
+# define DDSI_LOCATORSTRLEN INET_ADDRSTRLEN_EXTENDED
+#endif
+
 /* Core types */
 
 typedef struct ddsi_tran_base * ddsi_tran_base_t;
--- a/src/core/ddsi/include/dds/ddsi/q_bitset.h
+++ b/src/core/ddsi/include/dds/ddsi/q_bitset.h
@@ -13,41 +13,42 @@
 #define NN_BITSET_H
 
 #include <assert.h>
+#include <stdint.h>
 #include <string.h>
 
 #include "dds/ddsi/q_unused.h"
 
-inline int nn_bitset_isset (unsigned numbits, const unsigned *bits, unsigned idx)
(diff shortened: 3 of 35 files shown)


src/core/ddsc/src/dds_write.c
```diff
@@ -185,7 +185,7 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
     if (!config.whc_batch)
       nn_xpack_send (wr->m_xp, false);
     ret = DDS_RETCODE_OK;
-  } else if (w_rc == ERR_TIMEOUT) {
+  } else if (w_rc == Q_ERR_TIMEOUT) {
```
eboasson: I can see the pragmatism behind this, but just renaming ERR_TIMEOUT to Q_ERR_TIMEOUT is a bit too pragmatic to my taste. I'd rather the DDSI stack started using the same DDS_RETCODE_xxx codes instead of defining its own ones.

src/core/ddsi/include/dds/ddsi/q_bitset.h
```diff
@@ -13,40 +13,41 @@
...
 #include <assert.h>
+#include <stdint.h>
 #include <string.h>
 
 #include "dds/ddsi/q_unused.h"
 
-inline int nn_bitset_isset (unsigned numbits, const unsigned *bits, unsigned idx)
+inline int nn_bitset_isset (uint32_t numbits, const uint32_t *bits, uint32_t idx)
 {
   return idx < numbits && (bits[idx/32] & (1u << (31 - (idx%32))));
```
eboasson: The 1u now also needs to be (uint32_t)1: if uint32_t need not be equivalent to unsigned (and in particular, if unsigned can be narrower than uint32_t, using 1u might result in undefined behaviour here). Same holds for of course for all these constants in this file.

src/core/ddsi/include/dds/ddsi/q_time.h
```diff
@@ -53,9 +53,9 @@ int valid_ddsi_timestamp (nn_ddsi_time_t t);
 DDS_EXPORT nn_wctime_t now (void);       /* wall clock time */
 DDS_EXPORT nn_mtime_t now_mt (void);     /* monotonic time */
 DDS_EXPORT nn_etime_t now_et (void);     /* elapsed time */
-DDS_EXPORT void mtime_to_sec_usec (int * __restrict sec, int * __restrict usec, nn_mtime_t t);
-DDS_EXPORT void wctime_to_sec_usec (int * __restrict sec, int * __restrict usec, nn_wctime_t t);
-DDS_EXPORT void etime_to_sec_usec (int * __restrict sec, int * __restrict usec, nn_etime_t t);
+DDS_EXPORT void mtime_to_sec_usec (int32_t * __restrict sec, int32_t * __restrict usec, nn_mtime_t t);
```
eboasson: Switching to an unsigned type would deal gracefully with wrap around after 2106 or thereabouts (even if it is only used for printing). Secondly, the format strings need to updated as well — from %d to whichever type is used.

src/core/ddsi/include/dds/ddsi/ddsi_tran.h
```diff
@@ -26,6 +26,12 @@
 
 #define DDSI_TRAN_ON_CONNECT 0x0001
 
+#if DDSRT_HAVE_IPV6 == 1
+# define DDSI_LOCATORSTRLEN INET6_ADDRSTRLEN_EXTENDED
```
eboasson: It should be the maximum of all supported address families (which currently can include Ethernet as well). That's not a new issue and it isn't entirely trivial to get the correct value either, so I'm willing to let it be like this for now.

src/core/ddsi/src/q_nwif.c
```diff
@@ -39,7 +39,7 @@ static void print_sockerror (const char *msg)
 
 unsigned locator_to_hopefully_unique_uint32 (const nn_locator_t *src)
 {
-  unsigned id;
+  unsigned id = 0;
   if (src->kind == NN_LOCATOR_KIND_UDPv4 || src->kind == NN_LOCATOR_KIND_TCPv4)
     memcpy (&id, src->address + 12, sizeof (id));
   else
```
eboasson: Actually, the else branch shouldn't be IPv6 dependent: it should rather take all bytes & do its thing regardless of the locator type. IPv4 is special in that it fits in a 32-bit int.

src/core/ddsc/src/dds_write.c
k0ekk0ek: @eboasson, I had a feeling you'd say that. Agreed. I'll incorporate changing the `ERR_*` variables.

k0ekk0ek: On second thought, I'll skip this. Error codes in ddsi are negative and some are used in functions that return a positive integer for a count. To replace the `ERR_*` codes with `DDS_RETCODE_*` equivalents means that I need to change the `DDS_RETCODE_*` ones to be negative, which is a bit to much work to do simultaneously with making FreeRTOS a tier-1 platform.

src/core/ddsi/include/dds/ddsi/q_time.h
k0ekk0ek: For now I'll focus on getting FreeRTOS support done first. Switching to an unsigned int later on is the same amount of work, so I feel it's not a priority right now. I am wondering which format string you mean wrt these functions. I though I fixed all of them. Maybe I'm overlooking something?

src/core/ddsi/include/dds/ddsi/q_bitset.h
k0ekk0ek: Switched to using UINTxx_C.

src/core/ddsi/src/q_nwif.c
k0ekk0ek: This one is just to fix a compiler warning about `id` being uninitialized if IPv6 is not available. There's no change in logic. This situation should never arise because non UDPv4 or TCPv4 locators should not be used if nothing else is available on a given target. So the only thing that actually changed is that the result in a scenario that cannot happen is predictable. I propose we leave this as is for now since it is out-of-scope for my work on the FreeRTOS port.

k0ekk0ek:
@eboasson, don't know why Travis suddenly fails, but can you check the last commit on my *local* branch to verify the changes are correct? Travis and AppVeyor builds are successful there.

k0ekk0ek:
It seems someone merged something that contained a newly introduced *ERR_* prefixed error ;) Squashed the commits and force pushed them. Let's hope nothing is merged in before my merges :D

k0ekk0ek:
Maybe it's better to just merge #150, it builds on these changes.

eboasson:
Given that #150 has been merged, I'm closing this one.

PR #144, opened 2019-04-03
Add user data pointer to dds_entity

dennis-adlink:
To build API's (e.g. a C++11 API) on top op Cyclone's C99 API, bottom-up navigation is required, i.e. from a C99 handle to the corresponding C++ object. This is needed to facilitate things like listener-callbacks and lookup functionality (e.g. the get_subscriber call on the, which is supposed to return the Subscriber that used when creating a DataReader). This pull request adds a user_data pointer to the dds_entity type, so that a backref can be stored that points to the API object built on top of the C99 API.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -715,6 +715,61 @@ DDS_EXPORT dds_return_t
 dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
 
 /*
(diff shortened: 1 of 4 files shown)


rvloon:
That assumes that there will ever be one backref necessary. Wouldn't it make more sense to have a map available between the handle and the needed object? This way you would separate language support from the core objects. What if you needed multiple languages to be supported at the same time?

PR #152, opened 2019-04-11
Multi Process Testing framework

martinbremmer:
Some features and functionalities of Cyclone DDS can only be tested when there's communication between processes. Examples are durability, security, etc.
But 'normal' communication between processes should also be tested because of code paths that will otherwise not be executed (like calls to sockets). There are no tests for that yet.

This PR will introduce a framework for easy multi process testing and will remedy the lack of those basic communication tests.

It is basically a four parter.

1. Rudimentary Process Management.
This is the result of the previous PR #137. Except that ddsrt_proc_waitpid() and ddsrt_proc_waitpids() have an added ``DDS_RETCODE_NOT_FOUND`` return value.

2. Multi Process Testing framework.
The framework should provide the functionality to easily create multi process tests and can be found in the ``src/mpt/mpt`` directory.
It also contains a self test (``src/mpt/tests/self``) which is switched off by default. Use ``MPT_ENABLE_SELFTEST`` to switch it on.

3. Basic MPT tests.
A few basic multi process tests are introduced in ``src/mpt/tests/basic``. All tests use the same hello_publisher and hello_subscriber. The principle is that it is expected that the subscriber(s) receive the same text as the publisher(s) send.

4. ~~Moved expand_envvars.
MPT uses the expand_envvars functionality. To be able to build MPT (and without warnings) on Windows, expand_envvars had to be moved from ``util`` to ``ddsrt``.
This was part of [the plan](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/modules.md) anyways.~~
Edit: this was replaced after a rebase to master where it was also implemented.


martinbremmer:
For some reason, appveyor was unable to do some conan stuff. I can't relate the failure to my changes. I also don't know how and if I can restart the appveyor builds...

eboasson:
@martinbremmer I can't restart an AppVeyor build either, but fortunately we also have Windows builds on Travis.

The error as nothing to do with the code, it is some networking problem:

> cunit/2.1-3@bincrafters/stable: Not found in local cache, looking in remotes...
136cunit/2.1-3@bincrafters/stable: Trying with 'conan-center'...
137ERROR: HTTPSConnectionPool(host='conan.bintray.com', port=443): Max retries exceeded with url: /v1/ping (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x03F61890>: Failed to establish a new connection: [Errno 11001] getaddrinfo failed',))
138
139Unable to connect to conan-center=https://conan.bintray.com
140

Fortunately, I can decide to overlook a failed AppVeyor build :smile: But I haven't done my due diligence on the new bits in the pull requests yet.

eboasson:
@martinbremmer I have finally taken a more detailed look, including running it myself on my mac — and unfortunately I found some problems:
* if an MPT test *doesn't* specify a timeout, ctest ends up being told that it has to wait forever for the test to complete;
* that default timeout in the MPT library for waiting for a process is 1600s ... I understand the argument that it should be longer than the ctest timeout, but I think this is somewhat excessive;
* at least on macOS, ``waitpid`` is interrupted by SIGCHLD and so you need to either call ``waitpid`` from a SIGCHLD handler, or you have to restart the call if it was interrupted for any reason other than SIGALRM.
The first two of these issues directly affect the timely termination of a test suite, the third causes test failures to be reported when the system is actually functioning correctly. So these really need to be addressed.

martinbremmer:
I was (and am) under the impression that if you doen't specify a timeout (which will translate to a 0 timeout), then ctest will use its default, which is 1500 seconds.
That's also where the 1600 seconds comes from (just a value larger than 1500). 
However, I do agree that that is quite excessive, so I've reduced the default to 60 seconds.

Regarding the SIGCHLD vs waitpid problem, I've found a few websites that mentioned to set ``signal(SIGCHLD, SIG_DFL)`` if you don't want those triggers. This seems the default for most platforms, but apparently not all.

eboasson:
Thanks @martinbremmer for also rebasing it without complaining about my having moved all of the old ``util`` directory into ``ddsrt``. I do have one thing though: now ``expand_envvars`` is declared in both ``environ.h`` and in ``expand_envvars.h``.

I like ``environ.h`` rather better than the ``expand_envvars.h``, and I definitely don't want it declared in both. Can you please clean it up? Preferably by keeping ``environ.h`` and removing the other one!

If you could sort that out, then it will be my pleasure to merge it.

eboasson:
I'm trusting Travis to have done a good job and that the plug-in loading failure of maven on AppVeyor is simply the next one in a long line of failures to load a plug-in.

PR #153, opened 2019-04-12
add RPATH for compiled tools to find the libddsc.so

cwyark:
When assigning CMAKE_INSTALL_PREFIX into non-default path, e.g. /opt/CycloneDDS. 

```bash
$> cmake -Bbuild -H./src -DCMAKE_INSTALL_PREFIX=/opt/CyclineDDS 
$> sudo cmake --build build --target install 
```
Since libddsc.so is not the default library searching path in this situation, e.g. /opt/CycloneDDS/lib.
Those compiled tools like `ddsls` and `pubsub` are failed to find libddsc.so. 
Add relative rpath "$ORIGIN/../lib" for these targets to fix this problem.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,6 +155,11 @@ endif()
 include(FileIDs)
 include(GNUInstallDirs)
 include(AnalyzeBuild)
+if(APPLE)
+  set(CMAKE_INSTALL_RPATH "@loader_path/../${CMAKE_INSTALL_LIBDIR}")
+else()
+  set(CMAKE_INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}")
+endif()
 # Include Coverage before CTest so that COVERAGE_COMMAND can be modified
 # in the Coverage module should that ever be necessary.
 include(Coverage)
--- a/src/tools/ddsls/CMakeLists.txt
+++ b/src/tools/ddsls/CMakeLists.txt
@@ -11,11 +11,6 @@
 #
 add_executable(ddsls ddsls.c)
 target_link_libraries(ddsls ddsc)
-if(APPLE)
-  set_target_properties(ddsls PROPERTIES INSTALL_RPATH "@loader_path/../lib")
-else()
-  set_target_properties(ddsls PROPERTIES INSTALL_RPATH "$ORIGIN/../lib")
-endif()
 
 install(
   TARGETS ddsls
--- a/src/tools/ddsls/CMakeLists.txt
+++ b/src/tools/ddsls/CMakeLists.txt
@@ -11,6 +11,11 @@
 #
 add_executable(ddsls ddsls.c)
 target_link_libraries(ddsls ddsc)
+if(APPLE)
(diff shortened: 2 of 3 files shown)


eboasson:
Thanks @cwyark! As these things go, I ran into this the day after you created this pull request 😁

There are two complications in accepting this PR:
* anyone contributing to an Eclipse project must have signed the Eclipse Contributer Agreement (the ECA, this is what the "The author(s) of the pull request is not convered ..." is about). It is very simple to sign, it is just an online form
* I am not CMake expert whatsoever. I would probably have done the same thing as you propose, but I have a feeling there exists some CMake setting to get this result without having a duplication or platform-specific bits in there. @k0ekk0ek: can you please have a look?

k0ekk0ek:
Sorry, I got sick over the weekend. I'll have a quick look.

k0ekk0ek:
Personally I never install, so I never noticed the issue, but @cwyark it seems the library directory is not automatically added when installed. I more-or-less agree with the proposed solution. Not sure if adding a relative path is better than using an absolute path. Do you have an opinion on the matter @eboasson? Anyway, I propose to set CMAKE_INSTALL_RPATH instead. A patch is attached below.

```
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 19b0515..a211f28 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,6 +155,13 @@ endif()
 include(FileIDs)
 include(GNUInstallDirs)
 include(AnalyzeBuild)
+
+if(APPLE)
+  set(CMAKE_INSTALL_RPATH "@loader_path/../${CMAKE_INSTALL_LIBDIR}")
+else()
+  set(CMAKE_INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}")
+endif()
+
 # Include Coverage before CTest so that COVERAGE_COMMAND can be modified
 # in the Coverage module should that ever be necessary.
 include(Coverage)
```

eboasson:
@k0ekk0ek , @cwyark , I rather like the relative path. That way it works regardless of the installation location.

@cwyark : as I mentioned, I cannot just merge this because of the Eclipse contributor paperwork. I am happy to take your fix updated with the CMake wizardry of @k0ekk0ek, but then you will need to do the paperwork and update your branch accordingly.

I am also happy to do this by hand myself, but then it won't have your name on it anymore, I'm afraid. I can still mention you as the originator of the patch, of course.

So please let me know what you prefer.

cwyark:
@eboasson Finally, I just sign up the Eclipse contributor agreement and update the sign-off to the commits of the branch.

eboasson:
Hi @cwyark, I thought I'd check with you what your plan is with this pull request. Perhaps you have not noticed that the eclipsefdn/eca check fails again?

Secondly, @k0ekk0ek's suggestion remains and as I mentioned earlier I much prefer solving the problem in one place, as he suggests, rather than having to do this in each CMakeLists.txt that generates an executable.

I was hoping you'd incorporate his suggestion and get all the checks passing so that I could merge it.

PR #158, opened 2019-04-18
Remove dependency on jdk8 Chocolatey package

k0ekk0ek:
This pull request removes the dependency on the jdk8 Chocolatey package. The installation of jdk8 package failed because the archive wasn't/isn't available anymore. Luckily the Windows image has the Android SDK installed and OpenJDK 1.8 is available by default. Apart from fixing the issue with failing builds, it also shaves a couple of minutes from the build process and makes the `.travis.yml` a lot easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,13 +77,6 @@ before_install:
 # The real fix is to wait until Python and pip are both available on the
 # target. Until then download Conan from the official website and simply add
 # the extracted folder to the path.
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
And Travis builds are done. I guess we'll see if AppVeyor thinks my changes to `.travis.yml` are ok sometime tomorrow :yum:

PR #166, opened 2019-04-24
Add support for FreeRTOS and lwIP

k0ekk0ek:
This pull request add support for running on *[FreeRTOS](https://www.freertos.org/)* and adds support for using the *[lwIP](https://savannah.nongnu.org/projects/lwip/)* TCP/IP stack. A combination of these is often used on embedded targets, but lwIP can be used on general purpose platforms like Linux and Windows too and the changes to `src/ddsrt/CMakeLists.txt` now provide a good example on how platforms can be mixed and matched in the future. Note that the same mechanism can also be used to change e.g. crypto libraries per platform too.

I had (at least some) success running the code on the [simulator](https://github.com/k0ekk0ek/FreeRTOS-lwIP-Sim) that I slapped together last year. Note though that I did not use lwIP yet due to difficulty getting multicast to work. Also, some minor changes were required. I'll be sure to update the simulator as well (maybe provide a Docker image?) so that it can be used to verify pull requests against FreeRTOS as well.

An initial review would be very appreciated. Of course, if you don't have any review comments, feel free to simply merge :wink:.


k0ekk0ek:
For the future: [FreeRTOS+POSIX](https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_POSIX/index.html) seems interesting too.

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+      if (buf == NULL) {
+        rc = DDS_RETCODE_OUT_OF_RESOURCES;
+      } else {
+        cond->tasks.buf = buf;
+        cond->tasks.len = len;
+      }
+    }
+
+    /* Initialize memory that should not be used anymore/yet to zero. */
+    if (off < cond->tasks.len) {
```
eboasson: that needs to take realloc failure into account

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+
+static dds_retcode_t
+cond_push(ddsrt_cond_t *cond, TaskHandle_t task)
+{
+  size_t cnt;
+
+  assert(cond != NULL);
+  assert(task != NULL);
+
+  /* Ensure the task is not listed. */
```
eboasson: That's a somewhat misleading comment: you're searching for it, but then sort-of allows for the case where it is listed

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+  for (cnt = 0; cnt < cond->tasks.cnt && cond->tasks.buf[cnt] != task; cnt++) {
+    /* Do nothing. */
+  }
+
+  assert(cnt <= cond->tasks.cnt);
```
eboasson: I should think this assertion will never fail.

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+  /* Locate task index. */
+  for (cnt = 0; cnt < cond->tasks.cnt && cond->tasks.buf[cnt] != task; cnt++) {
+    /* Do nothing. */
+  }
+
+  if (cnt < cond->tasks.cnt) {
+    cond->tasks.cnt--;
+
+    if (cnt < cond->tasks.cnt) {
+      /* Ensure list of tasks is neatly packed. */
```
eboasson: If ordering doesn't matter (it doesn't seem to), it'd make more sense to swap the last one in. Actually yes, it does matter for fairness.

Still, all those ``memmove``s and reallocating every time ... turning it into a circular queue and shrinking it only when there are many available slots instead of just 1 might be a good idea. This stuff gets called a lot ...

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+    (void)ulTaskNotifyTake(1, 0);
+  }
+
+  ddsrt_mutex_unlock(&cond->mutex);
+
+  if (rc == DDS_RETCODE_OK) {
+    /* Wait for notification. */
+    switch (ulTaskNotifyTake(1, ticks)) {
+      case 0:
+        ddsrt_mutex_lock(&cond->mutex);
```
eboasson: Perhaps it would be wise to note here that by the time you call ``cond_pop`` it may already have been removed

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+  }
+  ddsrt_mutex_unlock(&rwlock->mutex);
+  return locked;
+}
+
+void ddsrt_rwlock_write(ddsrt_rwlock_t *rwlock)
+{
+  assert(rwlock != NULL);
+  ddsrt_mutex_lock(&rwlock->mutex);
+  while (rwlock->wtrs > 0) {
```
eboasson: So if I have three threads calling ``rwlock_write`` at the same time, one gets through, the other two will block here with ``wtrs==3``. Once the thread that won calls ``unlock``, the other two will see ``wtrs==2`` and so keep waiting forever. They'd need to check ``state`` as well.

I'm not too worried about performance or fairness right now. Better first get it to provide the exclusion without deadlocks. (And  especially if FreeRTOS+POSIX can be used most of the time.)

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+    stat = ddsrt_atomic_ld32(control);
+    /* Verify once control was initialized properly. */
+    assert(stat == ONCE_NOT_STARTED ||
+           stat == ONCE_IN_PROGRESS ||
+           stat == ONCE_FINISHED);
+
+    if ((stat & ONCE_FINISHED) != 0) {
+      /* The initialization function has been executed. No reason to block
+         execution of this thread. Continue. */
+      brk = 1;
```
eboasson: why not simply return?

src/ddsrt/src/sync/freertos/sync.c
```diff
@@ -0,0 +1,505 @@
...
+         executing it now) at the time of the load. If the atomic compare and
+         swap operation is successful, this thread will run the initialization
+         function. */
+      if (ddsrt_atomic_cas32(
+            control, ONCE_NOT_STARTED, ONCE_IN_PROGRESS) != 0)
+      {
+        /* Function must never block or yield, see reference manual. */
+        init_fn();
+
+        ret = (0 == ddsrt_atomic_cas32(
```
eboasson: cas returns false must be false ... ahum ... there's an easier way to write that!

You could also use a bitmask instead and use atomic_or32_ov instead of calling cas32

src/ddsrt/src/sync/freertos/sync.c
k0ekk0ek: Indeed, the ordering is to ensure every task gets it's turn.

I've made the code so that more than one slot must be free for a resize shrink operation, but choosing a larger number might be better. Also, I like the idea about using a circular queue. I'll update the implementation and the pull request.

k0ekk0ek:
@eboasson. I updated the condition variable and readers-writer lock implementations to use a cyclic buffer and added generic documentation regarding FreeRTOS and instructions on how to get it to compile for the FreeRTOS POSIX Port. There's one thing that still requires some work and that's the MPT test framework by @martinbremmer. Currently I've disabled building it, but I'll reenable it in an upcoming commit in this same pull request once I have a way to check if a given platform supports multi-process in a different module than ddsrt. Apart from that I believe the FreeRTOS port is done. Would you be so kind as to review my work? I would like to suggest taking another look at the condition variable and readers-writer lock implementations first.

Thanks in advance.

k0ekk0ek:
Thanks for taking the time to look at my work @eboasson. There's a couple of minor changes coming up, but I've successfully ran this version of Eclipse Cyclone DDS, built using Xilinx SDK 2018.2, on an UltraZed-EG and I can confirm this runs in the simulator too. After some polishing I think this is ready to be merged, hopefully there are other people that find it useful.

src/ddsrt/src/sync/freertos/tasklist.c
```diff
@@ -0,0 +1,397 @@
...
+  assert(list != NULL);
+  /* Should not be called if room is available. */
+  assert(list->cnt == list->len);
+
+  n = list->len + x;
+  if ((p = ddsrt_realloc(list->tasks, n * sizeof(*p))) == NULL) {
+    return -1;
+  }
+
+  /* Pack operation moved head to end of buffer on wrap around. Move head to
```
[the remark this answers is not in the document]
k0ekk0ek: I updated the comment. It's not necessary to pack the tasklist on a grow operation if it's wrapped as the end of the old buffer is just moved n positions to the back, resulting in n positions becoming free after the last one. It'll be packed automatically if the list is sparse and more space is required.

k0ekk0ek:
AppVeyor build succeeded already in my account: https://ci.appveyor.com/project/k0ekk0ek/cyclonedds/builds/24723669

eboasson:
@k0ekk0ek, thanks for all the hard work! There is one remaining point in my earlier comment: the support for triggering a "sockWaitset". Let me try to explain a little more.

There are two situations in which this triggering is required:
* Termination requires shutting down the receive threads, and so a thread waiting indefinitely for a packet to arrive would cause the stack to never shut down if no such packet arrives.
* If ManySocketsMode is true and a participant is created, or if TCP is being used, sockets are created dynamically. Those sockets get added to the set by calling "Add", followed by an explicit call to "Trigger" in case the way adding sockets to the set doesn't automatically cause the I/O multiplexing logic to wait for data to arrive on the new socket as well.

In both cases, there is no issue if a packet arrives shortly after the "Trigger" call anyway, nor is there one if a spurious wakeups follows shortly after. So I can see three routes of addressing this comment:
1. Implement it "properly" — without a doubt, the preferred approach, but I don't know FreeRTOS/lwIP well and have no idea how much work this is.
2. Document a limitation on the configurations supported under FreeRTOS/lwIP: while the conditions would be utterly unacceptable for a general-purpose platform, for small embedded targets they are in my view actually quite reasonable: (a) accept that a shutdown may block; (b) only use with UDP; (c) only use one participant or ensure ManySocketsMode = false.
3. Make "sockWaitsetWait" block for a limited duration — e.g., if it wakes up every 100ms, you know the new sources will be taken into account within 100ms, or shutdown will proceed after at most 100ms. The price is a lot spurious wakeups when they are not needed.

Obviously 1 is best, but I'd be happy to merge this if 2 or 3 were done (I personally prefer 2 over 3) now. In the latter case, a follow-up pull request implementing option 1 would be delightful!

k0ekk0ek:
Thanks @eboasson. I chose to go with 2 for now until I find the time to properly implement it.

eboasson:
Sensible choice, @k0ekk0ek ! 😁

I do hope you will find the time to figure out how to interrupt a wait and lift the limitation.

As everything built nicely before this latest commit and this latest commit is no more than a doc change, I don't see any value in waiting for all the builds to complete. Might as well merge it right away.

PR #176, opened 2019-05-06
Updated travis to include a coverity build run

ThijsSassen:
The coverity run will only be activated when done on a branch called coverity_scan. In order to avoid hitting the coverity scan quota.

Coverity requires a token 'COVERITY_SCAN_TOKEN' to submit builds. This token can be provided to travis-ci through a secure variable. The project owner needs to signup the project for coverity scan and then enter the secure token into the travis.yml. The secure variable can be generated by calling travis encrypt -r eclipse-cyclonedds/cyclonedds COVERITY_SCAN_TOKEN=xxx

An example of the results on my own branch can be found here:
https://scan.coverity.com/projects/thijssassen-cyclonedds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,17 @@ osx_xcode10_1: &osx_xcode10_1
 windows_vs2017: &windows_vs2017
   os: windows
 
-matrix:
+stages:
+    # Run Coverity only on a branch called coverity_scan
+    - name: Coverity
+      if: branch = coverity_scan
+
+jobs:
   include:
     - <<: *linux_gcc8
-      env: [ BUILD_TYPE=Debug, C_COMPILER=gcc-8, CXX_COMPILER=g++-8, USE_SANITIZER=none ]
+      env: [ BUILD_TYPE=Release, C_COMPILER=gcc-8, CXX_COMPILER=g++-8, USE_SANITIZER=none]
     - <<: *linux_gcc8
-      env: [ BUILD_TYPE=Release, C_COMPILER=gcc-8, CXX_COMPILER=g++-8, USE_SANITIZER=none ]
+      env: [ BUILD_TYPE=Debug, C_COMPILER=gcc-8, CXX_COMPILER=g++-8, USE_SANITIZER=none ]
     - <<: *linux_clang
       env: [ BUILD_TYPE=Debug, C_COMPILER=clang, CXX_COMPILER=clang++, USE_SANITIZER=address ]
     - <<: *linux_clang
@@ -53,6 +58,22 @@ matrix:
       env: [ ARCH=x86_64, BUILD_TYPE=Debug, GENERATOR="Visual Studio 15 2017" ]
     - <<: *windows_vs2017
       env: [ ARCH=x86_64, BUILD_TYPE=Release, GENERATOR="Visual Studio 15 2017" ]
+    - stage: Coverity
+      <<: *linux_gcc8
+      name: Coverity Build
+      env:
+        - BUILD_TYPE=Release
+        - C_COMPILER=gcc-8
+        - CXX_COMPILER=g++-8
+        - USE_SANITIZER=none
+        - COVERITY_SCAN_PROJECT_NAME="$TRAVIS_REPO_SLUG"
+        - COVERITY_SCAN_BRANCH_PATTERN="coverity_scan"
+        - COVERITY_SCAN_NOTIFICATION_EMAIL="$(git log -1 $TRAVIS_COMMIT --pretty=\"%aE\")"
+        - COVERITY_SCAN_BUILD_COMMAND="cmake --build ."
+        - COVERITY_SCAN_BUILD_COMMAND_PREPEND="cov-configure --comptype gcc --compiler gcc-8 --template"
+        # Encrypted COVERITY_SCAN_TOKEN env variable
+        # Generated using `travis encrypt -r eclipse-cyclonedds/cyclonedds COVERITY_SCAN_TOKEN=xxx`
+        - secure: "insert secret here"
 
 # Conan will automatically determine the best compiler for a given platform
 # based on educated guesses. The first check is based on the CC and CXX
@@ -71,7 +92,6 @@ before_install:
       eval "export CC=${C_COMPILER}";
       eval "export CXX=${CXX_COMPILER}";
     fi
-
 # Windows targets in Travis are still very much in beta and Python is not yet
 # available and installation of Python through Chocolaty does not work well.
 # The real fix is to wait until Python and pip are both available on the
@@ -119,6 +139,8 @@ script:
           -DUSE_SANITIZER=${USE_SANITIZER}
           -DBUILD_TESTING=on
           -G "${GENERATOR}" ../src
+  # execute coverity build, to let coverity integrate with the existing script code the following line is used to active the coverity build.
+  - if [[ -n $COVERITY_SCAN_PROJECT_NAME ]] ; then curl -s 'https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh' | bash || true ; fi
   - cmake --build . --config ${BUILD_TYPE} --target install
   - CYCLONEDDS_URI='<CycloneDDS><DDSI2E><Internal><EnableExpensiveChecks>all</EnableExpensiveChecks></Internal></DDSI2E></CycloneDDS>' ctest -T test -C ${BUILD_TYPE}
   - if [ "${USE_SANITIZER}" != "none" ]; then


k0ekk0ek:
@ThijsSassen: I have previously looked into this, but of course forgot most of the details already. Anyway before I look into this any further, I think I would much prefer the Coverity build to just run on master and have the secret be set in the Travis CI environment as opposed to exporting it as a *secret* in the .travis.yml. The reason for the latter being that users who fork it into their own GitHub repos will have to update the .travis.yml and that's an issue when pull requests come in. I'll try to find some time to do an actual proper review.

.travis.yml
```diff
@@ -53,6 +58,22 @@ matrix:
...
+      <<: *linux_gcc8
+      name: Coverity Build
+      env:
+        - BUILD_TYPE=Release
+        - C_COMPILER=gcc-8
+        - CXX_COMPILER=g++-8
+        - USE_SANITIZER=none
+        - COVERITY_SCAN_PROJECT_NAME="$TRAVIS_REPO_SLUG"
+        - COVERITY_SCAN_BRANCH_PATTERN="coverity_scan"
+        - COVERITY_SCAN_NOTIFICATION_EMAIL="$(git log -1 $TRAVIS_COMMIT --pretty=\"%aE\")"
```
k0ekk0ek: This should not be an email address randomly chosen from the git commit log. I'd prefer you use the cyclone-dev mailing list. If that's allowed. @eboasson might have a preference here?

.travis.yml
```diff
@@ -53,6 +58,22 @@ matrix:
...
+    - stage: Coverity
+      <<: *linux_gcc8
+      name: Coverity Build
+      env:
+        - BUILD_TYPE=Release
+        - C_COMPILER=gcc-8
+        - CXX_COMPILER=g++-8
+        - USE_SANITIZER=none
+        - COVERITY_SCAN_PROJECT_NAME="$TRAVIS_REPO_SLUG"
+        - COVERITY_SCAN_BRANCH_PATTERN="coverity_scan"
```
k0ekk0ek: I prefer this would just run on master. We don't have a coverity_scan branch and I doubt we're willing to go through the trouble of creating one, when it's just as easy to just run it on master.

.travis.yml
```diff
@@ -53,6 +58,22 @@ matrix:
...
+        - COVERITY_SCAN_NOTIFICATION_EMAIL="$(git log -1 $TRAVIS_COMMIT --pretty=\"%aE\")"
+        - COVERITY_SCAN_BUILD_COMMAND="cmake --build ."
+        - COVERITY_SCAN_BUILD_COMMAND_PREPEND="cov-configure --comptype gcc --compiler gcc-8 --template"
+        # Encrypted COVERITY_SCAN_TOKEN env variable
+        # Generated using `travis encrypt -r eclipse-cyclonedds/cyclonedds COVERITY_SCAN_TOKEN=xxx`
+        - secure: "insert secret here"
```
k0ekk0ek: I don't like the secret being in the `.travis.yml` for the aforementioned reason.

.travis.yml
```diff
@@ -53,6 +58,22 @@ matrix:
       env: [ ARCH=x86_64, BUILD_TYPE=Debug, GENERATOR="Visual Studio 15 2017" ]
     - <<: *windows_vs2017
       env: [ ARCH=x86_64, BUILD_TYPE=Release, GENERATOR="Visual Studio 15 2017" ]
+    - stage: Coverity
+      <<: *linux_gcc8
+      name: Coverity Build
+      env:
+        - BUILD_TYPE=Release
```
k0ekk0ek: In my humble option, this should probably be a debug build(?) Not sure if it really changes anything since Coverity simply intercepts the calls to the compiler.

.travis.yml
```diff
@@ -53,6 +58,22 @@ matrix:
...
+        - C_COMPILER=gcc-8
+        - CXX_COMPILER=g++-8
+        - USE_SANITIZER=none
+        - COVERITY_SCAN_PROJECT_NAME="$TRAVIS_REPO_SLUG"
+        - COVERITY_SCAN_BRANCH_PATTERN="coverity_scan"
+        - COVERITY_SCAN_NOTIFICATION_EMAIL="$(git log -1 $TRAVIS_COMMIT --pretty=\"%aE\")"
+        - COVERITY_SCAN_BUILD_COMMAND="cmake --build ."
+        - COVERITY_SCAN_BUILD_COMMAND_PREPEND="cov-configure --comptype gcc --compiler gcc-8 --template"
```
k0ekk0ek: I don't like this being fixed. In my version I made CMake automatically configure Coverity based on the compiler it finds (it works just as well with MSVC and Clang), but at the very least I expect you to use the `C_COMPILER` environment variable.

.travis.yml
```diff
@@ -33,12 +33,17 @@ osx_xcode10_1: &osx_xcode10_1
 windows_vs2017: &windows_vs2017
   os: windows
 
-matrix:
+stages:
```
k0ekk0ek: Not sure if build stages are really the way to go. As it's not multiple stages to achieve a result. Can't we simply add a job conditionally? Also, by using a *stage* the rest of the jobs automatically go under the Test stage. If adding a job conditionally isn't possible I'd at least expect that the name of the stage is determined by us.

.travis.yml
```diff
@@ -33,12 +33,17 @@ osx_xcode10_1: &osx_xcode10_1
...
+    # Run Coverity only on a branch called coverity_scan
+    - name: Coverity
+      if: branch = coverity_scan
```
k0ekk0ek: How would this work if we just want to build *master*?

.travis.yml
```diff
@@ -119,6 +139,8 @@ script:
           -DUSE_SANITIZER=${USE_SANITIZER}
           -DBUILD_TESTING=on
           -G "${GENERATOR}" ../src
+  # execute coverity build, to let coverity integrate with the existing script code the following line is used to active the coverity build.
+  - if [[ -n $COVERITY_SCAN_PROJECT_NAME ]] ; then curl -s 'https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh' | bash || true ; fi
```
k0ekk0ek: I really, really, REALLY, don't like it when people execute a script that they download over the internet without any verification of the contents. Apart from that, I also don't like to depend on a bash script. I believe integrating this into CMake would be better(?) It's a bit trickier though... We could also go for a CMake script that does more-or-less the same so people with Windows boxes can use the same logic. The `<build>/CMakeFiles/<cmake version>` subdirectory contains files named `CMakeCCompiler.cmake` and `CMakeCXXCompiler.cmake` we could use those to automatically configure Coverity based on the compiler path and compiler id, use most of the information already set by CMake to figure out the project name etc.

eboasson:
What @k0ekk0ek says.

That includes his remark on the email address, except that:
* the mailing list is cyclonedds-dev — whether this is allowed is something we will have to find out;
* I want it configured via a variable — if someone clones the repo and sets up coverity scan for his/her own work, the notifications for scans of those clones should not accidentally end up here.

About the branch: considering that the frequency of commits on master is currently pretty low, there is no point in creating a separate branch now. We can create that separate branch when the scan limit starts to hurt.

And absolutely no secrets in the repository.

.travis.yml
ThijsSassen: The dev list as email seems logical and will change that

.travis.yml
ThijsSassen: I will remove the secret and the token can be added in the travis-ci internal variable section. Doing this will break the windows builds see https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264/6 so for now I have to postpone this  PR as it will break windows builds.

.travis.yml
ThijsSassen: @k0ekk0ek can you share you version with me so this can be done properly?

.travis.yml
k0ekk0ek: Yeah, I thought about this... can't we export the encrypted version in an environment variable and then call decrypt? Turn it around?

.travis.yml
k0ekk0ek: Probably not the best idea to use the dev list either. Maybe just ask @eboasson where he would like the email to end up and grab the address from an environment variable or something...

.travis.yml
k0ekk0ek: You mean finish my version, issue a pull request on your branch, so that you can issue a pull request with the official repo? :wink:

ThijsSassen:
closing this because a better implementation is available

PR #177, opened 2019-05-07
Security Plugin API

kurtuluso:
We plan to add OMG DDS Security compliance to Eclipse Cyclone. The specification is documented here: https://www.omg.org/spec/DDS-SECURITY/

DDS Security is composed of the following components/modifications:
1. DDS Security plugin API
2. DDS Security plugins that implement the API
3. DDS Security Core Library that is used by the plugins and DDSI.
4. Changes in the DDSI that moderate the specified security model. 

The dependency diagram is below:
```

DDSI  ---->     DDS Security API (headers only)  <----- DDS Security Plugins

  |                    ^                                    |
  |                    |                                    |
  |                    |                                    |
   ------->     DDS Security Core  <-----------------------
  |                    |                                    |
  |                    |                                    |
  |                    |                                    |
  |                    v                                    |
   ------->          DDS_RT        <-----------------------
   
```
   
All security specific contents are planned to be under src/security.
As the first step, DDS Security API is added. 

The DDS Security plugin API consists of just a few header files. There are separate header files for each plugin: dds_security_api_authentication.h dds_security_api_cryptography.h and dds_security_api_access_control.c

The API functions and types are prepared from the IDL by adding DDS_Security_ namespace prefix to functions and data types. Instead of extending DDS builtin topic data types, separate DDS_Security_ data type is defined for the current type and the new secure data type.


eboasson:
Thanks for this first step in the contribution of an implementation of DDS Security for Cyclone! There are a few minor — or perhaps not so minor 😁 — points to be addressed before this contribution can be accepted, though.

To start with, there is the paperwork business. The eclipsefdn/eca check previously incorrectly indicated that all is well, but that has now been fixed: you need to have a valid Eclipse Committer Agreement on file and the commit message must end on a “Signed-off-by:” followed by the author (``git commit -s`` is your friend).

Then, the commit message itself: perhaps “CYC-203” is sufficiently clear to you, but I’m pretty sure it is not for most people. The aim with commit messages in Cyclone is to give a short description of the subject matter in the first line, followed by a more detailed description of what the changes entail and why. All within reason, of course. (This is pretty standard, I’d say.)

The reset of the commit message “Security core library removed, review logs were applied” means absolutely nothing to those not familiar with the code or the review logs. Clearly this one builds on the preceding one, but “DDS Security API has been added” doesn’t give any real information either.

I appreciate that you added the second commit rather than just amending the first, because such amending can somewhat mess up an ongoing review. But in general, two such commits ought to be merged into one to keep a clean history. It is generally worth it to spend the effort and clean up the commits by splitting/combining/reordering them as necessary to get to a state where commit logically lead to the end result — naturally within reason: I have no issue with the occasional monster commit.

In this case, I would say it would make sense to split it at least into the addition of ``ddsrt_strchrs`` and the rest. The other commits would have to be squashed — I don’t want to merge a PR with commit 1 adding a file and commit 2 deleting or moving it, if the two can still reasonably be squashed.

**``ddsrt_strchrs``**

Then there is the matter of ``ddsrt_strchrs`` itself: to me it looks suspiciously like a combination of the C standard library functions ``strspn``/``strcspn`` along with a differently encoded result. I notice the code that used it has been removed, but @martinbremmer’s comment suggests that it is likely to reappear again. Therefore, if you can convince me the addition is worth it, I’ll be inclined to accept it — but I must also say that, as it stands, I’m not convinced as the only use of it in the (now removed code) was for skipping a minor bit of whitespace:
```C
ptr = ddsrt_strchrs(str, " \t", false);
```
after which ``ptr`` was moreover assumed to be non-null. In this case, ``ptr = str + strspn(str, “ \t”)`` would do, too, though I’d think the idiomatic C here would be ``while(*str && isspace((unsigned char) *str) str++;`` — but I digress.

Given the need for skipping whitespace, presumably the input ``str`` is not sanitised yet and so presumably an all-whitespace string would be a possibility. In that case, you’d better also watch out for a null pointer. But I digress again … (I won’t comment further on things in the code that was added in the first commit and removed again in the second.)

The doc comment for ``ddsrt_strchrs`` besides makes references to ``OS_TRUE`` and ``OS_FALSE`` which do not exist in Cyclone. The implementation is far more complicated than necessary … I’m cool with accepting that ``strspn`` might well be viewed as exotic, but surely ``strchr`` is not. In between a hand-rolled ``strchr`` and various state variables to emulate a perfectly straightforward early return in a function free of side-effects … I find a rewrite would not be amiss.

(There is also the subtle issue that anything starting with “str” is reserved by the C standard … but that I am willing to overlook. In between all the identifiers reserved by C and POSIX, there’s precious little room for choosing sane names …)

**Security API proper**

What exactly is this security API here? It looks to me to be the plugin interface specification — i.e., a translation of the [IDL](https://www.omg.org/spec/DDS-SECURITY/20170901/dds_security_plugins_spis.idl) provided by the OMG. Is that correct?

The header files are in ``include/security`` instead of ``include/dds/security`` — so unlike all other header files of Cyclone. That won’t do.

Regarding the contents of those header files:

* Why all the
```C
struct dds_security_access_control_;
typedef struct dds_security_access_control_ dds_security_access_control;
```
nonsense with the underscores? In C you definitely don’t have to, and as far as I am aware, even though C++ implicitly “typedefs” the tag name, it is legal in C++ to do:
```C
struct dds_security_access_control;
typedef struct dds_security_access_control dds_security_access_control;
```
because the typedef doesn’t name a different type.

* I prefer function typedefs to have names that hint that it is a typedef, e.g., with suffix like “_t” (I know POSIX reserves names ending in “_t” but that’s a kind of infringement I tend to accept because I am of the opinion that the C and POSIX standard committees went too far in reserving parts of the name space).

* I am intrigued by the way the name spaces are mapped to the identifiers. The OMG IDL seems to put everything in DDS::Security and the security-specific types seem to be named in accordance with the IDL-to-C mapping. It doesn’t fit with Cyclone’s native API, but for a standardised plug-in interface that application code never has to deal with, that’s at least a defensible choice. But then why are the function typenames following a different rule in their name space prefix?

* I appreciate the effort that went into making this interface independent of the types used in the core library, but really … redefining all QoS policies *again*? Yes, there is a debate to be had on how to deal with the various different QoS policy names because of the decision in Cyclone to not simply follow the IDL-to-C mapping. By defining them here as-if in the ``DDS::Security`` namespace instead of simply in the ``DDS`` namespace, the argument to maintain compatibility with plug-ins written against OMG IDL interface spec is weakened, and that affects the point made above.

* The above holds even more so for the contents of ``dds_security_api_types.h``: Cyclone deliberate avoids “short/int/long/long long” business in its interfaces, favouring the C99 ``intN_t`` instead. So why would one ever want to typedef ``DDS_Security_long`` &c. types here _and_ map them to other things than the sensible ``stdint.h`` types? (And this being C, you can just do all your definitions and write code using ``intN_t`` and if someone else wants to use those impractical names, let them: typedefs are just aliases anyway.)

* ``DDS_Security_BuiltinTopicKey_t`` and ``DDS_Security_GuidPrefix_t`` are even more interesting cases: the first is not even compatible with the rest of Cyclone (I do see that it matches OpenSplice), and GUIDs/GUID prefixes are native to Cyclone and so there are types available for that anyway. Besides, array typedefs behave kinda weird in C because of the pointer decay rules when passed as an argument — and therefore ``sizeof(x)`` will return different sizes in different contexts.
Embedding it in a struct is a better solution (which, e.g., Cyclone’s native ``nn_guid_prefix_t`` does in spirit— it is a bit different because it treats GUIDs are arrays of 32-bit unsigned integers internally).

* There are many more things to remark on the contents of that types.h file — but I think you will have no problem imagining what those are, give the above.

**Relationship between security & the core library**

I gather that the idea is that there will be a stand-alone security library with its own type definitions, and that the Cyclone core library links against that. I really appreciate the effort and intention of introducing such a significant and potentially intrusive piece of code in such a low-impact way.

However, it does seem to require a lot of duplication of definitions, which in turn will require a lot of code to convert from one to the other. In terms of code organisation it would make more sense to rely a little more on the type definitions Cyclone already has.

Similarly, it would make sense to integrate the supporting code in the Cyclone library (I like keeping the idea of keeping the implementation (mostly) separate, but linking it together means one only ever needs to worry about a single shared library).

As it appears the security plug-in interface implementation will rely on functions in ``ddsrt`` and so from the perspective of linking the code, were one to put this security-related code in a separate library, there’d be a mutual dependency between the Cyclone core library and this security library. Solving that by duplicating ``ddsrt`` is a bad idea. Splitting current Cyclone into two libraries (``ddsrt`` and ``ddsc``) would solve it, but I personally prefer having one larger library. Combining multiple components into a single run-time library is just a mechanical process, after all.

**Next steps**

I am really excited that an implementation of DDS Security is now really coming to Cyclone and I am quite willing to overlook some things that I would ordinarily block, if I can see that it is by far the best route to getting it all integrated and working. Especially considering that for the time being the security implementation will live on a separate branch and some of these things can get sorted out before merging it into master.

In making such decisions, it would really help to have more background information in the PR description on why this set of commits makes sense as a separate PR, and on how it relates to what is to come. The addition and then immediate removal of some supporting code suggests that I may not be alone in being somewhat uncertain about what to expect.

There is also the possibility of using the cyclonedds-dev mailing list, if more extended discussions are required.

martinbremmer:
@eboasson, I skimmed through your comment and agree with most of it. Not sure how to fix some of it neatly though.

This is just a small clarification, a question and a thought.

Regarding
```
struct dds_security_access_control_;
typedef struct dds_security_access_control_ dds_security_access_control;
```
I remember having compile issues with your suggestion on some compiler or another. I'm just not sure which one anymore.

Regarding the security & core relationship.
The ddsc will depend on the security API to be able to call plugin functionality. When you let the security API depend on ddsc types, doesn't that create cross-dependencies with all the issues that entails?

Also Regarding the security & core relationship.
I haven't really thought this through yet, but this is how I imagined it to work (regarding linking and such). @k0ekk0ek, please correct me if I have some invalid assumptions regarding how cmake does its linking.
- The security API is just a bunch of header files (which will be bundled in a cmake target).
- The security core will be an cmake interface library. As far as I understand, a interface library will be linked into the result application/library that depend on it and will not be a stand alone library (mainly only object files).
- The ddsc will depend both on the security API and security core (linked into it).
- The security plugins ('standalone' libraries) will also depend on both the security API and security core (linked into them).
- Possible 3rd party plugins will only depend on the security API.

kurtuluso:
Thank you for your helpful long comment :)

- I've fixed the commit messages and merged the commits.

- Signed the Eclipse Committer Agreement.

- Removed the ddsrt_strchrs function because I'd already removed its user. But I noted your comments and I'll apply them when I commit for the core library.

Martin answered some of the issues about the API.

- You know the naming is strictly dependent on the spec, however as you've said, I think we can change the function typenames. We have a constraint that a security plugin should be used in both OpenSplice and Cyclone. We can slightly change the API but we need to ensure binary backward compatibility. We can discuss this together.

- I've used the intN definitions

- GuidPrefix & BuiltinTopicKey

GuidPrefix is mentioned in the DDS Security spec 7.3.3:

7.3.3 Constraints of the DomainParticipant GUID_t (GUID)
The DDS-RTPS specification [2] states that DDS DomainParticipant entities are identified by a
unique 16-byte GUID with type GUID_t. In this DDS-Security specification the type GUID_t refers
to the same type defined in clauses 8.4.2.1 and 9.3.1 of the DDS-RTPS specification [2]:
// From DDS-RTPS [2] clauses 8.4.2.1 and 9.3.1
```
typedef octet GuidPrefix_t[12];
struct EntityId_t {
octet entityKey[3];
octet entityKind;
};
struct GUID_t {
GuidPrefix_t prefix;
EntityId_t entityId;
};
```

BuiltinTopicKey was directly used in DDS Security 1.0, but there is no direct usage in 1.1 IDL. However, it is in current topic types which are extended for secure topic types. 

I think we can not change GuidPrefix_t, but we can change BuiltinTopicKey. However I propose not to change both as nothing is using/dependent on them now, we can decide on dependencies occur.

eboasson:
Thanks for making those changes, @kurtuluso. I’m going to give a combined response for the updated version and the comments from @martinbremmer, because they really all go together (and you have both been waiting long enough for a response to not delay it further by trying to keep things separate …)

@martinbremmer:

> Regarding:
> ```C++
> struct dds_security_access_control_;
> typedef struct dds_security_access_control_ dds_security_access_control;
> ```
>
> I remember having compile issues with your suggestion on some compiler or another. I'm just not sure which one anymore.

I just tried Clang and gcc 9.1 both in std=c++11 and pedantic mode and they seem perfectly fine with it:
```C++
struct y;
typedef struct y y;
struct y { int b; };
struct x { y a; };
typedef struct x x;
int main ()
{
  x z = { { 0 } };
  return z.a.b;
}
```

I have been able to find some remarks that suggest C++ doesn’t allow typedef’ing an incomplete type — but if that’s the case (and gcc and clang get it wrong with their -Wpedantic) and it turns out to be a problem, you can still use the tagged (``struct dds_security_access_control``) form and typedef it later.

> Regarding the security & core relationship.
> The ddsc will depend on the security API to be able to call plugin functionality. When you let the security API depend on ddsc types, doesn't that create cross-dependencies with all the issues that entails?

That depends. Truly internal implementation details: certainly. Committed external API: definitely not in a way that I consider problematic. Some don’t classify neatly, for example: a GUID in DDSI is a given — but because it is only a wire representation, you can define it in a number of ways internally. Abstracting them adds overhead of some kind, and some things are so stable that it seems unimaginable they will ever change. Usually, those do change … if only rarely.

The particularly important one here is the definition of the additional builtin topics and the QoS. The QoS are spec’d, with the security spec referencing the DCPS specification, so the types are the same *by definition*. That Cyclone’s C mapping chose to depart from the specified representation has some decent reasons, but it is trivial to write conversion to-and-from the standardised mapping.

A split where one representation is used for some built-in topics and another is used for other built-in topics really is not a good idea. So I am of the opinion that this security API should be adapted to match the Cyclone model. This actually makes life a lot easier: you just get a “dds_qos_t” object that you can query, and the various built-in topics unify.

For whose benefit exactly are those secure built-in topics? I presume it is only for the plug-in and the wire discovery, not for communicating with the applications. But who does the conversion to/from wire format? Just wondering — I haven’t seen any of that code yet!

To @kurtuluso’s point: “We have a constraint that a security plugin should be used in both OpenSplice and Cyclone” — that’s a tricky one. I can see the challenge of retaining [source] compatibility for the plugins and how that challenge may have come about … (I’m assuming "binary" compatibility was a typo.) Still, from the perspective of Cyclone, it is a rather quaint constraint, because this means Cyclone ends up with something that doesn’t fit nicely and yet that isn’t standards compliant either.

I’d say that a shim to maintain that particular backwards compatibility could be the answer, that way there’d be none of this compatibility stuff in the core APIs of Cyclone (I consider security to be a pretty “core” thing).

Regarding ``DDS_Security_GuidPrefix_t``: there is (I think I mentioned it) a real difference between the definition of a GUID in the DDSI spec and that of a GUID internally in Cyclone. On the wire and at the API level it conforms to the definition in the DDSI spec, the only difference is that internally it is treated not as octets but as ``uint32_t``s, along with the byteswapping that requires. I can see that the difference might affect the inner workings of a plugin — though nothing a shim couldn’t fix to maintain that compatibility. So that’s not a big deal.

Ordinarily, I wouldn’t allow just a bunch of includes that duplicate stuff and that have some other aspects I dislike into the repository. But DDS Security is a big chunk of work and can I understand the practical obstacles to rewriting everything — let alone to rewriting them first before adding anything that actually adds value to Cyclone.

It would be easier to have a well-founded judgement if these header files were actually used for something. That is: if the changes to the DDSI stack were there and that Security Core library were present, it would already be a lot more reasonable to hope for such a judgement. Having the default plug-in implementation would shine light on the other half of the interface.

The most logical model would be for the work to occur on a separate fork until something real can be added, but I dislike the invisibility of that. Also the smaller chunks are much more manageable for reviewing and discussion. Hence my willingness to create a separate branch … So except for the minor points discussed above — for which I would like a little better understanding why it really should be this way — I am hesitatingly willing to merge this PR into that separate branch, but it doesn’t imply further changes won’t be required later on for it to be merged to master, once the DDS Security implementation is complete. See where it goes, while having it safely stored in the main repository.

k0ekk0ek:
I've been following this pull request a bit and while I'm not an expert on the security api itself, I do know a thing or two about how the project is structured etc. I've noticed some oddities and some things that are not quite in line with the rest.

1. The security_api builds upon ddsrt directly. That really shouldn't be the case. ddsrt should not (currently) be considered a target in it's own right. It's an interface library because object libraries only work well since CMake version 3.13, and it's only a separate target because I want to be able to build tests that only verify the workings of the runtime layer. As such, I think the security core should be compiled as part of dds (libddsc.so) and not be kept separate. Now I understand that this could prove problematic if plugins want to use functionality that resides within ddsrt, because you'd have cyclic dependencies, but until I've seen what the plugins use, it's hard to say. If they really use functionality in ddsrt and cyclic dependencies are an issue, then ddsrt should become a real library on which the ddsc and security_api targets build.

2. The name security_api is not in line with the rest of Eclipse Cyclone DDS. If you noticed, most targets are named dds and then something. There's ddsc for the C api, there's ddsi for well... DDSI, there's ddsrt for the runtime layer, there'll be a ddsts for typesupport and a ddscxx for the C++ api. I would prefer the security api to be named something like ddssec. Especially if it'll become a real library (which is what I think the plugins require), in which case it would result in something like libsecurity_api.so or something. I'd prefer if it resulted in libddssec.so.

3. I've spent a great deal of time stripping out all the various different error codes and make it so that we simply use DDS_RETCODE_ constants everywhere. This pull request reintroduces separate error codes and that's something I really don't approve of. The security error codes start at an offset of 100 and should nicely integrate with the other codes in dds/ddsrt/retcode.h. The messages should be retrievable using dds_strretcode if you ask me. @eboasson, any thoughts? (Another example of why dds and ddsrt shouldn't be separated).

4. The security_api target export CMAKE_CURRENT_BINARY_DIR/include as an include location, but it doesn't exist. Please remove that.

5. If the security_api just becomes part of ddsc, and it should in my opinion, then I'd prefer you propagate the naming scheme as introduced in ddsrt etc and name the header files e.g. `dds/ddssec/auth.h` or something instead of `dds/security/dds_security_api_authentication.h`.

6. Most of the code does not adhere to the *standards* that we currently have in Eclipse Cyclone DDS. I know @eboasson mentioned this before too, so I won't go into this any further.

7. All the duplicate type definitions. Is it really necessary to typedef a float, or typedef a short while we've just worked our way towards fixed type integers? There are platforms where these things are actually different types. If customers want to print variables there, they must now use PRId16 for the short, which is just a little weird. Most of all I dislike the introduction of another type to keep durations. Even the runtime layer uses the dds_time_t and dds_duration_t types that are plain int64_t integers and this pull request reintroduces a duration type with second and nanosecond members. Isn't it just a lot easier to build on what's already there? I'll let @eboasson worry about this, but it just doesn't feel like a nice integrated package in it's current form.

8. Please update `docs/dev/modules.md`. I've taken great care in making sure that everyone can read up on how the various modules are structured and would appreciate the document is kept up to date. Apart from that it'll also show that the current design introduces all kinds of weird inter-module dependencies. I bet that at some point it'll just become clear that the security api doesn't need to be a separate library, but we'll see.

I'm sure I can find more, but @eboasson already OK'd this. Just keep in mind my remarks above and we'll see what's actually needed once the plugins start making it into the security branch.

eboasson:
Thank you @k0ekk0ek for pitching in :smile:

> 1. The security_api builds upon ddsrt directly. That really shouldn't be the case. ddsrt should not (currently) be considered a target in it's own right. It's an interface library because object libraries only work well since CMake version 3.13, and it's only a separate target because I want to be able to build tests that only verify the workings of the runtime layer. As such, I think the security core should be compiled as part of dds (libddsc.so) and not be kept separate. Now I understand that this could prove problematic if plugins want to use functionality that resides within ddsrt, because you'd have cyclic dependencies, but until I've seen what the plugins use, it's hard to say. If they really use functionality in ddsrt and cyclic dependencies are an issue, then ddsrt should become a real library on which the ddsc and security_api targets build.

Good point ... my CMake-fu being what it is, I did what was to be expected of me and completely overlooked this. This indeed needs addressing.

> 2. The name security_api is not in line with the rest of Eclipse Cyclone DDS. If you noticed, most targets are named dds and then something. There's ddsc for the C api, there's ddsi for well... DDSI, there's ddsrt for the runtime layer, there'll be a ddsts for typesupport and a ddscxx for the C++ api. I would prefer the security api to be named something like ddssec. Especially if it'll become a real library (which is what I think the plugins require), in which case it would result in something like libsecurity_api.so or something. I'd prefer if it resulted in libddssec.so.

I actually expect the security support functions to end up in the "regular" Cyclone library, but I'm reserving judgement a bit because there is still so much about the security implementation that is unknown to me.

Secondly, as I wrote in my comment yesterday, in my view there is a balancing act here. An implementation of DDS Security in Cyclone is extremely valuable and a lot of work. To what extent then is it wise to require a lot of rework before accepting anything, if that rework can also be postponed?

Creating legacy and creating rework in the future is not a good thing, but in this case it also has to be weighed against the expected effect on the time frame in which DDS Security support becomes available in Cyclone. So I'm willing to go with an in-between solution: accept some deviations now, but only in a separate branch. I think that likely is a quicker route to having something "available" that works.

Acceptance into master and becoming part of a released version is another matter: then those issues need to be addressed.

@kurtuluso: of course there is also the option of doing the rework today, I'm not requiring that it be postponed! I'm just willing to bend the rules a bit temporarily.

> 3. I've spent a great deal of time stripping out all the various different error codes and make it so that we simply use DDS_RETCODE_ constants everywhere. This pull request reintroduces separate error codes and that's something I really don't approve of. The security error codes start at an offset of 100 and should nicely integrate with the other codes in dds/ddsrt/retcode.h. The messages should be retrievable using dds_strretcode if you ask me. @eboasson, any thoughts? (Another example of why dds and ddsrt shouldn't be separated).

To me it depends a little bit on where those error codes show up. As a matter of principle I agree with you, but or entirely separate side-interface, it may be more appropriate to use different error codes than adding very specific single-use ones to ddsrt. With the limited visibility into how these exactly get used, I think the argument I gave above is applicable.

> 4. The security_api target export CMAKE_CURRENT_BINARY_DIR/include as an include location, but it doesn't exist. Please remove that.

@kurtuluso : can you please sort that out?

> 5. If the security_api just becomes part of ddsc, and it should in my opinion, then I'd prefer you propagate the naming scheme as introduced in ddsrt etc and name the header files e.g. dds/ddssec/auth.h or something instead of dds/security/dds_security_api_authentication.h.

I agree in principle, but see above. @kurtuluso : I think that the security work needs to maintain a list in the branch where any identified required rework (such as this) is tracked.

> 6. Most of the code does not adhere to the standards that we currently have in Eclipse Cyclone DDS. I know @eboasson mentioned this before too, so I won't go into this any further.
>
> 7. All the duplicate type definitions. Is it really necessary to typedef a float, or typedef a short while we've just worked our way towards fixed type integers? There are platforms where these things are actually different types. If customers want to print variables there, they must now use PRId16 for the short, which is just a little weird. Most of all I dislike the introduction of another type to keep durations. Even the runtime layer uses the dds_time_t and dds_duration_t types that are plain int64_t integers and this pull request reintroduces a duration type with second and nanosecond members. Isn't it just a lot easier to build on what's already there? I'll let @eboasson worry about this, but it just doesn't feel like a nice integrated package in it's current form.

I believe that tension is mostly caused by the DDS spec being what it is ... Cyclone has its own native API, I think the security plug-in interface should match that (just like the current one matches the standard DCPS C API in its use of the IDL-to-C mapping) and then have a shim to allow the standard interface to be used.

I'm willing to move forward on this separate branch without having all that in place, in the interest of getting a working implementation for that.

> 8. Please update docs/dev/modules.md. I've taken great care in making sure that everyone can read up on how the various modules are structured and would appreciate the document is kept up to date. Apart from that it'll also show that the current design introduces all kinds of weird inter-module dependencies. I bet that at some point it'll just become clear that the security api doesn't need to be a separate library, but we'll see.

Agreed. @kurtuluso I should have noticed that and raised it myself, but will you please do this?

> I'm sure I can find more, but @eboasson already OK'd this. Just keep in mind my remarks above and we'll see what's actually needed once the plugins start making it into the security branch.

I hope my responses above clarify the meaning of that "OK" ...

kurtuluso:
> 1. "security_api builds upon ddsrt directly",
There is no dependency from security_api to ddsrt. security API consists of only header files and does not include any ddsrt header file. 
As shown in the diagram above, ddsi and plugins are dependent on security core. So, I can not say directly that security_core will be in ddsc.

> 2. The name security_api is not in line with the rest of Eclipse Cyclone DDS. 
The current notation only includes the first letters of the words. In this notation, it should be ddss which indicates nothing meaningful just like ddssec. I understand ddsc, ddscxx and ddsi which are either complete abbreviation (ddsi) or meaningful word after 'dds'.
security_api will have no library target however we currently use typically a notation like libdds_security_auth.so (for authentication plugin). 
I think it is not a good idea to use cryptographic names like libddssa.so  (for security authentication plugin) and libddssac.so (for security access control plugin)
Anyway, we need a naming guideline for giving a name to components/files until we merge to master.

> 3. I've spent a great deal of time stripping out all the various different error codes and make it so that we simply use DDS_RETCODE_ constants everywhere. 
There is a section in DDS Security Spec: 7.2.9 Additional DDS Return Code: NOT_ALLOWED_BY_SECURITY
It says there is only one return code should be added to DDS return codes. So, they are not "return codes" but security specific error codes. However, we can discuss this when merging to master: https://github.com/kurtuluso/cyclonedds/projects/1#card-21850288

4: DONE

5,6. We are repeating the dds word in folder names but not in file names. Actually "dds security" is appropriate as a component name however, we can discuss making abbreviation later. 
We need a naming guideline; a democratic one.
 https://github.com/kurtuluso/cyclonedds/projects/1#card-21850593

7. Yes, it's different but it should not be unexpected; it is in dds spec (DCPS IDL): 
```
struct Duration_t {
   long sec;
   unsigned long nanosec;
 };

```
@eboasson you're right; shim would be a good idea if we can lock the plugin library API as spec and use cyclone-way types on the cyclone side. We can discuss how to do it for dynamic loading.

eboasson:
> @k0ekk0ek 
> > The name security_api is not in line with the rest of Eclipse Cyclone DDS.

@kurtuluso :
> The current notation only includes the first letters of the words. In this notation, it should be ddss which indicates nothing meaningful just like ddssec. I understand ddsc, ddscxx and ddsi which are either complete abbreviation (ddsi) or meaningful word after 'dds'.
> security_api will have no library target however we currently use typically a notation like libdds_security_auth.so (for authentication plugin).
> I think it is not a good idea to use cryptographic names like libddssa.so (for security authentication plugin) and libddssac.so (for security access control plugin)
> Anyway, we need a naming guideline for giving a name to components/files until we merge to master.

Since you want guidance: "ddssec" to me conveys sufficient information, which "ddss" does not. A full "security_api" is a too long to suit my tastes (which is why it is so different from other component names).

> We need a naming guideline; a democratic one.

I'm all for a "meritocratic" one. That's perfectly compatible with a "democratic" one, because the latter does not necessarily imply universal suffrage with equal voting power for all. Indeed, a formalistic view of things would allow even a Louis XIV-style definition of voting power — "L'État, c'est moi" — to be compatible with a democracy, but such a definition is anathema to me.

A certain consistency in the naming of components matters to me more than my own preference for short names, but it'll take a better argument than those raised so far in the discussion on whether "security_api" is or is not a good name for me to change my view.

I'm adding this remark here because the discussion today exists here. There is no implication that the naming needs to be changed for the pull request to be accepted. The discussion can be continued elsewhere, now or at some other time.

In fact: all points have been adequately addressed as far as I am concerned, except for one:

> I think that the security work needs to maintain a list in the branch where any identified required rework (such as this) is tracked.

I do see a few "cards", but I specifically requested that they be recorded in the branch to ensure they are recorded in the Eclipse-owned copy of the repository. Now, if GitHub offers a nicer way that also keeps them independent of whatever is on master, that's good, too. Anything goes, as long as there is: (1) an easy-to-view and easy-to-edit list in this repository/GitHub account, that (2) does not sit among the "normal" issues.

kurtuluso:
Commits have been merged
TODO list has been created: docs/dev/todo_list.md

PR #180, opened 2019-05-09
IDL Compiler: Restructuring TypeTree

FransFaase:
In this commit the TypeTree has been restructured such that it is easier to use. For the representation of arrays, the TypeTree no longer follows the syntax, but takes a more semantic approach, where an array is represented in a similar way as a sequence. (See typesupport.md for more details.)

The files have been structured in accordance with the changes in the abstraction layer. The type walker has been removed, because it seemed not to give much benefits with respect to the restructured TypeTree. The tests of the IDL parser no longer make use of the stringify function, but rather test the resulting TypeTrees directly, which are also more thoroughly. Test for potential memory leaks when parsing is terminated midway due to syntax errors, have been implemented.

The useability of the TypeTree has been tested in the various tests and an implementation of a C99 back-end, the latter of which is not included in this pull-request.


src/tools/idlc/src/tt_create.c
```diff
@@ -0,0 +1,517 @@
...
+struct dds_context {
+  ddsts_type_t *root_type;
+  ddsts_type_t *cur_type;
+  ddsts_type_t *type_for_declarator;
+  array_size_t *array_sizes;
+  dds_retcode_t retcode;
+  ddsts_identifier_t dangling_identifier;
+};
+
+extern bool dds_new_base_type(dds_context_t *context, ddsts_flags_t flags, ddsts_type_t **result)
```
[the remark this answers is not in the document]
FransFaase: I did consider using dds_retcode_t. However, the generated parser, from which these functions are called, cannot return a dds_retcode_t. It returns one of the integer values: 0 (succes), 1 (parser error), or 2 (out-of-memory). So, the only way to get the return code outside the parser is by storing it in the context. The functions are only intended to be called from idl.y. I thought it would result in cleaner code to store the return code in the context inside the function, than having to do this in the idl.y file.

docs/dev/typesupport.md
```diff
@@ -73,6 +72,7 @@ The diagram above provides a very minimal overview of the different parts
 involved with translating language agnostic type definitions into language
 native type definitions. The diagram clearly shows the importance of the
 intermediate format, hereafter to be referred to as the *TypeTree*.
+XML stands for the XML representation of the (binary) TypeObject representation.
```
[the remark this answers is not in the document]
FransFaase: I have clarified this and added a link to the DDS-XTYPES specification.

src/ddsts/CMakeLists.txt
```diff
@@ -9,61 +9,37 @@
...
-target_include_directories(ddsts PUBLIC "include")
-target_include_directories(ddsts PUBLIC "src")
-target_include_directories(ddsts PRIVATE "${GENERATED_SOURCE_DIR}")
+set(sources
+    "src/typetree.c")
 
+add_library(ddsts ${headers} ${sources})
 target_link_libraries(ddsts PRIVATE ddsc)
 
+# Create a dummy export header. generate_export_header can only be used with
```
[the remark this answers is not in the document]
FransFaase: It seems this is not needed when all uses of DDSTS_EXPORT are replaced by DDS_EXPORT. as you suggested in your next comment.

src/tools/idlc/src/idl.l
```diff
@@ -52,7 +50,7 @@ int yyerror(YYLTYPE *yylloc, yyscan_t yyscanner, dds_ts_context_t *context, char
 %option noyywrap
 %option nounistd
 %option never-interactive
-%option prefix="dds_ts_parser_"
+%option prefix="dds_parser_"
```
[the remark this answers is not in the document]
FransFaase: I have changed the dds_ prefix into ddsts_ and also changed some DDS_ defines into DDSTS_ and added some prefixes where they should have been used.

docs/dev/typesupport.md
```diff
@@ -183,3 +189,146 @@ union type {
...
+           +-- array size '6'  |
+               |               |
+               +---------------+
+```
+
+Note that the type 'short' is shared between the two declarations and
+that a far more complex type could have been used instead.
+During the freeing of the TypeTree, some mechanism is needed to determine
+if a tree is shared or not. Although, reference counting is commonly used
+for this, we decide to use a special flag for this. In the above example
```
eboasson: @FransFaase There is a reason reference counts are commonly used — could you please explain a little bit more about the reason why a single bit suffices?

E.g., is it that there is a clear notion of ownership, that all other references have a bit set to indicate that they are not, and that there is a mechanism that guarantees the owning references always outlasts all other references?

FransFaase: The current implementation assumes that the TypeTree are constructed once, during parsing the IDL-grammar or, in the future, one of the representation described in DDS-XTYPES, and never modified with respect to their structure.
The functions defined in typetree.h for constructing the TypeTree still lack a description of the ownership rules. The implementation of the function (in typetree.c) currently lack any checks on incorrect usage. If you think it is worth the effort to add these, I am happy to add them.

eboasson: @FransFaase: I'd really appreciate it if you'd add a brief description of the ownership rules. Checking for usage ... I'd say that's worth it if it is a small amount of work.

FransFaase: I have added some documentation to the functions in typetree.h
Every ddsts_type_t struct can be at most owned by one other ddsts_type_t. And if and only if, the struct is owned than the 'parent' member of ddsts_typespec_t points to its owner. The ownership relation between ddsts_type_t is a tree, where the root is the only struct with 'parent' equal NULL. When deleting the tree (from the root), it is sufficient to have only one flag, to mark all references that are not ownership relations, to take care that all structs are freed exactly once.

eboasson: Thanks @FransFaase. So that leaves only being unable to build on macOS and Windows. Would you be able to make it build on those platforms? It looks like it's simply a dependency on the "ddsc" library when generating the "ddsts" library.

eboasson: @FransFaase, I looked at the build problem a little bit better, and everything indeed points to the problem being that it doesn't link with "ddsc" when generating the "ddsts" shared library. While on Linux undefined symbols in a shared library is acceptable under the default settings, on macOS and Windows it isn't.

This in turn matches src/ddsts/CMakeLists.txt: there is no dependency on ddsc. Adding
```CMake
target_link_libraries(ddsts PRIVATE ddsc)
```
solves the problem on macOS but my CMake knowledge is entirely insufficient to know whether this introduces all kinds of other hidden issues. @k0ekk0ek, perhaps you can shed some light on this?

FransFaase: @eboasson, yes, that seems to be the solution, because ddsrt is made available in the ddsc (dynamic) library.

PR #184, opened 2019-05-16
Added code coverage target

ThijsSassen:
Added support for code coverage, the cmake files come from https://github.com/RWTH-HPC/CMake-codecov

To enable coverage support You can mark targets explicitly for coverage by adding your target with add_coverage(). This call must be done in the same directory as your `add_executable()` or `add_library()` call:
`add_executable(some_exe foo.c bar.c)`
`add_coverage(some_exe)`

`add_library(some_lib foo.c bar.c)`
`add_coverage(some_lib)`

When adding this to cyclonedds i also re arranged the travis.yml so things that belong to the specific platforms are only done there and made general functions that can be included in the specific targets.

In order maintain the code coverage setting the repository owner has to create a free codecov account which can be done here: https://codecov.io/. The [default](https://docs.codecov.io/docs/coverage-configuration) codecov settings of a minimum 70% coverage are applied. When this PR is approved it can also be added to the repository checks to track the coverage results changes of a PR.

Codecov results: https://codecov.io/gh/eclipse-cyclonedds/cyclonedds


k0ekk0ek:
Thanks for the pull request @ThijsSassen. This is something we still need to get sorted and this is a good step in the right direction. Of course, since this is an Eclipse project we need to have permission to include the CMAKE-codecov files. Can you find out if other Eclipse projects already use, or alternatively, find out which IP process applies here? Maybe @eboasson knows it from the top of his head? Apart from that, we'll also need to figure out if we're allowed to use the codecov.io service, but more importantly if it's a service that we actually want to use. I haven't looked into that, maybe @eboasson has some thoughts about this too? At the very least, we'll have to discuss what we want to do about "code coverage rules".

So, no remarks regarding the CMake files. I think it's the proper way to do this kind of stuff. I'll look at the changes to the `.travis.yml` shortly. One of things that caught my attention is the fact that you download a script from the codecov.io service and execute it without any verification. As I wrote in my comments with pull request #176, I really don't like that sort of thing. The script might contain malicious code without us knowing. So that's a definite no, but I'm sure we can fix that...

k0ekk0ek:
@eboasson: The Eclipse Hono project seems to be using codecov.io too, so it's likely that we could get the webmaster to enable it. An example of how the integration with codecov.io can look for a pull request can be found here: http://m.jxglzyxy.cn/github_/eclipse/hono/pull/1150. How a pull request is allowed/denied based on coverage can be tweaked in .codecov.yml (missing from this request). Let me know if you'd value the addition of code coverage statistics. If you do, do you think you can help out with getting the third-party CMake modules through IP review? Thanks.

@ThijsSassen:
* I don't agree with how you rearranged the `.travis.yml` file. I like the idea of a per platform install block, that really makes sense, but the other steps should be shared imho as it makes things a little easier to follow. In your version you have shared blocks that are included in the platform specific
* I'm missing the .codecov.yml file that configures how codecov behaves for this repository.
* I'm missing instructions on token variables etc that need to be exported.
* The FindX.cmake files are not written by you and as such need to go through IP review before we're allowed to add them. I do like the way they work/are constructed, so no comments on that.
* I'd like it if the Python module you're using now is just replaced by a CMake script that uploads the code to codecov. The original Bash script you used is pretty complex, but that's only because it tried to do everything for everyone. I think if you were to write a script dedicated to our situation it would be fairly simple.

If you want I can help you out, although I'd much prefer you'd finish pull request #175. Personally I think having static analysis is more important than code coverage statistics...

eboasson:
@ThijsSassen, @k0ekk0ek made some sensible comments and because of that I hadn't really looked closely at this PR. I don't foresee any issue with obtaining approval for the IP issues and therefore I don't see any point in requesting approval that until the other comments have been addressed. (I do apologise for not making this comment earlier.)

Do you intend to go forward with this and address the points that @k0ekk0ek raised?

k0ekk0ek:
@ThijsSassen I think the current `.travis.yml` is in itself a really good example of how things could be done for this PR too. Would it make sense to at least update this PR so that it doesn't have merge conflicts? I added the `install_coverity` and `submit_to_coverity_scan` anchors for the Coverity Scan service you can probably draw inspiration from those...

ThijsSassen:
Once i have time i'll re introduce this PR

PR #186, opened 2019-05-17
Dynamic lib loading

kurtuluso:
Platform independent dynamic loading feature is required for loading DDS Security plugins. It is also required for unit tests of the plugins.

In this pull request, the following functions are added to dds_rt:
**ddsrt_dlopen**: loads the dynamic shared object
**ddsrt_dlclose**: informs the system that the library, identified by 'handle', is no longer needed
**ddsrt_dlsym**: gets the memory address of a symbol from a loaded library
**ddsrt_dlerror**: gets the most recent library related error

The functions are implemented for windows and posix platforms that support dynamic loading.

### DIFF
--- a/src/ddsrt/tests/CMakeLists.txt
+++ b/src/ddsrt/tests/CMakeLists.txt
@@ -10,9 +10,11 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 include(CUnit)
+include(GenerateExportHeader)
 
 set(sources
     "atomics.c"
+    "dynlib.c"
     "environ.c"
     "heap.c"
     "ifaddrs.c"
@@ -65,4 +67,48 @@ set_target_properties(
 set(process_app_name "${CMAKE_CURRENT_BINARY_DIR}/process_app${CMAKE_EXECUTABLE_SUFFIX}")
 configure_file(
         "process_test.h.in" "${CMAKE_CURRENT_BINARY_DIR}/include/process_test.h" @ONLY)
+
+
+
+# Create a separate shared library that will be used to
+# test dynamic library loading.
+set(test_lib_name "dltestlib")
+add_library(${test_lib_name} SHARED dllib.c)
+# Force the lib to be at the same location, no matter what platform or build type.
+set_target_properties(
+        ${test_lib_name}
+        PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY                ${CMAKE_CURRENT_BINARY_DIR}
+        RUNTIME_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_CURRENT_BINARY_DIR}
+        RUNTIME_OUTPUT_DIRECTORY_RELEASE        ${CMAKE_CURRENT_BINARY_DIR}
+        RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO ${CMAKE_CURRENT_BINARY_DIR}
+        RUNTIME_OUTPUT_DIRECTORY_MINSIZEREL     ${CMAKE_CURRENT_BINARY_DIR}
+        LIBRARY_OUTPUT_DIRECTORY                ${CMAKE_CURRENT_BINARY_DIR}
+        LIBRARY_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_CURRENT_BINARY_DIR}
+        LIBRARY_OUTPUT_DIRECTORY_RELEASE        ${CMAKE_CURRENT_BINARY_DIR}
+        LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO ${CMAKE_CURRENT_BINARY_DIR}
+        LIBRARY_OUTPUT_DIRECTORY_MINSIZEREL     ${CMAKE_CURRENT_BINARY_DIR} )
+# Use proper export for this test lib.
+generate_export_header(${test_lib_name} BASE_NAME LIB_TEST)
+target_include_directories(${test_lib_name} PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
+# Let the cunit application know the location and name of the library.
+file(TO_NATIVE_PATH "${CMAKE_CURRENT_BINARY_DIR}" test_lib_native_dir)
+file(TO_NATIVE_PATH "/" test_lib_sep)
+string(REPLACE "\\" "\\\\" test_lib_dir ${test_lib_native_dir})
+string(REPLACE "\\" "\\\\" test_lib_sep ${test_lib_sep})
+configure_file("dl_test.h.in" "${CMAKE_CURRENT_BINARY_DIR}/include/dl_test.h" @ONLY)
+# Let ctest set the proper library path when executing library tests.
+unset(test_lib_tests)
+process_cunit_source_file("dynlib.c" test_lib_header test_lib_suites test_lib_tests)
+foreach(libtest ${test_lib_tests})
+  string(REPLACE ":" ";" libtest ${libtest})
+  list(GET libtest 0 suite)
+  list(GET libtest 1 test)
+  set(libtestname "CUnit_${suite}_${test}")
+  if("${CMAKE_HOST_SYSTEM}" MATCHES ".*Windows.*")
+    set_property(TEST ${libtestname} APPEND PROPERTY ENVIRONMENT "${test_lib_native_dir}")
+  else()
+    set_property(TEST ${libtestname} APPEND PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${test_lib_native_dir};$ENV{LD_LIBRARY_PATH}")
+  endif()
+endforeach()
 
--- a/src/cmake/modules/FindMaven.cmake
+++ b/src/cmake/modules/FindMaven.cmake
@@ -18,8 +18,13 @@ if(DEFINED ENV{M2_HOME})
 endif()
 
 # Chocolatey installs packages under C:\ProgramData\chocolatey.
-if(ENV{ProgramData} AND IS_DIRECTORY "$ENV{ProgramData}/chocolatey/bin")
-  list(APPEND _mvn_paths "$ENV{ProgramData}/chocolatey/bin")
+if(NOT "$ENV{ProgramData}" STREQUAL "")
+  if(IS_DIRECTORY "$ENV{ProgramData}/chocolatey/bin")
+    list(APPEND _mvn_paths "$ENV{ProgramData}/chocolatey/bin")
+  endif()
+  if(IS_DIRECTORY "$ENV{ProgramData}/chocolatey/bin")
+    list(APPEND _dirs "$ENV{ProgramData}/chocolatey/lib/maven")
+  endif()
 endif()
 
 # Maven documentation mentions intalling maven under C:\Program Files on
--- a/src/ddsrt/CMakeLists.txt
+++ b/src/ddsrt/CMakeLists.txt
@@ -66,6 +66,7 @@ list(APPEND headers
   "${include_path}/dds/ddsrt/misc.h"
   "${include_path}/dds/ddsrt/io.h"
   "${include_path}/dds/ddsrt/process.h"
+  "${include_path}/dds/ddsrt/dynlib.h"
   "${include_path}/dds/ddsrt/strtod.h"
   "${include_path}/dds/ddsrt/strtol.h"
   "${include_path}/dds/ddsrt/types.h")
@@ -102,7 +103,7 @@ list(APPEND sources
 # network stack. In order to mix-and-match various compilers, architectures,
 # operating systems, etc input from the build system is required.
 foreach(feature atomics cdtors environ heap ifaddrs random rusage
-                sockets string sync threads time md5 process)
+                sockets string sync threads time md5 process dynlib)
   if(EXISTS "${include_path}/dds/ddsrt/${feature}.h")
     list(APPEND headers "${include_path}/dds/ddsrt/${feature}.h")
     file(GLOB
@@ -176,6 +177,7 @@ target_sources(ddsrt INTERFACE ${sources})
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 find_package(Threads REQUIRED)
 target_link_libraries(ddsrt INTERFACE Threads::Threads)
+target_link_libraries(ddsrt INTERFACE ${CMAKE_DL_LIBS})
 
 if(WIN32)
   target_link_libraries(ddsrt INTERFACE wsock32 ws2_32 iphlpapi bcrypt)
--- a/src/ddsrt/include/dds/ddsrt/dynlib.h
+++ b/src/ddsrt/include/dds/ddsrt/dynlib.h
@@ -1,0 +1,143 @@
+
+/*
+ * Copyright(c) 2006 to 2018 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef DDSRT_LIBRARY_H
+#define DDSRT_LIBRARY_H
+
+#include "dds/export.h"
+#include "dds/ddsrt/types.h"
+#include "dds/ddsrt/retcode.h"
+#include "dds/ddsrt/attributes.h"
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+
+//typedef void *ddsrt_dynlib_t;
+typedef struct ddsrt_dynlib *ddsrt_dynlib_t;
+
+
+/**
+ * @brief Load a dynamic shared library.
+ *
+ * The function ddsrt_dlopen() loads the dynamic shared object (shared library)
+ * file, identified by 'name', sets the handle parameter for the loaded library and
+ * returns the result with dds return code.
+ *
+ * If the 'translate' boolean is true, this function will first try to open the
+ * library with a translated 'name'. Translated in this context means that if
+ * "mylibrary" is provided, it will be translated into libmylibrary.so,
+ * libmylibrary.dylib or mylibrary.dll depending on the platform.
+ * This translation only happens when the given name does not contain
+ * a directory.
+ * If the function isn't able to load the library with the translated name, it
+ * will still try the given name.
+ *
+ * @param[in]   name        Library file name.
+ * @param[in]   translate   Automatic name translation on/off.
+ * @param[out]  handle      Library handle that will be assigned after successfull operation. It is assigned to NULL if loading fails.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             Library handle was successfully loaded.
+ * @retval DDS_RETCODE_BAD_PARAM
+ *             There is an invalid input in the parameter list
+ * @retval DDS_RETCODE_ERROR
+ *             Loading failed.
+ *             Use ddsrt_dlerror() to diagnose the failure.
+ */
+DDS_EXPORT dds_retcode_t
+ddsrt_dlopen(
+    const char *name,
+    bool translate,
+    ddsrt_dynlib_t *handle) ddsrt_nonnull_all;
+
+/**
+ * @brief Close the library.
+ *
+ * The function ddsrt_dlclose() informs the system that the
+ * library, identified by 'handle', is no longer needed.
+ * will get the memory address of a symbol,
+ * identified by 'symbol', from a loaded library 'handle'.
+ *
+ * @param[in]   handle      Library handle.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             Library handle was successfully closed.
+ * @retval DDS_RETCODE_ERROR
+ *             Library closing failed.
+ *             Use ddsrt_dlerror() to diagnose the failure.
+ */
+DDS_EXPORT dds_retcode_t
+ddsrt_dlclose(
+    ddsrt_dynlib_t handle);
+
+/**
+ * @brief Get the memory address of a symbol.
+ *
+ * The function ddsrt_dlsym() will get the memory address of a symbol,
+ * identified by 'symbol', from a loaded library 'handle'.
+ *
+ * @param[in]   handle      Library handle.
+ * @param[in]   symbol      Symbol name.
+ * @param[out]  address     The memory address of the loaded symbol (void*).
+ *
+ * @returns  A dds_retcode_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             Symbol was found in the loaded library.
+ *             Address parameter is ready to use.
+ * @retval DDS_RETCODE_ERROR
+ *             Symbol was not found.
+ *             Use ddsrt_dlerror() to diagnose the failure.
+ */
+DDS_EXPORT dds_retcode_t
+ddsrt_dlsym(
+    ddsrt_dynlib_t handle,
+    const char *symbol,
+    void **address);
+
+/**
+ * @brief Get the most recent library related error.
+ *
+ * The function ddsrt_dlerror() will return the most recent error of the operating system
+ * in human readable form.
+ *
+ * If no error was found, it's either due to the fact that there
+ * actually was no error since init or last ddsrt_dlerror() call,
+ * or due to an unknown unrelated error.
+ *
+ * As error reporting function can be used for different purposes, dssrt_dlerror
+ * function should be called immediately after calling ddsrt_dlopen or ddsrt_dlsym
+ * function.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             Most recent library related error returned.
+ * @retval DDS_RETCODE_NOT_FOUND
+ *             No library related error found.
+ */
+DDS_EXPORT dds_retcode_t
+ddsrt_dlerror(
+    char *buf,
+    size_t buflen);
+
+#if defined (__cplusplus)
+}
+#endif
+
+#endif /* DDSRT_LIBRARY_H */
--- a/src/ddsrt/src/dynlib/posix/dynlib.c
+++ b/src/ddsrt/src/dynlib/posix/dynlib.c
@@ -1,0 +1,95 @@
+/*
+ * Copyright(c) 2006 to 2018 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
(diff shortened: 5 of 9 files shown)


martinbremmer:
Some comments:
1. It would have been nice if this PR only contained one or a few commits. Now it contains the difference between master and security branch as well.
1. Make ddsrt_dlopen() and ddsrt_dlsym() more consistent. Either let both return dds_retcode_t or let both return a pointer.
1. The handle param comment in the ddsrt_dlopen() header is not correct.
1. Add a small explanation to DDS_RETCODE_BAD_PARAM retval comment in the ddsrt_dlopen() header. Even if it seems a bit redundant.
1. Why not return DDS_RETCODE_BAD_PARAM in ddsrt_dlopen() when handle is NULL? Be aware that handle is set first thing. If it's NULL, it'll crash. Please add tests that checks what happens when either (or both) name and/or handle is NULL.
1. ddsrt_dlopen() checks for NULL pointer (at least for the name argument). All the other functions just assert. Please make it more consistent. Either let ddsrt_dlopen() assert with NULL as well, or add NULL checks to the other functions (and add tests to check NULL usage if you choose the latter).

eboasson:
Thanks @kurtuluso for the pull request and @martinbremmer for reviewing it. I'd just like to point out that I am hopeful that the minor details left on #177 will be resolved soon and that it can be merged shortly. That would make it trivial to address the first point.

kurtuluso:
1. Actually, there is only one commit but the security main branch is not up to date. And thıs wıll not be changed until be'ng prepared to merge to master
2.  DONE
3. DONE
4. DONE
5. DONE
6. DONE

kurtuluso:
@eboasson 
This is ready for review (all pass in local travis: https://travis-ci.org/kurtuluso/cyclonedds/builds/536755903?utm_source=github_status&utm_medium=notification) but it still shows conflict.
I've rebased to updated security branch and readded my changes but still, it is reporting conflict in the pull request.
I think it needs your touch.

src/ddsrt/include/dds/ddsrt/library.h
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * The function ddsrt_dlclose() informs the system that the
+ * library, identified by 'handle', is no longer needed.
+ * will get the memory address of a symbol,
+ * identified by 'symbol', from a loaded library 'handle'.
+ *
+ * @param[in]   handle      Library handle.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
```
martinbremmer: The fact that ddsrt_dlclose() can return DDS_RETCODE_BAD_PARAMETER, should be mentioned.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,118 @@
...
+    return retcode;
+}
+
+dds_retcode_t
+ddsrt_dlsym(
+    ddsrt_lib handle,
+    const char *symbol,
+    void **address)
+{
+    *address = NULL;
```
martinbremmer: If address is NULL, this'll crash.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,118 @@
...
+dds_retcode_t
+ddsrt_dlerror(
+    char *buf,
+    size_t buflen)
+{
+  const char *err = dlerror();
+  if (err == NULL) {
+    return DDS_RETCODE_NOT_FOUND;
+  }
+  snprintf(buf, buflen, "%s", err);
```
martinbremmer: If buf is NULL, this'll crash.

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,115 @@
...
+
+dds_retcode_t
+ddsrt_dlerror(char *buf, size_t buflen)
+{
+  /* Hopefully (and likely), the last error is
+   * related to a Library action attempt. */
+  DWORD err = GetLastError();
+
+  /* Get the string that is related to this error. */
+  if (ddsrt_strerror_r(err, buf, buflen) == DDS_RETCODE_OK) {
```
martinbremmer: This'll assert (debug) or likely crash (release) when buf is NULL.

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+  printf("\n%s", buffer);
+}
+
+CU_Test(ddsrt_library, dlopen_null)
+{
+  char buffer[256];
+  dds_retcode_t r;
+  ddsrt_lib l;
+
+  r = ddsrt_dlopen(NULL, false, &l);
```
martinbremmer: I also expect a test with the library handle being NULL.

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+
+  f_set(set_int);
+  get_int = f_get();
+  CU_ASSERT_EQUAL(set_int, get_int);
+
+  r = ddsrt_dlclose(l);
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_OK);
+}
+
+CU_Test(ddsrt_library, dlclose_null)
```
martinbremmer: Why is this called dlclose_null? There's only one test with NULL. Please split this up in multiple tests.

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+    CU_ASSERT_EQUAL(r, DDS_RETCODE_OK);
+
+    r = ddsrt_dlclose( l ); /*already closed handle */
+    CU_ASSERT_EQUAL(r, DDS_RETCODE_ERROR);
+
+    r = ddsrt_dlclose( NULL );
+    CU_ASSERT_EQUAL(r, DDS_RETCODE_BAD_PARAMETER);
+
+}
+
```
martinbremmer: Where are the ddsrt_dlerror() that supply NULL to their arguments?

src/ddsrt/include/dds/ddsrt/library.h
```diff
@@ -0,0 +1,274 @@
...
+ * @param[in]   translate   Automatic name translation on/off.
+ * @param[out]  handle      Automatic name translation on/off.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             Library handle was successfully loaded.
+ * @retval DDS_RETCODE_BAD_PARAM
+ *             There is an invalid input in the parameter list
+ *
```
martinbremmer: Remove this white-line.

src/ddsrt/include/dds/ddsrt/library.h
```diff
@@ -0,0 +1,274 @@
...
+ * The function ddsrt_dlerror() will return the most recent library
+ * related error in human readable form.
+ *
+ * If no error was found, it's either due to the fact that there
+ * actually was no error since init or last ddsrt_dlerror() call,
+ * or due to an unknown unrelated error.
+ *
+ * @returns A dds_retcode_t indicating success or failure.
+ *
```
martinbremmer: When NULL checks are added to ddsrt_dlerror(), then DDS_RETCODE_BAD_PARAMETER should be added to this header.

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+  r = ddsrt_dlsym(l, "get_int", &f);
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_OK);
+  CU_ASSERT_PTR_NOT_NULL(f);
+  TEST_ABORT_IF_NULL(f, "ddsrt_dlsym(l, \"get_int\") failed.");
+
+  r = ddsrt_dlclose(l);
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_OK);
+}
+
+CU_Test(ddsrt_library, dlsym_bad)
```
martinbremmer: Please rename this to dlsym_null

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+
+  r = ddsrt_dlopen(TEST_LIB_NAME, true, &l);
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_OK);
+  CU_ASSERT_PTR_NOT_NULL(l);
+  TEST_ABORT_IF_NULL(l,"ddsrt_dlopen() failed. Is the proper library path set?");
+
+  r = ddsrt_dlsym(l, NULL, &f);
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_BAD_PARAMETER);
+
+  r = ddsrt_dlsym(NULL, "UnknownSym", &f);
```
martinbremmer: Please use a valid symbol name when you only want to check what happens the the library handle is NULL.

src/ddsrt/tests/dl.c
```diff
@@ -0,0 +1,224 @@
...
+  CU_ASSERT_EQUAL(r, DDS_RETCODE_BAD_PARAMETER);
+
```
martinbremmer: I'd also expect a test that checks what happens when the symbol argument is NULL.

martinbremmer:
I don't have any more remarks.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+      dlerror();    /* Clear any existing error */
+      if ( (translate) && (strrchr(name, '/') == NULL)) {
+        /* Add lib and suffix to the name and try to open. */
+#if __APPLE__
+        static const char suffix[] = ".dylib";
+#else
+        static const char suffix[] = ".so";
+#endif
+        char* libName = ddsrt_malloc(3 + strlen(name) + sizeof(suffix));
+        sprintf(libName, "lib%s%s", name, suffix);
```
k0ekk0ek: Better not use `sprintf` here, but `snprintf`. If only to avoid compiler warnings. `ddsrt_asprintf` would be an option to, that way you can also ditch the `ddsrt_malloc`.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+  dds_retcode_t retcode = DDS_RETCODE_OK;
+
+
+  if ( name == NULL || handle == NULL ){
+      retcode = DDS_RETCODE_BAD_PARAMETER;
+
+  } else {
+      *handle = NULL;
+      dlerror();    /* Clear any existing error */
+      if ( (translate) && (strrchr(name, '/') == NULL)) {
```
k0ekk0ek: I understand the `strrchr` is there to check if it's not a path, but that might not be a good enough check(?). It might just be that the user means a file in the current directory, so if there's no `/` in the name, perhaps do a `stat` as well? Maybe @eboasson has something on this?

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+
+
+dds_retcode_t
+ddsrt_dlclose(
+    ddsrt_lib handle)
+{
+    dds_retcode_t retcode;
+
+    if( handle == NULL ){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
```
k0ekk0ek: Just use an early return here. Perhaps just settle on an `assert` or just don't do anything if `handle` is `NULL`.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+    ddsrt_lib handle,
+    const char *symbol,
+    void **address)
+{
+    dds_retcode_t retcode = DDS_RETCODE_OK;
+
+    if ( symbol == NULL || handle == NULL || address == NULL){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
+    } else {
+        *address =  dlsym(handle, symbol);
```
k0ekk0ek: Use a local variable and only assign `address` on success.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+dds_retcode_t
+ddsrt_dlerror(
+    char *buf,
+    size_t buflen)
+{
+
+    const char *err;
+    dds_retcode_t retcode = DDS_RETCODE_OK;
+
+    if( buf == NULL ){
```
k0ekk0ek: An `assert` would be better in this case.

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,124 @@
...
+    dds_retcode_t retcode = DDS_RETCODE_OK;
+
+
+    if ( name == NULL || handle == NULL ){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
+
+    } else {
+        *handle = NULL;
+        GetLastError(); /* Clear any existing error */
+        if ((translate) && (strrchr(name, '/') == NULL )
```
k0ekk0ek: Same as with the POSIX comments.

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,124 @@
...
+
+dds_retcode_t
+ddsrt_dlclose(
+    ddsrt_lib handle)
+{
+    dds_retcode_t retcode;
+
+    if( handle == NULL ){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
+    }
```
k0ekk0ek: Please choose a style and stick with it.

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,124 @@
...
+            retcode = DDS_RETCODE_ERROR;
+        }
+    }
+
+    return retcode;
+}
+
+
+dds_retcode_t
+ddsrt_dlerror(char *buf, size_t buflen)
```
k0ekk0ek: This doesn't do the same thing as the `ddsrt_dlerror` implementation in the POSIX layer above. `GetLastError` is used by much more of the api. It's better to use a thread-local variable to store the error when it occurred in one of the functions. Of course just dropping the `ddsrt_dlerror` call entirely is also a possibility. @eboasson do you have a preference here?

src/ddsrt/include/dds/ddsrt/library.h
```diff
@@ -0,0 +1,277 @@
...
+
+#include "dds/export.h"
+#include "dds/ddsrt/types.h"
+#include "dds/ddsrt/retcode.h"
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+typedef void *ddsrt_lib;
```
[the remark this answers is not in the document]
kurtuluso: I've added _t suffix but which _standard_ are we talking about? Is there a document to adhere? I do not remember now but I saw types without _t suffix.
Windows: 

```
typedef void *PVOID;
typedef PVOID HANDLE;
typedef HANDLE HINSTANCE;
typedef HINSTANCE HMODULE;
HMODULE LoadLibraryA(
  LPCSTR lpLibFileName
);

```

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,124 @@
...
+{
+    dds_retcode_t retcode = DDS_RETCODE_OK;
+
+
+    if ( name == NULL || handle == NULL ){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
+
+    } else {
+        *handle = NULL;
+        GetLastError(); /* Clear any existing error */
```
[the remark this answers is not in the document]
kurtuluso: User should run the GetLastError if there is an error.
I've removed the line that has reset-purpose. As stated in the function explanation, the user should check error only if he/she gets DDS_RETCODE_ERROR.

src/ddsrt/src/library/windows/library.c
```diff
@@ -0,0 +1,124 @@
...
+ddsrt_dlopen(
+    const char *name,
+    bool translate,
+    ddsrt_lib *handle)
+{
+    dds_retcode_t retcode = DDS_RETCODE_OK;
+
+
+    if ( name == NULL || handle == NULL ){
+        retcode = DDS_RETCODE_BAD_PARAMETER;
```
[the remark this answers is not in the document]
kurtuluso: early return is not a good practice, I try single return as much as I can. If you have a previous decision, I can change it.
I remember this practice in JOINT STRIKE FIGHTER AIR VEHICLE C++ CODING STANDARDS 

> AV Rule 113 (MISRA Rule 82, Revised)
Functions will have a single exit point.
Rationale: Numerous exit points tend to produce functions that are both difficult to
understand and analyze.
Exception: A single exit is not required if such a structure would obscure or otherwise
significantly complicate (such as the introduction of additional variables) a function’s control
logic. Note that the usual resource clean-up must be managed at all exit points.

src/ddsrt/src/library/posix/library.c
kurtuluso: I think it is safer to reset the given argument to NULL if there is an error and the dlsym already sets it to NULL in case of error.

src/ddsrt/src/library/windows/library.c
kurtuluso: done

src/ddsrt/src/library/windows/library.c
kurtuluso: Yes, you're right. However, GetLastError stores the error message for each thread. So, if the user calls the ddsrt_dlerror just after calling the other library functions, he/she will get the error message properly.

src/ddsrt/src/library/windows/library.c
kurtuluso: what do you mean?

src/ddsrt/src/library/posix/library.c
kurtuluso: As we provide these functions to other components, we decided to convert asserts to NULL checks.
If you think that it is impossible to get buf == NULL, I can change it.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+ddsrt_dlopen(
+    const char *name,
+    bool translate,
+    ddsrt_lib *handle)
+{
+  dds_retcode_t retcode = DDS_RETCODE_OK;
+
+
+  if ( name == NULL || handle == NULL ){
+      retcode = DDS_RETCODE_BAD_PARAMETER;
```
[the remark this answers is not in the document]
eboasson: no checking and ddsrt_nonnull_all would be even better

src/ddsrt/src/library/posix/library.c
eboasson: @k0ekk0ek this is actually the sensible thing to do: this way you can always avoid the translation by specifying ./file. I'd prefer to see "strchr" rather than "strrchr" though. (Also for windows.)

src/ddsrt/src/library/posix/library.c
eboasson: As this would be the _only_ use of ``sprintf`` ... yes, please replace it.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+
+  } else {
+      *handle = NULL;
```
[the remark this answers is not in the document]
eboasson: agree

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+#endif
+        char* libName = ddsrt_malloc(3 + strlen(name) + sizeof(suffix));
+        sprintf(libName, "lib%s%s", name, suffix);
+        *handle = dlopen(libName, RTLD_GLOBAL | RTLD_NOW);
+        ddsrt_free(libName);
+
+
+      }
+
+      if (*handle == NULL && retcode == DDS_RETCODE_OK) {
```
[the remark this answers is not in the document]
eboasson: I do actually care about the layout in general and really dislike a bunch of blank lines before closing a block ... there are plenty of other details that I really prefer differently but those do irk me so much.

Checking ``*handle`` makes sense (that's simply so a successful dlopen on the "translated" name is handled correctly). The ``retcode`` check makes no sense here.

src/ddsrt/src/library/posix/library.c
```diff
@@ -0,0 +1,127 @@
...
+        /* name contains a path,
+         * (auto)translate is disabled or
+         * dlopen on translated name failed. */
+        *handle = dlopen(name, RTLD_GLOBAL | RTLD_NOW);
```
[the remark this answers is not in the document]
eboasson: if only the rest of the code were perfectly consistent on that point ... the documentation comment ought to indicate what the effect on the out parameter is in case of failure

src/ddsrt/src/library/posix/library.c
eboasson: agree

src/ddsrt/src/library/posix/library.c
eboasson: I'm have a bigger problem with the coding style (this is not the only place) of declaring-and-initialising a return value to "OK" and then changing it on error. *If* one insists on single return and initialising all variables at declaration time as "safety" measures, *then* the logical thing would be to initialise a return code to "ERROR".

src/ddsrt/src/library/posix/library.c
eboasson: I don't think @k0ekk0ek 's point is that it is impossible to get a null pointer for buf; it is rather that there is a good argument to make consider it illegal to do so, i.e., that a precondition is that buf != NULL. That means an assert, preferably a compiler annotation that it must not be a null pointer. The only callers will be internal and there seems to be no good reason to ever call it with a null pointer for buf. Checking for a null pointer will only hide errors.

src/ddsrt/src/library/windows/library.c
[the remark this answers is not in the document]
eboasson: @kurtuluso : please read what you quoted:

> A single exit is not required if such a structure would obscure or otherwise
significantly complicate (such as the introduction of additional variables) a function’s control
logic. Note that the usual resource clean-up must be managed at all exit points.

An early return here reduces the amount of state on needs to track mentally (no null pointers — so no need to account for null pointers anywhere in the function anymore). No state has been changed yet, so an early return is the better option.

src/ddsrt/src/library/windows/library.c
eboasson: There are various possibilities — could be whitespace in combination with parenthesis ...

Also this is another of those functions where early returns and ternary expressions would be easier to read.

src/ddsrt/src/library/windows/library.c
eboasson: That's not what the documentation blurb says ("Get the most recent library related error.") You're now suggesting that it should be extended with "and you may not have intervening calls to any Win32 functions on the same thread". While I agree that commonly one would expect dlerror to be called immediately, there is something to be said in favour of implemting the interface properly.

src/ddsrt/tests/CMakeLists.txt
```diff
@@ -10,9 +10,11 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 include(CUnit)
+include(GenerateExportHeader)
 
 set(sources
     "atomics.c"
+    "dl.c"
```
eboasson: just like "library" is a bit vague, "dl" is a bit vague, too

src/ddsrt/include/dds/ddsrt/library.h
```diff
@@ -0,0 +1,141 @@
...
+ *             Symbol was found in the loaded library.
+ *             Address parameter is ready to use.
+ * @retval DDS_RETCODE_BAD_PARAM
+ *             There is an invalid input in the parameter list
+ * @retval DDS_RETCODE_ERROR
+ *             Symbol was not found.
+ *             Use ddsrt_dlerror() to diagnose the failure.
+ */
+DDS_EXPORT dds_retcode_t
+ddsrt_dlsym(
```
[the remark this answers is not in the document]
kurtuluso: We've mapped posix and windows functions in ddsrt. Which posix function are you talking about?

src/ddsrt/src/library/posix/library.c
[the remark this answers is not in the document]
kurtuluso: In DDS Security case, the parameters are coming from the configuration file and the user of this function probably get these parameters on the fly (not static), so I think it is better to check here.

src/ddsrt/src/library/posix/library.c
kurtuluso: The initial result value is OK because, in this approach, there can be different error codes; the correct one will be assigned according to conditions. If there had only one error code and several valid (OK) return codes, I would have used ERROR initialisation at the beginning.

src/ddsrt/src/library/posix/library.c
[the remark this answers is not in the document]
kurtuluso: Parameter documentation has been updated:
> @param[out]  handle      Library handle that will be assigned after successfull operation. It is assigned to NULL if loading fails.

src/ddsrt/src/library/posix/library.c
kurtuluso: @eboasson What is the parameter handling policy in Cyclone?
I will refactor all parameter handlings according to the policy.

src/ddsrt/include/dds/ddsrt/library.h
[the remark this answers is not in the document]
eboasson: dlsym — POSIX and Windows guarantee function pointers and data pointers are interchangeable, but C does not.

src/ddsrt/src/library/windows/library.c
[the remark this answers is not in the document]
kurtuluso: It is ~20 lines of function in total. Using a single return here does not make it complicated or introduce new additional variables. Of course, early return always reduces the amount of state on needs to track mentally but I think it does not fit the exception.
Anyway, I will not continue the discussion anymore, I'm just adding a return.

src/ddsrt/src/library/windows/library.c
kurtuluso: I've updated the description

src/ddsrt/tests/CMakeLists.txt
kurtuluso: dynlib

src/ddsrt/src/library/posix/library.c
kurtuluso: I've changed all parameter checks to assert. We need an assert policy in the coding guideline. (Of course, we need a guideline first)

eboasson:
@kurtuluso there are still various things that I want to see changed, but I am getting the feeling that that would ultimately just end up in the absence of any pull requests for a security implementation. So, let's do it this way: I'll merge it into the security branch in the hope of seeing interesting things in the near future. Merging from the security branch into master is then going to be a painful affair again, but it will then (hopefully) be worth the pain.

Would you be so kind as to cherry-pick the CMake change I did for the Windows builds? There is Windows-specific code in here, so it has get through the CI build successfully before I can merge it.

Regarding your question on the use of assertions: the traditional guideline applies. Assertions are for checking invariants — and preconditions are invariants, too, so assertions are fine for checking parameters on internal functions. External interfaces — whether that be network, files, parameters in API functions — need to be validated with proper error handling, and so for those checks, assertions may not be used.

The ddsrt library is a platform abstraction for use by the Cyclone implementation, as well as any brave souls who decide to use it in their application. Misuse of ddsrt is therefore an affair punishable with exceptions and/or aborts raised by failed assertions.

The reason behind this is that one of the key tricks when writing software is to minimise the amount of state to be kept track of and to avoid having to handle cases that cannot in reality exist. Handling situations where a precondition was not met introduces more state than recognising that by definition such a situation may never occur, and therefore more complexity and a higher likelihood of further failures or mistakes.

At all times, static analysis is preferred over dynamic checks, so if the compiler or an easily accessible static analysis tool can guarantee the precondition, that is preferred over an assertion. Unforunately, Microsoft's SAL annotations are too platform specific to really be of use in this project; I still need to experiment with Frama-C.

PR #194, opened 2019-06-03
IDL-compiler: Support for topic keys

FransFaase:
This commit adds support for topic keys, which are needed for a minimal implementation of the C99 back-end.

Both '@key' as the '#pragma keylist' method for defining topic keys are supported. It is possible to define an inline struct as a key. If some of its members are defined as keys, only those members actually are keys. If non of its members are defined as keys, all members are considered to be keys. The keys are stored with the structure as a linked list with pointers to the member declarations of the struct. The tests have been extended to cover the topic keys.

The grammar has been adapted such that no spaces are allowed between identifiers and the '@' and '::' symbols. This is needed to parse a member declaration such as '@key ::a::s x;' correctly.

### DIFF
--- a/src/ddsts/include/dds/ddsts/typetree.h
+++ b/src/ddsts/include/dds/ddsts/typetree.h
@@ -53,7 +53,7 @@
 
 #define DDSTS_TYPES                    ((1L<<(26+1))-1L)
 #define DDSTS_TYPE_OF(O)               ((O)->type.flags & DDSTS_TYPES)
-#define DDSTS_IS_TYPE(O,T)             (DDSTS_TYPE_OF(O) == (T))
+#define DDSTS_IS_TYPE(O,T)             ((DDSTS_TYPE_OF(O) & (T)) != 0)
 #define DDSTS_IS_DEFINITION(O)         ((DDSTS_DEFINITIONS & (O)->type.flags) != 0)
 
 #define DDSTS_UNBOUND                  (1L<<27)
@@ -109,6 +109,11 @@ struct ddsts_typespec {
   ddsts_type_t *next;       /* pointer to the next sibling */
   void (*free_func)(ddsts_type_t*);
 };
+
+typedef struct {
+  ddsts_type_t *first;     /* pointer to the first element */
+  ddsts_type_t **ref_end;  /* pointer to the pointer where a next element can be appended */
+} ddsts_type_list_t;
 
 /* Base type specification (base_type_spec) */
 typedef struct {
@@ -261,7 +266,7 @@ ddsts_create_map(ddsts_type_t *key_type, ddsts_type_t *value_type, unsigned long
 typedef struct ddsts_module ddsts_module_t;
 struct ddsts_module {
   ddsts_typespec_t type;
-  ddsts_type_t *members;
+  ddsts_type_list_t members;
   ddsts_module_t *previous; /* to previous open of this module, if present */
 };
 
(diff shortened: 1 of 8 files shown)


eboasson:
@FransFaase, that's a useful addition indeed. I don't see anything that is obviously incorrect, and given that this is destined for a separate branch, intermediate commits are fine. I do have a number of questions, and I would like you to consider them before I merge the commit in. None of them is a show stopper, but ultimately, I would prefer to see them addressed — or to be convinced that I am overlooking something.

* free_struct: stylistically, I really dislike ``for`` statements where the second or third component is missing … It’s perfectly well-defined and rewriting it to a ``while`` is not going to improve it other than that I would like it better. I’m noting it, perhaps you can do me the pleasure of avoiding it next time.
* ddsts_struct_add_key: I generally dislike appending items to singly linked lists in this way. In this case, I believe it is reasonable to assume the number of key fields will be small, and that case it because a non-issue. Perhaps a comment to this effect would be wise.
Secondly, what if I try to add the same member twice? It looks to me it would be listed twice — and a cursory look at the rest of the preprocessor doesn’t suggest it is checked for.
* Regarding the representation of the annotations: I would much prefer an enumerated type covering all the annotations that are understood by the preprocessor. It currently appears to rely on string constants, which is much more likely to create unpleasant surprises.
* idl.l: rename NOWS to NO_WS perhaps? It is not like underscores to separate words aren’t already used, and it would take a few milliseconds less to parse for the unsuspecting reader. That said, I do wonder if the “NOWS” lexer state would not be better replaced by simply matching “@ident” as one thing? I admit I haven’t studied the implementation, the grammar nor the set of annotations well enough to know whether they could be handled that way, but it does seem an obvious and possibly simpler approach.

FransFaase:
@eboasson, with respect to your comments:

- Appending items to singly linked lists also happens in ddsts_module_add_member and ddsts_struct_add_member, where is a bigger issue, as the number of members is likely to be much larger than the number of keys. It is a matter, as so often, between time and space. In case of an IDL-compiler, which usually executed infrequently, time is less of an issue than space. But when the TypeTree going to be used for process XType representations, time might become much more critical. I propose, the introduction of a ddsts_type_list_t type, which contains a pointer to the head of a singly linked list, and a pointer to the pointer where next type element can be appended. This type will then be used for the 'members' struct members. I propose to add a comment with respect to keys as you suggested. Keys will not run in the hundreds, whereas modules with hundreds of members (elements) are to be expected.

- Checks on the repeating keys are lacking. I will add a check for this. It also has been proposed that the '@key' and '#pragma keylist' should not be mixed. I propose to add a check that if a '@key' has been used in a struct, the use of '#pragma keylist' will result in an error. It is also possible to implement a global check that either one of the methods is used.

- For the topic keys, only a very limited part of the annotations has been included in the grammar. For the support of '@key' it would have been enough to only support the combination of '@' and an identifier, where the grammar specifies a scoped name. Scoped names are only relevant when also user defined annotations are implemented. In that case, something like '@::mymodule::myannotation' would be valid. Also, annotations usually have at least one argument. In case there is one argument with a default value, it can be left out. '@key' has a Boolean argument with default value 'true', so '@key' is a shorthand for '@key(true)'. I decided to support scoped names instead of just an identifier after '@', because I wanted to figure out how to implement the no-white-space requirement with yacc/bison. I propose to delay the implementation with an enumerate type, and possibly a pointer to ddsts_type_t for user defined annotations, when implementing the full set of (user defined) annotations.

eboasson:
@FransFaase — thanks for the clarifications and modifications. No futher comments at this point :)

PR #197, opened 2019-06-13
Do not reset dds_sample_rejected_status_t::last_reason when getting status

martinbremmer:
Getting a status should reset its ```change_count``` variables. But only those.
Calling ```dds_get_sample_rejected_status()``` would cause the ```last_reason``` variable to be reset as well. This reset has been removed.

This ```change_count``` reset is also applicable for listener triggers. But the listener already left the ```last_reason``` alone when resetting ```dds_sample_rejected_status_t```.

Tests were added.


eboasson:
Thanks, @martinbremmer. I think one could argue both ways, because in my view "last_reason" really is quite meaningless if "change_count" is 0 (either nothing happened yet, or you already dealt with it). Because of that, the discrepancy between the listener behaviour and that of ``dds_get_sample_rejected_status`` is more of an issue to me than the resetting of "last_reason". Now, which is the more sensible option? 🤔

Given that the spec is explicit about resetting the "change_count" fields and doesn't say a word about resetting "last_reason", and a quick look at other open source implementations suggests that not resetting it is at least an accepted strategy, I'll go with your option.

martinbremmer:
Thanks

martinbremmer:
BTW, I considered 'last_reason' in the same way as handles in a few other states (statuses?), which aren't reset either.

PR #200, opened 2019-06-17
IDL-compiler: C99 back-end

FransFaase:
This pull-request contains an implementation of the C99 back-end (limited to the grammar constructs implemented in the front-end) together with tests.

Some modifications have been made in the ddsts. In typetree.h the type ddsts_call_path_t has been added. This is not part of the TypeTree but needed in utility functions that depend on the path through which a type is reached. An example of such a function, is the function ddsts_declaration_is_key, that can be used to find out if a member declaration of an (embedded) struct is defined as a key. The file type_walk.h with the function ddsts_walk has been added.

The implementation of the back-end makes use of the type gen_ostream as a generic kind of output stream with function pointers for opening and closing 'files'. In the tests this is used to capture the generated output into a buffer. It could, for example, also be used to place the generated output in a specified target directory.
The file gen_c99.h declares two functions for generating the C99 back-end given the name of the input file and a pointer to a type.

In the front-end a small problem with freeing has been solved in case of an error. For this the semantics of ddsts_free_type have been changed, such that it returns an error when a pointer to a part of a type is passed. A check for using a forward struct (with no struct definition) in a member declaration has been added.


FransFaase:
For some reason the Travis check is not showing. Maybe because I resubmitted this pull-request because I had to correct the author email address for the eclipsefdn/eca check. The associated Travis build for the commit is: https://travis-ci.org/FransFaase/cyclonedds/builds/546780506

k0ekk0ek:
Hi @FransFaase! Haven't looked at the changes just yet (will probably do so this evening), but the Travis shows a green check mark, so no worries there.

k0ekk0ek:
Hi @FransFaase. Didn't come around to reviewing this last Wednesday, sorry for that. I remember us having a conversation about introducing an `ostream` vs just using `stdio`. I still don't like the output stream much, I would prefer being able to simply use `FILE` handles, but I suggest not letting that stop this code from being merged as we can go back and change it if necessary. The user probably won't notice and having a working compiler is more important. (I forgot how much time the Java version needs with downloading all the packages etc :sweat_smile:). I just tried it with the HelloWorldData.idl, but that shows some (probably minor issues).

* If I run the compiler on HelloWorldData.idl that's located in a subdirectory, in this case `hwd.1`, the include guard looks as follows `#ifndef _DDSL_HWD.1/HELLOWORLDDATA_H_`.
* Given the same situation, the source file tries to include the following file `hwd.1/HelloWorldData.h`. Given that it's located in the same directory I'd propose just including `HelloWorldData.h`.

I expected to have to change the `#pragma keylist` line, but much to my surprise it just worked, so kudos for that. I still have to do a thorough review (I'll continue right away), but I just wanted to ask how much of the features that the Java compiler supports are already in this compiler?

FransFaase:
Hi @k0ekk0ek. Thanks for reviewing. With respect to what is covered by the compiler. It is basically down to modules, structs, basic types, strings, sequence, and arrays. Unions, enumerate types, typedefs, and const expressions are missing (and maybe some more as well).
Embedded structs are supported. They @key annotation is also supported.

src/tools/idlc/src/gen_c99.c
```diff
@@ -0,0 +1,2063 @@
...
+  for (i = 0; i < file_name_len - 4; i++)
+    result[i] = (char)toupper(file_name[i]);
+  result[i] = '\0';
+  *ref_result = result;
+  return DDS_RETCODE_OK;
+}
+
+/* Generate copyright headere */
+
+static void write_copyright_header(ddsts_ostream_t *ostream, const char *target_file_name, const char *source_file_name)
```
k0ekk0ek: I propose to rename this function as it doesn't actually write any copyright information and it might give people a scare. That does remind me that we may have to include something that says code generated by the compiler is licensed under X or something? @eboasson, any thoughts?

src/tools/idlc/src/gen_c99.c
```diff
@@ -0,0 +1,2063 @@
...
+  ddsrt_ctime(time_now, time_descr, DDSRT_RFC3339STRLEN);
+
+  output(ostream,
+         "/****************************************************************\n"
+         "\n"
+         "  Generated by Eclipse Cyclone DDS IDL to C Translator\n"
+         "  File name: $T\n"
+         "  Source: $S\n"
+         "  Generated: $D\n"
+         "  Cyclone DDS: V0.1.0\n"
```
k0ekk0ek: Should be *Eclipse Cyclone DDS* instead of *Cyclone DDS*.

src/tools/idlc/src/gen_c99.c
eboasson: My personal view is that the output should not contain any copyright claim or license, it is only a mechanically transformed version of the input with a pretty straightforward and mostly standardized mapping. Moreover, we cannot know what the input copyright/license is, and therefore a copyright/license statement in the output would be problematic at best.

FransFaase: I change the name into write_file_header. The function used to print some copyright message, but it already did not do that anymore.

src/tools/idlc/src/gen_c99.c
FransFaase: I will change this.

FransFaase:
With respect to the remarks by @k0ekk0ek:

- I have added registered a log function. It does not print the file name, also because I suspect some changes when preprocessing is implemented and '#line' statements need to be parsed.
- I removed DDSTS_AT_TOKEN
- I renamed NOWS into NO_WS and removed it before #pragma. Maybe another lexer state is needed to enforce that #pragma is the first thing on the line (except some white space characters).
- Removed check on yylval->identifier. There used be some memory allocation in the line above. This allocation now has moved to idl.y to fix a problem with not freeing the allocated memory in case of an error.
- I am not in favour of changing YY_ABORT into YY_ERROR. I guess that many checks will need to be implemented in tt_create.c and that many test case need to be added to test all kind of recovery situations. (It is only rarely that I myself look beyond the first syntax error.)
- I fixed the problem with hexadecimal_number
- I tried to place yyerror in the %requires section, but at the place where it appeared in the header, YYLTYPE was not defined yet.

Travis fails at this moment for OSX. Something goes wrong with installing bison 3.4 (or using). Because of this bison 2.7 is used, which returns (as expected) an error.

k0ekk0ek:
Hi @FransFaase! Did you try incorporating the switch to Xcode that @eboasson made some days ago? There were some problems with macOS image being outdated. Also, did you fix the issues with the defines that I mentioned in my first comment? As for your comment on the `%requires` section: `%code provides` is the probably the place to do it then as per the [Bison manual](https://www.gnu.org/software/bison/manual/html_node/_0025code-Summary.html#g_t_0025code-Summary).

FransFaase:
Hi @k0ekk0ek, I have fixed the build problems by using the Xcode 10_2. I spend some time cleaning up the idl.l, idl.y, and parser.c files with respect to double or unneccessary declarations.

I have not fixed the problem with respect to running the compiler on IDL files in subdirectories. There is probably also a problem on using file names with non ANSI characters. Because the rules for characters in file names and  some other aspects are operating system dependend, It is possible that some of these problems will be solved by the preprocessor. Otherwise, it might be an idea to add some functions to the abstraction layer with respect to splitting file paths in parts and 'normalizing' file names for use in #define or #include. There was recently a user use case where one and the same file name was used in different directories and where these files could be included together. For this case, it might be needed to include the path in the #define.

k0ekk0ek:
@eboasson, would it be possible for you to restart some of the Travis CI builds? Maven failed to download some dependencies :weary:. Just shows how much we need the replacement :smile:.

@FransFaase, I'll have a look this evening. As for the include file problem, I guess I can live with it for now.

eboasson:
Yay! The jobs were not only restarted successfully, but also completed
successfully.

I agree that it would be unwise to spend forever trying to get the file
name mapping right for all possible edge cases before getting the core
functions to work. Some care obviously is worth it, but not to the
exclusion of progress on the other aspects.

On 26 Jun 2019, at 13:55, Jeroen Koekkoek <notifications@github.com> wrote:

@eboasson <https://github.com/eboasson>, would it be possible for you to
restart some of the Travis CI builds? Maven failed to download some
dependencies 😩. Just shows how much we need the replacement 😄.

@FransFaase <https://github.com/FransFaase>, I'll have a look this evening.
As for the include file problem, I guess I can live with it for now.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/eclipse-cyclonedds/cyclonedds/pull/200?email_source=notifications&email_token=AEBSPRMFR5IZX4OXCUTTB6TP4NKLNA5CNFSM4HYXE76KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODYTIZQY#issuecomment-505842883>,
or mute the thread

k0ekk0ek:
@FransFaase, luckily I had your old branch laying around locally so I could easily check the difference. In the future, please don't force push after having changed 3k lines. It takes a while to go through all of it again... Having pointed that out, this pull request looks good enough to merge into the *xtypes* branch. @eboasson, if you don't have any remarks, please be so kind as to merge the changes.

PR #202, opened 2019-06-19
Add PacketCapture for received messages in UDP

joxoby:


joxoby:
The error in the build seems to be code unrelated

k0ekk0ek:
@joxoby you're (probably) right as the `virtualenv` package can't be found on the macOS images. It'll take some tinkering to fix issue... I'll have a look to see what's causing this.

eboasson:
Hi @joxoby , the macOS build failed because of homebrew suddenly considering the Xcode 10.1 image of travis too old — I've solved it the easy way by switching to the Xcode 10.2 image that now exists. I don't see why there would be an issue on macOS with your change, and indeed it builds fine on my laptop.

So, it is my pleasure to merge it. Thank you!

Issue #213, opened 2019-07-05
dds_instance_get_key() returns  DDS_RETCODE_ILLEGAL_OPERATION when applied on readconditions or queryconditions


TheFixer:
Hi,

I have been playing around with Cyclone a bit and stumbled on the following. 

One of the nice things is that dds_read() and dds_take() can be applied to a readcondition or querycondition. For example, you can write something like
`
samplecount = dds_read (rc, samples, info, MAX_SAMPLES, MAX_SAMPLES);
`
where rc is readcondition. If there is data matching the condition specified in the readcondition, then this data will be returned in the samples list, and info list contains the sampleinfo of the data that is retrieved. For example, the handle tto the instance of the the j-th sample can be retrieved as
info[j].instance_handle.

To get the key of the handle you can use dds_instance_get_key(), but unfortunately this function requires a reader. I could add extra code to retrieve the reader, but this is a bit silly since I have the readcondition, and a readcondition is already tied to the reader.  I therefore would have expect that I could do somehting like
`
ret = dds_instance_get_key(rc, ...)
`
but this returns DDS_RETCODE_ILLEGAL_OPERATION.

Although this is not a bug, I believe it is more convenient and user friendly to allow dds_instance_get_key() to use a readcondition instead of requiring the reader. It is safe to do this because a readcondition is tied to a single reader anyway. The same holds for querycondition.

Allowing this would make the usage of dds_instance_get_key() a bit more user friendly in the context of readcondition and querycondition, and would lead to slightly less application code. 

In an attempt to see whether this indeed works I slightly changed the code and managed to get this working with very limited effort. As I did the code change already I am preparing a pull request, so you can decide yourself whether or not this is a useful contribution.

eboasson:
Implemented by #214

Issue #212, opened 2019-07-03
Intermittent ddsrt_select_timeout failure on macos


martinbremmer:
I noticed that ddsrt_select_timeout fails sometimes on macos on Travis. I can't remember seeing it on another platform.
I tried reproducing it locally (ubuntu 16.04), but was unable to.

k0ekk0ek:
Hi @martinbremmer. Saw the same failure on macOS some days ago. I don't currently have a macOS machine at my disposal (I seem to recall the ADLINK office has one of those nice little Mac minis though :wink:), but I'll try to find out what causes the test to fail. Thanks for the report!

martinbremmer:
@k0ekk0ek, ah yes. I forgot about that macmini.

I've cloned today's master onto it and ran the ddsrt_select_timeout test 5101 times. It didn't fail once.

(rhetorical) Is it a timing issue on Travis or has it magically been fixed in one of the latest commits? I don't know. So, I've tried the master of about 1 month old (aa2715f4fe2e6bebfab9936ec5d369a4becd74d4) and ran the test a thousand times. Again, no failure.

eboasson:
I strongly suspect it to be a timeout that's set just a little bit too short — the intermittent failures started some time ago without any change to the code, and besides I never have any problem locally on my laptop, &c. — but I haven't spent the time needed to prove that it must be either that or a bug in macOS ...

eboasson:
I haven't seen this particular failure anymore after my tweaking the timeout a bit. Given the observed failure frequency, I am quite sure I would have seen it at some point given the pretty large numbers of builds done since. Therefore, I think it can safely be considered "fixed".

PR #214, opened 2019-07-05
#213 - Allow readconditions and queryconditions as valid entities for…

TheFixer:
This pull requests adds the ability to use dds_instance_get_key() on readconditions and queryconditions. This is just for convenience and may lead to slightly less application code. Without this pull request dds_instance_get_key() can only be used on readers.

The pull requests consists of:

- A small modification in the logic to dds_instance_get_key() to handle readconditions and queryconditions
- A change in the description of this function to describe this new functionality
- Test cases that demonstrate that dds_instance_get_key() can be applied on readconditions and queryconditions.

Signed-off-by: TheFixer <thefixer@iteazz.com>

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -2939,7 +2939,7 @@ dds_instance_lookup(dds_entity_t entity, const void *data);
 /**
  * @brief This operation takes an instance handle and return a key-value corresponding to it.
  *
- * @param[in]  entity Reader or writer entity.
+ * @param[in]  entity Reader, writer, readcondition or querycondition entity.
  * @param[in]  inst   Instance handle.
  * @param[out] data   pointer to an instance, to which the key ID corresponding to the instance handle will be
  *    returned, the sample in the instance should be ignored.
--- a/src/core/ddsc/src/dds__types.h
+++ b/src/core/ddsc/src/dds__types.h
@@ -173,8 +173,9 @@ dds_return_t dds_entity_deriver_dummy_delete (struct dds_entity *e);
 dds_return_t dds_entity_deriver_dummy_set_qos (struct dds_entity *e, const dds_qos_t *qos, bool enabled);
 dds_return_t dds_entity_deriver_dummy_validate_status (uint32_t mask);
 
+
 inline dds_return_t dds_entity_deriver_close (struct dds_entity *e) {
-  return dds_entity_deriver_table[e->m_kind]->close (e);
+  return (dds_entity_deriver_table[e->m_kind]->close) (e);
 }
 inline dds_return_t dds_entity_deriver_delete (struct dds_entity *e) {
   return dds_entity_deriver_table[e->m_kind]->delete (e);
--- a/src/core/ddsc/src/dds__whc.h
+++ b/src/core/ddsc/src/dds__whc.h
@@ -18,7 +18,7 @@
 extern "C" {
 #endif
 
-struct whc *whc_new (int is_transient_local, unsigned hdepth, unsigned tldepth);
+struct whc *whc_new (int is_transient_local, uint32_t hdepth, uint32_t tldepth);
 
 #if defined (__cplusplus)
 }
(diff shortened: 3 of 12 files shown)


eboasson:
I totally agree with the extension of the interface — instance handles are valid in the context of a specific reader/writer, and a read or query condition specifies that context just as much as the reader itself does.

The change in the implementation is also fine, no comments there.

However, I do have a small problem with your test (and I'm not talking about having added calls to ``dds_delete`` for some but not all of the entities you added: ``dds_delete`` always deletes the entity and all its children anyway).

The issue I have is that you get the instance handle by calling "register_instance", then assume it is also valid in the context of the reader. I don't agree with that assumption, and am of the opinion that the test's succeeding is a bug. This despite having designed this bit of Cyclone to assign the same instance handle to the same key value (even across topics if the topics have the same representation).

I think you would have to actually write a sample so that the instance ends up in the reader. You wouldn't have to read it: you know the instance handle is the same anyway (thanks to the way instance handles are assigned).

There is little substance behind that opinion; the only I know of is that it allows testing whether the instance exists without altering the state of the reader/writer. Whether that is enough is debatable. Alternative opinions welcome.

(The PR build issue was just the intermittent troublemaker "select_timeout" on macOS Debug (#212) and also ppuserdata that ping-pongs between processes using QoS changes. I suspect in both cases the intermittency is caused by problems in the test rather than in what they test; in any case, they're completely unrelated to your PR.)

TheFixer:
Thanks @eboasson for the remarks. I fully agree with your remark that it is invalid to assume that the handle obtained by calling register_instance() can be used in the reader context, and I am happy to fix this.

I have been struggling a bit with your remark " You wouldn't have to read it: you know the instance handle is the same anyway". It seems to suggest that it is possible to retrieve the instance handle to the instance holding the data that is valid in the context of a reader without ever doing a read or take on the reader.  Is this correct, and how would you do that? I have been playing around with dds_get_instance_handle() but this does not seem to give me a handle to the instance holding the data.

eboasson:
> I have been struggling a bit with your remark " You wouldn't have to read it: you know the instance handle is the same anyway". It seems to suggest that it is possible to retrieve the instance handle to the instance holding the data that is valid in the context of a reader without ever doing a read or take on the reader.

It is not that you can _retrieve_ the handle, it is that you _know_ the handle, and (absent complications like resource limits and latencies) can conclude from a successful write of a sample for an instance with instance handle H that all matching readers will have instance H in their reader history cache.

You _know_ the handle, because by design there is a single global map from (f(T),key) to instance handle (in ddsi_tkmap.c). Here T is the topic, and f(T) is such that for two topics U and V from different sertopic/serdata implementations, f(U) ≠ f(V). For U and V from the same sertopic/serdata implementation, it depends on the implementation. Currently there are three implementations in existence:

1. The "default" one (ddsi_ser{topic,data}_default.c), which doesn't care about the topic whatsoever, and considers just the key value. The point of this is to help with managing different topics that are related through an equi-join — you know they are mapped to the same instance handle. (In its implementation it simply compares the serialised form of the keys, so it can consider topics with wildly different key types to have the same key value. I don't think that should be a problem, but it can be changed if needed.)
2. The one for built-in topics. The built-in topics are represented completely differently from the normal ones (they contain QoS objects for one), and moreover they are materialised only in the readers, but not in the respective writer history caches, as a trick to reduce memory use if there are no readers for them. All (current) built-in topics use a GUID as key, so even though the implementation ignores the topic, each instance is unique anyway.
3. In https://github.com/atolab/rmw_cyclonedds, there is one specifically for dealing with ROS2's internal memory formats. ROS2 doesn't do key values, so it gets away with ignoring this detail 😁.

(Just as an example: GUIDs are ultimately arrays of 16 bytes. If a builtin-topic for GUID a,b,c,d,...,p has instance handle H, then a sample from a default topic with an array of 16 bytes for its key with a value of a,b,c,d,...,p will have an instance handle J≠H; and if there is then a sample with another default topic, &c., it will have that sample handle J.)

Note that once the last sample/instance with a particular instance handle is removed, the table entry is removed and the next time that instance's key value shows up, it gets a new instance handle. One that, furthermore, is eternally unique.

And since I'm writing this anyway ... note also that the instance handles are for all practical purposes pretty much a number drawn from a uniform random generator without reuse. Thus, the identity function on instance handles is a sensible choice for a hash function. That should make life even easier from managing such an equi-join.

Does that answer your question?

TheFixer:
@eboasson Yes, that does answer my question. Based on your remarks I prepared a new pull request for this issue. Thanks.

PR #216, opened 2019-07-05
Fixed failing FreeRTOS target due to recent code refactors

ThijsSassen:
When trying to compile FreeRTOS I ran into errors. After inspection it turns out that the recent code refactor caused this. 

One of the errors was
`include/lwip/sockets.h:556:51: error: 'const struct dds_entity_deriver' has no member named 'lwip_close'; did you mean 'close'?
 #define close(s)                                  lwip_close(s)
                                                   ^
src/core/ddsc/src/dds__types.h:178:47: note: in expansion of macro 'close'
   return dds_entity_deriver_table[e->m_kind]->close (e);`

As the lwip header redefines close and I cannot control this so I renamed the close to fnclose.
The rest of the changes are pretty straight forward.

### DIFF
--- a/src/core/ddsc/src/dds__types.h
+++ b/src/core/ddsc/src/dds__types.h
@@ -173,8 +173,9 @@ dds_return_t dds_entity_deriver_dummy_delete (struct dds_entity *e);
 dds_return_t dds_entity_deriver_dummy_set_qos (struct dds_entity *e, const dds_qos_t *qos, bool enabled);
 dds_return_t dds_entity_deriver_dummy_validate_status (uint32_t mask);
 
+
 inline dds_return_t dds_entity_deriver_close (struct dds_entity *e) {
-  return dds_entity_deriver_table[e->m_kind]->close (e);
+  return (dds_entity_deriver_table[e->m_kind]->close) (e);
 }
 inline dds_return_t dds_entity_deriver_delete (struct dds_entity *e) {
   return dds_entity_deriver_table[e->m_kind]->delete (e);
--- a/src/core/ddsc/src/dds__whc.h
+++ b/src/core/ddsc/src/dds__whc.h
@@ -18,7 +18,7 @@
 extern "C" {
 #endif
 
-struct whc *whc_new (int is_transient_local, unsigned hdepth, unsigned tldepth);
+struct whc *whc_new (int is_transient_local, uint32_t hdepth, uint32_t tldepth);
 
 #if defined (__cplusplus)
 }
--- a/src/core/ddsi/include/dds/ddsi/q_whc.h
+++ b/src/core/ddsi/include/dds/ddsi/q_whc.h
@@ -73,8 +73,8 @@ typedef void (*whc_free_t)(struct whc *whc);
 /* max_drop_seq must go soon, it's way too ugly. */
 /* plist may be NULL or ddsrt_malloc'd, WHC takes ownership of plist */
 typedef int (*whc_insert_t)(struct whc *whc, seqno_t max_drop_seq, seqno_t seq, struct nn_plist *plist, struct ddsi_serdata *serdata, struct ddsi_tkmap_instance *tk);
-typedef unsigned (*whc_downgrade_to_volatile_t)(struct whc *whc, struct whc_state *st);
-typedef unsigned (*whc_remove_acked_messages_t)(struct whc *whc, seqno_t max_drop_seq, struct whc_state *whcst, struct whc_node **deferred_free_list);
+typedef uint32_t (*whc_downgrade_to_volatile_t)(struct whc *whc, struct whc_state *st);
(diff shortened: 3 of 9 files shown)


eboasson:
Thanks @ThijsSassen, that shows the importance of always building all platforms ...

I am happy with "the rest of the changes" but I am not happy simply renaming "close" to "fnclose" and calling it a day. At the very least, all functions in that struct should use the same name, but I what I really find objectionable is to have to change a name because lwip is been nasty enough to define it as a macro.

Parenthesising "dds_entity_deriver_table[e->m_kind]->close" would still be preferable to renaming it (and will stop the preprocessor from expanding the lwip macro).

But ultimately, before the code code gets changed, I want to see if there is a real need for having that macro in scope — there is a good chance the include files can be restructured so that include/lwip/sockets.h is not pulled into the core code. That would be correct fix.

ThijsSassen:
@eboasson I adjusted to code to match your comments.
I looked at the include structure but could not see a proper way to avoid it.
For now I parenthesized the close call as what I should call a temporarily work around.
The real fix should be to create a proper abstraction for the BSD socket API calls as we now ran into a problem with the close call but this can also happen for the others like read/write.
Maybe @k0ekk0ek  has some good ideas on how to do this as he was the one to add the freertos abstraction. 
I saw in the [esp-idf](https://github.com/espressif/esp-idf) project who had the [same ](https://github.com/espressif/esp-idf/issues/273) problem that they solved it with a [vfs layer](https://github.com/espressif/esp-idf/tree/master/components/vfs)

k0ekk0ek:
@ThijsSassen, @eboasson, the `lwip/sockets.h` header is required because it contains all kinds of type definitions needed by the core. e.g. `struct sockaddr_in`.

I don't agree with the statement: "The real fix should be to create a proper abstraction for the BSD socket API calls...". There's a proper abstraction in place with `ddsrt_socket_t` and the available set of socket functions within `ddsrt`. Introducing abstractions for well-known types like `struct sockaddr_in` doesn't sound like the best plan... As for the vfs layer approach, from what I gathered by browsing the comments, they're trying to solve a different problem.

The problem is that `close` and others are macros defined by the "platform". In this case, lwIP. The fix, in my opinion, would simply be to require the platform to not define `close` as a macro. lwIP has the options `LWIP_POSIX_SOCKETS_IO_NAMES` and `LWIP_COMPAT_SOCKETS` available for that. I would require those to be *off*. Then in the implementation, when LWIP_SOCKET is defined, use the `lwip_` prefixed functions.

Of course, for now the problem is solved by simply adding parentheses...

eboasson:
@k0ekk0ek and @ThijsSassen, I have a slightly different perspective than either of you, it appears. The core code in principle deals with "locators" and "connections" — the first is simply what the protocol calls it, the second is a bit of a misnomer because of the typical use of a connectionless transport.

The transport code implementation uses the sockets and abstracts them — or should abstract them — in a way that matches the particular use of the network by the core, i.e., it is not a one-on-one mapping of sockets to some socket-like interface, but a translation from a (mostly) datagram-oriented core with various needs and behaviours to an abstract networkt. In the case of UDP this happens to be a very straightforward mapping to sockets. For, e.g., TCP that mapping is more complicated, and that it happens to involve sockets and IP addresses is (should be) irrelevant to the core. (Not that the TCP code doesn't require a rewrite for other reasons ...)

There are still some remnants where the core code deals with IP addresses directly, or at least has a bit more knowledge of them than it ought to have. There are perhaps also some places where it looks behind the transport abstraction and suddenly deals with sockets. It's come quite a long way already, but these last steps also need to be taken. That's regardless of this particular problem. Taking those steps unfortunately takes time.

If disabling ``LWIP_POSIX_SOCKETS_IO_NAMES`` and ``LWIP_COMPAT_SOCKETS`` is very little work, I would much prefer that as a step in the right direction. Would it be possible to have a look at the impact of that first?

k0ekk0ek:
@eboasson. Typical case of miscommunication(?). I fully agree that it should be "abstracted" better in the sense that it should be hidden from the "core" wrt to `ddsi`. I was referring to the way it's "abstracted" in `ddsrt`.

Setting `LWIP_COMPAT_SOCKETS` in `lwipopts.h` is a proper way of not enabling the defines wrt to `ddsrt`. That being said, socket specific code should indeed only be needed in the UDP, TCP and SSL transports.

k0ekk0ek:
Actually the proper solution would be a combination of what @eboasson suggested and simply not depending on `LWIP_COMPAT_SOCKETS` in `posix/socket.c`. The reason being that users might depend on the POSIX socket calls in their application. Since this pull request, apart from the workaround, fixes a number of other things. I'd propose to merge and create a number of *enhancement* issues to:

1. Better isolate the networking code in `ddsi`,
2. not using POSIX socket calls in posix/socket.c (simply define a macro that forwards to `lwip_*`) and
3. have a Xilinx SDK build on Travis CI.

Of course, any help in those areas would be appreciated @ThijsSassen ;)

PR #217, opened 2019-07-06
Vintage edition: Solaris 2.6 support

eboasson:
Subject says all 😎

(and a fix for two or three minor things discovered along the way)


k0ekk0ek:
Haven't looked at the port code for Solaris 2.6 (and I'm not going too), but I agree with the rest of the changes. Feel free to merge after the conflicts have been resolved.

eboasson:
> Haven't looked at the port code for Solaris 2.6 (and I'm not going too), but I agree with the rest of the changes.

Sensible, but thanks for having look at the non Solaris 2.6 specific changes 😎

PR #222, opened 2019-07-16
Improve multicast defaults

eboasson:
This PR changes two defaults related to the use of multicast:

* It checks the kernel's information on the network interface to determine whether the chosen interface is a WiFi or any other type. On a WiFi network the use of multicast is (by default) restricted to participant discovery only.
WiFi pretends to be like an Ethernet, but the reliability of multicast is dramatically lower. As far as I know, pretty much any interference will cause packet loss, and often the loss comes in bursts. Experience is that behaviour on a WiFi is much, much better when avoiding multicast except for that one purpose.
* The second is to not assume that a retransmit request is likely to be followed in very short order by retransmit requests from other nodes, and that therefore the retransmits should not be multicast by default.

I've managed to check the WiFi/wired detection on macOS and Linux (though only with straightforwad network configuration). The code it uses on Windows seems like it ought to be correct, but it would be much appreciated if someone with a Windows box could give it a try.

The other commit is just a slight lenthening of a timeout in a test and the removal of a fixed port number. Those are the only two obvious things that may have caused the intermittent failure of the "select timeout" test (#212).

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_config.h
+++ b/src/core/ddsi/include/dds/ddsi/q_config.h
@@ -184,7 +184,7 @@ struct prune_deleted_ppant {
   int enforce_delay;
 };
 
-/* allow multicast bits: */
+/* allow multicast bits (default depends on network type): */
 #define AMC_FALSE 0u
 #define AMC_SPDP 1u
 #define AMC_ASM 2u
@@ -194,6 +194,7 @@ struct prune_deleted_ppant {
 #else
 #define AMC_TRUE (AMC_SPDP | AMC_ASM)
 #endif
+#define AMC_DEFAULT 0x80000000u
 
 /* FIXME: this should be fully dynamic ... but this is easier for a quick hack */
 enum transport_selector {
--- a/src/core/ddsi/include/dds/ddsi/q_nwif.h
+++ b/src/core/ddsi/include/dds/ddsi/q_nwif.h
@@ -28,6 +28,7 @@ struct nn_interface {
   nn_locator_t netmask;
   uint32_t if_index;
   unsigned mc_capable: 1;
+  unsigned mc_flaky: 1;
   unsigned point_to_point: 1;
   char *name;
 };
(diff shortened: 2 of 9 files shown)


rotu:
What happens if the publisher is not WiFi but the recipient is? I think you'll have the same performance degradation.

eboasson:
> What happens if the publisher is not WiFi but the recipient is? I think you'll have the same performance degradation.

That should actually be ok: the protocol works by advertising addresses that it is willing to receive data on (for the WiFi ones that would just be unicast) and by selecting the ones it is willing to send data to (for the WiFi ones, gain just unicast). Thus, the network should end up relying only on unicast for communicating with the WiFi ones.

It is a good point, though: by default the multicast address used for data is the same as for discovery. So, the WiFi nodes will get some packets via multicast, and it would be wise to verify that those do not trigger retransmit requests.

The other edge case is if there are nodes that are connected over Ethernet to a WiFi device that integrates into the larger network. That seems rather unusual, and so requiring a configuration file to overrule the default seems reasonable to me.

rotu:
Great! I'm glad you understand this, because I don't :-). I think it would be nice to detect questionable network characteristics, like long round-trip-times on a per-client basis.

eboasson:
> I think it would be nice to detect questionable network characteristics, like long round-trip-times on a per-client basis.

Absolutely. Detecting and gracefully handling can do with some TLC ...

eboasson:
I did a few more checks:
* On Windows, too, detecting WiFi/not-WiFi works as expected;
* A mix of AllowMulticast=spdp and AllowMulticast=true works fine with the limited equipment I have at hand;
* A Linux VM on VMware on macOS using bridged networking will show the WiFi as a wired network, so that results in the "wrong" setting (but with the other WiFi nodes all set correctly, the system still behaves as desired);
* A Linux VM on VMware on macOS using NAT networking will mess up things quite significantly becasue its (discovery) multicasts make it into the real world, but its advertised unicast address is unreachable ...

It looks to me like changing the defaults in this way does constitute an improvement. It doesn't change the need for improvements in the handling of problems, but I think that should be considered another problem.

@k0ekk0ek : are you ok with my changes?

k0ekk0ek:
@eboasson, no real comments. Personally, I would've checked for `[ '\t']` instead of using `isspace` as it's locale dependent and you check for `'\n'` in other places, but that doesn't matter in practice and would be nitpicking. Also, it might fail on Linux if an alias interface is used (I haven't tried)? But that too, would be nitpicking.

I would however just skip the first two header lines in the Linux code. I think currently you're only skipping one(?). Would make the `while` loop somewhat simpler too.

Maybe replace the `strncpy` by `ddsrt_strlcpy` for macOS?

eboasson:
> Personally, I would've checked for [ '\t'] instead of using isspace as it's locale dependent and you check for '\n' in other places, but that doesn't matter in practice and would be nitpicking.

Locale dependent stuff is nasty indeed — I'd better change it.

> I would however just skip the first two header lines in the Linux code. I think currently you're only skipping one(?). Would make the while loop somewhat simpler too.

It actually skips the first two lines: first it reads to the newline in state SKIP_HEADER_1, then it skips to the second newline in SKIP_TO_EOL, and only then does it start checking for names. Why Linux doesn't offer an easy way is a mystery to me ...

Perhaps I'll add a comment line :)

> Also, it might fail on Linux if an alias interface is used (I haven't tried)

I would assume the kernel would list aliases as well, but I too am inclined to ignore that detail. After all, it only affects the default setting.

> Maybe replace the strncpy by ddsrt_strlcpy for macOS

Might as well do that, too.

Issue #165, opened 2019-04-24
Five Pub(five Topic), and Five Sub;  It will lost some msg sometimes


Billlloves:
Hi! , I created two apps. A created a participant , 5 publishers(different topic for each one).  B created a participant, 5 sub(5 diff topics) . Sometimes not all of the subs can get the msg.  
What happened there?

eboasson:
Hi, that's precious little information, but there are two typical ways for running into this problem:
* One is to accidentally use best-effort instead of reliable communication: the default QoS setting for topics and for data readers is best-effort in the DDS spec — the setting for data writers is reliable, though! — and so it is quite possible that you accidentally created a best-effort data path. Actually I am very tempted to change the default in Cyclone to always default to reliable communications — I should make a ticket for that and see what the community thinks.
Typically, if the subscribers are fast enough to keep up with the writers, the amount of data loss with a best-effort setting is negligible with wired networks.
* The other is to have a reliability setting of KEEP_LAST with a depth of (typically) 1. This is also the default setting. With a KEEP_LAST history, newer data overwrites older data for the same key value, so if your reader doesn't keep up with your writer (for whatever reason) you'll lose some samples.
In this, "not keeping up" could also be a writer that publishes a few samples rather quickly and the reader reads only once per second.

I'd check those two things first, if I were you.

Billlloves:
Hi, thanks for giving me those two things. 
The implement is like this：
![image](https://user-images.githubusercontent.com/39365642/56703943-d4a8a980-673d-11e9-9dd6-3fe1fd40d172.png)

After they fond  each other, each pub only send one msg to the sub. Sometimes only two subs can get the msg, but all the pubs had sent the msgs.
ps: All of them are using default Qos.

eboasson:
@Billlloves, in my reading of the specification, there is no full handshake between a reader and a writer and for volatile data you're therefore only guaranteed to receive the data once (1) the writer has matched the reader and (2) the reader has matched the writer. Until both are the case, the data may still be dropped.

In practice, you can fairly easily account for (1) in the publishing process, but (2) is a rather troublesome thing because it is in a different process. This is not a problem DDS — and so Cyclone — should dump on the application.

For example if you would do:
```
wr = dds_create_writer
wait until there is a matching reader
dds_write(wr, ...)
```
you'd expect that matching reader to receive that one sample. However, if that reader has not yet discovered the writer, it can still drop the sample.

[...]

PR #223, opened 2019-07-19
Multi domain support

eboasson:
@k0ekk0ek, I have a little welcome gift for you! "Some minor changes", one might say.

This PR allows multiple domains to be used simultaneously (so having a single process with some participants in domain A while also having some in domain B, &c.). The big change is that where the code used to rely on global variables for various parts of its state — which entities exist, what sockets are being used, what the configuration settings are, etc. — they are now all derived (ultimately anyway) from references embedded in the entities.

Currently this leads to some more derefences in the hot paths (in particular also for the tracing) and so work is still to be done to recover the (minor) performance loss. There are also various places where the changes were in a more-or-less mechanical fashion and where a more elegant technique would be possible, too. It is my intent to address these issues in later pull requests, simply because maintaining a change this intrusive while other work changes make there way in is _very_ time-consuming.

Testing has been limited: both as to the exercising of this branch in exotic configurations, and as to the actual multi-domain support, there are only two tests that exercise it. One is a trivial one that creates two participants in different domains and deletes them. The other takes the multi-process test for verifying that changes to the TOPIC_DATA QoS on a topic propagate correctly to built-in topics of the readers and writers across the network, and runs two of them in parallel in different domains. That exercises a fair amount of the machinery.

(There are also some unrelated things in the PR, see below.)

**Some details**

Participant creation/destruction drives domain creation/destruction, which in turn drives starting/stopping the DDSI stack. The separation of the two layers has been maintained (including the somewhat ugly names like "q_globals"), with the former "gv" variable now having become embedded in the dds_domain object, and the former "config" object now embedded in "gv". The various DDSC entities refer to the dds_domain one; the various DDSI ones refer to the "gv" object embedded in that domain object.

Logging/tracing has been split: the global mechanism is still there and used for errors/warnings during domain creation, but there is now an additional interface (DDS_CLOG) that takes a logging configuration as an additional argument. Very nearly all the old DDS_LOGs and DDS_TRACEs have been replaced, with the logging configuration coming from the domain configuration. As a large fraction of those traces have one of a handful of variables in scope, there are special variants to reduce the number of (sometimes long) dereference chains in the source.

The thread table is still global, but the garbage collector restricts itself to looking at those threads that are part of the domain that the GC services. (It would be straightforward to share the GC between the domains, but reducing the separation between them may or may not be wise.) To do so and to deal with application threads that use objects from different domains at various times in their life, marking a thread awake now also sets the pointer to the "gv" object (because the GC is part of the lower layer, it is the "gv" pointer rather than the pointer to the "dds_domain" object).

The thread liveliness monitor hasn't yet been changed to look only at the domain it is part of — I was lazy ... I actually first changed it to be global, then moved it back to a per-domain thing when I realised that otherwise I would not have access to the logging configuration for the domain ...

All transport stuff (tcp, udp, raweth) is now also completely free of global variables, and store the "gv" pointers internally in the connections. This incidentally also makes it much easier to improve the support for simultaneous use of different transports, or to support IPv4 and IPv6 in parallel a lot easier. TLS support is almost ok, but there is a problem in some of the TLS callbacks with not having the required pointers data available ... That's not an issue when used in a single-domain configuration.

The RHC implementation is now also abstracted properly, so there is no longer a global "plugin" thingummy with pointers to functions that the DDSI stack needs. Instead, an RHC implementation simply refers to a vtable. The burden of the indirect calls is at least at source level removed by having some additional (inline) functions.

**Unrelated things**

This PR also fixes a number of unrelated bugs. Those can be split out and moved into a separate PR, but most of the issues were discovered — or more precisely, analyzed, because they were triggered by flakiness in tests — and fixed after the major changes went in. So splitting it out involves an significant amount of additional work.

* MPT QoS tests were flaky in part because of a too sloppy synchronization between the processes;
* They were also in part flaky because of a potential for stuttering in SEDP messages caused by processing historical SEDP samples after already having processed later ones. This was not an issue for entities that are merely created/deleted, but it is an issue in combination with QoS changes;
* There was a risk of built-in topic samples being generated from a QoS object while a change to that same object was underway. The current "fix" involves a global mutex, I'm still debating whether a single lock is better than a lock per entity or not. In any case, the standard entity lock cannot be used for this;
* The fix for #165 changed the way incoming samples are processed when some readers are out-of-sync. Since that fix, what was processed as if it were but a single sample could in reality be the first in a chain of samples. That occasionally (if very rarely) triggered an assert;
* Another really rare assertion in merging plist/qos was ultimately caused by modifying the default plist object for participants after starting the DDSI stack. The assertion triggered because not all fields had been copied ... As it happens, it set some defaults that shouldn't be set for remote participants, so that got addressed as well;
* The cause of the flaky ddsrt_select_timeout test seems to have been simpy not waiting long enough for the background thread to timeout. It now waits a bit longer, that seems to help.


k0ekk0ek:
I've started the review. It's going to take me *some* time, so it's going to be done in various phases. My first question would be: Why a logging configuration per-domain? What comes to mind is the ability to set different categories per-domain which is of course comes in handy for debugging...

Btw, I'm going through this per-commit, so rewriting history to keep the log clean is probably better done when the review is done :)

eboasson:
Hi @k0ekk0ek, I'm very happy that you're willing to even start on a review of this monster!

I fully agree with the plan to study this one commit at a time — it's one of the reasons I've done my best to make a reasonably clean commit history where each commit transitions from a working version to another working version (should be anyway, I may not have tried each and every one of them after all the shuffles, merges and splits). Rewriting the history now would turn reviewing it from a sheer impossbile task to a truly impossible one. I was planning on addressing issues in new commits and, once it is approved, clean up the history & verify that the code base at the end is identical.

> My first question would be: Why a logging configuration per-domain? What comes to mind is the ability to set different categories per-domain which is of course comes in handy for debugging...

Spot on: that's main reason. Especially when I consider some of the problems I have had to analyse over the years and the sizes of the detailed traces, it seems quite worthwhile to be able to set different levels of detail per domain.

Whether it is valuable to have multiple files is something I am not sure about, but that's a pretty trivial thing once you decide to have different logging configurations. Having a single file that collects the output for multiple domains is not yet properly supported, and seems eminently useful. There are other things still to be done in the configuration, like actually having different configs without having to reset CYCLONEDDS_URI.

k0ekk0ek:
Reviewed up to the "Abstract RHC" commit last night. Turns out, the logic hasn't changed and if you diff the original `dds_rhc.c` and `dds_rhc_default.c` the diff is far less intimidating. I like that it's actually properly abstracted now (including the read function etc) instead of a so-so "plugin" which wasn't really a plugin, so I fully agree with everything up to and including the second commit (the rest I'll deal with today). Two minor remarks:

1. There's a mention of `DDS_CLOG` in there already. That won't work, but seeing this is going to be one merge commit, I leave it up to you to decide if you want to go through the effort.
2. Maybe a mention in the source code of why the abstraction is a good idea would be helpful to other people?

k0ekk0ek:
I went over the complete diff again and did a build+test run. One extra thing I noticed:

`new_proxy_group` and `delete_proxy_group` implementations have been removed from `q_entity.c`. Probably need to remove them from `q_entity.h` as well? Maybe the struct definition can be removed entirely? Maybe proxypp_groups_treedef can be removed as well as it doesn't seem to be used anywhere at first glance?

And I got build errors and warnings compiling the `helloworld` example. `#include "ddsc/dds_public_impl.h"` should be `#include "dds/ddsc/dds_public_impl.h". Not sure if that was introduced by this PR, but maybe you can incorporate it while you're at it :)

```
[ 85%] Building C object examples/helloworld/CMakeFiles/HelloworldPublisher.dir/publisher.c.o
In file included from /home/k0ekk0ek/Projects/cyclonedds/src/examples/helloworld/publisher.c:2:
/home/k0ekk0ek/Projects/cyclonedds/src/examples/helloworld/HelloWorldData.h:11:10: fatal error: ddsc/dds_public_impl.h: No such file or directory
   11 | #include "ddsc/dds_public_impl.h"
```

After that I got the following warnings:

```
[ 87%] Building C object examples/helloworld/CMakeFiles/HelloworldSubscriber.dir/subscriber.c.o
In file included from /home/k0ekk0ek/Projects/cyclonedds/src/examples/helloworld/subscriber.c:2:
/home/k0ekk0ek/Projects/cyclonedds/src/examples/helloworld/HelloWorldData.h:13:15: warning: extra tokens at end of #ifndef directive
   13 | #ifndef _DDSL_../SRC/EXAMPLES/HELLOWORLD/HELLOWORLDDATA_H_
      |               ^
/home/k0ekk0ek/Projects/cyclonedds/src/examples/helloworld/HelloWorldData.h:14:9: warning: ISO C99 requires whitespace after the macro name
   14 | #define _DDSL_../SRC/EXAMPLES/HELLOWORLD/HELLOWORLDDATA_H_
      |         ^~~~~~
```

That concludes the review. Most of the changes are pretty straightforward and are simply the quickest route to multi-domain support. Some off it could use some polishing as you have indicated yourself too, but I agree with the fact that this should go in first because a changeset this large is hard to keep in-sync. I do think that logging should be changed somewhat and that it should be possible for a user to at least pass a configurtion filename from the program, preferebly it would also be possible to pass the configuration as a string, which would make it easier for users on embedded platforms. Curious to hear what you think about the latter two points though.

eboasson:
Thank you, @k0ekk0ek! I am well aware that an commit of this size is quite an effort to review. And as you have spotted a number of issues that I agree still need addressing, it was clearly not a wasted effort either.

I'll respond to all of your ocmments below:

> (Re abstracting the RHC:) There's a mention of DDS_CLOG in there already. That won't work, but seeing this is going to be one merge commit, I leave it up to you to decide if you want to go through the effort.

Although it isn't really that important to clean that up, I'll fix it anyway. It was actually my intent to just do rebase & merge (after resolving the conflicts) — but if you'd prefer a merge commit for such a massive one, that's fine with me, too.

> Maybe a mention in the source code of why the abstraction is a good idea would be helpful to other people?

Some note might indeed be sensible. It is pretty much a standard thing, but the way it could (and probably should) be exploited later on is a different story. That might be better done as part of a note in docs/dev, though.

> The configuration feels a little off. CYCLONEDDS_URI is still used, but you can't have multiple URIs, so it's actually not possible to configure something per-domain(?) I think that with this PR it must be possible to specify the configuration in a string, or (my preference), it should be possible to specify things per-domain in the configuration file itself (with a few exceptions)?

I fully agree that this bit isn't sorted yet. My reasoning was that it doesn't break anything in current use and so it need not necessarily be solved before merging.

It seems to me that one good step would be to extend the API with a function to explicit create a domain, and where the configuration is specified explicitly (i.e., call ``dds_create_domain`` with a config in a string parameter).

Having a single file with multiple configurations with a way of specifying what domain ids the configs apply to would be a good thing to. That will probably replacing the current Domain/Id element with an "id" attribute. Preferably in a backwards-compatible way ...

That API extension would be straightforward, the config file changes will be harder. So in my view it would make sense to do the former before merging, the latter later.

> The only thing I think should be implemented differently is the per-domain logging configuration. I agree that filtering on a per-domain base is very useful, but having separate files is something I don't like. What file will be logged to if you have two domains and don't specify it in DDS_LOG? I'd prefer being able to specify just a mask per domain that can be used in the ddsc code (and future API's) as well. Also, the domain id should be passed to the sink if available. In the past there were some users that wanted to have this functionality.

I think whether or not one wants separate files is really dependent on the situation, so I think it should be possible to do anything, and that it should be easy have a single log-file per domain where desired, with everything else going into a common one. Neither a single config nor the current state will do that ...

A custom sink implementation indeed takes care of allowing everything. I need to ponder this a bit. In any case, you would still need some per-domain state for logging configuration to pass in the id, and that means it won't be as simple as:

> Since the domain id must be a number between 0 and 230(?), how about just having a static array with 231 elements and allow setting a mask per domain. Then update the category to an uint64 and assign the higher bits to the log categories so that you can write DDS_LOG(<domain_id> | DDS_LC_X | DDS_LC_Y, ...)? Don't know how I'd do DDS_ERROR etc though.

Besides, the 230 limit really is the consequence of 16-bit port numbers in the default port mapping. The port mapping can be configured, and other transports might not have a 16-bit limitation, so it'd be better not to use it.

> (.gitignore) Just curious why this is added.

Thanks for catching that one! An ordinary mistake ... too many parallel build directories here, and a subdirectory of the repo means less typing. There was never any intention of making that part of the repo.

> ``gv`` is moved into the domain later on, but is it really necessary to have a threadmon thread per domain? I might be overlooking something here(?) Is it desirable to have an actual DDSI stack per-domain. I'm thinking it might be worthwhile to have only a couple of threads that service the ddsi stack, but know how to handle multiple domain?

I had the same thought (there is even a commit for moving it!), then it moved back because of the logging configurations. I'm sure there is a better way.

My view of multiple domains is that one generally does it because one wants to be independent, and in that case, I think it makes more sense to default to having a set of threads per domain. Small footprint systems would benefit from sharing the service threads, and it would perhaps be a good thing to allow sharing them. I'd prefer to view that as a possible enhancement in the future.

(Fixing "threadmon" is better done before merging it.)

> new_proxy_group and delete_proxy_group implementations have been removed from q_entity.c. Probably need to remove them from q_entity.h as well? Maybe the struct definition can be removed entirely? Maybe proxypp_groups_treedef can be removed as well as it doesn't seem to be used anywhere at first glance?

Thanks for spotting that. That's was completely dead code, but I should've removed those declarations as well.

> And I got build errors and warnings compiling the helloworld example. #include "ddsc/dds_public_impl.h" should be `#include "dds/ddsc/dds_public_impl.h". Not sure if that was introduced by this PR, but maybe you can incorporate it while you're at it :)

I had no intention of doing that here, but yes, might as well fix it here.

> That concludes the review. Most of the changes are pretty straightforward and are simply the quickest route to multi-domain support. Some of it could use some polishing as you have indicated yourself too, but I agree with the fact that this should go in first because a changeset this large is hard to keep in-sync. I do think that logging should be changed somewhat and that it should be possible for a user to at least pass a configurtion filename from the program, preferebly it would also be possible to pass the configuration as a string, which would make it easier for users on embedded platforms. Curious to hear what you think about the latter two points though.

That I, too, believe those points need some more attention before merging.

eboasson:
@k0ekk0ek, I’ve force-pushed my changes because of pulling out some completely independent commits and merging them as PR #231, and because of solving conflicts with master. Given that I was force-pushing anyway, I elected to do a small fixup in one commit (“eliminate domain-specific global variables”) that extends the changes also to code that is only used when SSM is unsupported.

Other than that, the commits that are still there are unchanged and the changes resulting from the review comments are all in new ones, as well as some other ones that I discovered along the way. I do intend to clean up the set of commits, but I hope this way it is still be possible to more-or-less continue the review instead of having to start over.

## Key changes

### Configuration

*The* change in this update is in the configuration and tracing. The configuration has been reworked once again (apologies to all who have to keep changing their configurations to get rid of the pesky deprecation notices), and now allows specifying multiple configurations in a single file, skipping configurations for non-matching domain ids. Multiple files was already a possibility, and a domain specification of “any” (still the default) will apply to regardless of the domain id.

The structural change is that the “Id” child-element of the “Domain” element has become an attribute of the same Domain element, and that moreover everything else now resides under the Domain element. So, where the old default basically was:
```XML
<CycloneDDS>
  <Domain>
    <Id>any</Id>
  </Domain>
  <General>
    <NetworkInterfaceAddress>auto</NetworkInterfaceAddress>
    <!-- etc -->
  </General>
  <!-- etc -->
</CycloneDDS>
```
it is now:
```XML
<CycloneDDS>
  <Domain id=“any”>
    <General>
      <NetworkInterfaceAddress>auto</NetworkInterfaceAddress>
      <!-- etc -->
    </General>
    <!-- etc -->
  </Domain>
</CycloneDDS>
```
a little bit of magic in the config file processing takes care of accepting the old one (even when it still uses the old “DDSI2E” element!) and generating reasonable warnings.

The rule for the domain id is now:
* if the application specifies anything other than ``DDS_DOMAIN_DEFAULT`` in ``dds_create_participant``, that’s what will be used, and any configuration specification for that domain id as well as any for domain id “any” will be processed, with later entries overwriting older ones;
* if, however, the application specifies ``DDS_DOMAIN_DEFAULT`` it will use the lowest-numbered domain already present in the process if such a thing exists, and otherwise the first numerical domain id it encounters and stick with that. If none is encountered, it ends up using 0.
It sounds more complicated than it is, I think. The “lowest-numbered” bit is a bit odd, but it avoids having to configure just to learn the domain id, then possibly discover it already exists. Using ``DDS_DOMAIN_DEFAULT`` in a multi-domain applications seems like a real edge-case to me.

And it is even backwards compatible :smile:.

Note that I do need to update the configurator tool before it can be merged it.

### Logging/tracing

The changes remove the separate log/trace sinks from the per-domain configuration. The domain now *only* specifies trace flags and a trace file, but what actually happens is ultimately determined by the sinks that the application can install itself. I had to really change ``src/ddsrt/log.c`` because of the different sources of tracing flags, and I removed some of the optimisations, instead determine a bit more for each log/trace message. The reason is that the global log settings affect the per-domain ones, and so either the global one would have to be cached (and kept in sync) in the domain, or the test done for each trace message would have to look at two sets of flags. I’d rather incur a slightly higher cost for outputting messages if it means a slightly lower cost for the many traces that don’t yield any output. (I probably should add ``__builtin_expect`` as well.)

Trace files are per-domain, because I couldn’t find a sufficiently simple and robust way of determining whether two configurations name the same file or not (you to have to allow for filesystem changes and for changing directory), and so one must still take care to use different trace files per domain with the default trace sink. If you want everything combined, then that should be possible — famous words, I haven’t actually tried it — using a custom sink.

To make it easier to specify keep log files separate, there are now two pseudo-environment variables (real ones take precedence): ``CYCLONEDDS_PID`` and ``CYCLONEDDS_DOMAIN_ID``. These have the obvious meaning. The only problematic case is using ``CYCLONEDDS_DOMAIN_ID`` before the domain id is set, as variable expansion is immediate on reading the element. Still:
```XML
<CycloneDDS>
  <Domain id=“any”>
    <Tracing>
      <OutputFile>cyclonedds.log${CYCLONEDDS_DOMAIN_ID:+.}${CYCLONEDDS_DOMAIN_ID}</OutputFile>
  </Domain>
</CycloneDDS>
```
would give a reasonably sane behaviour.

The domain id (if known) is also printed in log/trace messages. 

## Other changes

“Replace global QoS lock by per-entity QoS lock” is a bit out-of-place here, but it came about when I realised that QoS matching also has a problem. I am still undecided whether I like it enough this way to stick with it.

## Unaddressed review comments

> > (Re abstracting the RHC:) There's a mention of DDS_CLOG in there already. That won't work, but seeing this is going to be one merge commit, I leave it up to you to decide if you want to go through the effort.
> Although it isn't really that important to clean that up, I'll fix it anyway. It was actually my intent to just do rebase & merge (after resolving the conflicts) — but if you'd prefer a merge commit for such a massive one, that's fine with me, too.

I have not been able to find the problem — in fact, all commits compiled cleanly …

> > Maybe a mention in the source code of why the abstraction is a good idea would be helpful to other people?
> Some note might indeed be sensible. It is pretty much a standard thing, but the way it could (and probably should) be exploited later on is a different story. That might be better done as part of a note in docs/dev, though.

TBD (eventually …)

> > And I got build errors and warnings compiling the helloworld example. #include "ddsc/dds_public_impl.h" should be `#include "dds/ddsc/dds_public_impl.h". Not sure if that was introduced by this PR, but maybe you can incorporate it while you're at it :)
> I had no intention of doing that here, but yes, might as well fix it here.

You’ll have to help me here … I can’t find the problem.

## Commit info

### new, in response to review comments

* Support multiple domains in configuration
* Rework logging for multi-domain
* Use valid reference to globals for printing error
* Make domainId and port numbers uint32_t
* Build no-SSL version on CI as well
* Replace global QoS lock by per-entity QoS lock
* Include domain id in trace/log
* Use a single thread monitor for all domains

### (very nearly) unchanged since the review

* Add multi-domain version of topic_data test
* Use separate plist for local process settings
* Sync generation of built-in data with QoS changes
* Check sequence number of SEDP messages
* Make logging config per-domain
* Pacification of native compiler on OpenIndiana
* Take domain into account in GC, not in threadmon
* Allow coexisting domains and do a minimal test
* Eliminate domain-specific global variables
* Liveliness monitoring to dds_global
* Remove use of globals from config processing
* DDSI stack init/fini in domain_create/free
* Move deleted participants admin into "gv"
* Cleanup duplicate deinit code in UDP support
* Remove global temporaries from config parsing
* Remove unused "durability service" global variables
* Abstract RHC interface

### merged in PR #231

* Parallelize some more on Travis and Appveyor
* Reduce likelihood of ddsrt_select_timeout failure
* Assertion rsample_is_singleton fails
* Make qos MPTs more robust & parallelize tests
* C99 forbids flex array members in nested structs
* Ignore EPERM on sendmsg

k0ekk0ek:
@eboasson, I looked at this somewhat yesterday. The configuration file changes seem like the obvious way to go, so no comments there. The logging changes seem fine as well. We'll have to think of something for `ddsc` maybe and I have some reservations about offering the logging function in the `dds` namespace, but that's something we can worry about later. I'll check the other changes sometime later today.

k0ekk0ek:
@eboasson, I have no further comments. Feel free to merge.

ThijsSassen:
@eboasson I ran this branch for freertos and i failed with the following error:

In file included from /home/jenkins/workspace/CycloneDDS-FreeRTOS/src/core/ddsc/src/dds_reader.c:21:0:
/home/jenkins/workspace/CycloneDDS-FreeRTOS/src/core/ddsc/src/dds__rhc.h: In function 'dds_rhc_read':
/home/jenkins/workspace/CycloneDDS-FreeRTOS/src/core/ddsc/src/dds__rhc.h:73:93: error: macro "read" passed 8 arguments, but takes just 3
   return rhc->common.ops->read (rhc, lock, values, info_seq, max_samples, mask, handle, cond);

This looks again like a conflict with the hard coded read macro in the freertos header files.

eboasson:
@ThijsSassen, while I do best not to break things, I cannot help doing so for platforms that have such unreasonable header files and that I don't have a setup for that allows me to check. As it appears to be rather important to you, perhaps you could provide a build environment that we can add to the automated builds? That would be highly appreciated.

If you'd provide a minimal change that fixes it, that's fine with me (I suppose a pair of parentheses) will do. A better step would be to find out how the "ddsrt" includes could be restructured to eliminate these problems for once and for all. Such an effort, too, would be highly appreciated.

ThijsSassen:
@eboasson I created a patch [multi-domain-freertos.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3508487/multi-domain-freertos.txt) that makes the freertos build compile again.

Im working on getting a freertos build environment integrated into ddscyclone.

eboasson:
@k0ekk0ek: no need to worry that I may be trying to mess up your weekend: I've merely cleaned up the set of commits a bit so there is less of the do-undo-redo and fixup varieties. The diff of 9a54adec07e97004f2397345b99c524819d94587 with the old b290f6b95683c74718597cb29f52309bc8aab4d7 is empty and I've run them all through a build/test cycle on Linux to be sure the rebase doesn't introduce any problems (there is just ``CUnit_dds_log_only_log_file`` that fails for a few of the intermediate commits).

ThijsSassen:
@eboasson I saw you adopted the FreeRTOS work around but you did not change the DDS_TRACE call in sync.c which causes the build to still fail, any reason why you didn't?

eboasson:
@ThijsSassen, you mentioned a compilation problem caused by FreeRTOS defining a function-like macro named "read", so I assumed the other change had nothing to do with the compilation problem. And while I don't have a major issue with changing those ``TRACE``s to ``ERROR``s, I would like to understand why that change is necessary before I make it.

ThijsSassen:
@eboasson as DDS_TRACE no long exists I changed to to error as in both cases we are in an error state but after looking at it again I changed it to DDS_FATAL as an abort is performed after the message is logged so fatal would be more suitable.  see patch [here](https://github.com/eclipse-cyclonedds/cyclonedds/files/3514909/trace_fix.txt)

martinbremmer:
@eboasson, just FYI, dds.h still contains "DDS_EXPORT dds_domainid_t dds_domain_default (void);".

PR #224, opened 2019-07-29
Reorganize repository

eboasson:
## Warning

This PR changes the way you have to invoke CMake. The top-level CMakeLists.txt moved from ``src`` to the top-level directory. I have tried to make sure it gives a reasonable error message when you invoke it on ``src``.

It also changes the way it deals with OpenSSL. Firstly, option is renamed from ``DDSC_ENABLE_OPENSSL`` to ``ENABLE_SSL`` (the old one is still respected). Secondly, it will now simply include OpenSSL support if OpenSSL is found and leave it out if it isn't found. The option is now simply a means for disabling looking for OpenSSL.

## Changes

* Move the project top-level CMakeLists.txt to the root of the project (this also allows building Cyclone as part of ROS2 without any special tricks)

* Clean up the build options:

  * ENABLE_SSL:    whether to check for and include OpenSSL support if a library can be found (default = ON); this used to be called DDSC_ENABLE_OPENSSL, the old name is deprecated but still works
  * BUILD_DOCS:    whether to build docs (default = OFF)
  * BUILD_TESTING: whether to build test (default = OFF)

* Collect all documentation into top-level "docs" directory;

* Move the examples to the top-level directory;

* Remove the unused and somewhat misleading pseudo-default cyclonedds.xml;

* Remove unused cmake files


k0ekk0ek:
Finally, a sane layout! No comments. I take it you'll wait until the other PRs are merged?

eboasson:
No need to wait: it is very nearly independent of the other PRs 😎

CMakeLists.txt
```diff
@@ -0,0 +1,214 @@
...
+execute_process(COMMAND git submodule init
+                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+execute_process(COMMAND git submodule update
+                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+
+# Set module path before defining project so platform files will work.
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/Modules")
+set(CMAKE_PROJECT_NAME_FULL "Eclipse Cyclone DDS")
+set(PROJECT_NAME "CycloneDDS")
+project(${PROJECT_NAME} VERSION 0.1.0)
```
dirk-thomas: I would suggest to not use a variable here. You can just write `project(CycloneDDS VERSION 0.1.0)` and the CMake variable `PROJECT_NAME` will be set by CMake automatically.

What is the advantage of this? By looking into the CMake code and finding the `project()` call the project name can be extracted since it is a string literal (e.g. by tools like `colcon`).

dirk-thomas: I created #235 doing that.

PR #225, opened 2019-07-31
Generate documentation using the FindSphinx.cmake module

k0ekk0ek:
This PR uses the `FindSphinx.cmake` (and `doxygen_add_docs`) module that I added recently to build the documentation instead of using all kinds of custom commands and targets. And since we're cleaning up, it also removes a lot of cruft from the repository:
* make files and Visual Studio solution for the HelloWorld example (nobody maintains it anyway)
* pdf generation (better to have something up on [Read the Docs](https://readthedocs.org/))

In the future users can build the documentation with `cmake --build . --target docs` and the example documentation with `cmake --build . --target examples_docs`.

What's added by this is that CMake now simply throws an error if *Doxygen*, *Sphinx* or the *breathe module for Sphinx* is not installed.


eboasson:
It looks good, but there is a problem on Ubuntu 18.04:

```
erik@ubuntu:~/cyclonedds/CC$ cmake -DBUILD_DOCS=ON ..
-- Found Doxygen: /usr/bin/doxygen (found version "1.8.13") found components:  doxygen missing components:  dot
CMake Error at cmake/Modules/FindSphinx.cmake:88 (message):
  Versions for sphinx-build (1.6.7)and sphinx-quickstart ()do not match
  examples/CMakeLists.txt:19 (find_package)
```
and that is because ``quickstart --version`` outputs:
```
erik@ubuntu:~/cyclonedds/CC$ sphinx-quickstart --version
Sphinx v1.6.7
```
and the regex in FindSphinx.cmake doesn't allow for the "v". With:
```
diff --git a/cmake/Modules/FindSphinx.cmake b/cmake/Modules/FindSphinx.cmake
index b4296b5..0497489 100644
--- a/cmake/Modules/FindSphinx.cmake
+++ b/cmake/Modules/FindSphinx.cmake
@@ -26,7 +26,7 @@ macro(_Sphinx_find_executable _exe)
       RESULT_VARIABLE _result
       OUTPUT_VARIABLE _output
       OUTPUT_STRIP_TRAILING_WHITESPACE)
-    if(_result EQUAL 0 AND _output MATCHES " ([0-9]+\\.[0-9]+\\.[0-9]+)$")
+    if(_result EQUAL 0 AND _output MATCHES " v?([0-9]+\\.[0-9]+\\.[0-9]+)$")
       set(SPHINX_${_uc}_VERSION "${CMAKE_MATCH_1}")
     endif()
 
```
it works (even on macOS!) So feel free to merge it after making that change.

k0ekk0ek:
Fixed the bug. I think I've actually seen it before, but simply forgot to incorporate the fix. Anyway it's done now and I'll merge it once Travis is done. This works on all the platforms btw, even Windows if you install the modules using `pip`.

martinbremmer:
@k0ekk0ek, "also removes [...] make files and Visual Studio solution for the HelloWorld example".
I'd like to point out that these things are mentioned and used on various places in the "Getting Started Guide".

k0ekk0ek:
Thanks @martinbremmer! I've created an issue for now and will sort it out later.

PR #226, opened 2019-08-01
IDL-Compiler: adding unions to front-end

FransFaase:
Added support for unions to the grammar and the front-end. Due to limited support for values, only unsigned integers, the switch types are limited. Range checks have not been implemented yet. (Probably better to implement range checking when range checking for expressions is implemented.) Tests (both positive and negative) for the front-end have been added.


k0ekk0ek:
@FransFaase, I decided to merge in master into the xtypes branch first. Can you rebase please? I'll probably look at this PR sometime this weekend. Thanks!

FransFaase:
@k0ekk0ek, thanks for merging master into xtypes. I thought about asking for this, because it probably will solve the build problems on the Windows platforms on Travis. I will rebase and also fix a bug that I found while working on the support for unions in the C99 back-end.

FransFaase:
@k0ekk0ek, I have done the rebase. Extended the test parser/union with a test to verify that a union type can be used for a member of a struct (the above mentioned bug) and fixed the code.

k0ekk0ek:
@FransFaase, sorry I didn't get around to it over the weekend. I haven't finished the review yet, but I've attached an example of use for Bison that might be of some help for future PRs(?) It feels like the current parser is more-or-less a natural fit, but in order to be sure I'd have to dive a little deeper into it myself.

[shift-reduce.zip](https://github.com/eclipse-cyclonedds/cyclonedds/files/3468839/shift-reduce.zip)

src/tools/idlc/src/tt_create.c
```diff
@@ -548,6 +567,231 @@ extern void ddsts_struct_empty_close(ddsts_context_t *context, ddsts_type_t **re
...
+    case DDSTS_CHAR:           return "char"; break;
+    case DDSTS_WIDE_CHAR:      return "wide char"; break;
+    case DDSTS_BOOLEAN:        return "boolean"; break;
+    case DDSTS_OCTET:          return "octet"; break;
+    case DDSTS_INT8:           return "int8"; break;
+    case DDSTS_UINT8:          return "uint8"; break;
+    case DDSTS_FLOAT:          return "float"; break;
+    case DDSTS_DOUBLE:         return "double"; break;
+    case DDSTS_LONGDOUBLE:     return "long double"; break;
+    case DDSTS_FIXED_PT_CONST: return "fixed point const"; break;
```
k0ekk0ek: Why the `break`s if you have a `return`?

src/tools/idlc/src/tt_create.c
```diff
@@ -548,6 +567,231 @@ extern void ddsts_struct_empty_close(ddsts_context_t *context, ddsts_type_t **re
...
+  ddsts_literal_t casted_value;
+  dds_return_t rc = cast_value_to_type(value, context->cur_type->union_def.switch_type, &casted_value);
+  if (rc != DDS_RETCODE_OK) {
+    context->retcode = rc;
+    return false;
+  }
+  for (ddsts_type_t *cases = context->cur_type->union_def.cases.first; cases != NULL; cases = cases->type.next) {
+    for (ddsts_union_case_label_t *label = cases->union_case.labels; label != NULL; label = label->next) {
+      if (equal_values(&label->value, &casted_value)) {
+        DDS_ERROR("Same case label used in more than one case.\n");
```
k0ekk0ek: I don't see any mention of location in the source code, while `%locations` is enabled. I think that should be changed.

src/ddsts/include/dds/ddsts/typetree.h
```diff
@@ -72,7 +76,7 @@ typedef char *ddsts_identifier_t;
  * IDL elements, such as the constant definition and the case labels.
  */
 
-typedef uint32_t ddsts_flags_t;
+typedef uint64_t ddsts_flags_t;
```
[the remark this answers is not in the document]
FransFaase: Yes, I agree with you, that with respect to the types, it is not really used as a bitmask. (Note that it is a bitmask in combination with DDSTS_REFERENCE_1 and DDSTS_REFERENCE_2.) I see that some bits could be saved by, for example, by adding a flag for FORWARD and combine FORWARD_STRUCT and FORWARD_UNION as a combination of FORWARD and respectively STRUCT and UNION. Note that this would save just a single bit (add one to remove two). With the number of grammar elements still to be added, I am afraid that 32 bits is not going to be enough anyway. At the moment, the code implements a union of types with a bitwise union, which would be no longer possible if a bitmask would be used (of course, depending on how the bitmask is organized). Anyway, if we would create a bitmask for the types and it would still fit inside 32 bits, when all grammar element are added (or change it into an enum), we would still need another bitmask for the references. A uint8_t could be used for that. But on each compiler that implements a 32-alignment for struct members that would not reduce the memory footprint for tree nodes.

src/ddsts/include/dds/ddsts/typetree.h
```diff
@@ -429,6 +433,87 @@ ddsts_create_declaration(ddsts_identifier_t name, ddsts_type_t *decl_type, ddsts
...
+ * @returns A dds_return_t indicating success or failure.
+ */
+DDS_EXPORT dds_return_t
+ddsts_create_union_forward_dcl(ddsts_identifier_t name, ddsts_type_t **result);
+
+/* Union declaration (union_def)
+ */
+typedef struct {
+  ddsts_typespec_t type;
+  ddsts_flags_t switch_type;
```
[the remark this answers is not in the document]
FransFaase: At the moment, all scoped names are resolved during the parsing, and not stored as such in the Type Tree. So far, this seems to work in the remainder of the code. If you know of a case where this is not going to work, I am happy to hear about it.

src/ddsts/src/typetree.c
```diff
@@ -55,8 +55,21 @@ static void free_type_ref(ddsts_type_t *type, ddsts_type_t *parent, ddsts_flags_
...
+static void init_children(ddsts_type_list_t *list)
 {
+  list->first = NULL;
+  list->ref_end = &list->first;
+}
+
+static void children_add(ddsts_type_list_t *list, ddsts_type_t *child)
+{
+  *list->ref_end = child;
+  list->ref_end = &child->type.next;
```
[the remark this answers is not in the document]
FransFaase: Yes, it might look a little confusing, but it saves having to deal with special case that the list is empty.

rvloon:
Why not use a struct with named elements and a :<number_of_bits> specifier and let the compiler figure out the packing? Whether it ends up as 32, 48 and 64 bits is then irrelevant and it tends to lead to more readable code, without using more space than required.

Ronald
On 26/08/2019 11:01:19, Frans Faase <notifications@github.com> wrote:
@FransFaase commented on this pull request.
In src/ddsts/include/dds/ddsts/typetree.h [https://github.com/eclipse-cyclonedds/cyclonedds/pull/226#discussion_r317514598]:
> @@ -72,7 +76,7 @@ typedef char *ddsts_identifier_t; * IDL elements, such as the constant definition and the case labels. */ -typedef uint32_t ddsts_flags_t; +typedef uint64_t ddsts_flags_t;
Yes, I agree with you, that with respect to the types, it is not really used as a bitmask. (Note that it is a bitmask in combination with DDSTS_REFERENCE_1 and DDSTS_REFERENCE_2.) I see that some bits could be saved by, for example, by adding a flag for FORWARD and combine FORWARD_STRUCT and FORWARD_UNION as a combination of FORWARD and respectively STRUCT and UNION. Note that this would save just a single bit (add one to remove two). With the number of grammar elements still to be added, I am afraid that 32 bits is not going to be enough anyway. At the moment, the code implements a union of types with a bitwise union, which would be no longer possible if a bitmask would be used (of course, depending on how the bitmask is organized). Anyway, if we would create a bitmask for the types and it would still fit inside 32 bits, when all grammar element are added (or change it into an enum), we would still need another bitmask for the references. A uint8_t could be used for that. But on each compiler that implements a 32-alignment for struct members that would not reduce the memory footprint for tree nodes.

src/tools/idlc/src/tt_create.c
FransFaase: Yes, I know it is useless, but I understood that it is good practice to always end a case with a break. It might also prevent some compiler warnings, for compilers that check on a fall-through but do not check if the break statement can be reached (which in this simple case is rather trivial, but not if there a hunderd of lines of code in a case).

src/tools/idlc/src/tt_create.c
FransFaase: The location is used for syntax errors reported by yacc. It is not yet used for errors (warnings) generated here. It would require the location to be passed to each function called from idl.y, in case there is not another method of sharing it through the context. One reason for to dispose this, is also that it might have to be changed again when preprocessing has been implemented, because we probably also would like to include the file name in the location. Either the lexer or the parser will need to process the %line directives produced by the preprocessor.

k0ekk0ek: I'm currently in the process of integrating the preprocessor into `idlc`. I don't think [location tracking](https://www.gnu.org/software/bison/manual/html_node/Tracking-Locations.html) will change with regards to anything in `tt_create.c`. Having said that, I do agree changes will eventually be required to `tt_create.*` for the compiler, yes. For the first iteration of `idlc`, I'll try to keep the changes to the "rest" to a minimum, but I really think you should have a look at the example I provided in `shift-reduce.zip` above. I may be overlooking something, but I think some things could be done in a more pragmatic way, maybe eliminating the need to pass locations to any function outside `idl.y` all together. I wouldn't focus on that too much for now though, it's probably a lot easier to have a working tool first to see how things like this can be implemented best.

FransFaase:
@k0ekk0ek, I did have a look at the example in the shift-reduce.zip and it did give me some ideas for improving the code. For example, the processing of annotations now uses a stack implemented in tt_create, where it is probably simpler to use the 'stack' of the yacc parser to store those annotations until they are needed. (Freeing all memory in case of errors exactly once, has proven to be tricky.) At the moment annotations are not stored in the type tree, but that might change in the future.

Personally, I prefer clean interfaces and separation of concerns. I define a clean interface as an interface that hides all unnecessary implementation details, even if this would make the implementations a little more complex. I see tt_create as having the responsibility to perform semantic checks and call the functions defined in TypeTree.h to create the type tree. I foresee that to check the complex scoping rules of IDL, it might be needed to construct some temporary data structures in tt_create, which will not be part of the type tree at all. I assume that the semantic checks implemented in tt_create will not be needed when processing the xtypes representation exchanged between distributed instances of CycloneDDS. Of course, it could be possible to move all the code from tt_create into idl.y and call the functions to create the type tree directly. But I wonder if that would make the code easier to read. The type tree functions now presume that the tree is mostly constructed in a top-down manner, and suspect that some bottom-up construction functions would be needed to fully implement the example you gave. In the end this could lead to that there are multiple ways to construct the type tree. Something, that I do not find very elegant.

I do think that your example is useful with respect to processing annotations and the pragma's.

src/tools/idlc/src/tt_create.c
k0ekk0ek: @FransFaase, the whole process is taking a little longer than I anticipated, but I've created a fairly complete demonstration of how to use locations (including filenames) [here](https://github.com/k0ekk0ek/lloc). I've also included basics for the `#pragma` directive. I'm hoping to get locations in this week and then move on to doing the rest of the integration of the preprocessor. Anyway, maybe the example will give you an idea of how it's going to work/look.

src/ddsts/include/dds/ddsts/typetree.h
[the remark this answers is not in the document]
e-hndrks: This might become an issue when dealing with a discriminator based on an enum. Of course you could choose to store just the values of the enum, but it might make sense to actually store the labels. For example, if you want to generate the type descriptor for a union, you might want to refer to the enum labels instead of to the values they represent.

k0ekk0ek:
For future reference: I solved the conflicts in this PR and am squash-merging this into xtypes as my future work will be based on it. Thanks again @FransFaase! Your contributions are appreciated.

FransFaase:
@k0ekk0ek, I happy to see this PR to be merged and that the work on the IDL compiler and Xtypes is being continued. I will keep an eye on this project. I wish you much luck.

PR #227, opened 2019-08-01
Merge master into xtypes

k0ekk0ek:
This pull request merges the master branch into xtypes and fixes some compiler warnings. Normally I would've rebased, but seeing this is a public branch other users probably like a merge better. A rebase is probably done once the compiler is done.

Please consider pulling these changes so that I (and others of course) can continue working on an up-to-date branch.


k0ekk0ek:
@eboasson not sure what to do here? You've probably dealt with this situation before?

eboasson:
@k0ekk0ek, I presume you are wondering about the paperwork problem. I don't quite know how that came to be: I am pretty sure I only merged _after_ the ECA check passed, and indeed Eclipse has a valid ECA on file for the real email address of the author that's also in the Signed-off-by line. The "Author" line seems to be a github-generated email address, and the mismatch is what I think triggers the error.

I think it would be best to first check how t deal with it, what causes it and how to avoid it. Perhaps the incubation@eclipse.org list would be a good place to ask first. I'm sure others have seen this too.

k0ekk0ek:
@eboasson, sorry for the cryptic message. I was indeed referring to the IP verification failure. I've contacted incubation@eclipse.org and have received an answer. The problem occurs with other Eclipse projects as well and has been [reported before](https://bugs.eclipse.org/bugs/show_bug.cgi?id=496885#c18), but has not been solved. To quote Wayne Beaton's answer:

> The workaround while we sort this out is to check for the CLA yourself. There's a "validate" link on the CLA page [1]. The GitHub hook won't stop you from pushing if you believe that everything is in order. 
> 
> [1] https://www.eclipse.org/legal/CLA.php

The solution therefore seems to check the failing commit manually and simply merge. Do you agree with that approach?

eboasson:
@k0ekk0ek, absolutely. As I wrote earlier, I did the check just to be sure. So please, go ahead and merge.

PR #232, opened 2019-08-06
Fix acknack size & crash

eboasson:
One of the two fixes a mistake in calculating the bitmap sizes (and hence ACKNACK, &c) sizes that broke interoperability. The other fixes a crash when a fragmented sample needs to be dropped from the defragmenter but the first fragment of the sample has not arrived yet.


eboasson:
@k0ekk0ek, this fixing a rather arcane detail that I broke the other day, plus a rather arcane detail that I suspect caused its first crash after 8 years of being in hiding, it seems to me that it is not actually fair to ask you to approve or disapprove of the changes.

But I would appreciate it if you'd let me know whether you'd prefer to review them anyway. If not, I'll go ahead and merge them.

k0ekk0ek:
@eboasson, feel free to merge.

PR #233, opened 2019-08-07
Fallback to zip installer for non Linux/Windows builds

ThijsSassen:
When building an installer for a non Windows/Linux target i.e. FreeRTOS a CPACK error occurs due to missing generator setting. This PR adds a fallback to zip for these situations.


ThijsSassen:
2 windows builds fail due to issues with installing maven

PR #235, opened 2019-08-15
set project name from string literal

dirk-thomas:
See https://github.com/eclipse-cyclonedds/cyclonedds/pull/224/files#r314394345

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,8 +50,8 @@ execute_process(COMMAND git submodule update
 # Set module path before defining project so platform files will work.
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/Modules")
 set(CMAKE_PROJECT_NAME_FULL "Eclipse Cyclone DDS")
-set(PROJECT_NAME "CycloneDDS")
-project(${PROJECT_NAME} VERSION 0.1.0)
+# The following line will set the variable PROJECT_NAME
+project(CycloneDDS VERSION 0.1.0)
 
 # Set some convenience variants of the project-name
 string(REPLACE " " "-" CMAKE_PROJECT_NAME_DASHED "${CMAKE_PROJECT_NAME_FULL}")


k0ekk0ek:
Hi @dirk-thomas. Thanks for the PR! Seems like a good idea to do it like that. I've restarted the failed windows build and I'm merging it in after it, and the Eclipse IP check pass. If you click details, it'll say why it has failed, though I have a hunch that it's because of the noreply email address. If you insist on not using the address you used with Eclipse, I could check if there's a way around it. Thanks again!

dirk-thomas:
> If you insist on not using the address you used with Eclipse, I could check if there's a way around it.

That would be great.

I created an Eclipse account for this but with my company email address. In the Eclipse account profile I can't add additional email addresses. I also tried to add my GitHub username to my profile (in the hope that would help) by the profile form fails to do so (I sent an email to Eclipse about that).

So any help to make it pass for this PR (and potentially others in the future) would be very much appreciated.

k0ekk0ek:
I can check manually, but I'd need the email address of your Eclipse account, or the url of your Eclipse profile and a way to tie that to your GitHub profile. Seems like the Eclipse people are working on a solution judging by bug [496855](https://bugs.eclipse.org/bugs/show_bug.cgi?id=496885). Alternatively, you could force push and perhaps use the address you used with you Eclipse account? Perhaps it's even enough if you change the signed-off-by footer. Of course, other ideas are welcome, just need to verify the ECA has been signed.

dirk-thomas:
> the url of your Eclipse profile

https://accounts.eclipse.org/users/dthomas65x

> a way to tie that to your GitHub profile

Well, if the profile form would just allow me to enter my GitHub username :wink:

> Seems like the Eclipse people are working on a solution judging by bug 496855.

Not sure that one applies here. There is no merge commit. Also the ticket is three years old so not very promising...

k0ekk0ek:
If I search Eclipse users with you GitHub username it doesn't show me that user profile and the user profile doesn't have a GitHub account linked to it(?) At least, there's no GitHub icon there (which is there when I use another browser to view my own profile). I basically need some proof that the two go together (sorry to make you go through all this trouble for a two-line commit). I'll have to check the project handbook to see what the exact rules are.

dirk-thomas:
> If I search Eclipse users with you GitHub username it doesn't show me that user profile and the user profile doesn't have a GitHub account linked to it(?)

As mentioned above the form to edit my Eclipse profile is broken. Therefore I can't add my GitHub username, website or any other information.

I would suggest to iterate with Eclipse to fix that part and I am happy to add the information to my Eclipse account.

k0ekk0ek:
Hi @dirk-thomas. I've checked the handbook and the section [Git Commit Records](https://www.eclipse.org/projects/handbook/#resources-commit) states that *Author* and *Signed-off-by* should both use the email address use to sign the ECA. I think the fasted way to resolve the ECA issue is if you change those in your commit and force-push. Alternatively you could change the address you used to register with Eclipse, that is what one of the Cyclone DDS team members did. I don't know if the profile page will currently let you do that. If not, let me know and I too will inform/ask Eclipse.

dirk-thomas:
With feedback from the Eclipse support I was able to update my profile. It now contains the GitHub username so it should be possible to correlate them.

Basically I had to do the profile change in an incognito window since otherwise the password manager of the browser was autofilling some hidden fields of the form.

k0ekk0ek:
The fact that the profile page has a link to your GitHub account together with the reference to said page in this conversation indeed show the two go together (to some extend). However, the Eclipse Project Handbook explicitly states:

> The email address of the author must match the email address on the Eclipse Foundation Account.

I'm afraid I have to insist that you correct the information in the commit before I can hit merge.

eboasson:
While you're no doubt quoting the book correctly, when not so long ago we had an Author line set to a github address, we asked and were told (in my understanding) it is acceptable if in such a case a manual check of the address in the Signed-off-by line shows a valid ECA on file.

So I would think the obstacle here is now simply the Signed-off-by line, not the author line.

dirk-thomas:
I certainly won't update the email in the signature. But I can delete my Eclipse account and create a new one with the email matching the signature. Especially if that helps in the future.

I would really expect Eclipse to solve this in a way that e.g. allows to specify multiple email addresses in a single account as GitHub does it.

dirk-thomas:
Unfortunately I won't be able to create an account with the email from the signature or change the email of my existing account to it. Simply because Eclipse requires to send a confirmation email which isn't possible for this no-reply address.

I would encourage you to bring this up with Eclipse so that they support this use case. The whole idea of the no-reply email address from GitHub is to not expose an email address in public repositories.

I am fine to just close this PR for now since the change is not really significant. Please feel free to create a "similar" patch.

PR #238, opened 2019-08-22
Prevent participant creation with unknown domain id.

martinbremmer:
Consider the situation where a configuration mentions only specific domain(s) (non-any).
The creation of a participant with a different domain id (non-default) should fail, but doesn't.

This pull request fixes that.

p.s. Also removed the redundant dds_domain_default() function declaration from dds.h (https://github.com/eclipse-cyclonedds/cyclonedds/pull/223#issuecomment-523519153).

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -51,17 +51,6 @@ extern "C" {
 
 struct ddsi_serdata;
 
-/**
- * @brief Returns the default domain identifier.
- *
- * The default domain identifier can be configured in the configuration file
- * or be set through an evironment variable ({DDSC_PROJECT_NAME_NOSPACE_CAPS}_DOMAIN).
- *
- * @returns Default domain identifier
- */
-DDS_EXPORT dds_domainid_t dds_domain_default (void);
-
-
 #define DDS_MIN_PSEUDO_HANDLE ((dds_entity_t) 0x7fff0000)
 
 /* @defgroup builtintopic_constants Convenience constants for referring to builtin topics
--- a/src/core/ddsc/src/dds_domain.c
+++ b/src/core/ddsc/src/dds_domain.c
@@ -73,7 +73,7 @@ static dds_return_t dds_domain_init (dds_domain *domain, dds_domainid_t domain_i
         enough) */
 
   (void) ddsrt_getenv ("CYCLONEDDS_URI", &uri);
-  domain->cfgst = config_init (uri, &domain->gv.config, domain_id);
+  domain->cfgst = config_init (uri, &domain->gv.config, (domain_id == DDS_DOMAIN_DEFAULT) ? UINT32_MAX : domain_id);
   if (domain->cfgst == NULL)
   {
     DDS_ILOG (DDS_LC_CONFIG, domain_id, "Failed to parse configuration XML file %s\n", uri);
--- a/src/core/ddsc/tests/participant.c
+++ b/src/core/ddsc/tests/participant.c
@@ -75,7 +75,7 @@ CU_Test(ddsc_participant, create_multiple_domains)
   dds_return_t status;
   dds_domainid_t domain_id;
 
-  ddsrt_setenv("CYCLONEDDS_URI", "<Tracing><Verbosity>finest</><OutputFile>multi-domain-1.log</></>");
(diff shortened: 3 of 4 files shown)


eboasson:
@martinbremmer: it requires a bit of explanation for me to understand why that is a problem. There is — and should be, in my view — a default configuration and it seems rather logical to me that the default configuration will be used if it isn't overridden.

What problem does it solve to add a rule that says: thou shalt explicitly state that the default configuration is to be used in case there is a configuration that overrides some default settings for some particular domains?

I can see one might want to restrict domain creation to a specific set of domain ids, but that's a different problem.

Also: I was rather more in favour of eliminating ``DDS_PROJECT_NAME_NOSPACE_CAPS`` and similar noise from the code than of retaining it. It isn't like "Eclipse Cyclone DDS" will suddenly be renamed into "Eclipse Brobdingnag DDS", and really doesn't make anything easier to read or search through.

martinbremmer:
@eboasson, thanks for the reaction.

I basically have two reasons for trying to get this in.

1) Not preventing the creation of a participant in this case is quite a behavioral change (it was always prohibited).
But the multi-domain is a large behavioral change regarding participant creation anyway. So it's not a big nor strong argument.

2) Someone created a nice configuration and then creates a participant with a domain that is not configured (unwittingly). The creation succeeds.
But now it acts different from what was expected and it can take a while before it is realized that the participant is created with the wrong domain id and thus using default configuration.
I just think that preventing the creation makes the behaviour of cyclonedds more intuitive and less prone to opaque problems.


Regarding DDS_PROJECT_NAME_NOSPACE_CAPS.
I agree that the project name of "Eclipse Cyclone DDS" would probably not change suddenly. Also, if you did that, you could just change the hardcoded stuff as well.
The problem when it is hardcoded, however, emerges when using cyclonedds as a submodule for a different project. Then the project name changes to the parent projects' name.
This is not really a problem for cyclonedds, but it would be nice if it supports being used as a submodule.

eboasson:
> Not preventing the creation of a participant in this case is quite a behavioral change (it was always prohibited). But the multi-domain is a large behavioral change regarding participant creation anyway. So it's not a big nor strong argument.

You are right that it is a behavioural change. The question is whether it matters: all old combinations of config file & domain id specified in ``dds_create_participant`` that used to work behave as before.

Those cases that failed, failed because something that is eminently reasonable to do (and that is supported by most DDS implementations) was simply impossible. If you can only have a single configuration, it makes no sense whatsoever to *ignore* what is in the configuration if the ids don't match.

When you can have multiple domains, it is far more logical that you simply get what you ask for, and also more in line with other implementations.

> Someone created a nice configuration and then creates a participant with a domain that is not configured (unwittingly). The creation succeeds.
> But now it acts different from what was expected and it can take a while before it is realized that the participant is created with the wrong domain id and thus using default configuration.
> I  just think that preventing the creation makes the behaviour of cyclonedds more intuitive and less prone to opaque problems.

Possibly — but the trouble here is that one needs to guess at the likelihood of the mistake of passing an unintended domain id to ``dds_create_participant``. We have no means of knowing. Moreover, it depends on whether you view the configuration file as more than simply as overriding some of the defaults (and thus not restricting anything).

My choice was the freewheeling one, but as I wrote, I can see that one might want to limit creation of domains to those that have been configured. But I'd rather see that as an option under the CycloneDDS tag. That still leaves open the question what you'd want to do if there are multiple files ...

Also, I really tried to make it easy to use a combination of Domain elements for id "any" in combination with elements for id ≠ "any". For example, so that you can set some common options in "any", then set — and possibly even override some — in a domain-specific one.

The way your change checks things, ``domid_accepted`` would be dependent on the placement of "any"s in the configuration file. That in itself would be pretty confusing, too.

martinbremmer:
@eboasson, ok, thanks.

I have to think about it a bit, but I'll probably close this PR.

Just for my information, in what way does ```domid_accepted``` depend on the "any" placement?

Also, what do you think about a warning (if config(s) only contains specific ids and the specific id of the create does not match any of them)? That'll reduce my concern nr2 quite a bit.

eboasson:
> Just for my information, in what way does domid_accepted depend on the "any" placement?

It looks to me like an "any" at the end would always caused ``domid_accepted`` to be true, that's all.

> Also, what do you think about a warning (if config(s) only contains specific ids and the specific id of the create does not match any of them)? That'll reduce my concern nr2 quite a bit.

I think an option would make sense, something along the likes of CycloneDDS/AllowedDomains that can be set to "any" or "configured" with some way of specifying you'd like a warning rather than an error.

martinbremmer:
@eboasson, thanks for the explanation. Closing this PR.

PR #239, opened 2019-08-23
Set project name via string literal

k0ekk0ek:
This pull request sets the project name via a string literal so 3rd party tools can easily extract it. The fix was proposed by @dirk-thomas in PRs #235 and details about why were mentioned in #224. Apart from that this PR enables extra warnings for Clang builds and adds a `DDSRT_WARNING_CLANG_OFF` and a `DDSRT_WARNING_CLANG_ON` to disable and re-enable them as required in the code. e.g. for test cases that verify the api behaves correctly on bad parameters.

Please consider pulling these changes. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,17 +41,10 @@ FUNCTION(PREPEND var prefix)
    SET(${var} "${listVar}" PARENT_SCOPE)
 ENDFUNCTION(PREPEND)
 
-# Update the git submodules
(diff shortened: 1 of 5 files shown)


eboasson:
@k0ekk0ek: should I be happy that you fixed the same warnings in the tests that I just did, or annoyed that I will now have to delete my commit for those fixes? Happy seems to be the better option! 😁

PR #244, opened 2019-09-02
Cyclone configuration as string

FransFaase:
For targets that do not support ddsrt_setenv and ddsrt_getenv, an alternative method is needed to supply an application specific configuration. One way to implement this, is to add a function for creating a domain with a string arguments, which needs to be called before any call to dds_create_participant for given domain identifier.
    
The function dds_create_domain has been added, which has as arguments a domain identifier and a configuration string. In case the string starts with '<' it is parsed as XML fragment for the configuration, otherwise it is taken as the name of the file containing the configuration. (This is similar to how the value of the environment variable is treated.)
    
Two tests have been added. One limits the number of participants to two and verifies that creating a third participant fails. The other tests checks incorrect calls to dds_create_domain.
    
An assert in dds_handle_delete has been weakened.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -726,7 +726,7 @@ dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
  * If no configuration file exists, the default domain is configured as 0.
  *
  *
- * @param[in]  domain The domain in which to create the participant (can be DDS_DOMAIN_DEFAULT). Valid values for domain id are between 0 and 230. DDS_DOMAIN_DEFAULT is for using the domain in the configuration.
+ * @param[in]  domain The domain in which to create the participant (can be DDS_DOMAIN_DEFAULT). DDS_DOMAIN_DEFAULT is for using the domain in the configuration.
  * @param[in]  qos The QoS to set on the new participant (can be NULL).
  * @param[in]  listener Any listener functions associated with the new participant (can be NULL).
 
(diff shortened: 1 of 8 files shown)


eboasson:
Thanks @FransFaase, but I'm not too happy with it. Naturally, I would almost say, considering the lengths to which I have gone in reworking domains in #246. There is plenty in that PR that one could argue about, and so I am certainly not planning on just squashing it in, but I think that any arguments on that PR will be about implementation details, naming or other changes affecting the API — but not so much about the conceptual change.

When considered in the light of that PR, the solution you chose (adding it as an argument to the participant, then adding rules for what is good and what is not) seems a prima facie a worse solution that creating a function ``dds_create_domain`` that simply takes such an argument and creates the domain if it does not yet exist.

FransFaase:
Hi @eboasson, I understand your considerations. I did consider the solution you are proposing, but I remember that at that time, there was not anything with respect to domains in the (public) API. I will have a look at it and try to come up with a better solution.

FransFaase:
@eboasson, I have updated the pull request and editted the comment according to the API change you have suggested.

eboasson:
Thanks, @FransFaase, I like this much better. The bit I don't like is that if you do ``dds_create_domain(1,config1)`` and then ``dds_create_domain(1,config2)``, the second call will simply return the domain created in the first call, despite it having a different configuration. I'd say it would make sense to return an error if the domain already exists.

Secondly, I'm not so sure the combination of ``DDS_DOMAIN_DEFAULT`` and an explicit configuration makes all that much sense. There are all kinds of interesting possibilities for unexpected behaviour the way it is now. I'd say there is much less risk of nasty surprises if ``DDS_DOMAIN_DEFAULT`` is disallowed in ``dds_create_domain``.

Finally, to have a ``dds_create_domain`` and a ``dds_domain_create`` that are different in rather important aspects is not so practical. I understand the way this came to be, but I think it would be better to rename ``dds_domain_create`` to something that is more clearly an internal function.

FransFaase:
@eboasson, I have processed your comments. It is possible that dds_create_domain returns DDS_RETCODE_OUT_OF_RESOURCES. This is not mentioned in the description of the function. I found no other mentioning of DDS_RETCODE_OUT_OF_RESOURCES in said file. Should it still be mentioned explicitly?

eboasson:
@FransFaase , from a functional perspective, this looks good to me, but I do still have few things to nitpick about some details on how the changes are structured:

* ``ILLEGAL_OPERATION`` is pretty consistently used for trying to perform an operation on a type of “object” that doesn’t support that operation. So I think the choice of returning this error if the domain already exists is less than ideal. ``PRECONDITION_NOT_MET`` seems a more fitting choice.
* there is no (longer a) restriction on the value of the domain id, so please remove the "between 0 and 230".
* ``dds_domain_init`` now explicitly takes a configuration, which I like, but it may be a null pointer, which I like less. This especially true since ``dds_create_domain`` requires a configuration. (Such a change would also eliminate having two call sites to ``config_init``, something I might otherwise fall over.)
* 'Tis true, if the variable doesn't exist, ``ddsrt_getenv`` leaves its output parameter unchanged, and in this case that's a null pointer that is then specially handled in ``config_init``. But it would be more elegant to initialise it to an empty string and modify ``config_init`` to disallow a null pointer (and eliminate the test).
* In ``dds_domain_create_internal`` you now have ``use_existing`` and ``configfile`` but the value of ``configfile`` implies ``use_existing``. If you were to also require that ``configfile`` is always a non-null pointer here, it becomes a lot cleaner. Then you simply let ``dds_create_participant`` pass the result of ``ddsrt_getenv`` to ``dds_domain_create_internal`` and all is well.
* And for a final nitpick: I’d rename ``configfile`` to ``config``: it can be a file, empty, multiple files, XML, or even a mixture of XML fragments and files …

FransFaase:
@eboasson, I have processed your comments. I have removed the test on `configfile` (now renamed to `config`) being NULL from config_init and added a `ddsrt_nonnull` clause in the header. The code to retrieve the environment variable has been moved from `dds_domain_create_internal` to `dds_create_participant` as you suggested. This includes moving the neccessary `#include` statements.

eboasson:
Looks good to me, thank you!

PR #243, opened 2019-08-30
Sync security branch with master

martinbremmer:
I wanted to sync the security branch with the master branch. I did that by merging. This resulted in this (almost) impossible to review PR.

If there is a better (or more standard) way to sync a (upstream) branch with (upstream) master, then I'll gladly close this PR and do it that way. Otherwise, please consider merging this PR.

PR #245, opened 2019-09-03
Sync security branch with master

martinbremmer:
I wanted to sync the security branch with the master branch. I did that by merging.

For some background information, see also #243, which was the first try.


martinbremmer:
Hmm, the eclipse check fails on a commit that was in master: #202. At least, it seems to. I don't know how to fix or circumvent that...

eboasson:
@martinbremmer, yes that's caused by a problem with the ECA check tool. As the commit in this PR is identical to the commit that is in master, I do not see a reason to worry about it.

martinbremmer:
@eboasson, so you can still merge (after approval) if that check fails?

eboasson:
@martinbremmer , yes, it requires additional care, but it is not impossible.

(I also wonder what would happen if a commit from person X with a valid ECA was merged in, that the ECA was subsequently cancelled, and that after the cancellation one would try to merge it ...)

PR #248, opened 2019-09-05
Fix compilation errors for RMW build on Windows

dennis-adlink:
With this PR the compilation errors are fixed that occur when compiling CycloneDDS headers as part of the ROS2 rmw_cyclonedds_cpp module on Windows (using the msvc C++ compiler). The errors are fixed by adding explicit casts when calling operations on atomics and using a different syntax for compound literals (because the c99 syntax is not supported in msvc in C++ mode). I also suppressed some warnings (msvc only) related to emtpy array declaration in structs.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_vendor.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_vendor.h
@@ -19,6 +19,7 @@ typedef struct {
   uint8_t id[2];
 } nn_vendorid_t;
 
+
 /* All existing vendor codes have the major part equal to 1 (and this will probably be true for a long, long time) */
 #define NN_VENDORID_MINOR_RTI                0x01
 #define NN_VENDORID_MINOR_PRISMTECH_OSPL     0x02
@@ -38,8 +39,15 @@ typedef struct {
 #define NN_VENDORID_MINOR_ECLIPSE            0x10
 #define NN_VENDORID_MINOR_PRISMTECH_CLOUD    0x20
 
-#define NN_VENDORID_UNKNOWN ((nn_vendorid_t) {{ 0x00, 0x00 }})
-#define NN_VENDORID_ECLIPSE ((nn_vendorid_t) {{ 0x01, 0x10 }})
+#if defined(_WIN32) && defined(__cplusplus)
+#define NN_VENDORID_CAST
+#define NN_VENDORID_UNKNOWN {{ 0x00, 0x00 }}
+#define NN_VENDORID_ECLIPSE {{ 0x01, 0x10 }}
+#else
+#define NN_VENDORID_CAST (nn_vendorid_t)
+#define NN_VENDORID_UNKNOWN (NN_VENDORID_CAST {{ 0x00, 0x00 }})
+#define NN_VENDORID_ECLIPSE (NN_VENDORID_CAST {{ 0x01, 0x10 }})
+#endif
 
 #if defined (__cplusplus)
 extern "C" {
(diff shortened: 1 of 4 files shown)


eboasson:
Thanks @dennis-adlink, I really like effort being put into making sure the RMW layer build cleanly on Windows.

Although the change look correct to me (and, well, it passes the tests, so surely that it is proof of its correctness :P) I think it would be slightly more elegant to make some syntactic changes. The reason is basically that I dislike type casts. In this case, though, there are certainly other views possible.

For the vendor id stuff, I think I would do something like what you can find at the end of this comment. To me it is a little cleaner to have a macro that constructs a well-formed literal of the ``nn_vendorid_t`` type (or something that may be used in that way), then to have a ``NN_VENDORID_CAST`` macro. Note that I am simply assuming we can go a couple of years before the major code changes ...

Concerning the win32 atomics: given that the ``PTROP`` thing is there already, I am thinking that it would work out nicer if the 32- and 64-bit variants also went through a macro like that, rather than have all those casts written out.

Note: although I think it'll work out nicer, I haven't done the exercise. So what I am asking is that you consider it, and if you think it nonsense, explain to me why.

```C
#if defined(_WIN32) && defined(__cplusplus)
#define NN_VENDORID(vendor_equals) {{ 0x01, NN_VENDORID_MINOR_##vendor }}
#define NN_VENDORID_UNKNOWN {{ 0x00, 0x00 }}
#define NN_VENDORID_ECLIPSE NN_VENDORID (ECLIPSE)
#else
#define NN_VENDORID(vendor) ((nn_vendorid_t) {{ 0x01, NN_VENDORID_MINOR_##vendor }})
#define NN_VENDORID_UNKNOWN ((nn_vendorid_t) {{ 0x00, 0x00 }})
#define NN_VENDORID_ECLIPSE NN_VENDORID (ECLIPSE)
#endif

#if defined (__cplusplus)
extern "C" {
#endif

inline bool vendor_equals (nn_vendorid_t a, nn_vendorid_t b) {
  return ((a.id[0] << 8) | a.id[1]) == ((b.id[0] << 8) | b.id[1]);
}
inline bool vendor_is_eclipse (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (ECLIPSE));
}
inline bool vendor_is_rti (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (RTI));
}
inline bool vendor_is_opensplice (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (PRISMTECH_OSPL));
}
inline bool vendor_is_twinoaks (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (TWINOAKS));
}
inline bool vendor_is_eprosima (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (EPROSIMA));
}
inline bool vendor_is_cloud (nn_vendorid_t vendor) {
  return vendor_equals (vendor, NN_VENDORID (PRISMTECH_CLOUD));
}
inline bool vendor_is_eclipse_or_opensplice (nn_vendorid_t vendor) {
  return vendor_is_eclipse (vendor) | vendor_is_opensplice (vendor);
}
inline bool vendor_is_prismtech (nn_vendorid_t vendor) {
  return (vendor_equals (vendor, NN_VENDORID (PRISMTECH_OSPL)) ||
          vendor_equals (vendor, NN_VENDORID (PRISMTECH_LITE)) ||
          vendor_equals (vendor, NN_VENDORID (PRISMTECH_GATEWAY)) ||
          vendor_equals (vendor, NN_VENDORID (PRISMTECH_JAVA)) ||
          vendor_equals (vendor, NN_VENDORID (PRISMTECH_CLOUD)));
}
inline bool vendor_is_eclipse_or_prismtech (nn_vendorid_t vendor) {
  return vendor_is_eclipse (vendor) || vendor_is_prismtech (vendor);
}
```

PR #250, opened 2019-09-09
Added property policy.

martinbremmer:
Add property policy to QoS and parameter list.

The property policy has to be (de)serialized as part of the parameter list of a message. There is already generic code for that for other message parameters. However, that is for relatively simple structures.

The property policy is somewhat more elaborate. In short, it contains two sequences of structures ({name,string,propagate bool} and {name,buf,propagate bool}), which isn't supported by the generic (de)serialization. The other complication is the propagation boolean. First, it's not present on the wire. Second, when the value is 'false', then the complete struct should not be present on the wire either.

To support this, the generic parameter handling is extended.
* A XbPROP is added to the pserop to identify the propagation boolean within structures.
* The generic parameter handling functions are extended to be able to handle XbPROP.
* A propagate_generic function is added to check a struct if it contains the XbPROP boolean and what the value is.
* Various property_policy functions (like deser, unalias, etc) are added.
* Various struct_seq functions (like deser, unalias, etc) are added to support the property_policy functions. These functions expect a pserop array that represents the specific struct.
* dds_property_desc and dds_binaryproperty_desc pserop arrays are added to be used with the struct_seq functions.
* ser_generic() has been split up so that the data part can be used in ser_struct_seq().
* Few generic parameter handling functions now update the given offset, which makes using them in the struct_seq functions easier.
* Because the propagation boolean is not sent over the wire, the mem size of the struct on the wire is smaller than the one in memory. This means that the sequence buffer can not be aliased, but has to be specifically allocated. The contents, however, can be aliased. Because of this, a specific copy function for the property policy had to be created to make sure that the copy also has it's own sequence buffer. Otherwise, the fini and unaliase functions won't properly work.
* PID_PROPERTY_LIST is added to the piddesc piddesc_omg array that contains the parameter handling information. The specific property functions are referenced to make sure they're used instead of the generic ones.

I've tested it manually, please see the attached patch file for that.
[test.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3590524/test.patch.txt)

### DIFF
--- a/src/core/ddsi/src/q_plist.c
+++ b/src/core/ddsi/src/q_plist.c
@@ -38,6 +38,7 @@
 #include "dds/ddsrt/avl.h"
 #include "dds/ddsi/q_misc.h" /* for vendor_is_... */
 
+
 /* I am tempted to change LENGTH_UNLIMITED to 0 in the API (with -1
    supported for backwards compatibility) ... on the wire however
    it must be -1 */
@@ -103,6 +104,7 @@ enum pserop {
   Xo, Xox2, /* octet, 1 .. 2 in a row */
   Xb, Xbx2, /* boolean, 1 .. 2 in a row */
   XbCOND, /* boolean: compare to ignore remainder if false (for use_... flags) */
+  XbPROP, /* boolean: propagation onto the wire indication (not on the wire itself) */
   XG, /* GUID */
   XK /* keyhash */
 } ddsrt_attribute_packed;
@@ -126,6 +128,7 @@ struct piddesc {
       dds_return_t (*fini) (void * __restrict dst, size_t * __restrict dstoff, struct flagset *flagset, uint64_t flag);
       dds_return_t (*valid) (const void *src, size_t srcoff);
       bool (*equal) (const void *srcx, const void *srcy, size_t srcoff);
+      void (*cpy) (void * __restrict dst, const void * __restrict src);
     } f;
   } op;
   dds_return_t (*deser_validate_xform) (void * __restrict dst, const struct dd * __restrict dd);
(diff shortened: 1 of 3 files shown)


src/core/ddsi/src/q_plist.c
```diff
@@ -126,6 +128,7 @@ struct piddesc {
       dds_return_t (*fini) (void * __restrict dst, size_t * __restrict dstoff, struct flagset *flagset, uint64_t flag);
       dds_return_t (*valid) (const void *src, size_t srcoff);
       bool (*equal) (const void *srcx, const void *srcy, size_t srcoff);
+      void (*cpy) (void * __restrict dst, const void * __restrict src);
```
BartP6703: Isn't 'copy' i.s.o. 'cpy' better?

src/core/ddsi/src/q_plist.c
```diff
@@ -237,6 +250,340 @@ static bool equal_reliability (const void *srcx, const void *srcy, size_t srcoff
...
+  size_t offx = *srcoff;
+  size_t offy = *srcoff;
+  bool equal;
+  uint32_t i;
+
+  seqx = deser_generic_src (srcx, &offx, alignof (ddsi_octetseq_t));
+  seqy = deser_generic_src (srcy, &offy, alignof (ddsi_octetseq_t));
+
+  equal = (seqx->length == seqy->length);
+  structx = seqx->value;
```
BartP6703: Move to the if (equal) block. Then the assignment is not done when not equal.

eboasson:
Thanks @martinbremmer for contributing this. I know it is a rather unpleasant area to have to mess around with ... While in general I think your approach is quite sensible, I do have some remarks:

> The property policy is somewhat more elaborate. In short, it contains two sequences of structures ({name,string,propagate bool} and {name,buf,propagate bool}), which isn't supported by the generic (de)serialization. The other complication is the propagation boolean. First, it's not present on the wire. Second, when the value is 'false', then the complete struct should not be present on the wire either.

My sense is that it would be more practical to put the "propagate" field at the beginning of the struct. I don't really like the addition of the various "struct" functions, and would prefer it if the existing "generic" code were to be extended to handle these more complex types. As those functions process the fields in order, putting the "propagate" flag as the first field should make it much easier to handle that vile thing.

> * A propagate_generic function is added to check a struct if it contains the XbPROP boolean and what the value is.

If I'm not mistaken, the propagate thing is only ever used as the first field in a struct that's only used as an element in a sequence. So if one were to do things a bit differently and add a generic sequence (non-nested: that's enough for now and has fewer complications with memory allocations and deserialisation failures), then that sequence handling code would need to check for XbPROP, but pretty much nothing else would.

> * ser_generic() has been split up so that the data part can be used in ser_struct_seq().
> * Few generic parameter handling functions now update the given offset, which makes using them in the struct_seq functions easier.

Thanks ☺️

> * Because the propagation boolean is not sent over the wire, the mem size of the struct on the wire is smaller than the one in memory. This means that the sequence buffer can not be aliased, but has to be specifically allocated. The contents, however, can be aliased. Because of this, a specific copy function for the property policy had to be created to make sure that the copy also has it's own sequence buffer. Otherwise, the fini and unaliase functions won't properly work.

I don't think that's correct: the string sequence already has the same problem. I think here, too, it suffices to say that the sequence of T is always allocated and owned by plist/qos object, but that the elements (of type T) alias for strings and octet sequences. Fini frees the sequence and optionally the contents, unalias always creates a deep copy.

I did a quick experiment to see how hard it would be to add such (non-nested) sequences, and changed all (three) sequence-of-strings to use it. Perhaps it would be worth having a look at [it](https://github.com/eboasson/cyclonedds/tree/plist-sequence) to see whether it would suffice for these property things as well?

(Note: I have tested it only a little bit with the partition QoS ... so beware!)

martinbremmer:
Because of the comments and specifically the (non-nested) sequences handling prototype, it was easier for me to close this PR and just create a new one: #254.

PR #254, opened 2019-09-16
Added property policy.

martinbremmer:
Add property policy to QoS and parameter list.

A previous PR was created regarding the property policy: #250.
Because of the comments and specifically the (non-nested) sequences handling prototype, it was easier for me to just create a new PR.

The property policy has to be (de)serialized as part of the parameter list of a message. There is already generic code for that for other message parameters. However, that is for relatively simple structures.

The property policy is somewhat more elaborate. In short, it contains two sequences of structures ({name,string,propagate bool} and {name,buf,propagate bool}), which already isn't supported by the generic (de)serialization. The other complication is the propagation boolean. First, it's not present the wire. Second, when the value is 'false', then the complete struct should not be present on the wire either.

To support this, the generic parameter handling is extended.
* A XbPROP is added to the pserop to identify the propagation boolean within structures.
* A XQ is added to the pserop to identify a (non-octet, non-nested) sequence (replacing the XZ string sequence).
* The generic parameter handling functions are extended to be able to handle XbPROP and XQ.
* Some generic parameter handling functions are split into an 'entry' function and a 'partial' function.
* PID_PROPERTY_LIST is added to the piddesc piddesc_omg array that contains the parameter handling information.

This is done in 3 commits:
1. Unused length opcode has been removed.
2. The plist handling has been extended with the sequence and propagate boolean handling.
3. The property policy types and handling has been added.

I've tested it manually, please see the attached patch file for that.
[prop_test.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3617110/prop_test.patch.txt)

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_plist.h
+++ b/src/core/ddsi/include/dds/ddsi/q_plist.h
@@ -243,6 +243,9 @@ struct nn_rdata;
 DDS_EXPORT unsigned char *nn_plist_quickscan (struct nn_rsample_info *dest, const struct nn_rmsg *rmsg, const nn_plist_src_t *src);
 DDS_EXPORT const unsigned char *nn_plist_findparam_native_unchecked (const void *src, nn_parameterid_t pid);
 
+DDS_EXPORT void nn_free_property_policy (dds_property_qospolicy_t *property);
+DDS_EXPORT void nn_duplicate_property (dds_property_t *dest, const dds_property_t *src);
+
 #if defined (__cplusplus)
 }
 #endif
--- a/src/core/ddsi/include/dds/ddsi/q_plist.h
+++ b/src/core/ddsi/include/dds/ddsi/q_plist.h
@@ -243,9 +243,6 @@ struct nn_rdata;
 DDS_EXPORT unsigned char *nn_plist_quickscan (struct nn_rsample_info *dest, const struct nn_rmsg *rmsg, const nn_plist_src_t *src);
 DDS_EXPORT const unsigned char *nn_plist_findparam_native_unchecked (const void *src, nn_parameterid_t pid);
 
-DDS_EXPORT void nn_free_property_policy (dds_property_qospolicy_t *property);
-DDS_EXPORT void nn_duplicate_property (dds_property_t *dest, const dds_property_t *src);
-
 #if defined (__cplusplus)
 }
 #endif
--- a/src/core/ddsi/include/dds/ddsi/q_xqos.h
+++ b/src/core/ddsi/include/dds/ddsi/q_xqos.h
@@ -35,7 +35,8 @@ typedef ddsi_octetseq_t dds_groupdata_qospolicy_t;
 typedef struct dds_property {
   /* The propagate boolean will not be send over the wire.
    * When the value is 'false', the complete struct shouldn't be send.
-   * It has to be the first variable within the structure. */
+   * It has to be the first variable within the structure because it
+   * is mapped to XbPROP in the serialiser. */
   unsigned char propagate;
   char *name;
   char *value;
@@ -49,7 +50,8 @@ typedef struct dds_propertyseq {
 typedef struct dds_binaryproperty {
   /* The propagate boolean will not be send over the wire.
    * When the value is 'false', the complete struct shouldn't be send.
-   * It has to be the first variable within the structure. */
+   * It has to be the first variable within the structure because it
+   * is mapped to XbPROP in the serialiser. */
   unsigned char propagate;
   char *name;
   ddsi_octetseq_t value;
--- a/src/core/ddsi/include/dds/ddsi/q_xqos.h
+++ b/src/core/ddsi/include/dds/ddsi/q_xqos.h
@@ -32,6 +32,39 @@ typedef ddsi_octetseq_t dds_userdata_qospolicy_t;
 typedef ddsi_octetseq_t dds_topicdata_qospolicy_t;
 typedef ddsi_octetseq_t dds_groupdata_qospolicy_t;
 
+typedef struct dds_property {
+  /* The propagate boolean will not be send over the wire.
+   * When the value is 'false', the complete struct shouldn't be send.
+   * It has to be the first variable within the structure. */
+  unsigned char propagate;
+  char *name;
+  char *value;
+} dds_property_t;
+
+typedef struct dds_propertyseq {
+  uint32_t n;
+  dds_property_t *props;
+} dds_propertyseq_t;
+
+typedef struct dds_binaryproperty {
+  /* The propagate boolean will not be send over the wire.
+   * When the value is 'false', the complete struct shouldn't be send.
+   * It has to be the first variable within the structure. */
+  unsigned char propagate;
+  char *name;
+  ddsi_octetseq_t value;
+} dds_binaryproperty_t;
+
+typedef struct dds_binarypropertyseq {
+  uint32_t n;
+  dds_binaryproperty_t *props;
+} dds_binarypropertyseq_t;
+
+typedef struct dds_property_qospolicy {
+  dds_propertyseq_t value;
+  dds_binarypropertyseq_t binary_value;
+} dds_property_qospolicy_t;
+
 typedef struct dds_durability_qospolicy {
   dds_durability_kind_t kind;
 } dds_durability_qospolicy_t;
@@ -212,6 +245,7 @@ typedef struct dds_ignorelocal_qospolicy {
 #define QP_PRISMTECH_SUBSCRIPTION_KEYS       ((uint64_t)1 << 25)
 #define QP_PRISMTECH_ENTITY_FACTORY          ((uint64_t)1 << 27)
 #define QP_CYCLONE_IGNORELOCAL               ((uint64_t)1 << 30)
+#define QP_PROPERTY_LIST                     ((uint64_t)1 << 31)
 
 /* Partition QoS is not RxO according to the specification (DDS 1.2,
    section 7.1.3), but communication will not take place unless it
@@ -263,6 +297,7 @@ struct dds_qos {
   /*x xR*/dds_subscription_keys_qospolicy_t subscription_keys;
   /*x xR*/dds_reader_lifespan_qospolicy_t reader_lifespan;
   /* x  */dds_ignorelocal_qospolicy_t ignorelocal;
+  /*xxx */dds_property_qospolicy_t property;
 };
 
 struct nn_xmsg;
--- a/src/core/ddsi/src/q_plist.c
+++ b/src/core/ddsi/src/q_plist.c
@@ -94,7 +94,6 @@ enum pserop {
   XSTOP,
   XO, /* octet sequence */
   XS, /* string */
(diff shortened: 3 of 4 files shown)


martinbremmer:
Travis build https://travis-ci.org/eclipse-cyclonedds/cyclonedds/jobs/585656720 failed because of a test failure: ```mpt_ppuserdata_qos_rwtopicdataM```.
I can't relate that failure to my changes (travis on my fork succeeded). I also can't restart the build.

eboasson:
@martinbremmer I agree that test failure is very unlikely to have anything to do with your changes. Hmm ... flaky tests ... except in this case the test output itself is fine, so it is shutdown issue.

src/core/ddsi/src/q_plist.c
```diff
@@ -521,50 +562,74 @@ static dds_return_t deser_generic (void * __restrict dst, size_t * __restrict ds
   return DDS_RETCODE_BAD_PARAMETER;
 }
 
-static size_t ser_generic_size (const void *src, size_t srcoff, const enum pserop * __restrict desc)
+static void ser_generic_size_partial (size_t *dstoff, const void *src, size_t srcoff, const enum pserop * __restrict desc)
```
eboasson: I wouldn't call this "partial": the only other "partial" is in "fini" and there it is actually about doing part of the type description. This is "embeddable" or something ...

src/core/ddsi/src/q_plist.c
```diff
@@ -521,50 +562,74 @@ static dds_return_t deser_generic (void * __restrict dst, size_t * __restrict ds
...
+    return src->length;
+  /* and if it does start with XbPROP, only those for which it is true are included */
+  uint32_t count = 0;
+  for (uint32_t i = 0; i < src->length; i++)
+    if (src->value[i * elem_size])
+      count++;
+  return count;
+}
+
+static dds_return_t ser_generic_partial (char * const data, size_t *dstoff, const void *src, size_t srcoff, const enum pserop * __restrict desc)
```
eboasson: The same with the "partial".

src/core/ddsi/src/q_plist.c
```diff
@@ -717,21 +784,23 @@ static dds_return_t unalias_generic (void * __restrict dst, size_t * __restrict
...
       case Xb: case Xbx2: SIMPLE (Xb, unsigned char); break;
       case XbCOND: SIMPLE (XbCOND, unsigned char); break;
+      case XbPROP: SIMPLE (XbPROP, unsigned char); break;
       case XG: SIMPLE (XG, nn_guid_t); break;
       case XK: SIMPLE (XK, nn_keyhash_t); break;
+      case XQ: COMPLEX (XQ, ddsi_octetseq_t, if (x->length) {
+        size_t elem_off = 0;
+        x->value = ddsrt_memdup (x->value, x->length * ser_generic_srcsize(desc + 1));
+        for (uint32_t i = 0; i < x->length; i++) {
+          unalias_generic (x->value, &elem_off, desc + 1);
```
eboasson: Same offset calculation issues as in "fini"

src/core/ddsi/src/q_plist.c
```diff
@@ -951,6 +1027,15 @@ static dds_return_t dvx_reader_favours_ssm (void * __restrict dst, const struct
 }
 #endif
 
+/* pserop representation of dds_propertyseq_t. */
```
eboasson: I'm not so keen on doing these with macros, given that they are all used exactly once.

src/core/ddsi/src/q_plist.c
```diff
@@ -2228,7 +2314,7 @@ const unsigned char *nn_plist_findparam_native_unchecked (const void *src, nn_pa
   const nn_parameter_t *par = src;
   while (par->parameterid != pid)
   {
-    if (pid == PID_SENTINEL)
+    if (par->parameterid == PID_SENTINEL)
```
eboasson: Thanks for catching that one 😁

src/core/ddsi/src/q_plist.c
```diff
@@ -2763,6 +2849,29 @@ void nn_log_xqos (uint32_t cat, const struct ddsrt_log_cfg *logcfg, const dds_qo
   });
   DO (PRISMTECH_ENTITY_FACTORY, { LOGB1 ("entity_factory=%u", xqos->entity_factory.autoenable_created_entities); });
   DO (CYCLONE_IGNORELOCAL, { LOGB1 ("ignorelocal=%u", xqos->ignorelocal.value); });
+  DO (PROPERTY_LIST, {
```
eboasson: You know, I don't mind the addition of the property list here so much as that I mind my own laziness in not rewriting it when I added those tables ... most of all this special-purpose printing code could just be done using the same tables as the serialisation, comparison, &c.

It would be especially nice if it would print into a string buffer rather than using the tracing macros, because then the implementation of printing the content of parameter-list samples would become trivial. And the current printing of QoS values could just go away.

I don't think it reasonable to require all those changes in this PR, but if you were to decide to do it anyway, it would be appreciated!

src/core/ddsi/include/dds/ddsi/q_plist.h
```diff
@@ -243,6 +243,9 @@ struct nn_rdata;
 DDS_EXPORT unsigned char *nn_plist_quickscan (struct nn_rsample_info *dest, const struct nn_rmsg *rmsg, const nn_plist_src_t *src);
 DDS_EXPORT const unsigned char *nn_plist_findparam_native_unchecked (const void *src, nn_parameterid_t pid);
 
+DDS_EXPORT void nn_free_property_policy (dds_property_qospolicy_t *property);
```
eboasson: I presume you have a plan for these two functions?

src/core/ddsi/include/dds/ddsi/q_xqos.h
```diff
@@ -32,6 +32,39 @@ typedef ddsi_octetseq_t dds_userdata_qospolicy_t;
 typedef ddsi_octetseq_t dds_topicdata_qospolicy_t;
 typedef ddsi_octetseq_t dds_groupdata_qospolicy_t;
 
+typedef struct dds_property {
+  /* The propagate boolean will not be send over the wire.
+   * When the value is 'false', the complete struct shouldn't be send.
+   * It has to be the first variable within the structure. */
```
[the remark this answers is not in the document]
martinbremmer: Coming up.

src/core/ddsi/src/q_plist.c
```diff
@@ -344,17 +378,27 @@ static void fini_generic_partial (void * __restrict dst, size_t * __restrict dst
...
       case XG: SIMPLE (XG, nn_guid_t); break;
       case XK: SIMPLE (XK, nn_keyhash_t); break;
+      case XbPROP: SIMPLE (XbPROP, unsigned char); break;
+      case XQ:
+        /* non-nested, so never a need to deallocate only some of the entries and no complications
+           in locating the end of the sequence element description */
+        COMPLEX (XQ, ddsi_octetseq_t, {
+          size_t elem_off = 0;
+          for (uint32_t i = 0; i < x->length; i++) {
+            fini_generic_partial (x->value, &elem_off, desc + 1, desc_end, aliased);
```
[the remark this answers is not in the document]
martinbremmer: My mistake was thinking that the calls to deser_generic_src() takes care of that, but that's only for the alignment/padding elements within the structures, not 'between' them.  

The reason my manual test works is because the first element within the (binary)property is the string, causing a 4byte alignment automatically.

I've extended the manual test with the same kind of struct you proposed in your comment. That shows the problem clearly.

Element offset calculation revert coming up.

src/core/ddsi/src/q_plist.c
```diff
@@ -521,50 +562,74 @@ static dds_return_t deser_generic (void * __restrict dst, size_t * __restrict ds
   return DDS_RETCODE_BAD_PARAMETER;
 }
 
-static size_t ser_generic_size (const void *src, size_t srcoff, const enum pserop * __restrict desc)
+static void ser_generic_size_partial (size_t *dstoff, const void *src, size_t srcoff, const enum pserop * __restrict desc)
 {
-  size_t dstoff = 0;
+  bool suppressed = false;
```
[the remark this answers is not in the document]
martinbremmer: Will remove.

src/core/ddsi/include/dds/ddsi/q_plist.h
martinbremmer: Yes, but it makes more sense to introduce them when I use them (maybe they'll change slightly). So, I'll remove them from this PR.

src/core/ddsi/src/q_plist.c
martinbremmer: There's a similar PR in the works that add some msg parameters to plist (iso xqos). Maybe I'll do it in that PR. 

@eboasson, question though.
Currently, strings like "ignorelocal=", "lifespan=", etc are added to the log.
Don't you lose that kind of information when using the tables?

martinbremmer:
Updated test patch that contains the improved alignment/padding test.
[prop_test2.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3631357/prop_test2.patch.txt)

eboasson:
Thanks, @martinbremmer, it looks good to me. I do think it would be better to add a small test for this code (it is actually pretty easy to run the code on a set of different type descriptors, given a bit of refactoring to make the relevant functions and types externally visible). The primary reason is that I'd like to also have some test that proves erroneous input is flagged and discarded.

That's something that has been missing before and that bothered me a bit, so I propose I'll give that a shot based on this PR. If that works fine, I'll merge both.

martinbremmer:
@eboasson, ok, thanks.
Can you give a timeframe how long you think that'll take?
I just ask because a have two PRs (almost) on the shelf that depend on this one being merged. One of those could benefit from the testing you propose though.

eboasson:
@martinbremmer, I was thinking of doing it right now, I'm hoping it'll be sorted later today, but it might well run into tomorrow.

With all these changes to this code — and again, this is not caused by your changes but rather by my own change to a table-driven one — I just want to have a sanity check in there. I don't foresee any problems.

eboasson:
@martinbremmer: I spoke too soon ... my test caught a problem in the ``XbPROP`` equality test — as well as in the GUID and keyhash equality checks. Fortunately, the only equality tests it currently performs are on QoS settings, so the GUID and keyhash bugs are not affecting behaviour (yet).

I've pushed it all on a branch of mine: https://github.com/eboasson/cyclonedds/tree/plist-deser-test, that builds on this PR. There is a memory leak I still need to look into, and besides I want the test and my fixes on master as well, and that won't all happen today.

Are there any further changes to this code required for implementing DDS Security? Because if the addition of ``XbPROP`` is the only change required, then it might make the most sense to just move it into master already, instead of having to deal with a divergence.

martinbremmer:
@eboasson, the coming changes (that will impact this part of the code a bit) is how PIDs are used as index in ```piddesc_omg_index``` and the ```piddesc_omg``` is extended.
But no new ```pserop``` nor 'specialized' functions will be introduced.

eboasson:
Work integrated as part of #260.

PR #255, opened 2019-09-17
Add support for #line directive

k0ekk0ek:
This PR adds support for `#line` directives to `idlc` and fixes how `#pragma` directives are processed. Additionally `yy_decl.h` has been removed in favor of using a `%code provides` block and `ddsrt_strndup` has been added.

@FransFaase, can you take a look the the modifications to `tt_create.c`, or (if you feel like it) review the all the changes?


eboasson:
@k0ekk0ek I decided to have a look at this PR because it adds something valuable but seems to just sit here, ignored and unloved. I don't quite see why it would be so unloved, I don't see much that is wrong with it.

The only comment I have is that at least GCC appends some useful flags to the ``#line`` directives (after the filename), see, e.g., https://stackoverflow.com/questions/33089168/what-do-the-numbers-mean-in-the-preprocessed-i-files-when-compiling-c-with-gcc that allow tracking which file included which for nicer diagnostics. Without them, you could still make an informed guess as cyclical dependencies are quite uncommon :smile:

So if you are still happy with the PR, then please merge it.

src/ddsrt/src/string.c
```diff
@@ -162,3 +162,21 @@ ddsrt_strdup(
...
+
+char *
+ddsrt_strndup(const char *str, size_t len)
+{
+  char *s;
+  size_t n = 0;
+
+  assert(str != NULL);
+
+  while (str[n] != '\0' && n < len) { n++; }
```
e-hndrks: You could replace this line with a n =strnlen(str, len); which might be better optimized than your 1-character at a time algorithm.

k0ekk0ek:
@eboasson, I'm going to merge this. It indeed adds a nifty feature that's certainly needed in the future, at least for integration of the preprocessor. Since it specifically adds support for the C99 `#line` directive (because we use mcpp), I don't think the support for GCC specific line-directives is worth-delaying the merge for. If someone wants that feature, I'm certainly not going to object, but for now, I'm not going to implement it.

src/ddsrt/src/string.c
k0ekk0ek: I agree, but since it's an optimization, I think it can be tackled later.

PR #256, opened 2019-09-17
Switch from colcon.pkg to package.xml and add build-time dependencies.

rotu:
This allows pulling dependencies with rosdep and building with Colcon.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -6,11 +6,17 @@
     <description>Eclipse Cyclone DDS is a very performant and robust open-source DDS implementation. Cyclone DDS is developed completely in the open as an Eclipse IoT project.</description>
     <maintainer email="cyclonedds-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
     <license>Eclipse Public License 2.0</license>
+
     <url type="website">https://projects.eclipse.org/projects/iot.cyclonedds</url>
+    <url type="bugtracker">https://github.com/eclipse-cyclonedds/cyclonedds/issues</url>
+    <url type="repository">https://github.com/eclipse-cyclonedds/cyclonedds</url>
 
     <buildtool_depend>cmake</buildtool_depend>
-    <buildtool_depend>maven</buildtool_depend>
     <buildtool_depend>java</buildtool_depend>
+    <buildtool_depend>maven</buildtool_depend>
+    <depend>openssl</depend>
+    <test_depend>libcunit-dev</test_depend>
+    <doc_depend>python3-sphinx</doc_depend>
 
     <export>
         <build_type>cmake</build_type>
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>cyclonedds</name>
+    <version>0.1.0</version>
+    <description>Eclipse Cyclone DDS is a very performant and robust open-source DDS implementation. Cyclone DDS is developed completely in the open as an Eclipse IoT project.</description>
+    <maintainer email="cyclonedds-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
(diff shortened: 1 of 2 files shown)


rotu:
@dirk-thomas probably a better approach than vendoring this package in https://github.com/atolab/rmw_cyclonedds/pull/34
@cottsay If there was some reason to prefer the colcon.pkg, let me know.

package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>cyclonedds</name>
+    <version>0.1.0</version>
+    <description>Eclipse Cyclone DDS is a very performant and robust open-source DDS implementation. Cyclone DDS is developed completely in the open as an Eclipse IoT project.</description>
+    <maintainer email="cyclonedds-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
+    <license>Eclipse Public License 2.0</license>
+    <url type="website">https://projects.eclipse.org/projects/iot.cyclonedds</url>
```
dirk-thomas: It would be good to also add a bugtracker and repository URL: like https://github.com/ros/catkin/blob/461cd3c217f952f908a3cba56927660764b9ae62/package.xml#L13-L14

package.xml
```diff
@@ -0,0 +1,19 @@
...
+
+    <buildtool_depend>cmake</buildtool_depend>
```
dirk-thomas: Nitpick: we try to keep any kind of enumeration like these `buildtool_depend` in alphabetical order since it makes any kind of insert/remove deterministic and helps users visually browse the list.

package.xml
```diff
@@ -0,0 +1,19 @@
...
+    <buildtool_depend>maven</buildtool_depend>
+    <buildtool_depend>java</buildtool_depend>
```
dirk-thomas: Based on the dependencies needed for the CI jobs (https://github.com/ros2/ros_buildfarm_config/blob/ff009db2c0cc6ac61b9acd2e504676b460ca76af/eloquent/ci-nightly-cyclonedds.yaml#L13-L15) don't you also need `libcunit-dev`?

package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>cyclonedds</name>
+    <version>0.1.0</version>
```
dirk-thomas: I am not sure what versioning scheme you are aiming for. Maybe 0.1.0 might be confusing with the same number being advertised on the website being from March / the tag being from February.

Just a note - it can stay as is. Before a new release is made this needs to be bumped and the repo tagged.

rotu: I agree it should be updated. I pulled this version number from the top-level CMakeLists.txt

rotu:
> If you are willing to add a `package.xml` file to this repo this is certainly the preferred version.

Agreed. I figured there was some reason there wasn't one, but I'll let the repo owners explain that when they reject this PR ;-)

eboasson:
> > If you are willing to add a package.xml file to this repo this is certainly the preferred version.
> Agreed. I figured there was some reason there wasn't one, but I'll let the repo owners explain that when they reject this PR ;-)

😄

@rotu you're correct that the "repo owners" have an "issue" with this, but you might not be entirely correct in assuming they'll reject it. The issue @k0ekk0ek and I have with adding this ``package.xml`` file to the root of the repository is that ``package.xml`` is a very generic name and yet it is ROS-specific — well, that's what the ``http://download.ros.org/schema/package_format3.xsd`` implies. Fortunately, the contents of the file are a perfectly reasonable generic package descriptor.

So although including it doesn't seem quite right for a project with wider aims than primarily supporting ROS2, we're willing to add it. Perhaps one day it'll give rise to a conflict, but I am pretty sure that the cost of solving such a conflict if it arises won't be much higher than the cost today of avoiding the possibility of it ever arising. So that is insufficient reason to make life more miserable today.

PR #258, opened 2019-09-24
Fixes prompted by running Clang's static analyzer and Coverity

eboasson:
What the title subjects: there are a few *serious* bugs, like an inconsistency in locking/unlocking in not-invoking a "data on readers" listener, but most of the them are pretty benign — but needed to be addressed nonetheless, in my view anyway.

This set of changes also has a pretty radical effect on number of issues reported by Coverity and brings the issue density down to more acceptable levels.

I've kept them as separate commits for the benefit of the reviewer, but I propose to squash all trivial ones prior to merging.

### DIFF
--- a/src/core/ddsc/src/dds_entity.c
+++ b/src/core/ddsc/src/dds_entity.c
@@ -403,14 +403,17 @@ dds_return_t dds_delete_impl_pinned (dds_entity *e, enum delete_impl_state delst
     ddsrt_mutex_unlock (&p->m_mutex);
   }
 
-  /* Do some specific deletion when needed.  Bootstrapping and its inverse are always a
-     tricky business, and here it is no different: deleting the pseudo-top-level object
-     tears down all kinds of stuff that is supposed to remain in existence (like the
-     entire platform abstraction) and so it must be the final call.  Thus, we rely on it
-     to call "dds_entity_final_deinit_before_free" and return a special error code. */
+  /* Do some specific deletion when needed */
   ret = dds_entity_deriver_delete (e);
   if (ret == DDS_RETCODE_NO_DATA)
+  {
+    /* Bootstrapping and its inverse are always a tricky business, and here it is no different:
+       deleting the pseudo-top-level object tears down all kinds of stuff that is supposed to
+       remain in existence (like the entire platform abstraction) and so it must be the final
+       call.  Thus, we rely on it to call "dds_entity_final_deinit_before_free" and return a
+       special error code that we don't pass on to the caller. */
     ret = DDS_RETCODE_OK;
+  }
   else if (ret != DDS_RETCODE_OK)
   {
     if (parent_to_delete)
@@ -423,6 +426,8 @@ dds_return_t dds_delete_impl_pinned (dds_entity *e, enum delete_impl_state delst
     dds_free (e);
   }
 
+  assert (ret == DDS_RETCODE_OK);
+  (void) ret;
   return (parent_to_delete != NULL) ? dds_delete_impl_pinned (parent_to_delete, DIS_IMPLICIT) : DDS_RETCODE_OK;
 }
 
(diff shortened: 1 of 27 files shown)


src/core/ddsc/src/dds_entity.c
```diff
@@ -403,14 +403,16 @@ dds_return_t dds_delete_impl_pinned (dds_entity *e, enum delete_impl_state delst
...
 
-  /* Do some specific deletion when needed.  Bootstrapping and its inverse are always a
-     tricky business, and here it is no different: deleting the pseudo-top-level object
-     tears down all kinds of stuff that is supposed to remain in existence (like the
-     entire platform abstraction) and so it must be the final call.  Thus, we rely on it
-     to call "dds_entity_final_deinit_before_free" and return a special error code. */
+  /* Do some specific deletion when needed */
   ret = dds_entity_deriver_delete (e);
   if (ret == DDS_RETCODE_NO_DATA)
-    ret = DDS_RETCODE_OK;
```
k0ekk0ek: I trust you know what you're doing here as I'm still unfamiliar with the *pseudo-top-level* mechanism.

src/core/ddsc/src/dds_reader.c
```diff
@@ -127,13 +126,13 @@ void dds_reader_data_available_cb (struct dds_reader *rd)
...
       lst->on_data_on_readers (sub->m_hdllink.hdl, lst->on_data_on_readers_arg);
 
-      ddsrt_mutex_lock (&rd->m_entity.m_observers_lock);
       ddsrt_mutex_lock (&sub->m_observers_lock);
       sub->m_cb_count--;
       sub->m_cb_pending_count--;
       ddsrt_cond_broadcast (&sub->m_observers_cond);
     }
     ddsrt_mutex_unlock (&sub->m_observers_lock);
+    ddsrt_mutex_lock (&rd->m_entity.m_observers_lock);
```
k0ekk0ek: Looks fine to me. Probably the hierarchy is unimportant for *m_observers_lock*, but just wondering unlock of subscriber isn't better done after locking the reader? (the opposite would of course be true when locking the subscriber and unlocking the reader)

src/core/ddsc/src/dds_entity.c
eboasson: Trust but verify, always a good procedure. I have noticed there is still a problem with that "pseudo-top-level" mechanism in that a failed domain creation can result in an initialised library and that there are some cases where that would break things — they're edge cases, but that's not a reason to ignore them.

On reflection, I also disagree with the way delete and the entity reference count play together. If I call delete, the handle should be gone, simple as that. The only type of object that ever needs to linger is the topic (if it is still used in readers or writers), but even so, the handle can be closed. That confusion has been present from the initial commit, and it does play a role in the this edge case in dealing with this top-level object.

But I digress: the change here is because the analysers rightly complained that this is dead code, and so replacing it by a comment won't affect the behaviour. But there are other ways of addressing that issue, and the one where the assignment stays and ``ret`` is then actually used is better.

I'll change it.

src/core/ddsc/src/dds_reader.c
eboasson: You can go up the hierarchy, never down. (So the few functions that do need to go down, like pushing listeners or changing QoS jump through hoops to be able to resume iterating over the children after having unlocked the parent prior to locking a child.)

For once I did even add a comment 😀:
```C
  ddsrt_mutex_t m_observers_lock;   /* locking parent->...->m_observers_lock while holding it is allowed */
```

No state of the reader is touched between unlocking the reader's ``m_observers_lock`` and re-locking it. Indeed, the entire function is more-or-less a no-op if you end up in the "data on readers" case. Locking the reader after unlocking the subscribers is ok as far as mutual exclusion on the reader's internal state is concerned.

Then there is the ordering of callbacks. Listener invocations on a single entity have always been serialised. This happens through the ``m_cb_count`` field, which is itself protected by ``m_observers_lock``. The reader has incremented its ``m_cb_count`` field, therefore no other thread will make it past the barrier at the top of the function until the end up of this function.

Finally — for what I can think of, I may be mistaken — functions like setting listeners and status masks use the same counts to ensure they are serialised with respect to listener invocations, so such changes won't be happening either.

And thus, I can't think of observable behavioural differences depending on whether ``m_observers_lock`` is re-locked after unlocking the subscriber, or before re-locking the subscriber. And then I think the current structure is cleaner.

PR #260, opened 2019-09-26
parameter list update: properties, fixes, unit tests

eboasson:
This PR is a based on #254, which introduced the property list parameter, but:
* extended with some unit tests to exercise things a bit,
* a few bug fixes to the parameter list handling (caught by these unit tests, some old, all fortunately are in paths not currently used in Cyclone),
* with a further modification to support an optional "tail" of fields so that one can have some required fields followed by some that need not be present, and
* ported from the "security" to the "master" branch.

Regarding the change of destination branch: the tests and the fixes are valuable today, and the addition of the property lists is trivial given the improvements in the general machinery. So it altogether seems to make the most sense to replace #254 by this PR, and to then merge master into the security branch to get these features. That seems like it would be less bookkeeping.

@martinbremmer do you agree? And would you be willing to go over these changes once more for a sanity check?

### DIFF
--- a/src/core/ddsi/src/q_plist.c
+++ b/src/core/ddsi/src/q_plist.c
@@ -98,10 +98,10 @@ struct piddesc {
   size_t plist_offset;   /* offset from start of nn_plist_t */
   size_t size;           /* in-memory size for copying */
   union {
-    /* descriptor for generic code: 11 is enough for the current set of
+    /* descriptor for generic code: 12 is enough for the current set of
        parameters, compiler will warn if one ever tries to use more than
-       will fit; on non-GCC/Clang and 32-bits machines */
-    const enum pserop desc[11];
+       will fit */
+    const enum pserop desc[12];
     struct {
       dds_return_t (*deser) (void * __restrict dst, size_t * __restrict dstoff, struct flagset *flagset, uint64_t flag, const struct dd * __restrict dd, size_t * __restrict srcoff);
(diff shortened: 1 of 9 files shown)


src/core/ddsi/src/q_plist.c
```diff
@@ -1091,7 +1140,7 @@ static const struct piddesc piddesc_omg[] = {
   QP  (DEADLINE,                            deadline, XD),
   QP  (LATENCY_BUDGET,                      latency_budget, XD),
   QP  (LIVELINESS,                          liveliness, XE2, XD),
-  QP  (PROPERTY_LIST,                       property, XQ, XbPROP, XS, XS, XSTOP, XQ, XbPROP, XS, XO, XSTOP),
+  QP  (PROPERTY_LIST,                       property, XQ, XbPROP, XS, XS, XSTOP, Xopt, XQ, XbPROP, XS, XO, XSTOP),
```
martinbremmer: Maybe add a small comment that Xopt is necessary for the property_list difference between DDS RTPS and DDS Security?

martinbremmer:
@eboasson, I don't have a problem with the #260->master->security route and closing #254 when that's done.

I didn't find a issue with the code either (except for one nitpicking thing about a comment).

src/core/ddsi/src/q_plist.c
eboasson: Good point

PR #261, opened 2019-09-27
Merge master into security

martinbremmer:
Merging master into security branch, mostly to get the property policy upgrade (see #254 and #260).


martinbremmer:
Same eclipse error as previous master->security merge:
```Anonymous (45659984+dennis-adlink@****.noreply) is not covered by the necessary legal agreements in order to proceed.```

eboasson:
The failing ECA check is caused by GitHub putting a user name into a squash-and-merge commit ``Author`` heading instead of using the original author. The originals were fine when I squashed them, so merging.

PR #263, opened 2019-09-27
Added DDS Security msg parameters.

martinbremmer:
Added remaining DDS Security related message parameters.

Operations on the message parameters are done with the available pserop operations.

However, the Security related PID values start at 0x1001. At some points in the plist code, pids are used as indexes into an array. This would mean that the array suddenly has to get a lot larger and most of it would be unused. Therefore, the Security related PIDs are translated to indexes that are just behind the 'normal' PID indexes.

I've manually tested it as an extra sanity check: 
[plist_test.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3663308/plist_test.patch.txt)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,30 +133,32 @@ windows_vs2017: &windows_vs2017
 jobs:
   include:
     - <<: *linux_gcc8
-      env: [ ARCH=x86_64, ASAN=none, BUILD_TYPE=Debug, SSL=YES, GENERATOR="Unix Makefiles", COVERITY_SCAN=true ]
+      env: [ ARCH=x86_64, ASAN=none, BUILD_TYPE=Debug, SSL=YES, SECURITY=YES, GENERATOR="Unix Makefiles", COVERITY_SCAN=true ]
       if: type = cron
     - <<: *linux_gcc8
-      env: [ ARCH=x86_64, ASAN=none, BUILD_TYPE=Debug, SSL=YES, GENERATOR="Unix Makefiles" ]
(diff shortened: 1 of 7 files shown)


eboasson:
Thanks @martinbremmer for continuing the process of getting the bits and pieces needed for implementing security in. I agree with the PR for the most part, but there are a few details where I would like to see it improved:

* I would prefer ``IS_SECURITY_PID`` (and similar) to be implemented as inline functions (simply because they are typed);
* It is important not only that Cyclone gains support for DDS Security, but also that it can be built without: for platforms where openssl is not available and for cases where the footprint matters most. It would therefore be best to make all the security-specific parts conditional on, e.g., ``DDSI_INCLUDE_SECURITY`` (in the style of the various other "INCLUDE" macros — see, e.g., ``q_feature_check.h``).  Note that the "property list" implementation of the previous PR is on the border, but that one we will probably also want to start using outside security, so it makes sense to have that one in all builds.
* Obviously making security support optional requires building both variants on the CI system. We need to avoid the build matrix to grow exponentially, but one additional Linux/clang/address-sanitizer build would give most of the coverage.

martinbremmer:
@eboasson, I agree with your comments.

I have just one remark though. The DDS Security support is not dependant on OpenSSL. The plugins are, but the support is not. In principle, people could create their own plugins without the use of OpenSSL and DDSI should be able to support that (if security support is enabled).

martinbremmer:
I've mimicked the SSL compilation option for Security.

eboasson:
> The DDS Security support is not dependant on OpenSSL. The plugins are, but the support is not

Of course. Silly me!

PR #264, opened 2019-09-28
macOS pre-10.13 support

eboasson:
macOS 10.13 is the first version to provide fmemopen and so building Cyclone (with BUILD_TESTING enabled) fails on older versions. That's somewhat annoying given that ROS2 very much supports macOS 10.12 still. This PR simply turns the handful of logging tests that depend on it into no-ops on macOS 10.12 and older.

It also adds a missing ``#include <AvailabilityMacros.h>`` needed for properly detecting whether ``clock_gettime_nsec_np`` exists. That only affects macOS 10.11 and older.

This should fix https://github.com/ros2/ros2/pull/787#issuecomment-536115180

### DIFF
--- a/src/ddsrt/src/time/darwin/time.c
+++ b/src/ddsrt/src/time/darwin/time.c
@@ -13,6 +13,9 @@
 #include <errno.h>
 #include <time.h>
 #include <sys/time.h>
+#if __APPLE__
+#include <AvailabilityMacros.h>
+#endif
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_12
 #include <mach/mach_time.h>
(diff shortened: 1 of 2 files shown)


eboasson:
@k0ekk0ek about that review request: the bit I'm particularly interested in is whether you agree with turning those ``dds_log`` tests into no-ops in those one case. Obviously the more elegant solutions would be to ignore those tests in ``CUnit.cmake``, but that seems like it would require a fair amount of effort. Emulating ``fmemopen`` for this one case seems rather silly ...

The checks I tested by targetting different versions of the SDK.

k0ekk0ek:
I guess I can live with not testing it on versions before 10.13 (not that it would be hard to do to create something based on `funopen`). Tests are run on Travis and `FILE` handles aren't likely to change much I guess. Feel free to hit *merge* whenever you feel like it.

PR #265, opened 2019-10-01
Fix building tests on macOS 10.12

eboasson:
* Clang 9 does not yet define MAC_OS_X_VERSION_10_13

  Which breaks the test for macOS versions older than 10.13 when
  building on an older system.

### DIFF
--- a/src/ddsrt/src/time/darwin/time.c
+++ b/src/ddsrt/src/time/darwin/time.c
@@ -13,11 +13,9 @@
 #include <errno.h>
 #include <time.h>
 #include <sys/time.h>
-#if __APPLE__
 #include <AvailabilityMacros.h>
-#endif
 
-#if MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_12
+#if !(defined MAC_OS_X_VERSION_10_12 && MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_12)
 #include <mach/mach_time.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
@eboasson, if you must, I guess you can merge, but I don't like the chosen solution. I'd much prefer that you'd install Conan and install CUnit using that. I feel taking the quick route with things like this is a slippery slope (one more exception to cover and no guarantee that each build target has Conan). An installer exists for Windows, but does not for macOS. The Homebrew installation on the xcode9 images is probably too old? Do you happen to know if Python is available on the image? If so, doing a shallow `git` clone of the Conan repo and setting the path could be a way to go...

eboasson:
@k0ekk0ek, I can’t say I’m surprised by your opinion — and I can’t say I disagree with you either. The problem is that there seem to be few good solutions. The options I have considered:
* the proper way, updating homebrew and using Conan is possible (certainly the first half is, the second is an educated guess) — but this increases the build times by a lot, and long build times are a constant source of annoyance;
* a Cron job for nightly builds — that would eliminate the issue I have with the duration of the builds, and would still ensure that breakage of macOS 10.12 builds go undetected for long, but wouldn’t avoid breaking things;
* this “slippery slope” one — I think there is a good chance it’ll be a while before it becomes a problem, but it sure is icky;
* Travis caching of homebrew + Conan might be an option, I haven’t investigated that one because I wanted to verify something today ...;
* I supposed a custom image might be possible — it seems a lot of work, but at least it would be an effort that would pay off in the long run.

I don’t see any need to merge the Travis configuration shortly in any case. I do think it would be good to get the code changes in “soon” (like tomorrow) because it needs to be picked up by the ROS2 nightly builds to get confirmation that it solves the one problem those changes are meant to solve. It is easy enough to (re)arrange commits and PRs in such a way that only the code change goes in and yet have confirmation the code changes do result in a something that works.

And all that for what really is an obsolete platform ...

k0ekk0ek:
@eboasson, code changes look fine. Feel free to hit merge if you strip out the changes to the build instructions. Personally, I'd prefer the caching option if it solves the problem... but let's investigate/discuss that at a later time.

eboasson:
Reduced the PR to just the code changes, the proposed Travis changes are now in #266.

PR #266, opened 2019-10-02
macOS 10.12 Travis build

eboasson:
This PR is the part of #265 removed by reducing it to just the code changes needed for building the tests on macOS 10.12. That was done to be able to merge the code changes and make progress on automated ROS2 mac builds, without importing an arguably "overly pragmatic" approach to the CI builds.

Copy-paste from the relevant comment by @k0ekk0ek on #265:

> I don't like the chosen solution. I'd much prefer that you'd install Conan and install CUnit using that. I feel taking the quick route with things like this is a slippery slope (one more exception to cover and no guarantee that each build target has Conan). An installer exists for Windows, but does not for macOS. The Homebrew installation on the xcode9 images is probably too old? Do you happen to know if Python is available on the image? If so, doing a shallow git clone of the Conan repo and setting the path could be a way to go...

Hopefully a clean solution will be possible, but it looks like it may take some time to get there. For the time being I'm putting this icky, hacky solution in its own PR so that it doesn't get lost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,33 @@ osx_xcode10_3: &osx_xcode10_3
     - pyenv activate conan
     - pip install conan --upgrade
 
+# macOS 10.12 / Xcode 9 is the macOS platform for ROS2, and hence quite important.  Travis
+# supports it, but the homebrew version on that image is outdated, and updating it takes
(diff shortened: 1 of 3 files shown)


eboasson:
Obsolete because of #267

PR #268, opened 2019-10-02
Added DDS Security XML configuration elements.

martinbremmer:
DDS Security needs to be configured. For this, the following XML elements are added to the cyclonedds XML configuration:
```
Domain/DDSSecurity/Authentication/Library/#text
Domain/DDSSecurity/Authentication/Library[@path]
Domain/DDSSecurity/Authentication/Library[@initFunction]
Domain/DDSSecurity/Authentication/Library[@finalizeFunction]
Domain/DDSSecurity/Authentication/IdentityCertificate/#text
Domain/DDSSecurity/Authentication/IdentityCA/#text
Domain/DDSSecurity/Authentication/PrivateKey/#text
Domain/DDSSecurity/Authentication/Password/#text
Domain/DDSSecurity/Authentication/TrustedCADirectory/#text
Domain/DDSSecurity/AccessControl/Library/#text
Domain/DDSSecurity/AccessControl/Library[@path]
Domain/DDSSecurity/AccessControl/Library[@initFunction]
Domain/DDSSecurity/AccessControl/Library[@finalizeFunction]
Domain/DDSSecurity/AccessControl/PermissionsCA/#text
Domain/DDSSecurity/AccessControl/Governance/#text
Domain/DDSSecurity/AccessControl/Permissions/#text
Domain/DDSSecurity/Cryptographic/Library/#text
Domain/DDSSecurity/Cryptographic/Library[@path]
Domain/DDSSecurity/Cryptographic/Library[@initFunction]
Domain/DDSSecurity/Cryptographic/Library[@finalizeFunction]
```

Detailed explanations of these elements have been added to q_config.c.

The configuration values are added to the Property policy of the internal QoS, so that the plugins can get their settings (in the near future) through 'normal' means.

Note, the DDS Security specification states that these configuration values should be part of the Property policy within the Participant QoS in the API. The Property policy is part of the internal QoS, but not part of the API yet. This API extension of the QoS is planned to be added in the (near) future. See also https://github.com/martinbremmer/cyclonedds/blob/mpt/docs/dev/dds_security_effort.md#Move-configuration

### DIFF
--- a/src/core/ddsc/tests/config.c
+++ b/src/core/ddsc/tests/config.c
@@ -158,6 +158,12 @@ static bool patternmatch(const char* pattern, const char* string)
     }
 }
 
+/*
+ * The 'found' variable will contain flags related to the expected log
+ * messages that were received.
+ * Using flags will allow to show that when message isn't received,
+ * which one it was.
+ */
 static uint32_t found;
 static void logger(void *ptr, const dds_log_data_t *data)
 {
@@ -167,6 +173,33 @@ static void logger(void *ptr, const dds_log_data_t *data)
             found |= (uint32_t)(1 << i);
         }
     }
+}
+
+CU_Test(ddsc_config, security_non, .init = ddsrt_init, .fini = ddsrt_fini) {
+
+    /* There shouldn't be traces that mention security. */
+    const char *log_expected[] = {
+      "*Security*",
+      NULL
+    };
+
+    dds_entity_t participant;
+
+    /* Set up the trace sinks to detect the config parsing. */
+    dds_set_log_mask(DDS_LC_FATAL|DDS_LC_ERROR|DDS_LC_WARNING|DDS_LC_CONFIG);
+    dds_set_log_sink(&logger, (void*)log_expected);
+    dds_set_trace_sink(&logger, (void*)log_expected);
+
+    /* Create participant with an empty security element. */
+    found = 0;
+    ddsrt_setenv(URI_VARIABLE, "<Tracing><Verbosity>finest</></>");
+    participant = dds_create_participant(DDS_DOMAIN_DEFAULT, NULL, NULL);
+    ddsrt_setenv(URI_VARIABLE, "");
+    CU_ASSERT_FATAL(participant > 0);
+    dds_delete(participant);
+
+    /* No security traces should have been provided. */
+    CU_ASSERT_FATAL(found == 0x0);
 }
 
 CU_Test(ddsc_config, security_empty, .init = ddsrt_init, .fini = ddsrt_fini) {
@@ -243,8 +276,6 @@ CU_Test(ddsc_config, security_missing, .init = ddsrt_init, .fini = ddsrt_fini) {
 
     dds_entity_t participant;
 
-    ddsrt_setenv(URI_VARIABLE, sec_config);
-
     /* Set up the trace sinks to detect the config parsing. */
     dds_set_log_mask(DDS_LC_FATAL|DDS_LC_ERROR|DDS_LC_WARNING|DDS_LC_CONFIG);
     dds_set_log_sink(&logger, (void*)log_expected);
@@ -265,7 +296,7 @@ CU_Test(ddsc_config, security_missing, .init = ddsrt_init, .fini = ddsrt_fini) {
 #endif
 }
 
-CU_Test(ddsc_config, security, .init = ddsrt_init, .fini = ddsrt_fini) {
+CU_Test(ddsc_config, security_all, .init = ddsrt_init, .fini = ddsrt_fini) {
 
     /* Expected traces when creating participant with the security elements. */
     const char *log_expected[] = {
@@ -280,7 +311,7 @@ CU_Test(ddsc_config, security, .init = ddsrt_init, .fini = ddsrt_fini) {
       "config: Domain/DDSSecurity/Authentication/IdentityCA/#text: testtext_IdentityCA_testtext*",
       "config: Domain/DDSSecurity/Authentication/PrivateKey/#text: testtext_PrivateKey_testtext*",
       "config: Domain/DDSSecurity/Authentication/Password/#text: testtext_Password_testtext*",
-      "config: Domain/DDSSecurity/Authentication/TrustedCADirectory/#text:*",
+      "config: Domain/DDSSecurity/Authentication/TrustedCADirectory/#text: testtext_Dir_testtext*",
       "config: Domain/DDSSecurity/AccessControl/Library/#text: dds_security_ac*",
       "config: Domain/DDSSecurity/AccessControl/Library[@path]: dds_security_ac*",
       "config: Domain/DDSSecurity/AccessControl/Library[@initFunction]: init_ac*",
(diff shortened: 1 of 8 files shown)


src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,228 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
...
+    ddsrt_setenv(URI_VARIABLE, sec_config);
+
+    /* Set up the trace sinks to detect the config parsing. */
+    dds_set_log_mask(DDS_LC_FATAL|DDS_LC_ERROR|DDS_LC_WARNING|DDS_LC_CONFIG);
+    dds_set_log_sink(&logger, (void*)log_expected);
+    dds_set_trace_sink(&logger, (void*)log_expected);
+
+    /* Create participant with an empty security element. */
+    found = 0;
+    ddsrt_setenv(URI_VARIABLE, sec_config);
```
kurtuluso: How can it work without `<CycloneDDS><Domain>` parent elements?

src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,228 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
...
+    dds_entity_t participant;
+
+    /* Set up the trace sinks to detect the config parsing. */
+    dds_set_log_mask(DDS_LC_FATAL|DDS_LC_ERROR|DDS_LC_WARNING|DDS_LC_CONFIG);
+    dds_set_log_sink(&logger, (void*)log_expected);
+    dds_set_trace_sink(&logger, (void*)log_expected);
+
+    /* Create participant with an empty security element. */
+    found = 0;
+    ddsrt_setenv(URI_VARIABLE, "<DDSSecurity/>");
```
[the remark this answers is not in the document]
martinbremmer: It's never too late to learn :)

src/core/ddsc/tests/config.c
martinbremmer: I don't know. It's a configuration feature that 'just works'. I scanned the documentation, but could not find an explanation of the usage. Maybe @eboasson can explain a bit more about how the newer features of the configuration works?

src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,228 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
...
+
+    const char *sec_config =
+      "<Tracing><Verbosity>finest</></>"
+      "<DDSSecurity>"
+        "<Authentication>"
+          "<Library path=\"dds_security_auth\" initFunction=\"init_authentication\" finalizeFunction=\"finalize_authentication\" />"
+          "<IdentityCertificate>testtext_IdentityCertificate_testtext</IdentityCertificate>"
+          "<IdentityCA>testtext_IdentityCA_testtext</IdentityCA>"
+          "<PrivateKey>testtext_PrivateKey_testtext</PrivateKey>"
+          "<Password>testtext_Password_testtext</Password>"
```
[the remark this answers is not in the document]
martinbremmer: I will add a test.

src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,228 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
...
+    }
+}
+
+static uint32_t found;
+static void logger(void *ptr, const dds_log_data_t *data)
+{
+    char **expected = (char**)ptr;
+    for (uint32_t i = 0; expected[i] != NULL; i++) {
+        if (patternmatch(expected[i], data->message)) {
+            found |= (uint32_t)(1 << i);
```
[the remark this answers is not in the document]
martinbremmer: I will add a comment to the code.

src/core/ddsc/tests/config.c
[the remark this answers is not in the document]
martinbremmer: Extended the tests.

src/core/ddsc/tests/config.c
eboasson: @martinbremmer yes, that'd be worth documenting ... I suppose the existing documentation is limited to the PR descriptions ...

Anyway: yes, if an XML fragment from the environment doesn't start with ``<CycloneDDS>``, it will insert it; and if it then doesn't start with ``<Domain>`` it will insert ``<Domain id="any">``. In this mode, it will also accept ``</>`` as a universal closing tag and close any open elements on end-of-input.

It doesn't do any of that trickery when from a file.

src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,423 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
          "</"DDS_PROJECT_NAME">") == DDS_RETCODE_OK);
     CU_ASSERT_FATAL(dds_create_domain(2, "") == DDS_RETCODE_PRECONDITION_NOT_MET);
 }
+
+static bool patternmatch(const char* pattern, const char* string)
```
eboasson: The DDSI core has a function ``ddsi2_patmatch`` that does exactly this, I would be preferable to use that one instead. (Note: it'll have to be ``DDS_EXPORT``ed because of Windows ...)

src/core/ddsc/tests/config.c
```diff
@@ -118,3 +118,423 @@ CU_Test(ddsc_config, incorrect_config, .init = ddsrt_init, .fini = ddsrt_fini) {
...
+    dds_entity_t participant;
+
+    /* Set up the trace sinks to detect the config parsing. */
+    dds_set_log_mask(DDS_LC_FATAL|DDS_LC_ERROR|DDS_LC_WARNING|DDS_LC_CONFIG);
+    dds_set_log_sink(&logger, (void*)log_expected);
+    dds_set_trace_sink(&logger, (void*)log_expected);
+
+    /* Create participant with an empty security element. */
+    found = 0;
+    ddsrt_setenv(URI_VARIABLE, "<Tracing><Verbosity>finest</></>");
```
eboasson: Recently, a function named ``dds_create_domain`` was added that makes it possible to explicitly create a domain entity with the configuration passed in as a parameter. So if you use that one rather than ``dds_create_participant`` there is no longer a need to mess around with ``ddsrt_setenv``.

``dds_create_domain`` returns a handle that you can pass to ``dds_delete`` so the change is pretty trivial.

src/core/ddsi/src/q_entity.c
```diff
@@ -516,6 +516,69 @@ dds_return_t new_participant_guid (const ddsi_guid_t *ppguid, struct q_globals *
   nn_plist_copy (pp->plist, plist);
   nn_plist_mergein_missing (pp->plist, &gv->default_local_plist_pp, ~(uint64_t)0, ~(uint64_t)0);
 
+#ifdef DDSI_INCLUDE_SECURITY
```
eboasson: The ``new_participant_guid`` function (and not only that one, there are a bunch more with similar names) are already rather unwieldy. Adding 60-odd lines just to copy some properties makes it worse still. Would you mind refactoring this so that the code to fill the property list becomes a separate function?

src/core/ddsc/tests/config.c
martinbremmer: > `dds_create_domain` returns a handle that you can pass to `dds_delete`.

This is unfortunately not true (yet).

martinbremmer:
Two test failures:
1) CUnit_ddsrt_thread_create_and_join 
failed on ARCH=x86_64 ASAN=none BUILD_TYPE=Debug SSL=YES SECURITY=YES GENERATOR="Visual Studio 15 2017 Win64"
2) CUnit_ddsc_waitset_torture
failed on ARCH=x86_64 ASAN=address BUILD_TYPE=Debug SSL=YES SECURITY=YES GENERATOR="Unix Makefiles"

I can not relate those failures to my changes. Also, these platforms succeeded on my 'travis': https://travis-ci.com/martinbremmer/cyclonedds

eboasson:
@martinbremmer I agree that there is no clear relationship between your changes and those failures. Current hypothesis is that ``ddsrt_thread_create_and_join`` simply can take a very long time on a very busy machine, so we might try increasing the timeout.

The ``ddsc_waitset_torture`` one I have never seen before, but that definitely needs looking into. I'm rather happy that it failed this time (even with a bit of info), given that there is a problem there. I've saved the stack traces.

src/core/ddsi/src/q_plist.c
```diff
@@ -2760,6 +2760,55 @@ dds_qos_t * nn_xqos_dup (const dds_qos_t *src)
...
+    }
+    if (cfg->authentication_properties.trusted_ca_dir && (strlen(cfg->authentication_properties.trusted_ca_dir) != 0))
+    {
+      fill_property(&(properties[idx++]), "dds.sec.auth.trusted_ca_dir", cfg->authentication_properties.trusted_ca_dir);
+    }
+
+    /* Add properties. */
+    xqos->present |= QP_PROPERTY_LIST;
+    xqos->property.value.n = idx;
+    xqos->property.value.props = properties;
```
kurtuluso: There is a leak here if the property list is not empty.
We are not merging the properties, we are replacing the properties here.

dennis-adlink: @kurtuluso this issue is addressed in PR #278

PR #269, opened 2019-10-02
Improve state of option docs and remove (some) old cruft

eboasson:
I have done a little experiment with the documentation of the settings, resulting in a little perl script that can take the configuration processing code and generate a markdown file with an overview of all the options as well as a RELAX NG compact form (RNC) specification of the configuration file. That RNC file can be translated into an XSD with standard tools.

The configuration processing is updated to allow referencing an XSD (that it otherwise ignores), so that a modern XSD-aware editor can be used to edit the configuration file and get auto-completion and a reasonable level of validation. (Cyclone always validates its configuration, so the XSD really is meant as a tool to make editing easier, nothing more.) The example configuration in the README has been updated to include a reference to the XSD file.

The Java-based configuration editor is removed. There is no point in having it when it does a worse job than, e.g., Visual Studio Code does ...

This should all be generated automatically from the source, but that's true for all the documentation. It would also be more practical to generate XSD directly, rather than having to go through another conversion tool. Moreover, it would be better to use a python because python is available on all the CI build systems, but perl is not. And finally, it would actually be best to not take the current C code as a starting point ...

Finally, as the processing is rather simplistic and doesn't take into account which options are compiled in and which are not. As a consequence, the old, broken, not-to-be-trusted protection against casual eavesdroppers (i.e. it does encrypt, but it doesn't come close to securing communications) showed up rather prominently. With DDS Security around the corner I can no longer think of any reason — however weak — to keep the old code around. So this pull request deletes that.

Opinions welcome. If there is general agreement that this is a sensible intermediate step, it'll go in, but I suspect there might be some ideas on how to do it better. If anyone reading this feels like taking a shot at providing a fully automated generation pipeline, that would most welcome!


martinbremmer:
@eboasson, nitpicking, but shouldn't this https://github.com/eclipse-cyclonedds/cyclonedds/blob/e87f1a1dbf2cc13c9cb8b6a05226bb908f94f950/src/core/ddsi/src/q_config.c#L55 also be deleted? (I couldn't do an inline comment).

eboasson:
@martinbremmer yes, it should be deleted. Thanks for catching it.

i-and:
Hi @eboasson,
what extension for Visual Studio Code with XSD validation support do you recommend?

eboasson:
Hi @i-and, I generally use emacs and Xcode — don't know what that says about me 🤔 — and so I can't really recommend anything for VSCode. My only experience with the latter for this is using this extension:
```
Name: XML
Id: redhat.vscode-xml
Description: XML Language Support by Red Hat
Version: 0.9.0
Publisher: Red Hat
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml
```
which is certainly nice, with documentation pop-ups and auto-completion.

(As a — probably entirely superfluous — aside: the ``https://github.com/eclipse-cyclonedds/cyclonedds/etc/cyclonedds.xsd`` URL doesn't work yet because it is still just a pull-request. But you can use a ``file://...`` URL instead to point it to a local file.)

rvloon:
If you want, I can have it hosted at eclipse.cyclonedds.org so you could make it an actually official URL.

Ronald

On 03/10/2019 10:48:10, eboasson <notifications@github.com> wrote:
Hi @i-and [https://github.com/i-and], I generally use emacs and Xcode — don't know what that says about me 🤔 — and so I can't really recommend anything for VSCode. My only experience with the latter for this is using this extension:
Name: XML Id: redhat.vscode-xml Description: XML Language Support by Red Hat Version: 0.9.0 Publisher: Red Hat VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml
which is certainly nice, with documentation pop-ups and auto-completion.
(As a — probably entirely superfluous — aside: the https://github.com/eclipse-cyclonedds/cyclonedds/etc/cyclonedds.xsd URL doesn't work yet because it is still just a pull-request. But you can use a file://... URL instead to point it to a local file.)

i-and:
Thanks, @eboasson, at first glance, this extension also seemed attractive to me )

PR #271, opened 2019-10-07
Fix compile error when using rusage functionality on FreeRTOS

ThijsSassen:
During testing of a FreeRTOS target i ran into compile issues, please allow the provided fix in.


eboasson:
Thanks!

Issue #270, opened 2019-10-06
Ignoring dds_fatal() and DDS_[IC]LOG(DDS_LC_FATAL) processing and a few other things


i-and:
1. There is no emergency exit of the application when calling DDS_FATAL(). 
The situation is the same with DDS_CLOG(DDS_LC_FATAL), provided that the 
Tracing/Verbosity parameter takes the default value (gv.config.tracemask=0). 
Apparently this is due to the fact that the log file.c by default, only 
DDS_LC_ERROR | DDS_LC_WARNING events are allowed globally:
~~~
static struct ddsrt_log_cfg_impl logconfig = {
  .c = {
    .mask = DDS_LC_ERROR | DDS_LC_WARNING,
    .tracemask = 0,
    .domid = UINT32_MAX
  },
  .sink_fps = {
    [LOG] = NULL,
    [TRACE] = NULL
  }
};
~~~

As a solution, I suggest adding the value DDS_LC_FATAL to the 'mask' initialization.

2. File ddsi_iid.c: 
~~~
void ddsi_iid_init (void)
{
  union { uint64_t u64; uint32_t u32[2]; } tmp;

#if ! DDSRT_ATOMIC64_SUPPORT
  ddsrt_mutex_init (&dds_iid.lock);
#endif

  for (size_t i = 0; i < sizeof (dds_iid.key) / sizeof (dds_iid.key[0]); i++)
    dds_iid.key[0] = ddsrt_random ();
...
~~~

2.1 Is there incomplete key(dds_iid.key[]) initialization with random values above?
2.2 Taking into account the fact that ddsrt includes emulation of 64-bit operations, 
it would be logical to exclude the use of DDSRT_ATOMIC64_SUPPORT (there are also q_entity.h, ddsi_uud.h).

3. In file attributes.h:
~~~
#if __clang__
# define ddsrt_gnuc (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
#else
# define ddsrt_gnuc (0)
#endif

#if __GNUC__
# define ddsrt_clang (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
#else
# define ddsrt_clang (0)
#endif
~~~

Perhaps '#if \__GNUC__' and '#if \__clang__' need to be swapped.

eboasson:
Thanks @i-and !

[...]

PR #272, opened 2019-10-08
DDS_FATAL, atomic64, iid init, ddsrt attributes

eboasson:
This PR addresses the various sub-issues reported in #270:

* DDS_FATAL not aborting in the default configuration;
* Proper initialisation of the instance id generator;
* Removal of code variants for platforms that do not support 64-bit atomic operations, now that ``ddsrt`` emulates them;
* Defining ``ddsrt_clang`` and ``ddsrt_gnuc`` to the version numbers of clang respectively gcc, rather than the reverse ...

There is now a test for DDS_FATAL aborting, but that one only actually tests on Linux and macOS because of the vagaries of testing for abnormal program termination.

### DIFF
--- a/src/core/ddsc/src/dds_domain.c
+++ b/src/core/ddsc/src/dds_domain.c
@@ -51,7 +51,6 @@ static const ddsrt_avl_treedef_t dds_domaintree_def = DDSRT_AVL_TREEDEF_INITIALI
 static dds_return_t dds_domain_init (dds_domain *domain, dds_domainid_t domain_id, const char *config)
 {
   dds_return_t ret = DDS_RETCODE_OK;
-  char * uri = NULL;
   uint32_t len;
   dds_entity_t domain_handle;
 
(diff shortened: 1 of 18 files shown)


eboasson:
@k0ekk0ek in anticipation of review comments, I've spent some time trying to get the ``DDS_FATAL`` test to work on Windows — msdn says that catching abort with a signal handler is possible — but it doesn't seem to work in all cases (it works well enough on Travis but not on AppVeyor, oddly enough). So I propose leaving it as is for now.

This being yet another case where a test needs to be disabled on some platforms, I think I'll have a look at https://github.com/eclipse-cyclonedds/cyclonedds/blob/0b57b74b93c2938249431950b64068e1de46d096/cmake/Modules/CUnit.cmake#L30-L42 and extend it to so we can skip such tests rather than claim they pass.

k0ekk0ek:
@eboasson, most of the changes seem just fine. The clang vs gcc is especially amusing :smile:. I'm fine with not testing `DDS_FATAL` on Windows. Feel free to merge.

PR #273, opened 2019-10-09
Participant generic message

martinbremmer:
This message will be used by various parts of DDS Security, but mainly the handshaking.

The DDSI native (de)serializer had to be extended to support the message.
- Added support for int64_t.
- Added support for nested sequences.

Then the Generic Participant Message pserop array and support functions were added.

When creating (or init) such a message, the given arguments are aliased.
When the message is filled through deserialization, it's also aliased.
This means we can use the deinit on a message, regardless of its contents source.

### DIFF
--- a/src/core/ddsi/CMakeLists.txt
+++ b/src/core/ddsi/CMakeLists.txt
@@ -123,6 +123,17 @@ PREPEND(hdrs_private_ddsi "${CMAKE_CURRENT_LIST_DIR}/include/dds/ddsi"
 target_sources(ddsc
   PRIVATE ${srcs_ddsi} ${hdrs_private_ddsi})
 
+if(ENABLE_SECURITY)
+  PREPEND(security_srcs "${CMAKE_CURRENT_LIST_DIR}/src"
+          q_security_msg.c)
+
+  PREPEND(security_hdrs "${CMAKE_CURRENT_LIST_DIR}/include/dds/ddsi"
+          q_security_msg.h)
+
+  target_sources(ddsc
+    PRIVATE ${security_srcs} ${security_hdrs})
+endif()
+
 target_include_directories(ddsc
   PRIVATE "${CMAKE_CURRENT_LIST_DIR}/include")
 
--- a/src/core/ddsi/CMakeLists.txt
+++ b/src/core/ddsi/CMakeLists.txt
@@ -125,10 +125,10 @@ target_sources(ddsc
 
 if(ENABLE_SECURITY)
   PREPEND(security_srcs "${CMAKE_CURRENT_LIST_DIR}/src"
-          q_security_msg.c)
+          ddsi_security_msg.c)
 
   PREPEND(security_hdrs "${CMAKE_CURRENT_LIST_DIR}/include/dds/ddsi"
-          q_security_msg.h)
+          ddsi_security_msg.h)
 
   target_sources(ddsc
     PRIVATE ${security_srcs} ${security_hdrs})
--- a/src/core/ddsi/include/dds/ddsi/ddsi_plist_generic.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_plist_generic.h
@@ -34,6 +34,7 @@ enum pserop {
   Xi, Xix2, Xix3, Xix4, /* int32_t, 1 .. 4 in a row */
   Xu, Xux2, Xux3, Xux4, Xux5, /* uint32_t, 1 .. 5 in a row */
   XD, XDx2, /* duration, 1 .. 2 in a row */
+  Xll,      /* int64_t */
   Xo, Xox2, /* octet, 1 .. 2 in a row */
   Xb, Xbx2, /* boolean, 1 .. 2 in a row */
   XbCOND, /* boolean: compare to ignore remainder if false (for use_... flags) */
@@ -51,6 +52,15 @@ inline bool pserop_seralign_is_1 (enum pserop op) {
      operator */
   assert (op != XbPROP && op != Xopt && op != XSTOP);
   return (op >= Xo && op <= XK);
+}
+
+inline bool pserop_seralign_is_8 (enum pserop op) {
+  /* NB: XbPROP is never serialized, so its alignment is irrelevant.  If ever there
+     is a need to allow calling this function when op = XbPROP, it needs to be changed
+     to taking the address of the pserop, and in that case inspect the following
+     operator */
+  assert (op != XbPROP && op != Xopt && op != XSTOP);
+  return (op == Xll);
 }
 
 DDS_EXPORT void plist_fini_generic (void * __restrict dst, const enum pserop *desc, bool aliased);
--- a/src/core/ddsi/include/dds/ddsi/ddsi_plist_generic.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_plist_generic.h
@@ -34,7 +34,7 @@ enum pserop {
   Xi, Xix2, Xix3, Xix4, /* int32_t, 1 .. 4 in a row */
   Xu, Xux2, Xux3, Xux4, Xux5, /* uint32_t, 1 .. 5 in a row */
   XD, XDx2, /* duration, 1 .. 2 in a row */
-  Xll,      /* int64_t */
+  Xl,       /* int64_t */
   Xo, Xox2, /* octet, 1 .. 2 in a row */
   Xb, Xbx2, /* boolean, 1 .. 2 in a row */
   XbCOND, /* boolean: compare to ignore remainder if false (for use_... flags) */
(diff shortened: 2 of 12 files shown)


src/core/ddsi/src/q_plist.c
```diff
@@ -310,6 +339,21 @@ static dds_return_t fini_locator (void * __restrict dst, size_t * __restrict dst
   return 0;
 }
 
+static const enum pserop* goto_sequence_end (const enum pserop * __restrict desc, const enum pserop * const desc_end)
```
eboasson: I'm not sold on this one (yet): the loops I put in to scan for the end of the (then still non-nested) sequence were never the prettiest thing. Perhaps you could consider changing it to, e.g., ``pserop_advance`` that returns ``desc+1`` if ``*desc==XQ`` and otherwise does what you do here; and then use that throughout instead of the ``desc++`` it uses now in all the loops?

(Do let me know if this description makes no sense!)

src/core/ddsi/src/q_plist.c
```diff
@@ -382,14 +428,14 @@ static void fini_generic_embeddable (void * __restrict dst, size_t * __restrict
       case XQ:
         /* non-nested, so never a need to deallocate only some of the entries and no complications
```
eboasson: The "non-nested" bit is no longer true

src/core/ddsi/src/q_plist.c
```diff
@@ -382,14 +428,14 @@ static void fini_generic_embeddable (void * __restrict dst, size_t * __restrict
...
            in locating the end of the sequence element description */
-        COMPLEX (XQ, ddsi_octetseq_t, {
+        COMPLEX (XQ, ddsi_octetseq_t, (void)0, {
```
eboasson: Two things: firstly, I suspect that it would be clearer if the macro expansion were done by hand here and all superfluous code were to be removed. Secondly, it is a bit unfortunate that it will end up processing trying to free each element, if it is a list of, say, integers. To some extent, that's unavoidable ... but there might be a nicer solution. Perhaps it would work out nicely if it were to return whether it had to any work and bail out early in this loop if there was nothing to be done.

src/core/ddsi/src/q_plist.c
```diff
@@ -563,13 +617,18 @@ static dds_return_t deser_generic (void * __restrict dst, size_t * __restrict ds
           }
```
eboasson: This is a (potential) memory leak now: if you have a sequence of sequence and you have to abort on the second or later elements of an outer sequence, memory allocated for the preceding elements of the outer sequence is leaked.

E.g., XQ,XQ,Xu,XSTOP,XSTOP and an (invalid) input of (uint32_t[]){2,1,31415,3}. You allocate the outer sequence buffer, then you recurse for the element, which allocates the inner sequence buffer, reads 1 element (31415). Then you allocate memory for the inner sequence buffer of the second element and abort. This would free the first and third allocation, but not the second: because the incomplete element is not freed in ``fini_generic_embeddable``.

src/core/ddsi/tests/security_msg.c
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+#include "CUnit/Theory.h"
+#include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/string.h"
+#include "dds/ddsi/q_security_msg.h"
+
+#if DDSRT_ENDIAN == DDSRT_BIG_ENDIAN
```
eboasson: Maybe put this in a separate header file so it needs to be defined only once?

eboasson:
> Actually, I would just declare an internal static array (of bytes), and put the actual alignment corresponding to the enum value in there. The only thing you need checking is an out of bounds condition, but otherwise you just return the appropriate index in the array. It doesn't get much more efficient than this. Unless of course there is enum-number trickery involved.

Which is the sort of solution I would definitely choose in Ada (the only mainstream programming language sensible enough to have properly typed array indices). But in C, that doesn't give you a compile-time check that you have covered all cases ...

Besides, I don't see why I would forgo on the compile time checks:

image

rvloon:
On 09/10/2019 18:46:17, eboasson <notifications@github.com> wrote:
Actually, I would just declare an internal static array (of bytes), and put the actual alignment corresponding to the enum value in there. The only thing you need checking is an out of bounds condition, but otherwise you just return the appropriate index in the array. It doesn't get much more efficient than this. Unless of course there is enum-number trickery involved.
Which is the sort of solution I would definitely choose in Ada (the only mainstream programming language sensible enough to have properly typed array indices). But in C, that doesn't give you a compile-time check that you have covered all cases ...
Besides, I don't see why I would forgo on the compile time checks:
Ronald van Loon: It doesn't cover the case where someone inadvertently would call f((enum x) -1) (yeah, that happens), and it uses 16 bytes instead of 4. Add in your own check, and you would end up with similar, but explicit code and a smaller footprint. Arguably, in this case it would not matter that much, but I don't really have that much faith in compilers doing the right thing, other than what I have told them to do. YMMV, as always.


[image] [https://user-images.githubusercontent.com/16984005/66502143-01226900-eac5-11e9-89e8-f4eb1f2a88a9.png]

eboasson:
@rvloon, thank you for trying to point out issues with my approach, but I'm afraid I have to correct you on a number of points:

> It doesn't cover the case where someone inadvertently would call f((enum x) -1) (yeah, that happens)

It happens, true, but it is a problem that is not very likely to happen in internal interfaces. And besides, there really is a bounds check in there.

> and it uses 16 bytes instead of 4. 

Compilers these days give you options ... ask gcc to optimise for size and it'll use bytes. Alternatively, use a return type of, e.g., ``unsigned char`` and you get bytes, too. (I do think the "optimisation" of using 4-byte entries instead of 1-byte entries is a mistake because I'm pretty sure the load-byte-and-zero-extend instruction is no more costly than a loading a word on x64 — but this is not universally true.)

> Add in your own check, and you would end up with similar, but explicit code and a smaller footprint. 

So ... no, it's all there. Moreover, inlining and constant propagation squeeze out a bit more, including replacing the function by a constant when circumstances allow. That latter bit is an optimisation they can't do when it is loaded from an array because of the vagaries of C. 

> Arguably, in this case it would not matter that much, but I don't really have that much faith in compilers doing the right thing, other than what I have told them to do. YMMV, as always.

One's mileage varies indeed ... I have had to write assembler for crazy machines because the compiler was no good ... But here, I'll think the compile-time checking combined with better optimisation opportunities with a roughly identical footprint makes for a pretty good combination.

src/core/ddsi/include/dds/ddsi/ddsi_plist_generic.h
```diff
@@ -34,6 +34,7 @@ enum pserop {
   Xi, Xix2, Xix3, Xix4, /* int32_t, 1 .. 4 in a row */
   Xu, Xux2, Xux3, Xux4, Xux5, /* uint32_t, 1 .. 5 in a row */
   XD, XDx2, /* duration, 1 .. 2 in a row */
+  Xll,      /* int64_t */
```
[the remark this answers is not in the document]
martinbremmer: In my brain, sizeof(long) == 4 and sizeof(long long) == 8. But reality doesn't always bent to the way my brain wants it to work.
I'll change it to ```Xl``` because I have to touch the code anyway.

src/core/ddsi/CMakeLists.txt
```diff
@@ -123,6 +123,17 @@ PREPEND(hdrs_private_ddsi "${CMAKE_CURRENT_LIST_DIR}/include/dds/ddsi"
 target_sources(ddsc
   PRIVATE ${srcs_ddsi} ${hdrs_private_ddsi})
 
+if(ENABLE_SECURITY)
```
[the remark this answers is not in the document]
martinbremmer: I'll rename the files.

src/core/ddsi/include/dds/ddsi/q_security_msg.h
```diff
@@ -0,0 +1,112 @@
...
+
+/*
+ * Aliased struct variables will not be freed.
+ */
+DDS_EXPORT void
+nn_participant_generic_message_deinit(
+   nn_participant_generic_message_t *msg);
+
+/*
+ * The arguments are aliased in the resulting message structure.
```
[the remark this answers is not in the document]
martinbremmer: The ```new``` is used in OpenSplice, but we can very easily not use it, making having only an ```init``` enough.

eboasson:
@martinbremmer please accept my apologies for the delay. I'm happy with the changes you made, although there is one thing I don't generally like and would kindly request that you change: that's the use of macros where a function would fit just as well. So, ``align4``, ``ser_generic_align4`` and the ``...8`` variants.

Don't get me wrong: I am all for creative and clever use of macros, where it significantly reduces boilerplate or repetitive code. But for what really is a simple function (like these) it is almost invariably better to use functions and get type checking and other goodies. It'll get inlined anyway.

I would appreciate it if you would make those small changes.

martinbremmer:
CUnit_ddsrt_thread_create_and_join
failed on
 ARCH=x86_64 ASAN=none BUILD_TYPE=Release SSL=YES SECURITY=YES GENERATOR="Visual Studio 15 2017 Win64"

eboasson:
Thanks for making that change! ☺️ I don't see why:

> CUnit_ddsrt_thread_create_and_join
> failed on
> ARCH=x86_64 ASAN=none BUILD_TYPE=Release SSL=YES SECURITY=YES GENERATOR="Visual Studio 15 2017 Win64"

should prevent my merging it now.
 
> I'll merge master, where is should be fixed, into security again: see #279.

Great

PR #276, opened 2019-10-14
Increased time-out for the ddsrt_thread_create_and_join test. 

dennis-adlink:
This test runs on Travis Windows x86 often for up to 4-6 seconds. In case of busy build machines it may exceed the time-out of 10 seconds (because of testing with lowest thread priority), so I've increased the time-out to 60s. 


k0ekk0ek:
Looks fine to me. Thanks!

PR #278, opened 2019-10-17
Add PropertyPolicy to QoS API for Security settings

dennis-adlink:
This PR adds the public API for PropertyQosPolicy including tests. This policy can be used to set the parameters for the DDS security implementation, as an alternative for using the xml configuration. Tests are also inlcuded for setting security properties and conflict resolving when both security configuration and qos properties are present. Finally, the pubsub tool is updated so that is handles this qos correctly.


src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,152 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
   qos->present |= QP_CYCLONE_IGNORELOCAL;
 }
 
+void dds_qset_prop (dds_qos_t * __restrict qos, const char * __restrict name, const char * __restrict value)
+{
+  bool found = false;
+  if (qos == NULL || name == NULL || value == NULL)
+    return;
+
+  if (!(qos->present & QP_PROPERTY_LIST) || qos->property.value.n == 0)
```
eboasson: There's no need to try and squeeze every last cycle out of these functions that can be squeezed out, so it makes more sense to have a general function for initialising the property list and setting ``QF_PROPERTY_LIST`` if it isn't set already. Then that function can be used for both regular & binary ones.

Then, you also don't need to treat the empty list case differently from the non-empty case, making this function (and its binary equivalent) quite a bit simpler.

src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,152 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
...
+  if (qos == NULL || !(qos->present & QP_PROPERTY_LIST) || qos->property.value.n == 0 || name == NULL)
+    return;
+
+  for (uint32_t i = 0; i < qos->property.value.n; i++)
+  {
+    if (strcmp (qos->property.value.props[i].name, name) == 0)
+    {
+      dds_free (qos->property.value.props[i].name);
+      dds_free (qos->property.value.props[i].value);
+      if (qos->property.value.n > 1)
```
eboasson: I wouldn't bother clearing the ``present`` flag when the last property is unset: an empty set is still a set. Whether reallocating memory is worth it is debatable ... but I also see no reason why it should be removed :)

src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,152 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
...
+    qos->present |= QP_PROPERTY_LIST;
+  }
+}
+
+void dds_qunset_prop (dds_qos_t * __restrict qos, const char * __restrict name)
+{
+  if (qos == NULL || !(qos->present & QP_PROPERTY_LIST) || qos->property.value.n == 0 || name == NULL)
+    return;
+
+  for (uint32_t i = 0; i < qos->property.value.n; i++)
```
eboasson: I would prefer using a lookup function that returns an index (or "not found" in some way) — and then use that in the setters as well.

src/core/ddsc/include/dds/ddsc/dds_public_qos.h
```diff
@@ -680,6 +746,68 @@ dds_qget_ignorelocal (
   const dds_qos_t * __restrict qos,
   dds_ignorelocal_kind_t *ignore);
 
+/**
+ * @brief Gets the names of the properties from a qos structure.
```
eboasson: The likelihood of mishaps is probably lower if it simply returns (0, ``NULL``) when the list is empty or missing, rather than leaving the out parameters untouched when it is missing.

src/core/ddsc/tests/qos.c
```diff
@@ -646,3 +649,214 @@ CU_Test(ddsc_qos, durability_service, .init=qos_init, .fini=qos_fini)
     CU_ASSERT_EQUAL_FATAL(p.max_samples_per_instance, g_pol_durability_service.max_samples_per_instance);
 }
 
+CU_Test(ddsc_qos, property, .init=qos_init, .fini=qos_fini)
+{
+    char * value = NULL;
+    char ** names = NULL;
+    uint32_t cnt = 0;
+    char * str;
+    const size_t len = 0x6400000; /* 100 mbyte */
```
eboasson: What's the point in allocation 100MB?

src/core/ddsc/tests/qos.c
```diff
@@ -646,3 +649,214 @@ CU_Test(ddsc_qos, durability_service, .init=qos_init, .fini=qos_fini)
...
+}
+
+CU_Test(ddsc_qos, bproperty, .init=qos_init, .fini=qos_fini)
+{
+    void * bvalue = NULL;
+    size_t size = 0;
+    char ** names = NULL;
+    uint32_t cnt = 0;
+    unsigned char * data;
+    const size_t data_sz = 0x6400000; /* 100 mbyte */
```
eboasson: Same here: why this much?

src/core/ddsi/src/q_plist.c
```diff
@@ -2768,44 +2768,62 @@ static void fill_property(dds_property_t *prop, const char *name, const char *va
...
-    dds_property_t *properties;
-    uint32_t idx = 0;
+  bool qos_cfg = false;
+  assert(cfg != NULL);
 
-    /* No property should be present yet. */
-    assert(!(xqos->present & QP_PROPERTY_LIST));
+  /* Add DDS Security configuration to the QoS as a Property policy,
+   * used by the security plugins to get their proper settings. */
+  if (!(xqos->present & QP_PROPERTY_LIST) || xqos->property.value.n == 0)
```
eboasson: On consideration, I think it is reasonable to enforce the settings from the configuration for any participant that does not set the security properties explicitly, because that makes it much harder to accidentally create a non-secured participant with a configuration that specifies security.

I do wonder why the presence of any security property is enough to override them all. What is the (intended) effect of a partial security policy? I would expect it would be rejected as an inconsistent QoS on creation of the participant. @martinbremmer, I suppose you may know off the top of your head if that is indeed the case? If not, why not?

This is something worth noting in the documentation or, at least, comments.

Apart from that, it would make more sense to
* initialise the properties to (0,``NULL``) if they're not set
* then (always) check for presence of the security properties (even if it were just initialised) — this is a trivial check because the list is empty
* then add the properties if they are needed

that shrinks the code while making it simpler.

martinbremmer: > What is the (intended) effect of a partial security policy? I would expect it would be rejected as an inconsistent QoS on creation of the participant. @martinbremmer, I suppose you may know off the top of your head if that is indeed the case?

Using security properties from a user QoS is new. We only supported security settings through the configuration.
When creating a participant, the security properties are (will be) used to load the plugins and initialize them. This will fail if the properties are not correct or missing[1]. This in turn will fail the creation of the participant. At least, that's what I had in mind.

[1] Parts will be tested by ddsi (mostly related to plugins loading), parts will be tested by the plugins (mostly related to certificates and such).

src/core/ddsc/src/dds_qos.c
dennis-adlink: I've added an init function, and adjusted the set functions to use it.

src/core/ddsc/src/dds_qos.c
dennis-adlink: I've removed the code for clearing the qp flag

src/core/ddsc/src/dds_qos.c
dennis-adlink: You're right, this simplifies the get/set/unset, thanks!

src/core/ddsc/include/dds/ddsc/dds_public_qos.h
dennis-adlink: Fixed

src/core/ddsc/tests/qos.c
dennis-adlink: I didn't realise that the value for `propagate` is currently always `0`, so the property name/values won't be serialized and sent over the wire and this 100 mbyte test does not make sense. I've removed this code from the test case.

src/core/ddsc/tests/qos.c
dennis-adlink: See above, code removed

src/core/ddsi/src/q_plist.c
dennis-adlink: I have simplified `nn_xqos_mergein_security_config` and added code in `new_participant_guid` to check if all required security parameters are set when using qos properties for security configuration. With this check added, I think it is reasonable to override the security config if any security property is present, because the check prevents creating an unsecured participant accidentally by using the reserved `dds.sec.` prefix in a property name.

src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,151 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
...
+  bool found = false;
+  if (qos == NULL || name == NULL || index == NULL || !(qos->present & QP_PROPERTY_LIST))
+    return false;
+  for (uint32_t i = 0; !found && i < qos->property.value.n; i++)
+  {
+    if (strcmp (qos->property.value.props[i].name, name) == 0)
+    {
+      *index = i;
+      found = true;
+    }
```
eboasson: Stylistically, I much prefer an early-out in such cases, eliminating the state and — more importantly — the mental state involved in avoiding the early-out. Styles are quite personal and early-outs can cause discussion almost as fierce as indentation or emacs-vs-vi, so all I'm asking is that you consider it in the future.

src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,151 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
...
+  }
+  return found;
+}
+
+static bool dds_qbprop_get_index (const dds_qos_t * qos, const char * name, uint32_t * index)
```
eboasson: I would merge the bulk of the two ``get_index`` functions: they are identical save for ``value.X`` vs ``binary_value.X``. That can be parameterised.

src/core/ddsc/src/dds_qos.c
```diff
@@ -335,6 +335,151 @@ void dds_qset_ignorelocal (dds_qos_t * __restrict qos, dds_ignorelocal_kind_t ig
...
+    qos->property.binary_value.props = dds_realloc (qos->property.binary_value.props,
+      (qos->property.binary_value.n + 1) * sizeof (*qos->property.binary_value.props));
+    qos->property.binary_value.props[qos->property.binary_value.n].propagate = 0;
+    qos->property.binary_value.props[qos->property.binary_value.n].name = dds_string_dup (name);
+    dds_qos_data_copy_in (&qos->property.binary_value.props[qos->property.binary_value.n].value, value, sz, false);
+    qos->property.binary_value.n++;
+  }
+}
+
+void dds_qunset_bprop (dds_qos_t * __restrict qos, const char * name)
```
eboasson: Similar to the ``get_index`` functions, I would combine the two ``unsets``. I'd probably do the same for the part of the ``set`` functions that extends the array, but I'm not entirely certain that would be worth the bother.

src/core/ddsc/src/dds_qos.c
```diff
@@ -565,3 +710,91 @@ bool dds_qget_ignorelocal (const dds_qos_t * __restrict qos, dds_ignorelocal_kin
...
+    return false;
+
+  props = (qos->present & QP_PROPERTY_LIST) && qos->property.value.n > 0;
+
+  if (n != NULL)
+    *n = props ? qos->property.value.n : 0;
+
+  if (names != NULL)
+  {
+    if (!props || qos->property.value.n == 0)
```
eboasson: ``qos->property.value.n == 0`` means ``props`` is false anyway, so that term serves no purpose.

src/core/ddsc/src/dds_qos.c
```diff
@@ -565,3 +710,91 @@ bool dds_qget_ignorelocal (const dds_qos_t * __restrict qos, dds_ignorelocal_kin
...
+  if (qos == NULL || name == NULL)
+    return false;
+
+  found = dds_qprop_get_index (qos, name, &i);
+  if (value != NULL)
+    *value = found ? dds_string_dup (qos->property.value.props[i].value) : NULL;
+  return found;
+}
+
+bool dds_qget_bpropnames (const dds_qos_t * __restrict qos, uint32_t * n, char *** names)
```
eboasson: Much like ... etc. — you can guess the rest 🙂

src/core/ddsi/src/q_entity.c
```diff
@@ -524,7 +524,24 @@ dds_return_t new_participant_guid (const ddsi_guid_t *ppguid, struct q_globals *
...
+                      DDS_SEC_PROP_ACCESS_PERMISSIONS_CA,
+                      DDS_SEC_PROP_ACCESS_GOVERNANCE,
+                      DDS_SEC_PROP_ACCESS_PERMISSIONS };
+
+      GVLOGDISC ("new_participant("PGUIDFMT"): using security settings from qos, ignoring configuration\n", PGUID (*ppguid));
+
+      /* check if all required security properties exist in qos */
+      for (size_t i = 0; i < sizeof(req) / sizeof(req[0]); i++)
+      {
+        if (!nn_xqos_has_prop (&pp->plist->qos, req[i], false))
```
eboasson: Is it really enough to only log this? I would expect it to refuse to create the participant because of, let's say, "inconsistent QoS settings".

src/core/ddsi/src/q_plist.c
```diff
@@ -2895,44 +2911,42 @@ static void fill_property(dds_property_t *prop, const char *name, const char *va
...
 
-    /* Fill properties. */
-    fill_property(&(properties[idx++]), "dds.sec.auth.identity_ca", cfg->authentication_properties.identity_ca);
-    fill_property(&(properties[idx++]), "dds.sec.auth.private_key", cfg->authentication_properties.private_key);
-    fill_property(&(properties[idx++]), "dds.sec.auth.identity_certificate", cfg->authentication_properties.identity_certificate);
-    fill_property(&(properties[idx++]), "dds.sec.access.permissions_ca", cfg->access_control_properties.permissions_ca);
-    fill_property(&(properties[idx++]), "dds.sec.access.governance", cfg->access_control_properties.governance);
-    fill_property(&(properties[idx++]), "dds.sec.access.permissions", cfg->access_control_properties.permissions);
+  /* check for existing security properties (name starts with dds.sec. conform DDS Security spec 7.2.4.1) */
+  if (!nn_xqos_has_prop (xqos, "dds.sec.", true))
```
eboasson: Stylistically, I much prefer conditionals where one branch is trivial and the other quite involved such that the short branch comes first. Like the early-outs, it reduces the mental state. But like the other comment, this one too is left to your discretion 🙂

src/core/ddsc/src/dds_qos.c
dennis-adlink: I agree to use early-out here because the function is only a few lines. I fixed this.

src/core/ddsc/src/dds_qos.c
dennis-adlink: Parameterisation to a single function is hard because of the two distinct types `dds_property_t` and `dds_binaryproperty_t`, this would required struct-length and offset of the name field as parameters to that function, which makes the code harder to understand. Because I also really didn't like the code duplication, I've created macros for get_index, unset and get_propnames

src/core/ddsc/src/dds_qos.c
dennis-adlink: Fixed

src/core/ddsc/src/dds_qos.c
dennis-adlink: See above. I didn't refactor the `set` functions, because I think it's not worth the effort here

src/core/ddsc/src/dds_qos.c
dennis-adlink: Fixed, see above

src/core/ddsi/src/q_entity.c
dennis-adlink: You're right, should return error code in that case. I've refactored the function and updated the test to check that participant creation fails in case of missing security properties.

src/core/ddsi/src/q_plist.c
dennis-adlink: Fixed

PR #279, opened 2019-10-17
Merge master into security

martinbremmer:
I've merged the master branch into the security branch.


eboasson:
@martinbremmer there are a number of conflicts (and also an oddity in ``.travis.yml`` that I just spotted looking at the conflicts, there is a duplicate linux 64-bit build with gcc!) that I could sort out just now, but I'm a little pressed for time.

Would you mind looking at them?

PR #280, opened 2019-10-19
Use PROJECT_NAME instead of CMAKE_PROJECT_NAME

rotu:
CMAKE_PROJECT_NAME refers to the top-level project name, not the most recent project. So any CMake project that pulls this in as a sub-project was in for a nasty surprise.
https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_NAME.html


rotu:
CI issues seem trivial, resolved in #281

eboasson:
There's always something to learn :smile: (especially with CMake) — thanks, @rotu!

PR #281, opened 2019-10-19
Fix CI not finding pip3

rotu:


eboasson:
Thanks for looking into this. This issue has bothered me for a while, but I hadn't gotten around to investigating it, I just noticed that deleting the cache and restarting the build fixed it. I'm not sure a single successful build is enough to declare victory, but it is certainly worth a try!

PR #282, opened 2019-10-19
Namespace the schema and add references in xml files

rotu:

### DIFF
--- a/etc/cyclonedds.xsd
+++ b/etc/cyclonedds.xsd
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified">
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns="https://cdds.io/config"
+           targetNamespace="https://cdds.io/config"
+           elementFormDefault="qualified">
   <xs:element name="CycloneDDS">
     <xs:annotation>
       <xs:documentation>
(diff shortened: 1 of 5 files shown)


rotu:
CI failures are because of config parsing problems. Will need to be fixed before this is merged

eboasson:
@rotu no worries about that — and as I wrote, I'm pretty sure those changes wouldn't affect the behaviour of the tests anyway. More interestingly, I have been looking at the parts missing in the XSD, and found (and fixed) the problem.

Perhaps you can take a look at https://github.com/eboasson/cyclonedds/tree/xsd-fix ? The key bit is of course that it adds the attributes missing because of a bug in my perl script, but I have found a way of adding the name spacing as well so that the automated conversion tools work.

I have been quite unhappy with the fact that the repository contains what are essentially generated files, but at the same time, having them easily available is quite useful and besides generating them depends on tools not everyone has (like perl, and worse "trang" for converting RNC to XSD).

So I have made an attempt at checking at build time that the committed documents are identical to a freshly generated set. I should at least be able to add the checking of ``cyclonedds.rnc`` and ``options.md`` on the Linux builds, and that means it becomes impossible to commit changes to the configuration file without also updating the reference docs.

However, the CMake that came out is rather ugly, and it also doesn't deal with the various directories too elegantly, so that needs some more work. Advice on how best to do it in "proper CMake style" would be much appreciated.

rotu:
I share your discomfort. Yes, the cmakelists could be better (it's not bad, though) but I don't know that it's the best solution. I would really like to reverse things so that `q_config.c` consumes instead of produces the schema.

eboasson:
> I would really like to reverse things so that q_config.c consumes instead of produces the schema.

That I wholeheartedly agree with! 🙂

eboasson:
Closing because of #313

PR #284, opened 2019-10-19
Rename header directory dds -> cyclonedds

### LINKED ISSUE #249: Headers installed at include/dds
Ubuntu 18.04
ros-dashing-cyclonedds/bionic,now 0.1.0-1bionic.20190829.220209 amd64 

CycloneDDS installs some of its headers at `/opt/ros/dashing/include/dds`. This makes it unclear what package these headers belong to.

It also prevent building TwinOaks CoreDX which also has a header at `include/dds/dds.h`. At compile time, this collision prevents CoreDX from building if you have installed `ros-dashing-cyclonedds`.

rotu:
Fix #249 
Code pointing to the old headers should still compile, but it will show an advisory message when built, suggesting to update #include paths.


i-and:
@eboasson,  maybe instead of "cyclonedds" use a shorter form? 

For example "#include "cyclonedds/ddsrt/log.h"" -> #include "cydds/ddsrt/log.h"".

rotu:
While better than "dds", I don't think "cydds" is as good an identifier. I want the top-level include folder to make it obvious what these files are from.

i-and:
From the point of view of the program code the prefix length exceeds the length of all subsequent components:
~~~C
#include "cyclonedds/ddsi/ddsi_iid.h"
#include "cyclonedds/ddsi/ddsi_tkmap.h"
#include "cyclonedds/ddsi/ddsi_serdata.h"
#include "cyclonedds/ddsi/ddsi_threadmon.h"
#include "cyclonedds/ddsi/q_entity.h"
#include "cyclonedds/ddsi/q_config.h"
#include "cyclonedds/ddsi/q_gc.h"
#include "cyclonedds/ddsi/q_globals.h"
~~~

For comparison the shortened version:
~~~C
#include "cydds/ddsi/ddsi_iid.h"
#include "cydds/ddsi/ddsi_tkmap.h"
#include "cydds/ddsi/ddsi_serdata.h"
#include "cydds/ddsi/ddsi_threadmon.h"
#include "cydds/ddsi/q_entity.h"
#include "cydds/ddsi/q_config.h"
#include "cydds/ddsi/q_gc.h"
#include "cydds/ddsi/q_globals.h"
~~~
Subjectively, it seems more organic. Although from the point of view of the file system, a directory named "cyclonedds" is certainly more informative. Are there any other naming options?

rvloon:
Actually,

I would like to propose - what I would say is - a more elegant solution. It also might speed up the build on slower systems.

It's a dependency approach. 

Each source-file (.c) has a public interface (.h), and potentially private or implementation-only interfaces (potentially in a different directory).

The public interface has potentially other dependencies, which need to be included first, and the same may apply to the private header file. The source file itself may have to pull in even more dependencies.

Keeping track of order of includes and even the path to the includes is cumbersome, tedious, error prone and inflexible. Worse, if you update or change the directory layout, you need to change all occurrences of the paths to point to the right directories.

Cue the dependency file. Every header-file uses special #defines to refer to the headerfiles it needs (you could extend this to be even more granular: just #define the types you need, but let's do the simpler case first)

So, if for example I need the ddsi_iid.h file, I would first create a dependency file:

#define D_CYCLONEDDS_DDSI_IID_H
#define D_CYCLONEDDS_DDSI_TKMAP_H

Similarly, for private headerfiles:

#define D_CYCLONEDDS_PRIVATE_SOMEPRIVATE_H

Then, the master-dependency file is as follows:

#ifdef D_SOMEHEADER_FILE
#define D_OTHER_DEPENDENCY_NEEDED_BY_THIS_FILE
...
#endif

The order is from more dependent to less dependent.

The second half is then:

#ifdef CYCLONEDDS_DDSI_IID_H
#include "<path/to/ddsi_iid.h"
#endif

and so on, for every define one include file.

This ensures that all headerfiles are included exactly once.
Of course, if multiple files depend on the same headerfile, the dependency will be specified multiple times, but this does not mean reparsing of the headerfile.

So, headerfile, will have corresponding .d file with dependencies (including the dependency on the headerfile itself)
Sourcefile will include the .d file, specify additional dependencies, then include the master .d file, which will ensure that all necessary headerfiles will be pulled in.

The nice thing is that it is possible to generate the master-file and the .d automatically from current source. I should have some scripts somewhere that do this, but I am sure Erik B would be able to whip something up in perl before I could find the stuff I wrote in 1991. You might find it in the source-distribution of Motif++ from the same era, if it still is around somewhere.

Ronald

Ronald
On 23/10/2019 22:21:56, i-and <notifications@github.com> wrote:
From the point of view of the program code the prefix length exceeds the length of all subsequent components:
#include "cyclonedds/ddsi/ddsi_iid.h" #include "cyclonedds/ddsi/ddsi_tkmap.h" #include "cyclonedds/ddsi/ddsi_serdata.h" #include "cyclonedds/ddsi/ddsi_threadmon.h" #include "cyclonedds/ddsi/q_entity.h" #include "cyclonedds/ddsi/q_config.h" #include "cyclonedds/ddsi/q_gc.h" #include "cyclonedds/ddsi/q_globals.h"
For comparison the shortened version:
#include "cydds/ddsi/ddsi_iid.h" #include "cydds/ddsi/ddsi_tkmap.h" #include "cydds/ddsi/ddsi_serdata.h" #include "cydds/ddsi/ddsi_threadmon.h" #include "cydds/ddsi/q_entity.h" #include "cydds/ddsi/q_config.h" #include "cydds/ddsi/q_gc.h" #include "cydds/ddsi/q_globals.h"
Subjectively, it seems more organic. Although from the point of view of the file system, a directory named "cyclonedds" is certainly more informative. Are there any other naming options?

rotu:
A couple thoughts:
1. The length of an include doesn't really matter as long as it fits on a line. You're using one line for includes anyway.
2. There are too many headers. I think reducing the number of public headers would do a lot for readability, especially since so few are needed: https://github.com/ros2/rmw_cyclonedds/search?q=%22include+dds%22&unscoped_q=%22include+dds%22 afaict most apps just need `dds.h`
3. I think a macro system is overkill. If something is dependent on include order, that seems like a bug in the headers.

rotu:
> I suspect that would be trickier with auto-generating the redirects, though.

It wouldn’t be tricky at all. Changing the library name *would* be hard, and I’m not sure how to do it gracefully

i-and:
> While we're on the subject of renaming things: @rotu and @i-and, wouldn't you agree that renaming the library to "cyclonedds" (so, e.g., libcyclonedds.so) makes sense?

Your proposed renaming of the library seems logical.

eboasson:
> Changing the library name would be hard, and I’m not sure how to do it gracefully. I do think it’s a good idea, though.

On Linux/macOS we could do a soft link, but on Windows I have no clue either. Those who use CMake properly never need to type the name, so if there is a way of deleting ``ddsc`` (or refusing to install if it exists) the consequence would be simply that old executables would fail to run.

I think that's a price worth paying for cleaning up the naming.

eboasson:
I finally accepted that there would have to be a big notice recommending people to throw away the "old" ddsc libraries and all would be well. Now, I find myself in a dispute with CMake once again.

Renaming the library is as straightforward as replacing essentially all occurrences of "ddsc" related to the library by "cyclonedds" in the CMake files. It seems like leaving an alias target like this:
```CMake
add_library(${PROJECT_NAME}::ddsc ALIAS cyclonedds)
```

(so basically the old one) should allow application CMake files to continue to refer to it as CycloneDDS::ddsc, but it doesn't seem to work. I must be misreading the CMake documentation:

> add_library(_name_ ALIAS _target_)
>
> Creates an Alias Target, such that _name_ can be used to refer to _target_ in subsequent commands. The _name_ does not appear in the generated buildsystem as a make target. The _target_ may not be a non-GLOBAL Imported Target or an ALIAS. ALIAS targets can be used as linkable targets and as targets to read properties from. They can also be tested for existence with the regular if(TARGET) subcommand. The <name> may not be used to modify properties of _target_, that is, it may not be used as the operand of set_property(), set_target_properties(), target_link_libraries() etc. An ALIAS target may not be installed or exported.

Or would it be ok to force everyone to update their CMake files? But if that's the case, then why even bother with providing forwarding header files for backwards compatibility?

Anyone has a better idea?

(See https://github.com/eboasson/cyclonedds/tree/pull-284)

PR #285, opened 2019-10-21
Removed duplicate build from travis

martinbremmer:
This was noticed (by @eboasson) when checking the master->security merge.


martinbremmer:
I've updated the file in github, which seems to confuse eclipse. I redid it: #286.

PR #286, opened 2019-10-21
Removed duplicate build from travis

martinbremmer:
This was noticed (by @eboasson) when checking the master->security merge: #279.


eboasson:
Thanks for fixing it, @martinbremmer.

PR #289, opened 2019-10-22
Stylistic CMake changes

rotu:
No functional impact intended

### DIFF
--- a/src/core/ddsc/CMakeLists.txt
+++ b/src/core/ddsc/CMakeLists.txt
@@ -87,11 +87,6 @@ generate_export_header(
     EXPORT_FILE_NAME "${CMAKE_CURRENT_BINARY_DIR}/include/dds/export.h"
 )
 
-target_include_directories(
-  ddsc PUBLIC
-    "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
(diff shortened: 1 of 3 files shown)


rotu:
Looks like the CI issues here are due to maybe CI caching some old versions of the python3 package?

```
$ rvm $brew_ruby do brew bundle --verbose --global

Warning: 'python3' formula is unreadable: python3: undefined method `undent' for #<String:0x00007f8b22a0aac8>

Warning: 'gpg-agent' formula is unreadable: undefined method `undent' for #<String:0x00007f8b235d03a8>

Error: python3: undefined method `undent' for #<String:0x00007f8b233762d8>
```

https://travis-ci.org/eclipse-cyclonedds/cyclonedds/jobs/601080579

eboasson:
Thanks @rotu! My CMake knowledge is pretty limited, all I can say is that it looks sensible and that I can't find any functional changes. So, I'm happy to merge it.

 > Looks like the CI issues here are due to maybe CI caching some old versions of the python3 package?

Either that, or it deletes something that should have been retained. Getting it build reliably on Travis on macOS 10.12 is proving challenging — and that's simply because the travis image is so far out of date that installing packages without updating homebrew is no longer possible, and updating it takes so long that it can cause Travis to timeout ...

But the 10.12 bit matters because of ROS2, so it has be investigated a bit further ...

PR #296, opened 2019-10-24
Protected discovery preparation.

martinbremmer:
This PR introduces protected discovery preparation. It does not handle the actual protection or handling any settings that determine if it should be protected or not (currently it's hardcoded to be not protected).
What it does introduce is the switching between the 'default' discovery endpoints and the 'security protected' discovery endpoints.

For this, all the the security endpoints are added (except the BuiltinParticipantVolatileMessageSecure ones, for reasons that will become apparent in the PR that'll add those).
For now, only the security endpoints that are related to the discovery are used. The others are just added.

Some of the function names are somewhat verbose (especially the ones starting with 'q_omg_security'). I don't mind changing them in the future. But for now, I'd like to retain them, so that comparisons between the ported and original code is easier.

Whether discovery through the security endpoints works or not, can be done by using the attached patch and then run the mpt_basic test manually.
[test.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3768018/test.patch.txt)


martinbremmer:
FYI, I accidentally pushed something unrelated to the branch of this pull request. I removed it again by force pushing only the previous commit.

eboasson:
@martinbremmer it is my pleasure to merge this PR, because I do understand the reasons for asking me to go easy on the naming of things 😁.

Still, I do have one question about the deleting of proxy participants: you check whether the proxy writer entity id matches the "secure" variant of the SPDP writer. Arguably, one should check that it's GUID prefix matches the proxy participant that is being deleted as well: otherwise any secure SPDP writer can delete any proxy participant.

Now, if an authenticated proxy participant sends a message to delete another proxy participant, that probably means that you were wrong to trust the former in the first place 🤔. So I don't see this as a reason to delay merging this PR, but I do think it is worth giving some thought to such scenarios.

martinbremmer:
@eboasson, nicely spotted and fixing this problem should be relatively easy. However, if you trust an 'evil' participant, you're up sh*t creek anyway. Deleting another participant is only one of the many offenses it can do in that case.
But, you're right, that shouldn't stop us from thinking about those scenarios.

PR #297, opened 2019-10-24
Moved bswap functions to ddsrt

dennis-adlink:
Moved the byte swapping functions from ddsi to ddsrt so that
these can be re-used in e.g. the security plugins and tests.

### DIFF
--- a/src/core/ddsi/src/q_bswap_inlines.c
+++ b/src/core/ddsi/src/q_bswap_inlines.c
@@ -9,13 +9,13 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
+#include "dds/ddsrt/bswap.h"
 #include "dds/ddsi/q_bswap.h"
 
-extern inline uint16_t bswap2u (uint16_t x);
-extern inline uint32_t bswap4u (uint32_t x);
-extern inline uint64_t bswap8u (uint64_t x);
-extern inline int16_t bswap2 (int16_t x);
-extern inline int32_t bswap4 (int32_t x);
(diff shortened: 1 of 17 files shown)


src/core/ddsi/src/q_bswap_inlines.c
```diff
@@ -9,13 +9,13 @@
...
+#include "dds/ddsrt/bswap.h"
 #include "dds/ddsi/q_bswap.h"
 
-extern inline uint16_t bswap2u (uint16_t x);
-extern inline uint32_t bswap4u (uint32_t x);
-extern inline uint64_t bswap8u (uint64_t x);
-extern inline int16_t bswap2 (int16_t x);
-extern inline int32_t bswap4 (int32_t x);
-extern inline int64_t bswap8 (int64_t x);
+extern inline uint16_t ddsrt_bswap2u (uint16_t x);
```
eboasson: The ``extern inline`` declarations are necessary to make the compiler emit a callable function needed whenever the inlining doesn't happen (like when optimisation is disabled). As there should be no dependency from ``src/ddsrt`` to ``src/core/ddsi``, these need to be moved to ddsrt (presumably, ddsrt/src/bswap.c).

dennis-adlink: You're right, that dependency is obviously wrong. I have moved the `extern inline` declarations to ddsrt, except for the `bswapSN` function, which is in ddsi's q_bswap.h (moved this extern to q_bswap.c)

eboasson:
Thanks for cleaning up this bit, @dennis-adlink!

PR #298, opened 2019-10-28
Use ddsrt_malloc for allocating iov array

MarcelJordense:
Signed-off-by: Marcel Jordense <marcel.jordense@prismtech.com>


eboasson:
@MarcelJordense — my apologies: I didn't look carefully enough. The ECA check fails because the "Author" line is using a different address from the "Signed-off-by". So if you just fix the former to match the latter and do a force push, I expect all will be well.

PR #299, opened 2019-10-28
Move IDLC to its own install component

rotu:
This allows you to express a dependency on it as `find_package(CycloneDDS REQUIRED COMPONENTS idlc)`


eboasson:
@rotu this all makes sense to me. I've left it alone for a bit because the title still says "WIP", and because my limited affinity with CMake makes me a bit hesitant to declare that it is good despite that title.

In short: is this really still "WIP" or did you just forgot to change the title?

rotu:
@eboasson I merely forgot to change the title

eboasson:
Thanks, @rotu!

PR #302, opened 2019-10-31
WIP: improve dds_create_domain behaviour

eboasson:
And a host of other things as a consequence ...:
* not screwing up refcounting when deleting a topic multiple times
* not leaking a reference to a domain created in dds_create_domain if a subsequent dds_create_participant failed

I think all of the refcounting and state tracking is now done in the "count & flags" word of the handle, so that the entity need no longer be locked in order to determine whether it is safe to use it, or that it can be thrown away if it is implicit.

All tests pass for me, but it is work-in-progress, so beware.

The reviewer — me ... 😳 — missed rather much when he approved #244 ...

@martinbremmer #300 is yours if you want it. As you can see by this, it had exceeded my tolerance for annoying things already, but I am more than happy for you to look into it further. If this (so far experimental) set of changes is of use to you, so much the better.

Issue #300, opened 2019-10-31
dds_create_domain() does not return an entity


martinbremmer:
The dds_create_domain() returns a return code instead of an entity. The result is that an domain entity was created that can not be deleted by the application*.

A side effect of this is that manually running cunit_ddsc test application does not work anymore. Rogue domains keep existing and interfering in various tests.
Running through ctest doesn't have this issue, because every test runs in its own process.

@eboasson, can you assign me to this issue please?

Note, see also https://github.com/eclipse-cyclonedds/cyclonedds/pull/268#discussion_r333608411

\* It can be deleted by calling dds_get_parent() on a related participant and delete the returned entity.

eboasson:
@martinbremmer I had become sufficiently annoyed by to try and figure out what would be a sane way of dealing with things, and trying to capture that in working code. I've created #302 for your (or anyone's, for that matter) perusal. Feel free to shoot that down, use it, or whatever.

eboasson:
Closed via #319. Thanks @martinbremmer for helping resolve this.

PR #304, opened 2019-11-04
WIP: topic creation/deletion behaviour.

martinbremmer:
Originally, this was a PR to improve the create domain feature. It was noticed that the topic creation/deletion behaviour changed compared to master. A quick change was suggested to solve that. However, that wasn't accepted. Currently, the master contains the revised behaviour (see #319).

This PR is changed to a WIP to not lose the history of the topic creation/deletion discussion nor lose the proposed (but declined) solution.

Original PR description:
> See issue #300.
> 
> This upgrade impacts the entity deletion quite a bit, as indicated in the WIP pull-request #302.
> 
> That #302 PR contains a WIP commit regarding dds_create_domain() and entity deletions.
> 
> This PR takes that commit, changed some function headers, added some tests and refactored the WIP commit a bit.
> 
> The main difference after the refactor is that only ```pinning for deletion``` determines if the ```closing``` flag should be set. It's not done in the ```reduce refcount``` functions anymore.
> By changing ```pinning for deletion``` a little bit more, the 'normal' entity delete sequence can be followed for (builtin) topics, domains and cyclonedds top entity (no need for specific repins).

### DIFF
--- a/src/core/ddsc/src/dds_entity.c
+++ b/src/core/ddsc/src/dds_entity.c
@@ -92,10 +92,7 @@ void dds_entity_drop_ref (dds_entity *e)
 {
   if (dds_handle_drop_ref (&e->m_hdllink))
   {
-    /* increment pin count unconditionally to satisfy the "pinned" requirement */
-    dds_handle_repin (&e->m_hdllink);
-    ddsrt_mutex_lock (&e->m_mutex);
-    dds_return_t ret = really_delete_pinned_closed_locked (e, DIS_EXPLICIT);
+    dds_return_t ret = dds_delete_impl(e->m_hdllink.hdl, DIS_EXPLICIT);
     assert (ret == DDS_RETCODE_OK);
     (void) ret;
   }
@@ -105,10 +102,7 @@ void dds_entity_unpin_and_drop_ref (dds_entity *e)
 {
   if (dds_handle_unpin_and_drop_ref (&e->m_hdllink))
   {
-    /* increment pin count unconditionally to satisfy the "pinned" requirement */
-    dds_handle_repin (&e->m_hdllink);
-    ddsrt_mutex_lock (&e->m_mutex);
-    dds_return_t ret = really_delete_pinned_closed_locked (e, DIS_EXPLICIT);
+    dds_return_t ret = dds_delete_impl(e->m_hdllink.hdl, DIS_EXPLICIT);
     assert (ret == DDS_RETCODE_OK);
     (void) ret;
   }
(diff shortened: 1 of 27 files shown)


src/core/ddsc/src/dds_entity.c
```diff
@@ -80,11 +80,34 @@ const ddsrt_avl_treedef_t dds_entity_children_td = DDSRT_AVL_TREEDEF_INITIALIZER
...
 void dds_entity_add_ref_locked (dds_entity *e)
 {
   dds_handle_add_ref (&e->m_hdllink);
 }
 
+void dds_entity_drop_ref (dds_entity *e)
+{
+  if (dds_handle_drop_ref (&e->m_hdllink))
+  {
+    dds_return_t ret = dds_delete_impl(e->m_hdllink.hdl, DIS_EXPLICIT);
```
eboasson: There was (some) method behind the madness of my "WiP" commit in setting the ``CLOSING`` flag on the handle in ``dds_handle_drop_ref`` (and ``dds_handle_unpin_and_drop_ref``). I'm no way trying to defend the body of this function (and the next) in that commit, that was simply an ugly hack along the way.

The way you do it here looks perfectly fine, but what it does allow is for some thread to increment the refcount after it dropped to 0 but before ``dds_delete_impl`` is called, if there is a path that allows another thread to do so. One has to wonder whether that's an actual problem, and there are two perspectives on it:

1. Is there any way in which that can happen? — If the answer is "no", then all is well.
2. If it happens, does it behave in an acceptable manner? — If it always behaves in an acceptable manner (not crashing for example) then all is well, too.

Re 1: for ``dds_entity_drop_ref``, it really is only a topic that can be passed in, and it would already have to be in the "deferred delete" state (otherwise the implied reference from the handle will keep it alive). I'm pretty sure it is possible, but it can be prevented by checking in ``dds_create_reader`` and ``dds_create_writer``. That is a manageable case.

``dds_entity_unpin_and_drop_ref`` is only used for the global "cyclone" entity and the domain entities. Those are also really special cases, sneak around the regular destruction path and have additional locking to prevent such games. I just haven't figured that one out yet ...

Re 2: For topics, it could cause the call to ``delete`` to return ``ALREADY_DELETED``, but I think nothing worse than that. One could change the assert, one could even decide that ``ALREADY_DELETED`` doesn't make sense and that it should always return ``OK`` anyway.

For the domain and cyclonedds objects, I think it is equivalent to simply calling ``dds_delete`` from the application code. It may therefore result in deleting a domain/participant that was being created in parallel, but I do think that afterward that new one either exists and functions normally, or doesn't exist at all. Given that it is somewhat daft use of the API, I'd say that is acceptable.

In short, your simplification looks reasonable; in this form it will work better than the code currently in master (e.g., calling ``dds_delete`` twice on a topic is deadly now); and the possibly problematic cases seem to be rather far out there in terms of realistic use.

It'd be interesting to try testing it a bit, for example by creating and deleting participants in some domain on multiple threads.

martinbremmer: I've added a torture test that showed that a wrong mutex was used when adding a participant to the domain as a child.

martinbremmer:
There's some behaviour change regarding creation/deletion of topics. I'm still looking into that.

The current solution direction will be to 'clone' a topic (creating a new handle) when a duplicate topic is created on a participant.

src/core/ddsc/src/dds_entity.c
martinbremmer: It still crashes intermittently when compiling with gcc.
Somehow, I can't reproduce it with clang.

martinbremmer: There are two torture tests. One triggers an exiting issue (the crash) that is somewhat outside the scope of this PR and is related to actually init/deinit the library. For that reason, I've disabled that torture test.

src/core/ddsc/src/dds_participant.c
```diff
@@ -126,9 +126,9 @@ dds_entity_t dds_create_participant (const dds_domainid_t domain, const dds_qos_
   pp->m_builtin_subscriber = 0;
 
   /* Add participant to extent */
-  ddsrt_mutex_lock (&dds_global.m_entity.m_mutex);
+  ddsrt_mutex_lock (&dom->m_entity.m_mutex);
```
eboasson: 😳

Thanks for catching that one!

martinbremmer:
I don't have strong feelings either way. It's just that the 'clone' option currently implemented is more intuitive (at least to me) when considering deletions.

It also makes my live a lot easier regarding C++. But as you said, that should be solvable.

eboasson:
I have merged #319 which is basically this one minus the last commit. That commit is interesting, so don't throw that away just yet please, @martinbremmer. (In fact, you could turn it into a WIP PR if you wish.)

PR #306, opened 2019-11-05
DDS Security built-in Cryptographic plugin

dennis-adlink:
This PR adds the built-in Cryptographic plugin that is part of the DDS Security implementation for Cyclone.

The Cryptographic plugin defines the types and operations necessary to support encryption, digest, message authentication codes, and key exchange for DDS DomainParticipants, DataWriters and DDS DataReaders.

Similar to other builtin plugins, the DDS Security cryptographic plugin is built as a shared library to allow dynamic library loading on runtime. This enables DDS participants to use specific plugin implementations with different configurations.

Although I think this initial version is a reasonable starting point to be merged in the security branch, some parts of the code will need refactoring:

* crypto_key_factory.c: crypto_factory_get_endpoint_relation returns arbitrary local-remote relation if no specific key for remote is found, which will not work in Cyclone because participants can have different security settings

* performance of encoding data can be improved by not copying plain_rtps_message to a new buffer (to enable this, crypto_cipher_encrypt_data should allow encrypting parts of a message)

* when decoding a message the message is split in several parts (header, body, footer, etc) and for this memory is allocated which is probably not necessary. Performance should be improved by removing these allocations and use pointers to the data instead.

This PR also includes some minor refactoring in security core and in the authentication plugin tests related to moving bswap functions to ddsrt. 


dennis-adlink:
@eboasson in the additional commits in this PR I've addressed all issues from your review comments, so this PR is now ready to be re-reviewed. The changes I've made since the initial commit for this PR:
- Fixed typo DDD_
- Fixed cmake project variable name
- Removed superfluous casts
- Moved memset to deserialization functions
- Improved support for using 128 bit keys (includes fix for serialised representation)
- Fixed atomic inc_nv -> inc_ov
- Refactored function that returned pointers to released refcounted object
- Improved locking for crypto table
- Erase key material when memory is freed

src/security/builtin_plugins/cryptographic/src/crypto_key_factory.c
```diff
@@ -0,0 +1,1872 @@
...
+  if (CRYPTO_TRANSFORM_HAS_KEYS(dst->transformation_kind) && !CRYPTO_TRANSFORM_HAS_KEYS(src_transform_kind))
+  {
+    ddsrt_free(dst->master_salt);
+    ddsrt_free(dst->master_sender_key);
+    ddsrt_free(dst->master_receiver_specific_key);
+  }
+  else if (CRYPTO_TRANSFORM_HAS_KEYS(src_transform_kind))
+  {
+    uint32_t key_bytes = CRYPTO_KEY_SIZE_BYTES(src_transform_kind);
+    if (!CRYPTO_TRANSFORM_HAS_KEYS(dst->transformation_kind))
```
eboasson: Is it a given that ``src`` and ``dst`` have the same key size if both have keys?

src/security/builtin_plugins/cryptographic/src/crypto_objects.c
```diff
@@ -0,0 +1,593 @@
...
+
+bool crypto_object_valid(CryptoObject *obj, CryptoObjectKind_t kind)
+{
+  return (obj && obj->kind == kind && obj->handle == (int64_t)(uintptr_t)obj);
+}
+
+static uint32_t crypto_object_hash(const void *obj)
+{
+  const CryptoObject *object = obj;
+  const uint64_t c = 0xF6D6E408505;
```
eboasson: I'm curious why you switched to such a small constant. With this constant the LSB of the handle affects approximately roughly the lowest 12 bits of the computed hash. The hash table implementation then takes the low order bits of the calculated hash, and so (unless I'm mistaken) for tables larger than ~2k buckets the hashing is really insufficient. (See also https://en.wikipedia.org/wiki/Universal_hashing)

eboasson:
@dennis-adlink thanks for improving so many details. Pursuant to policy for the security branch, I've limited myself to sampling the changes 😉 . Apart from two spots that I found I am ok with it. I'd appreciate it if you could have a look at those two.

src/security/builtin_plugins/cryptographic/src/crypto_objects.c
dennis-adlink: Thanks for catching this, I missed a parantheses in the macro definition and so I calculated the wrong value (I should have read some background info on hashing, then I'd notice this myself). I fixed this in my latest commit.

src/security/builtin_plugins/cryptographic/src/crypto_key_factory.c
dennis-adlink: No, it's not, so I've changed the code so that it always frees the memory and allocates new buffers (which has minor performance impact because `crypto_token_copy` is only called during handshake/initialization)

eboasson:
Thanks @dennis-adlink !

PR #308, opened 2019-11-05
Fixed coverity issues

BartP6703:
- Fixed a lot of 'unchecked return value' issues.
- Fixed 'illegal address computation' issue in do_packet() function by using pointer arithmetic i.s.o. using offsets.
- Allocate array of samples from heap i.s.o. using local variable on stack in throughput/subscriber example.

Travis job with compiler 'clang Xcode: xcode9 C' is failing, but that is a known problem.

### DIFF
--- a/examples/throughput/subscriber.c
+++ b/examples/throughput/subscriber.c
@@ -93,7 +93,7 @@ int main (int argc, char **argv)
   signal (SIGINT, sigint);
   process_samples(reader, maxCycles);
 
-  dds_set_status_mask (reader, 0);
+  (void) dds_set_status_mask (reader, 0);
   HandleMap__free (imap);
   finalize_dds (participant);
   return EXIT_SUCCESS;
@@ -154,7 +154,7 @@ static HandleEntry * retrieve_handle (HandleMap *map, dds_instance_handle_t key)
 static int do_take (dds_entity_t reader)
 {
   int samples_received;
-  dds_sample_info_t info [MAX_SAMPLES];
+  dds_sample_info_t *info = NULL;
   dds_instance_handle_t ph = 0;
   HandleEntry * current = NULL;
 
@@ -165,9 +165,14 @@ static int do_take (dds_entity_t reader)
 
   /* Take samples and iterate through them */
 
+  info = dds_alloc (sizeof(dds_sample_info_t) * MAX_SAMPLES);
+
   samples_received = dds_take (reader, samples, info, MAX_SAMPLES, MAX_SAMPLES);
   if (samples_received < 0)
+  {
+    dds_free( info );
     DDS_FATAL("dds_take: %s\n", dds_strretcode(-samples_received));
+  }
 
   for (int i = 0; !done && i < samples_received; i++)
   {
@@ -196,13 +201,14 @@ static int do_take (dds_entity_t reader)
       total_samples++;
     }
   }
+  dds_free (info);
   return samples_received;
 }
 
 static void data_available_handler (dds_entity_t reader, void *arg)
 {
   (void)arg;
-  do_take (reader);
+  (void) do_take (reader);
 }
 
 static int parse_args(int argc, char **argv, unsigned long long *maxCycles, char **partitionName)
--- a/src/core/ddsc/src/dds_stream.c
+++ b/src/core/ddsc/src/dds_stream.c
@@ -710,7 +710,7 @@ static void dds_stream_read (dds_istream_t * __restrict is, char * __restrict da
           case DDS_OP_VAL_4BY: *((uint32_t *) addr) = dds_is_get4 (is); ops += 2; break;
           case DDS_OP_VAL_8BY: *((uint64_t *) addr) = dds_is_get8 (is); ops += 2; break;
           case DDS_OP_VAL_STR: *((char **) addr) = dds_stream_reuse_string (is, *((char **) addr), 0); ops += 2; break;
-          case DDS_OP_VAL_BST: dds_stream_reuse_string (is, (char *) addr, ops[2]); ops += 3; break;
+          case DDS_OP_VAL_BST: addr = (void *) dds_stream_reuse_string (is, (char *) addr, ops[2]); ops += 3; break;
           case DDS_OP_VAL_SEQ: ops = dds_stream_read_seq (is, addr, ops, insn); break;
           case DDS_OP_VAL_ARR: ops = dds_stream_read_arr (is, addr, ops, insn); break;
           case DDS_OP_VAL_UNI: ops = dds_stream_read_uni (is, addr, data, ops, insn); break;
@@ -1127,7 +1127,7 @@ void dds_stream_read_key (dds_istream_t * __restrict is, char * __restrict sampl
       case DDS_OP_VAL_4BY: *((uint32_t *) dst) = dds_is_get4 (is); break;
       case DDS_OP_VAL_8BY: *((uint64_t *) dst) = dds_is_get8 (is); break;
       case DDS_OP_VAL_STR: *((char **) dst) = dds_stream_reuse_string (is, *((char **) dst), 0); break;
-      case DDS_OP_VAL_BST: dds_stream_reuse_string (is, dst, op[2]); break;
+      case DDS_OP_VAL_BST: dst = dds_stream_reuse_string (is, dst, op[2]); break;
       case DDS_OP_VAL_ARR:
         dds_is_get_bytes (is, dst, op[2], get_type_size (DDS_OP_SUBTYPE (*op)));
         break;
@@ -1728,16 +1728,16 @@ static const uint32_t *prtf_seq (char * __restrict *buf, size_t *bufsize, dds_is
   num = dds_is_get4 (is);
   if (num == 0)
   {
-    prtf (buf, bufsize, "{}");
+    (void) prtf (buf, bufsize, "{}");
     return skip_sequence_insns (ops, insn);
   }
   switch (subtype)
   {
     case DDS_OP_VAL_1BY: case DDS_OP_VAL_2BY: case DDS_OP_VAL_4BY: case DDS_OP_VAL_8BY:
-      prtf_simple_array (buf, bufsize, is, num, subtype);
+      (void) prtf_simple_array (buf, bufsize, is, num, subtype);
       return ops + 2;
     case DDS_OP_VAL_STR: case DDS_OP_VAL_BST:
-      prtf_simple_array (buf, bufsize, is, num, subtype);
+      (void) prtf_simple_array (buf, bufsize, is, num, subtype);
       return ops + (subtype == DDS_OP_VAL_STR ? 2 : 3);
     case DDS_OP_VAL_SEQ: case DDS_OP_VAL_ARR: case DDS_OP_VAL_UNI: case DDS_OP_VAL_STU: {
       const uint32_t jmp = DDS_OP_ADR_JMP (ops[3]);
@@ -1745,10 +1745,10 @@ static const uint32_t *prtf_seq (char * __restrict *buf, size_t *bufsize, dds_is
       bool cont = prtf (buf, bufsize, "{");
       for (uint32_t i = 0; cont && i < num; i++)
       {
-        if (i > 0) prtf (buf, bufsize, ",");
+        if (i > 0) (void) prtf (buf, bufsize, ",");
         cont = dds_stream_print_sample1 (buf, bufsize, is, jsr_ops, subtype == DDS_OP_VAL_STU);
       }
-      prtf (buf, bufsize, "}");
+      (void) prtf (buf, bufsize, "}");
       return ops + (jmp ? jmp : 4); /* FIXME: why would jmp be 0? */
     }
   }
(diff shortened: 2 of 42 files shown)


examples/throughput/subscriber.c
```diff
@@ -165,9 +165,14 @@ static int do_take (dds_entity_t reader)
 
   /* Take samples and iterate through them */
 
+  info = dds_alloc (sizeof(dds_sample_info_t) * MAX_SAMPLES);
```
i-and: From the point of view of performance and determinism it is better to avoid working with dynamic memory in the process of receiving messages. In this regard, the array on the stack looked preferable.

BartP6703: Performance is not one of the goals of this example anymore. Actually this example is too complex and should be replaced by a much more simple example.

eboasson: There is a middle way here, too, which is to keep the array around. The samples are a global, there's no reason ``info`` should be treated differently.

That said, @BartP6703's comment that these examples are great at showing how _not_ to program is entirely correct. So I can't be bothered by a suboptimal code change that satisfies Coverity if that's merely a precursor to improving the examples altogether.

src/core/ddsc/src/dds_stream.c
```diff
@@ -710,7 +710,7 @@ static void dds_stream_read (dds_istream_t * __restrict is, char * __restrict da
           case DDS_OP_VAL_4BY: *((uint32_t *) addr) = dds_is_get4 (is); ops += 2; break;
           case DDS_OP_VAL_8BY: *((uint64_t *) addr) = dds_is_get8 (is); ops += 2; break;
           case DDS_OP_VAL_STR: *((char **) addr) = dds_stream_reuse_string (is, *((char **) addr), 0); ops += 2; break;
-          case DDS_OP_VAL_BST: dds_stream_reuse_string (is, (char *) addr, ops[2]); ops += 3; break;
+          case DDS_OP_VAL_BST: addr = (void *) dds_stream_reuse_string (is, (char *) addr, ops[2]); ops += 3; break;
```
eboasson: That won't do: it introduces a dead store. (I am surprised Coverity doesn't complain about it!)

src/core/ddsc/src/dds_stream.c
```diff
@@ -1127,7 +1127,7 @@ void dds_stream_read_key (dds_istream_t * __restrict is, char * __restrict sampl
       case DDS_OP_VAL_4BY: *((uint32_t *) dst) = dds_is_get4 (is); break;
       case DDS_OP_VAL_8BY: *((uint64_t *) dst) = dds_is_get8 (is); break;
       case DDS_OP_VAL_STR: *((char **) dst) = dds_stream_reuse_string (is, *((char **) dst), 0); break;
-      case DDS_OP_VAL_BST: dds_stream_reuse_string (is, dst, op[2]); break;
+      case DDS_OP_VAL_BST: dst = dds_stream_reuse_string (is, dst, op[2]); break;
```
eboasson: Also introduces a dead store.

src/core/ddsi/include/dds/ddsi/q_radmin.h
```diff
@@ -98,7 +98,7 @@ struct nn_rmsg {
 
   struct nn_rmsg_chunk chunk;
 };
-#define NN_RMSG_PAYLOAD(m) ((unsigned char *) (&(m)->chunk + 1))
+#define NN_RMSG_PAYLOAD(m) ((unsigned char *) (m + 1))
```
eboasson: At the moment this should be ok because ``chunk``'s alignment should be >= ``struct nn_rmsg``'s alignment. One would expect this to always be the case, but this changes will break things in a very nasty way if one day it isn't.

So it should then at least have a static assertion added to ensure ``sizeof(m)`` equals ``offsetof(m,chunk)+sizeof(chunk)``

src/core/ddsi/src/q_debmon.c
```diff
@@ -381,7 +381,7 @@ struct debug_monitor *new_debug_monitor (struct q_globals *gv, int port)
   if (ddsi_listener_listen (dm->servsock) < 0)
     goto err_listen;
   dm->stop = 0;
-  create_thread (&dm->servts, gv, "debmon", debmon_main, dm);
+  (void) create_thread (&dm->servts, gv, "debmon", debmon_main, dm);
```
eboasson: No way. It is only an optional monitoring thing, but even so, it ought to check whether thread creation succeeded and gracefully handle failure, and certainly mustn't hide _not_ checking!

src/core/ddsi/src/q_entity.c
```diff
@@ -1807,7 +1807,7 @@ static void reader_add_connection (struct reader *rd, struct proxy_writer *pwr,
       /* FIXME: for now, assume that the ports match for datasock_mc --
        't would be better to dynamically create and destroy sockets on
        an as needed basis. */
-      ddsi_join_mc (rd->e.gv, rd->e.gv->mship, rd->e.gv->data_conn_mc, &m->ssm_src_loc, &m->ssm_mc_loc);
+      (void) ddsi_join_mc (rd->e.gv, rd->e.gv->mship, rd->e.gv->data_conn_mc, &m->ssm_src_loc, &m->ssm_mc_loc);
```
eboasson: No, casting away this return value is a mistake. An error here is definitely possible and should be handled properly. It probably means the entire call needs to fail, that might mean the proxy writer needs to be undiscovered or something like that, but it being difficult to figure out _how_ to handle an error is not a sufficient reason to pretend it can't occur.

src/core/ddsi/src/q_init.c
```diff
@@ -1403,7 +1403,7 @@ int rtps_start (struct q_globals *gv)
   }
   if (gv->listener)
   {
-    create_thread (&gv->listen_ts, gv, "listen", (uint32_t (*) (void *)) listen_thread, gv->listener);
+    (void) create_thread (&gv->listen_ts, gv, "listen", (uint32_t (*) (void *)) listen_thread, gv->listener);
```
eboasson: Again, casting away an error here is not right. And there is even a FIXME right there!

src/core/ddsi/src/q_receive.c
```diff
@@ -2896,7 +2896,7 @@ static bool do_packet (struct thread_state1 * const ts1, struct q_globals *gv, d
 
   if (conn->m_stream)
   {
-    MsgLen_t * ml = (MsgLen_t*) (buff + RTPS_MESSAGE_HEADER_SIZE);
+    MsgLen_t * ml = (MsgLen_t*) (hdr + 1);
```
eboasson: Why exactly is this change required? The address is the same, that's not the issue.

Secondly, there is actually a real bug a few lines below, which is that it doesn't verify the received message length is within bounds ...

src/core/ddsi/src/q_xmsg.c
```diff
@@ -1217,7 +1217,7 @@ void nn_xpack_sendq_init (struct q_globals *gv)
 
 void nn_xpack_sendq_start (struct q_globals *gv)
 {
-  create_thread (&gv->sendq_ts, gv, "sendq", nn_xpack_sendq_thread, NULL);
+  (void) create_thread (&gv->sendq_ts, gv, "sendq", nn_xpack_sendq_thread, NULL);
```
eboasson: This is some code that is more of an experiment than something that should remain. Still, the response to Coverity complaining that the result of attempting to create a thread is being ignored should not be to tell it to shut up in this way.

And should it be determined that this is, all considered, the right approach, it warrants a little note why it is so.

src/ddsrt/src/rusage/posix/rusage.c
```diff
@@ -109,7 +109,7 @@ ddsrt_getrusage_anythread (
     return DDS_RETCODE_ERROR;
   if ((fp = fopen (file, "r")) == NULL)
     return DDS_RETCODE_NOT_FOUND;
-  enum { ERROR, READ_HEADING, SKIP_TO_EOL, READ_VCSW, READ_IVCSW } state = READ_HEADING;
+  enum { ERROR = 1, READ_HEADING, SKIP_TO_EOL, READ_VCSW, READ_IVCSW } state = READ_HEADING;
```
eboasson: Why does ``ERROR`` have to be 1 instead of 0?

src/tools/pubsub/pubsub.c
```diff
@@ -280,7 +280,7 @@ static char *expand_envvars(const char *src0);
 // FIXME: This is the same as the expand function in util. Merge.
 static char *expand_env(const char *name, char op, const char *alt) {
     char *env = NULL;
-    ddsrt_getenv(name, &env);
+    (void) ddsrt_getenv(name, &env);
```
eboasson: Given how ``ddsrt_getenv`` works and the way ``name`` is set, ignoring the return value is certainly acceptable in principle. But with the FIXME just a few lines above it, I wonder why this solution was chosen. I would prefer to see the function removed altogether and the one from ``ddsrt`` used instead.

src/ddsrt/src/rusage/posix/rusage.c
BartP6703: In the path of this error in coverity states that the value of "state" must be at least 1. ERROR is the lowest value of state.

src/core/ddsi/src/q_debmon.c
BartP6703: I will add a check on the return code of create_thread.

src/core/ddsi/src/q_entity.c
BartP6703: I will check the return code and log a message.

src/core/ddsi/src/q_init.c
BartP6703: I will check the return code and log a message.

src/core/ddsc/src/dds_stream.c
BartP6703: Split the function into two different functions: one for string (returning the string) and one for bounded string (not returning anything).

src/core/ddsi/src/q_xmsg.c
BartP6703: Will check the return code and log a message.

src/core/ddsi/include/dds/ddsi/q_radmin.h
BartP6703: I will add the static assertion.

BartP6703: I will add the static assertion in the function where the macro is used.

src/core/ddsi/src/q_receive.c
BartP6703: The address is the same indeed, but the code is in my opinion easier to read (if hdr == buff then buff + sizeof(header_t) == hdr + 1.
I don't see the real bug. As far as I can see all the received messages are within bounds.

src/tools/pubsub/pubsub.c
BartP6703: As far as I can see the call to `expand_envvars` can be replaced by a call to `ddsrt_expand_envvars_sh`. All the local environment functions in pubsub.c can then be removed.

src/core/ddsi/src/q_receive.c
eboasson: The problem I alluded to is in this line:
```C
        sz = ddsi_conn_read (conn, buff + stream_hdr_size, ml->length - stream_hdr_size, false, NULL);
```

and is simply that there is no check that ``ml->length`` is ≥``stream_hdr_size``, nor that it is ≤ than ``maxsz``.

src/ddsrt/src/rusage/posix/rusage.c
eboasson: I get it. Coverity is correct in the sense that in ``switch(state)``, ``state`` ≠ ``ERROR``, so ≥1 given the way enumerated types in C work. The ``case ERROR`` is simply there to make it clear to the compiler (which typically doesn't perform the same level of analysis that Coverity does) that all possible cases have been considered.

One could eliminate the ``ERROR`` state entirely by doing ``{fclose(fp);return DDS_RETCODE_ERROR;}`` where it enters that state. But I like the structure with an ERROR state better here …

Just tell Coverity that this is intentional instead of changing the code.

src/core/ddsi/src/q_entity.c
```diff
@@ -1807,7 +1807,9 @@ static void reader_add_connection (struct reader *rd, struct proxy_writer *pwr,
       /* FIXME: for now, assume that the ports match for datasock_mc --
        't would be better to dynamically create and destroy sockets on
        an as needed basis. */
-      (void) ddsi_join_mc (rd->e.gv, rd->e.gv->mship, rd->e.gv->data_conn_mc, &m->ssm_src_loc, &m->ssm_mc_loc);
+      int ret = ddsi_join_mc (rd->e.gv, rd->e.gv->mship, rd->e.gv->data_conn_mc, &m->ssm_src_loc, &m->ssm_mc_loc);
+      if (ret < 0)
+        ELOGDISC (rd, "  unable to join\n");
```
eboasson: Not being able to join the multicast group is likely to be a big deal: you'd end up advertising to be reachable at a certain multicast address when in fact you aren't. The proper fallback would be not to advertise the multicast address for this reader, but that's a slightly more involved project. You're more than welcome to do that.

Otherwise, I think it should be at least be logged as an error (which normally goes to the log sink, i.e., stderr by default).

src/core/ddsi/src/q_init.c
```diff
@@ -1403,7 +1403,8 @@ int rtps_start (struct q_globals *gv)
   }
   if (gv->listener)
   {
-    (void) create_thread (&gv->listen_ts, gv, "listen", (uint32_t (*) (void *)) listen_thread, gv->listener);
+    if (create_thread (&gv->listen_ts, gv, "listen", (uint32_t (*) (void *)) listen_thread, gv->listener) != DDS_RETCODE_OK)
+      GVERROR ("rtps_start: can't create listener thread\n");
```
eboasson: Stopping everything is quite intricate (which no doubt is the reason that the error handling was omitted in the first place), and logging that something went badly wrong is a definite improvement over silence. But if you look at the start/stop sequence, ``rtps_stop`` undoes ``rtps_start`` and not much is needed to use ``rtps_stop`` for cleaning up in these two error cases. Looking at ``rtps_stop`` you'd think:

```C
ddsi_listener_free(gv->listener);
gv->listener=NULL;
rtps_stop();
```

would do the trick. That's far more elegant than simply printing an error.

src/core/ddsi/src/q_receive.c
```diff
@@ -2891,6 +2891,9 @@ static bool do_packet (struct thread_state1 * const ts1, struct q_globals *gv, d
   {
     return false;
   }
+
```
eboasson: This is good — but ``DDSRT_STATIC_ASSERTION`` would be even better: that turns it into a compile-time check, so that you can't build it if the two aren't the same.

eboasson:
Closed via #341.

PR #310, opened 2019-11-06
Added volatile security endpoints.

MarcelJordense:
Signed-off-by: Martin Bremmer <martin.bremmer@adlinktech.com>

Actually use max_seq_available

Signed-off-by: Martin Bremmer <martin.bremmer@adlinktech.com>

Some small updates.

Signed-off-by: Martin Bremmer <martin.bremmer@adlinktech.com>

Some small updates.

Signed-off-by: Martin Bremmer <martin.bremmer@adlinktech.com>

remove destination reader specific heartbeat control

Signed-off-by: Marcel Jordense <marcel.jordense@adlinktech.com>

remove unused code

Signed-off-by: Marcel Jordense <marcel.jordense@adlinktech.com>


src/core/ddsi/src/q_entity.c
```diff
@@ -2939,11 +2979,12 @@ static void new_writer_guid_common_init (struct writer *wr, const struct ddsi_se
   assert (wr->xqos->present & QP_RELIABILITY);
   wr->reliable = (wr->xqos->reliability.kind != DDS_RELIABILITY_BEST_EFFORT);
   assert (wr->xqos->present & QP_DURABILITY);
-  if (is_builtin_entityid (wr->e.guid.entityid, NN_VENDORID_ECLIPSE))
+  if (is_builtin_entityid (wr->e.guid.entityid, NN_VENDORID_ECLIPSE) ||
+      (wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER))
```
eboasson: ``NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER`` has the "source" component set to ``NN_ENTITYID_SOURCE_BUILTIN``, right? So why the additional condition?

src/core/ddsi/src/q_entity.c
```diff
@@ -3053,7 +3094,10 @@ static void new_writer_guid_common_init (struct writer *wr, const struct ddsi_se
     wr->whc_low = wr->e.gv->config.whc_lowwater_mark;
     wr->whc_high = wr->e.gv->config.whc_init_highwater_mark.value;
   }
-  assert (!is_builtin_entityid(wr->e.guid.entityid, NN_VENDORID_ECLIPSE) || (wr->whc_low == wr->whc_high && wr->whc_low == INT32_MAX));
+  assert (!(is_builtin_entityid(wr->e.guid.entityid, NN_VENDORID_ECLIPSE) &&
```
eboasson: ``(is_builtin_entityid(wr->e.guid.entityid, NN_VENDORID_ECLIPSE) && (wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER))`` is simply ``entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER`` ...

src/core/ddsi/src/q_entity.c
```diff
@@ -3505,7 +3549,8 @@ static dds_return_t new_reader_guid
   assert (rd->xqos->present & QP_RELIABILITY);
   rd->reliable = (rd->xqos->reliability.kind != DDS_RELIABILITY_BEST_EFFORT);
   assert (rd->xqos->present & QP_DURABILITY);
-  rd->handle_as_transient_local = (rd->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL);
+  rd->handle_as_transient_local = (rd->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL) ||
+                                  (rd->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_READER);
```
eboasson: Handling a volatile writer as transient-local deserves a comment.

src/core/ddsi/src/q_misc.c
```diff
@@ -42,6 +42,16 @@ int WildcardOverlap(char * p1, char * p2)
 }
 #endif
 
+int guid_prefix_eq (const ddsi_guid_prefix_t *a, const ddsi_guid_prefix_t *b)
+{
+  return a->u[0] == b->u[0] && a->u[1] == b->u[1] && a->u[2] == b->u[2];
+}
+
+int guid_eq (const struct ddsi_guid *a, const struct ddsi_guid *b)
```
eboasson: There is a ``entity_guid_eq`` in ``q_entity.c``. That one should be removed.

(And I dislike the superfluous parentheses.)

src/core/ddsi/src/q_receive.c
```diff
@@ -876,14 +882,40 @@ static int handle_AckNack (struct receiver_state *rst, nn_etime_t tnow, const Ac
         }
         else
         {
-          /* no merging, send directed retransmit */
-          RSTTRACE (" RX%"PRId64"", seqbase + i);
-          enqueued = (enqueue_sample_wrlock_held (wr, seq, sample.plist, sample.serdata, prd, 0) >= 0);
-          if (enqueued)
+          /* Is this a volatile reader with a filter?
+           * If so, call the filter to see if we should re-arrange the sequence gap when needed. */
+          if (prd->filter && (prd->filter(wr, prd, sample.serdata) == 0))
```
eboasson: ``prd->filter`` yields a boolean, so better use it as such. And surely there must be a more elegant way to compute ``gapstart``, ``gapend`` and the bitset without the duplication.

src/core/ddsi/src/q_receive.c
```diff
@@ -1190,44 +1229,73 @@ static int handle_Heartbeat (struct receiver_state *rst, nn_etime_t tnow, struct
     int refc_adjust = 0;
     nn_reorder_result_t res;
     gap = nn_rdata_newgap (rmsg);
-    if ((res = nn_reorder_gap (&sc, pwr->reorder, gap, 1, firstseq, &refc_adjust)) > 0)
+    int filtered = 0;
+
+    if (pwr->uses_filter && !guid_eq(&dst, &zero_guid))
```
eboasson: Better use ``is_null_guid``

src/core/ddsi/include/dds/ddsi/q_entity.h
```diff
@@ -369,6 +374,10 @@ struct proxy_writer {
   void *ddsi2direct_cbarg;
 };
 
+
+struct serdata;
+typedef int (*filter_function)(struct writer *wr, struct proxy_reader *prd, struct ddsi_serdata *serdata);
```
eboasson: The function typedefs tend to have the name of the function suffixed with ``_t``, ``_fn`` or ``_fn_t``. While one could argue that with such inconsistency there is no reason not to add another one, I think it would be better to name it ``filter_fn_t``.

src/core/ddsi/src/q_receive.c
```diff
@@ -2167,108 +2246,139 @@ static void handle_regular (struct receiver_state *rst, nn_etime_t tnow, struct
...
     struct nn_rdata *fragchain = nn_rsample_fragchain (rsample);
-    nn_reorder_result_t rres;
-
-    rres = nn_reorder_rsample (&sc, pwr->reorder, rsample, &refc_adjust, 0); // nn_dqueue_is_full (pwr->dqueue));
+    nn_reorder_result_t rres, rres2;
+    struct pwr_rd_match *wn;
+    int filtered = 0;
 
-    if (rres == NN_REORDER_ACCEPT && pwr->n_reliable_readers == 0)
+    if (pwr->uses_filter && !guid_eq(&dst, &zero_guid))
```
eboasson: ``is_null_guid``

src/core/ddsi/src/q_transmit.c
```diff
@@ -1024,6 +1068,117 @@ static int maybe_grow_whc (struct writer *wr)
...
+int write_sample_p2p_wrlock_held(struct writer *wr, seqno_t seq, struct nn_plist *plist, struct ddsi_serdata *serdata, struct ddsi_tkmap_instance *tk, struct proxy_reader *prd)
+{
+  struct q_globals * const gv = wr->e.gv;
+  int r;
+  nn_mtime_t tnow;
+  int rexmit = 1;
+  struct wr_prd_match *wprd = NULL;
+  seqno_t gseq;
+  struct nn_xmsg *gap = NULL;
+  seqno_t gapstart = -1, gapend = -1, segstart;
```
eboasson: ``seqstart``?

src/core/ddsi/src/q_transmit.c
```diff
@@ -1024,6 +1068,117 @@ static int maybe_grow_whc (struct writer *wr)
...
+        segstart = wprd->seq + 1;
+        memset (gapbits, 0, sizeof (gapbits));
+        for (gseq = segstart; gseq < seq; gseq++)
+        {
+          struct whc_borrowed_sample sample;
+          if (whc_borrow_sample (wr->whc, seq, &sample))
+          {
+            if (prd->filter(wr, prd, sample.serdata) == 0)
+            {
+              if (gapstart == -1)
```
eboasson: See also the comment in ``handle_AckNack``: three copies of the same logic really is too much. It would make sense to combine ``gapstart``, ``gapend``, ``gapnumbits`` and ``gapbits`` in a single struct and create a function to update that struct with following this logic. That can then be used in all call sites.

src/core/ddsi/src/q_transmit.c
```diff
@@ -1024,6 +1068,117 @@ static int maybe_grow_whc (struct writer *wr)
...
+            }
+            whc_return_sample (wr->whc, &sample, false);
+          }
+        }
+
+        if (gapstart > 0)
+        {
+          if (gapnumbits == 0)
+          {
+            /* Avoid sending an invalid bitset */
```
eboasson: DDSI 2.3 has made it legal to send an empty bitset. Cyclone doesn't yet declare itself to be DDSI 2.3 (though it pretty much is), and it makes sense to get rid of these crazy workarounds for what have been declared to be errors in the spec. (``handle_NackFrag`` has the same problem though ...)

src/core/ddsi/src/q_transmit.c
```diff
@@ -1024,6 +1068,117 @@ static int maybe_grow_whc (struct writer *wr)
...
+            gapnumbits = 1;
+            nn_bitset_set (gapnumbits, gapbits, 0);
+            gapend--;
+          }
+          gap = nn_xmsg_new (gv->xmsgpool, &wr->e.guid.prefix, sizeof (Gap_t), NN_XMSG_KIND_CONTROL);
+
+          if (nn_xmsg_setdstPRD (gap, prd) < 0) {
+               nn_xmsg_free (gap);
```
eboasson: Indentation

src/core/ddsi/src/q_xevent.c
```diff
@@ -578,6 +578,99 @@ static void handle_xevk_entityid (struct nn_xpack *xp, struct xevent_nt *ev)
...
+
+    for (m = ddsrt_avl_iter_first (&wr_readers_treedef, &wr->readers, &it); m; m = ddsrt_avl_iter_next (&it))
+    {
+      if (m->seq < m->lst_seq)
+        count++;
+    }
+
+    if (count > 0)
+    {
+      msg = ddsrt_malloc(count * sizeof(struct nn_xmsg *));
```
eboasson: I don't quite like counting readers and allocating memory here. I think it can be eliminated by iterating over the writers in a different way that doesn't require the writer to remain locked, using ``ddsrt_avl_lookup_succ``.

As unlocking the writer for each matching reader is a significant change, I'd like you to change it only if it still makes sense when you look into it a bit further. For now, this code is limited to when security is used, so it is not too big a deal. (But as this commit basically implements writer-side filtering, it probably won't be long before that becomes available to application as well 😄 .)

src/core/ddsi/src/q_entity.c
MarcelJordense: The test is incorrect it should exclude the NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER writer instead of allowing it.

src/core/ddsi/src/q_misc.c
MarcelJordense: I did not the entity_guid_eq because it references entities and guid_eq does not.

src/core/ddsi/src/q_xevent.c
MarcelJordense: Not yet addressed this issue. Will see if it can be changed.

src/core/ddsi/src/q_misc.c
eboasson: Good point — don't worry about it for now

eboasson:
@MarcelJordense, thanks for addressing all my nitpicking ☺️. Now, if you could remove the ``#if 0``'d version of ``handle_AckNack`` ... and do let me know what you conclude regarding the counting/allocation of readers in ``send_heartbeat_to_all_readers``.

src/core/ddsi/src/q_xevent.c
```diff
@@ -594,55 +593,49 @@ static void send_heartbeat_to_all_readers(struct nn_xpack *xp, struct xevent *ev
...
     struct wr_prd_match *m;
-    ddsrt_avl_iter_t it;
+    struct ddsi_guid last_guid = { .prefix = {.u = {0,0,0}}, .entityid = {0} };
 
     hbansreq = writer_hbcontrol_ack_required (wr, &whcst, tnow);
     t_next.v = tnow.v + writer_hbcontrol_intv (wr, &whcst, tnow);
 
-    for (m = ddsrt_avl_iter_first (&wr_readers_treedef, &wr->readers, &it); m; m = ddsrt_avl_iter_next (&it))
+    while ((m = ddsrt_avl_lookup_succ (&wr_readers_treedef, &wr->readers, &last_guid)) != NULL)
     {
```
eboasson: Methinks this is now an infinite loop: ``last_guid`` doesn't get updated. Adding ``last_guid = m->prd_guid;`` should solve that problem.

src/core/ddsi/src/q_xevent.c
```diff
@@ -581,11 +581,10 @@ static void handle_xevk_entityid (struct nn_xpack *xp, struct xevent_nt *ev)
 #ifdef DDSI_INCLUDE_SECURITY
 static void send_heartbeat_to_all_readers(struct nn_xpack *xp, struct xevent *ev, struct writer *wr, nn_mtime_t tnow)
 {
-  struct nn_xmsg **msg = NULL;
   struct whc_state whcst;
   nn_mtime_t t_next;
   int hbansreq = 0;
-  unsigned count = 0, n = 0, i;
+  unsigned count = 0;
```
eboasson: ``count`` is no longer actually used — I am guessing you forgot to remove it but perhaps you have a some other plan with it?

PR #313, opened 2019-11-06
Namespace the schema and add references in xml files — take 2

eboasson:
This PR is a rebase of #282 extended with the generation of cyclonedds.xsd, cyclonedds.rnc and options.md fixed to include some attributes missing because of a bug in the generator script.

Even though these XSD, RNC and MD are mechanically generated, the files are committed in the repository, because the tools are not necessarily available everywhere and because it is nice to always have them at hand. Unfortunately, it means the committed copies can become out of date with respect to the sources.

To guard against this, this PR adds build code to generate what it can (and if perl is available, that is at least some), and fails the build if the committed files are out-of-sync. This should only ever happen when the configuration tables in ``q_config.c`` are changed, which is not all that often (fortunately).

Note that @rotu 's comment https://github.com/eclipse-cyclonedds/cyclonedds/pull/282#issuecomment-548866780 that it would be **much** better to generate the C source from a decent input form, rather than this fragile reading of the C code, is something I very much agree with. However, changing that means some real work, and I don't want to delay fixing the XSD because of it.

@rotu do you agree?

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ point to it.  E.g. (on Linux):
 
     $ cat cyclonedds.xml
     <?xml version="1.0" encoding="UTF-8" ?>
-    <CycloneDDS xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://github.com/eclipse-cyclonedds/cyclonedds/etc/cyclonedds.xsd">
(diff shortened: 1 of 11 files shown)


rotu:
I 100% agree. The config parser code is gonna be a rough time to rewrite, and keeping the schemas fresh is a good enough solution for now.

PR #315, opened 2019-11-10
support delayed endpoint matching

MarcelJordense:
Signed-off-by: Marcel Jordense <marcel.jordense@adlinktech.com>

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
@@ -12,6 +12,7 @@
 #ifndef DDSI_HANDSHAKE_H
 #define DDSI_HANDSHAKE_H
 
+#include "q_unused.h"
 #include "q_entity.h"
 #include "ddsi_security_msg.h"
 
@@ -23,8 +24,6 @@ struct participant;
 struct proxy_participant;
 struct ddsi_handshake;
 struct dssi_hsadmin;
-
-#ifdef DDSI_INCLUDE_SECURITY
 
 enum ddsi_handshake_state {
     STATE_HANDSHAKE_IN_PROGRESS,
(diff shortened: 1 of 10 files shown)


src/core/ddsi/src/q_entity.c
```diff
@@ -532,6 +582,61 @@ static void add_security_builtin_endpoints(struct participant *pp, ddsi_guid_t *
...
+  }
+}
+
+static void disconnect_participant_secure(struct participant *pp)
+{
+  struct proxy_participant *proxypp;
+  struct ephash_enum_proxy_participant it;
+  struct q_globals *gv;
+
+  gv = pp->e.gv;
```
eboasson: I've slowly converged on writing ``struct q_globals * const gv = pp->e.gv`` instead of this more verbose layout. (I would generally prefer a more compact layout, but that's something I'm willing to overlook 😄 )

src/core/ddsi/src/q_entity.c
```diff
@@ -532,6 +582,61 @@ static void add_security_builtin_endpoints(struct participant *pp, ddsi_guid_t *
...
+    ephash_enum_proxy_participant_init (&it, gv->guid_hash);
+    while ((proxypp = ephash_enum_proxy_participant_next (&it)) != NULL)
+    {
+      /* Do not start handshaking when security info doesn't match. */
+      if ((proxypp->handshake_admin) && (q_omg_is_similar_participant_security_info(pp, proxypp)))
+      {
+        ddsi_hsadmin_lock(proxypp->handshake_admin);
+        handshake = ddsi_hsadmin_register_locked(proxypp->handshake_admin, pp, proxypp, timestamp, handshake_end_cb);
+        assert(handshake);
+        if (!handshake) {
```
eboasson: I'm pretty easy on the formatting, but to have different styles in a single function is a bit too much.

Secondly, if the ``assert(handshake)`` is correct, the test is superfluous; if the test is needed, then I'd be curious to know what will happen with this participant: surely without completing the handshake there will be no communication where some is expected/required and it gets to be really interesting if it succeeds for some but not all. Would it not make far more sense to refuse creation of the participant in such a case?

(And thirdly a small note: with a ``struct q_globals *gv`` in scope, you can use ``GVWARNING`` which is a bit easier.)

src/core/ddsi/src/q_entity.c
```diff
@@ -620,7 +725,22 @@ dds_return_t new_participant_guid (const ddsi_guid_t *ppguid, struct q_globals *
       if (ret != DDS_RETCODE_OK)
         goto new_pp_err_secprop;
     }
+
+    if (!q_omg_security_check_create_participant (pp, gv->config.domainId))
```
eboasson: I don't see why this check needs to be duplicated. If the ``pp->plist->qos`` doesn't specify a security config and there is a ``gv->config.omg_security_configuration`` it copies the latter in ``pp->plist->qos``. After that step, you are in the same state and can do with a single check.

src/core/ddsi/src/q_entity.c
```diff
@@ -2324,8 +2470,24 @@ static void connect_writer_with_proxy_reader (struct writer *wr, struct proxy_re
     writer_qos_mismatch (wr, reason);
     return;
   }
-  proxy_reader_add_connection (prd, wr);
-  writer_add_connection (wr, prd);
+  if (q_omg_security_check_remote_reader_permissions (prd, wr->e.gv->config.domainId, wr->c.pp))
```
eboasson: For next time: inverting and reordering the tests gives a much more compact if/else if/else 😄  pattern.

src/core/ddsi/src/q_entity.c
```diff
@@ -2644,6 +2822,143 @@ static void match_proxy_reader_with_writers (struct proxy_reader *prd, nn_mtime_
...
+  }
+
+  if (q_omg_security_match_remote_reader_enabled (wr, prd))
+  {
+    proxy_reader_add_connection (prd, wr);
+    writer_add_connection (wr, prd);
+  }
+}
+
+static void match_proxy_writer_with_participant_readers (struct proxy_writer *pwr, struct participant *pp, nn_mtime_t tnow)
```
eboasson: I suspect these are added to be able to match on completion of a handshake. I'd much rather see that done without the code duplication, but obviously iterating over all participant and then scanning all endpoints in this manner scales terribly badly.

There is a solution ... Cyclone would really benefit from more efficient matching in the discovery anyway, and I have some code lying around that would help. I'll try to create a PR for that "real soon now" and then this can be reworked a bit.

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+  int64_t permissions_hdl;
+  struct q_globals *gv;
+
+  DDSRT_UNUSED_ARG(timestamp);
+
+  assert(handshake);
+  assert(lpguid);
+  assert(ppguid);
+
+  gv = ddsi_handshake_get_globals(handshake);
```
eboasson: Based on the code in this PR it would make more sense for ``gv`` to be passed in, rather than extracted from the ``handshake`` object.

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+  assert(gv);
+
+  proxypp = ephash_lookup_proxy_participant_guid (gv->guid_hash, ppguid);
```
eboasson: I'd prefer early outs on lookup failure: it is not doing anything in those cases and it means less intendentation and less state to (mentally) track.

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+
+      case STATE_HANDSHAKE_OK:
+        DDS_CLOG (DDS_LC_DISCOVERY, &gv->logconfig, "handshake (lguid="PGUIDFMT" rguid="PGUIDFMT") succeeded\n", PGUID (*lpguid), PGUID (*ppguid));
+        update_proxy_participant_endpoint_matching(proxypp, pp);
+        ddsi_hsadmin_remove_from_fsm(proxypp->handshake_admin, handshake);
+        break;
+
+      case STATE_HANDSHAKE_TIMED_OUT:
+        DDS_CERROR (&gv->logconfig, "handshake (lguid="PGUIDFMT" rguid="PGUIDFMT") failed: (%d) Timed out\n", PGUID (*lpguid), PGUID (*ppguid), (int)result);
+        if (q_omg_participant_allow_unauthenticated(pp)) {
```
eboasson: This is intruiging: why even perform a handshake when it communication is allowed anyway? (I find the ``FAILED`` case even more intriguing. I should look it up in the spec, I know.)

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+}
+
+static int proxy_participant_create_handshake(struct q_globals *gv, struct proxy_participant *proxypp, nn_wctime_t timestamp)
+{
+  int r = 1;
+  struct participant *pp;
+  struct ddsi_handshake *handshake;
+  struct ephash_enum_participant est;
+
+  ddsi_hsadmin_lock(proxypp->handshake_admin);
```
eboasson: Is it really necessary to lock ``proxypp->handshake_admin`` for the duration of the call? It seems like it would make more sense to only lock it for the duration of ``ddsi_hsadmin_register_locked`` (see also ``connect_participant_secure``), in which case it looks like the ``lock``/``unlock`` pair of functions is superfluous.

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+  ephash_enum_participant_init (&est, gv->guid_hash);
+  while (((pp = ephash_enum_participant_next (&est)) != NULL)  && (r == 1)) {
+    /* Check if local participant is ready to communicate handshake messages. */
```
eboasson: Does this mean ``pp->bes`` can change? That will have some impact.

src/core/ddsi/src/q_entity.c
```diff
@@ -4000,8 +4456,46 @@ void new_proxy_participant
...
+#ifdef DDSI_INCLUDE_SECURITY
+  if (secure)
+  {
+    /* Secure participant detected: start handshake. */
+    if ((plist->present & PP_IDENTITY_TOKEN))
+    {
+      /* initialize the security attributes associated with the proxy participant */
+      q_omg_security_init_remote_participant(proxypp);
+      /* Check if this secure participant is allowed. */
+      if (proxy_participant_check_security_info(gv, proxypp)) {
```
eboasson: I have to think through the implications of creating the proxy participant and registering it in the GUID hash table before checking whether the security info is acceptable. (If the local participant checks against all proxies, and all proxies check against all locals, it is important to do it in this order in order to check all pairs, but making it visible possibly has other consequences as well. Possibly the parallelism needs to be eliminated.)

src/core/ddsi/src/q_entity.c
```diff
@@ -3894,6 +4209,132 @@ static void add_proxy_builtin_endpoints(
...
+
+static int proxy_participant_check_security_info(struct q_globals *gv, struct proxy_participant *proxypp)
+{
+  int r = 1;
+  struct participant *pp;
+  struct ephash_enum_participant est;
+
+  ephash_enum_participant_init (&est, gv->guid_hash);
+  while (((pp = ephash_enum_participant_next (&est)) != NULL) && (r == 1)) {
+    if (!q_omg_is_similar_participant_security_info(pp, proxypp)) {
```
eboasson: I have to think about this ... does it make sense to reject a proxy participant if there is a mismatch with a single participant? Should it not be rejected only when there is a mismatch with all participants?

The way the stack handles multiple participants, I can see that this may be necessary. That would be an argument to allow multiple instantiations of a single domain in a single process (which is something outside the scope of this PR, and of other security-related PRs).

src/core/ddsi/src/q_entity.c
```diff
@@ -4195,6 +4698,46 @@ static void delete_ppt (struct proxy_participant *proxypp, nn_wctime_t timestamp
...
+
+  DDS_CWARNING (&proxypp->e.gv->logconfig, "downgrade participant "PGUIDFMT" to non-secure\n", PGUID (proxypp->e.guid));
+
+  /* Remove security related endpoints. */
+  c = proxypp->endpoints;
+  while (c)
+  {
+    struct proxy_endpoint_common * next = c->next_ep;
+    struct entity_common *e = entity_common_from_proxy_endpoint_common (c);
+    if (is_builtin_security_endpoint(e->guid.entityid))
```
eboasson: If this is the only place ``is_builtin_security_endpoint`` gets used, it would make sense to create a table with entity ids to remove.

src/core/ddsi/include/dds/ddsi/q_entity.h
```diff
@@ -641,6 +654,10 @@ void rebuild_or_clear_writer_addrsets(struct q_globals *gv, int rebuild);
 
 void local_reader_ary_setfastpath_ok (struct local_reader_ary *x, bool fastpath_ok);
 
+void connect_writer_with_proxy_reader_secure(struct writer *wr, struct proxy_reader *prd, nn_mtime_t tnow);
```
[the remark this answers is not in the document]
MarcelJordense: In the case that the endpoint requires payload encoding or submessage encoding then communication can only start when the crypto tokens from the remote entity are available. In that case the matching is delayed and when the associated crypto tokens become available this function is used to complete the matching.

src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
```diff
@@ -0,0 +1,140 @@
...
+ * proxy participant to store the handshake information associated with each
+ * local participant.
+ */
+
+struct ddsi_hsadmin * ddsi_hsadmin_create(void);
+void ddsi_hsadmin_delete(struct ddsi_hsadmin *admin);
+void ddsi_hsadmin_clear(struct ddsi_hsadmin *admin);
+struct ddsi_handshake * ddsi_hsadmin_register_locked(struct ddsi_hsadmin *admin, const struct participant *pp, const struct proxy_participant *proxypp, nn_wctime_t timestamp, ddsi_handshake_end_cb_t callback);
+struct ddsi_handshake * ddsi_hsadmin_find(const struct ddsi_hsadmin *admin, const ddsi_guid_t *lguid);
+void ddsi_hsadmin_lock(struct ddsi_hsadmin *admin);
```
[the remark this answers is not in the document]
MarcelJordense: The lock/unlock function will be removed in the implementation of the handshake.

src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
```diff
@@ -0,0 +1,140 @@
...
+#endif
+
+struct ddsi_handshake;
+struct dssi_hsadmin;
+
+#ifdef DDSI_INCLUDE_SECURITY
+
+enum ddsi_handshake_state {
+    STATE_HANDSHAKE_IN_PROGRESS,
+    STATE_HANDSHAKE_TIMED_OUT,
```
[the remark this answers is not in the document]
MarcelJordense: Will reconsider which clock to use for the timestamps when the handshake is implemented.

src/core/ddsi/src/q_entity.c
MarcelJordense: The spec mentions the AuthenticationTimeoutWait time when this time expires the validation of the authentication will fail. In the current implementation when the handshake times out and the local participant allows unauthenticated communication then the matching continues as if the remote was unauthenticated. However it is probably better to retry the handshake at a later time. This will be taken into account with the implementation of the handshake.

src/core/ddsi/src/q_entity.c
MarcelJordense: Yes I also think that it is better to reject the proxy participant only when there is a mismatch with all the participants. In the current implementation there could be only one security configuration thus all participants would have the same security settings.

src/core/ddsi/src/ddsi_handshake.c
```diff
@@ -0,0 +1,187 @@
...
+  DDS_Security_HandshakeMessageToken *handshake_message_out;
+  DDS_Security_AuthRequestMessageToken local_auth_request_token;
+  DDS_Security_AuthRequestMessageToken *remote_auth_request_token;
+  DDS_Security_OctetSeq pdata;
+  DDS_Security_SharedSecretHandle shared_secret;
+  int handled_handshake_message;
+};
+
+struct ddsi_hsadmin {
+  struct ddsrt_h *hstab;
```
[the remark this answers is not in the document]
MarcelJordense: Removed struct ddsrt_h. The ddsi_hsadmin is not yet used and will be completed when the handshake is implemented.

src/core/ddsi/include/dds/ddsi/q_entity.h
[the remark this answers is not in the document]
MarcelJordense: When checking for a match it could be that although the match is allowed the crypto tokens which are used for the communication between those endpoints are not yet received. In that case the matching is delayed until the crypto tokens are received and then this function is called match those endpoints.

src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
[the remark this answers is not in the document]
MarcelJordense: I agree. I was planning to remove these lock/unlock functions when implementing the handshake functionality.  Is that acceptable?

src/core/ddsi/src/q_entity.c
```diff
@@ -532,6 +582,61 @@ static void add_security_builtin_endpoints(struct participant *pp, ddsi_guid_t *
...
+  {
+    timestamp = now();
+    ephash_enum_proxy_participant_init (&it, gv->guid_hash);
+    while ((proxypp = ephash_enum_proxy_participant_next (&it)) != NULL)
+    {
+      /* Do not start handshaking when security info doesn't match. */
+      if ((proxypp->handshake_admin) && (q_omg_is_similar_participant_security_info(pp, proxypp)))
+      {
+        ddsi_hsadmin_lock(proxypp->handshake_admin);
+        handshake = ddsi_hsadmin_register_locked(proxypp->handshake_admin, pp, proxypp, timestamp, handshake_end_cb);
```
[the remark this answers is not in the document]
MarcelJordense: Yes

src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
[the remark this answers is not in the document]
eboasson: Sure

src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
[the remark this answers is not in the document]
eboasson: Ok, that seems like a sensible approach

src/core/ddsi/src/q_entity.c
eboasson: That change to perform matching more efficiently went into master the other day

PR #319, opened 2019-11-14
Let dds_create_domain() return an entity (rebased and shrunk)

eboasson:
This is #304 rebased to current master and with multiple topic objects created by ``dds_create_topic`` in a single participant removed (as discussed in #304). Also with a bit of squashing and fixing up of a commit message.


eboasson:
(Checked against quite a few ROS2 tests as well on macOS and Linux.)

PR #329, opened 2019-11-19
Encoding preparations

martinbremmer:
This pull request does not actually introduce encoding and decoding. But it will introduce DDSI preparations to support that.

It's divided into three* commits to separate the three different encoding/decoding 'layers': payload, sub-message and RTPS message.

\* The fourth commit is just removing a redundant function call.

**Payload**
For this, the security attributes have to be send during the discovery to know if the writer has encoded the payload or not.

When data is send: refd_payload_iov is used and normally points to the 'normal' sample. When encoding is necessary, it'll point to a new buffer (refd_payload_encoded) causing the encoded payload to be used automatically.

When (fragmented) data is received, the sub-message header is extracted. Then the payload is decoded when necessary. That'll replace the original encoded payload within the sub-message. Now the complete receive stack can continue normally. Small problem is that the byteswap bit was part of the encoded payload, meaning that it can only be set after the decoding.

**Sub-message**
The plugin API forces that encoding the submessage has to be with a single buffer. This means that the payload has to be explicitly copied right after the submessage header (normally it's just the next iov element). The encoded submessage will replace the original one. This way, the send sequence can continue normally.

When receiving a submessage, it'll be decoded when necessary. Again, it'll replace the original encoded submessage so that the rest of the code can remain the same. 
Small caveat though. When handling the submessage, the code has to make sure that it was actually encoded previously, otherwise someone can inject clear submessages.

**RTPS message**
This has basically the same issues as the sub-message encoding and decoding. Only even more iov buffers when sending. Also, it needs participant keys iso reader/writer keys.

**Testing**
This has been tested manually by this [testpatch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3864416/testpatch.txt) patch that'll mock encryption and then run the mpt_basic tests.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_security_omg.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_security_omg.h
@@ -12,9 +12,11 @@
 #ifndef DDSI_OMG_SECURITY_H
 #define DDSI_OMG_SECURITY_H
 
-#include "dds/ddsi/q_entity.h"
 #include "dds/ddsi/q_plist.h"
+#include "dds/ddsi/q_entity.h"
 #include "dds/ddsi/q_globals.h"
+#include "dds/ddsi/q_radmin.h"
+#include "dds/ddsi/q_xmsg.h"
 
 #if defined (__cplusplus)
 extern "C" {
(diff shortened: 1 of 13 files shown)


src/core/ddsi/include/dds/ddsi/ddsi_security_omg.h
```diff
@@ -107,12 +151,326 @@ unsigned determine_publication_writer(const struct writer *wr);
  * @retval true   The proxy participant may be deleted.
  * @retval false  The proxy participant may not be deleted by this writer.
  */
-bool allow_proxy_participant_deletion(struct q_globals * const gv, const struct ddsi_guid *guid, const ddsi_entityid_t pwr_entityid);
```
eboasson: I had to look twice to realise that it tests whether deletion is allowed — were it named ``proxy_participant_deletion_allowed``, that would've been avoided. Perhaps you'd prefer to add "is" somewhere in the name, considering the functions declared further down in the file.

src/core/ddsi/src/q_xmsg.c
```diff
@@ -560,30 +681,74 @@ void nn_xmsg_add_entityid (struct nn_xmsg * m)
...
+  if (m->sec_info.use_rtps_encoding && !m->sec_info.dst_pp_handle)
+  {
+    struct proxy_participant *proxypp;
+    ddsi_guid_t guid;
+
+    guid.prefix = *gp;
+    guid.entityid.u = NN_ENTITYID_PARTICIPANT;
+
+    proxypp = ephash_lookup_proxy_participant_guid(gv->guid_hash, &guid);
+    if (proxypp)
```
eboasson: Does this mean the data will go out unencrypted if (somehow) the participant is no longer referenced by the hash table?

I am pretty sure this function only gets called when the writer still exists and that the API makes it impossible to remove the participant from the hash table (by ``delete_participant``) until all after all writers have been completely removed — and therefore that this doesn't "normally" occur. The lower level DDSI interfaces are a different story (I could call ``delete_participant`` while the writer still exists).

I'd say that, if the answer to the question is yes, it would be better to guarantee that it won't go out.

src/core/ddsi/src/q_xmsg.c
```diff
@@ -1012,6 +1180,109 @@ void nn_xpack_free (struct nn_xpack *xp)
   ddsrt_free (xp);
 }
 
+static ssize_t nn_xpack_send_rtps(struct nn_xpack * xp, const nn_locator_t *loc)
+{
+  ssize_t ret = -1;
+
+#ifdef DDSI_INCLUDE_SECURITY
+  /* Only encode when needed. */
+  if (xp->sec_info.use_rtps_encoding)
```
eboasson: Looks to me like it would be better to refactor this into a function

src/core/ddsi/src/q_xmsg.c
martinbremmer: The answer is 'no'.
It's just that instead of a specific destination, it's send (protected) in such a way that all known proxy participants can decode the RTPS message.
So, the effect is not as severe, but it should be handled anyway.

src/core/ddsi/include/dds/ddsi/ddsi_security_omg.h
martinbremmer: Renamed.

src/core/ddsi/src/q_xmsg.c
martinbremmer: The new secure_conn_write() now has an early exit when it expects to have a specific destination handle but doesn't.

src/core/ddsi/src/q_xmsg.c
martinbremmer: See the new secure_conn_write().

eboasson:
Thanks @martinbremmer !

PR #331, opened 2019-11-20
Add DOMAIN_ID, DOMAIN_TAG to config, discovery

eboasson:
This commits adds support for the DOMAIN_ID and DOMAIN_TAG parameters in
participant discovery, allowing multiple domains to share a port
number (a feature introduced in DDSI 2.3).  The tag can be configured
via Discovery/Tag.

This commit also introduces a setting Discovery/ExternalDomainId that
makes it possible to override the domain id on the network, both in what
is advertised in the DOMAIN_ID discovery parameter and in the
calculation of port numbers.  This way a single process can create two
independent domains that talk via the network, which is on occassion
useful in writing tests.

### DIFF
--- a/docs/manual/options.md
+++ b/docs/manual/options.md
@@ -101,7 +101,7 @@ The default value is: "lax".
 
 
 ### //CycloneDDS/Domain/Discovery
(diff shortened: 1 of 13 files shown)


eboasson:
(Added a comment explaining why DOMAIN_TAG must not be transmitted when the tag is the default of an empty string.)

PR #335, opened 2019-11-21
DDS Security built-in Access Control plugin

dennis-adlink:
This PR adds the build-in Access Control plugin that is part of the DDS Security implementation for Cyclone (note that the 2 commits for ddsrt changes are also in a separate PR #334 )

The Access Control Plugin API defines the types and operations necessary to support an access control mechanism for DDS Domain Participants.

Similar to other builtin plugins, the DDS Security access control plugin is built as a shared library to allow dynamic library loading on runtime. This enables DDS participants to use specific plugin 
implementations with different configurations.

This commit includes some basic tests for the access control functions. This initial version of the plugin does not support permissions expiry (not-valid-after date in permissions configuration).


src/security/core/src/dds_security_utils.c
```diff
@@ -1132,3 +1132,228 @@ DDS_Security_normalize_file(
 
 }
 
+dds_time_t
+DDS_Security_parse_xml_date(
```
eboasson: It would be better to be clearer about the time format that is acceptable ... if this (part of my own reverse-engineering notes) would help:
```
WHAT           LENGTH
yyyy-mm-dd         10
Thh:mm:ss           9
optional fraction of 1 .. 12 digits
.frac         2 .. 13
optional timezone:
Z | [+-]hh:mm   1 | 6
```

I'm quite surprised that the convoluted code is mostly correct, but as always with time conversion functions, there are some issues.

* Not accepting leap seconds on input
* Is truncation the intent when dealing with a fraction that has a sub-ns component? or should it be rounded instead? I'd expect rounding without any further documentation. I also don't like the way loop termination is handled.
* Year <= 2553 works for uint64_t timestamps, but not for int64_t; and so this can still overflow (and cause undefined behaviour)
* Notpicking: ``year % 400 == 0`` implies ``year % 100 == 0``
* Computation of the total number of leap years since 1-1-1970 doesn't account for (0 mod 100) and not (0 mod 400), which will cause a mishap in 2100.
* I don’t understand the purpose of the cast to ``int32_t`` in the ts_offset > 0 case

I couldn't be bothered to verify whether the final addition can still overflow.

I must say I am a little surprised that a hand-crafted conversion was used, rather than using ``strftime`` and ``mktime`` or something like that, but perhaps there was a good reason not to use those.

dennis-adlink:
@eboasson, in the last 4 commits I pushed on this PR, I've addressed all issues you mentioned in your review. For the `fn_match` and `DDS_Security_parse_xml_date` I've added a unit test and I think both are behaving correctly now.
The merge conflict with the security branch is trivial to solve, so to avoid having a merge commit in this PR, I suggest that you squash and rebase these commits on top of security if you agree with the changes

eboasson:
Thanks @dennis-adlink! The build failure this time is the usual maven issue. It used to be that I could restart builds, but no longer. So I am merging it anyway.

PR #340, opened 2019-11-25
Add timed callback dispatcher to security core

jrebelo:
Implement the timed callback dispatcher for the DDS Security core.

Signed-off-by: Joao Rebelo <jrebelo@s2e-systems.com>

### DIFF
--- a/src/security/core/src/dds_security_timed_cb.c
+++ b/src/security/core/src/dds_security_timed_cb.c
@@ -1,0 +1,329 @@
+/*
+ * Copyright(c) 2006 to 2019 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+
+#include "dds/ddsrt/atomics.h"
+#include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/log.h"
+#include "dds/ddsrt/sync.h"
+#include "dds/ddsrt/threads.h"
+#include "dds/ddsrt/time.h"
+#include "dds/ddsrt/fibheap.h"
+
+
+#include "dds/security/core/dds_security_timed_cb.h"
+
+
+struct dds_security_timed_dispatcher_t
+{
+    bool active;
+    void *listener;
+    ddsrt_fibheap_t events;
+};
+
+struct list_dispatcher_t
+{
+    struct list_dispatcher_t *next;
+    struct dds_security_timed_dispatcher_t *dispatcher;
+};
+
+struct event_t
+{
+    ddsrt_fibheap_node_t heapnode;
+    dds_security_timed_cb_t callback;
+    dds_time_t trigger_time;
+    void *arg;
+};
+
+static int compare_timed_cb_trigger_time(const void *va, const void *vb);
+static const ddsrt_fibheap_def_t timed_cb_queue_fhdef = DDSRT_FIBHEAPDEF_INITIALIZER(offsetof(struct event_t, heapnode), compare_timed_cb_trigger_time);
+
+static int compare_timed_cb_trigger_time(const void *va, const void *vb)
+{
+    const struct event_t *a = va;
+    const struct event_t *b = vb;
+    return (a->trigger_time == b->trigger_time) ? 0 : (a->trigger_time < b->trigger_time) ? -1 : 1;
+}
+
+struct dds_security_timed_cb_data {
+    ddsrt_mutex_t lock;
+    ddsrt_cond_t  cond;
+    struct list_dispatcher_t *first_dispatcher_node;
+    ddsrt_thread_t thread;
+    bool terminate;
+};
+
+
+static uint32_t timed_dispatcher_thread(
+        void *tcbv)
+{
+    struct list_dispatcher_t *dispatcher_node;
+    struct event_t *event;
+    dds_duration_t timeout;
+    dds_duration_t remain_time;
+    struct dds_security_timed_cb_data *tcb = (struct dds_security_timed_cb_data *)tcbv;
+
+    ddsrt_mutex_lock(&tcb->lock);
+    do
+    {
+        remain_time = DDS_INFINITY;
+        for (dispatcher_node = tcb->first_dispatcher_node; dispatcher_node != NULL; dispatcher_node = dispatcher_node->next)
+        {
+            /* Just some sanity checks. */
+            assert(dispatcher_node->dispatcher);
+            if (dispatcher_node->dispatcher->active)
+            {
+                do
+                {
+                    timeout = DDS_INFINITY;
+                    event = ddsrt_fibheap_min(&timed_cb_queue_fhdef, &dispatcher_node->dispatcher->events);
+                    if (event)
+                    {
+                        /* Just some sanity checks. */
+                        assert(event->callback);
+                        /* Determine the trigger timeout of this callback. */
+                        timeout = event->trigger_time - dds_time();
+                        if (timeout <= 0)
+                        {
+                            /* Trigger callback when related dispatcher is active. */
+                            event->callback(dispatcher_node->dispatcher,
+                                            DDS_SECURITY_TIMED_CB_KIND_TIMEOUT,
+                                            dispatcher_node->dispatcher->listener,
+                                            event->arg);
+
+                            /* Remove handled event from queue, continue with next. */
+                            ddsrt_fibheap_delete(&timed_cb_queue_fhdef, &dispatcher_node->dispatcher->events, event);
+                            ddsrt_free(event);
+                        }
+                        else if (timeout < remain_time)
+                        {
+                            remain_time = timeout;
+                        }
+                    }
+                }
+                while (timeout < 0);
+            }
+        }
+        // tcb->cond condition may be triggered before this thread runs and causes
+        // this waitfor to wait infinity, hence the check of the tcb->terminate
+        if (((remain_time > 0) || (remain_time == DDS_INFINITY)) && !tcb->terminate)
+        {
+            /* Wait for new event, timeout or the end. */
+            (void)ddsrt_cond_waitfor(&tcb->cond, &tcb->lock, remain_time);
+        }
+
+    } while (!tcb->terminate);
+
+    ddsrt_mutex_unlock(&tcb->lock);
+
+    return 0;
+}
+
+struct dds_security_timed_cb_data*
+dds_security_timed_cb_new()
+{
+    struct dds_security_timed_cb_data *tcb = ddsrt_malloc(sizeof(*tcb));
+    dds_return_t osres;
+    ddsrt_threadattr_t attr;
+
+    ddsrt_mutex_init(&tcb->lock);
+    ddsrt_cond_init(&tcb->cond);
+    tcb->first_dispatcher_node = NULL;
+    tcb->terminate = false;
+
+    ddsrt_threadattr_init(&attr);
+    osres = ddsrt_thread_create(&tcb->thread, "security_dispatcher", &attr, timed_dispatcher_thread, (void*)tcb);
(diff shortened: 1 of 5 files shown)


src/security/core/src/dds_security_timed_cb.c
```diff
@@ -0,0 +1,330 @@
...
+    d->active = true;
+
+    /* Start thread when not running, otherwise wake it up to
+     * trigger callbacks that were (possibly) previously added. */
+    if (g_thread_ptr == NULL ) {
+        g_thread_ptr = ddsrt_malloc(sizeof(*g_thread_ptr));
+        ddsrt_threadattr_t attr;
+        ddsrt_threadattr_init(&attr);
+        osres = ddsrt_thread_create(g_thread_ptr, "security_dispatcher", &attr, timed_dispatcher_thread, NULL);
+        if (osres == DDS_RETCODE_ERROR)
```
dennis-adlink: The test for failure of thread creation should be `osres != DDS_RETCODE_OK` (e.g. the Windows abstration could return `DDS_RETCODE_OUT_OF_RESOURCES` because it allocates memory for the thread context)

src/security/core/src/dds_security_timed_cb.c
```diff
@@ -0,0 +1,330 @@
...
+        {
+            fprintf(stderr, "Error: cannot create thread security_dispatcher");
+            exit(2);
```
dennis-adlink: Using the `DDS_FATAL` macro is the preferred way of handling a fatal error. This will result in a call to `dds_log()` which will call `abort()`

src/security/core/src/dds_security_timed_cb.c
jrebelo: I have replaced the condition as suggested.

src/security/core/src/dds_security_timed_cb.c
jrebelo: I replaced the two lines by DDS_FATAL.

jrebelo:
The tests that occasionally fail on some of the platforms are not related to the changes introduced.

src/security/core/src/dds_security_timed_cb.c
```diff
@@ -0,0 +1,330 @@
...
+
+struct dds_security_timed_dispatcher_t
+{
+    bool active;
+    void *listener;
+};
+
+struct queue_event_t
+{
+    struct queue_event_t *next;
```
eboasson: Implementing a priority queue with a linked list is "not so efficient". I'd much rather see it implemented using the priority queue in ddsrt (see ``fibheap.c``), just like in ``q_xevent.c`` or the lease management. It'd also eliminate most of the code, adding an event becomes a simple call to ``insert`` and checking/extracting the next event becomes ``ddsrt_fibheap_min``/``ddsrt_fibheap_extract_min``.

You would still have to keep track of the events associated with a dispatcher maintaining the current behaviour of ``dds_security_timed_dispatcher_free``, but you'd eliminate the scanning of the list of events.

jrebelo: The reason why it is currently implemented as a linked list seems to be because multiple dispatchers with multiple event can simultaneously coexist on one thread. There are two options: 1. If we replace the list by the heap we either need to keep a dispatcher list with each its own fibheap and iterate through the list for finding which events have timedout. Keeping one heap and multiple dispatchers in the way it is currently done is not really feasible since nodes in the middle of the tree might need to be removed. 2. There is only one dispatcher per thread and all the resources are allocated to that dispatcher and duplicated if more dispatchers are needed.

Would you say it is better to have the linked list as it currently is or to keep a list of dispatcher with each dispatcher having its own fibheap?

eboasson: I'm not sure I follow: right now there is a single priority queue of events (implemented as a sorted double-linked list), with the only unusual thing being that it skips any events associated with disabled dispatchers. (I don't know when/how/if/why that case is important, but I presume it is.)

Switching to a proper priority queue would mean never having events associated with a disabled dispatcher in the queue. Presumably one would let each disabled dispatcher have a list of events or a priority queue, to be added to the event queue when the dispatcher is enabled. Remains the question what to do when disabling a dispatcher. I'd just keep the current doubly-linked list to track events per dispatcher (minus the ordering).

That way the enable/disable operations get a bit more complicated because they need to add/remove the events from the priority queue, but the upside is that adding and processing become simpler.

jrebelo: Thanks for the feedback @eboasson. Unfortunately I also can't clarify what is the use case for enabling/disabling dispatchers but I also assume it is important to keep it.

If I understand your idea correctly, we would keep the information about events in two places, the priority queue and the unordered linked list. If the dispatcher is disabled, events remain in the linked list but are removed from the queue. Once the dispatcher is enabled they are added again to the queue from the linked list. Is my understanding correct?

The main issue I have with that approach is that when an event is triggered it will have to be removed both from the priority queue and the linked list. I assume this to be the most common case and handling it will increase complexity. The second thing is that removing elements from the queue by disabling the dispatcher might result in non-minimum elements having to be removed which can be a costly operation for the queue.

I would see as a better alternative to this having a list of dispatchers, each with its own priority queue.

eboasson: > Is my understanding correct?

Yes.

> The main issue I have with that approach is that when an event is triggered it will have to be removed both from the priority queue and the linked list. I assume this to be the most common case and handling it will increase complexity.

True; and I agree one would assume this to be a common case. I understand in principle how this contraption is/can be useful to the security implementation, but I don't know the details of how it will be used. That means any consideration about common cases versus uncommon case is a bit of a guess.

The model I proposed at least has the advantage that the only potentially expensive operations are enabling/disabling a dispatcher, and those seem to be unlikely to be common.

> The second thing is that removing elements from the queue by disabling the dispatcher might result in non-minimum elements having to be removed which can be a costly operation for the queue.

Removing an arbitrary element (in this implementation) means effectively a decrease-key + extract-min, so worst-case would be O(n log n) or thereabouts (all entries in the queue are owned by the dispatcher). That seems okayish.

> I would see as a better alternative to this having a list of dispatchers, each with its own priority queue.

That should work, too. Instead of a list you could also do a priority queue of active dispatchers, with each dispatcher scheduled based on the earliest event for that dispatcher.

Perhaps @MarcelJordense can shed some light on how this thing is intended to be used? It would be nice to have a better idea of that ...

MarcelJordense: Both the authentication and the access_control plugin provide an interface to set a listener. Note that on each plugin a single listener can be set. The listener is used to inform the user of the authentication and access_control, which is the ddsi core, of identity status changes (not implemented) or of certificates that are invoked (expired). The set_listener is called when initializing the plugin and removed when the plugin is unloaded. Thus the timed callback functionality should allow both the authentication and the access_control plugin to set callbacks on the expiry time of certificates and remove the timeout callback when the corresponding certificate is removed when for example a local or remote participant is deleted.

stkimmer: @MarcelJordense Thanks for the context clarification. Could you also clarify if the enabling/disabling of dispatcher is used or needed? The current interface allows creating/deleting dispatchers and enabling/disabling. The latter makes the implementation much harder.

src/security/core/src/dds_security_timed_cb.c
```diff
@@ -0,0 +1,330 @@
...
+    struct queue_event_t *prev;
+    struct dds_security_timed_dispatcher_t *dispatcher;
+    dds_security_timed_cb_t callback;
+    dds_time_t trigger_time;
+    void *arg;
+};
+
+
+static ddsrt_mutex_t g_lock;
```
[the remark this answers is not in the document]
stkimmer: I would be better to use ddsrt_once indeed. However here is a counting of dispatchers included that allows automatic starting and stopping of the thread for created dispatchers. As such the user does not need to call any init/deinit. Could you tell what is broken with it?

src/security/core/src/dds_security_timed_cb.c
MarcelJordense: Yes enabling/disabling of dispatchers is needed. A dispatcher is associated with a listener that is set on either the authentication and access_control plugin. Thus a dispatcher is enabled when the plugin is initialized and disabled when the plugin is removed.

src/security/core/src/dds_security_timed_cb.c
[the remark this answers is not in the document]
eboasson: What's broken is just the usual with such loops: calling init in parallel to fini you can end up with a schedule where init re-initializes everything while fini is still cleaning up; you need memory barriers to ensure the initialisation is actually globally visible before "initialized" can be read as true in line 101; "initialized" is used as-if atomic reads/writes are supported and as-if the compiler won't reorder it ...

If you tie it to the per-domain or to the global initialisation and teardown you avoid all those problems here. I'd do it in the per-domain ones and moreover put the global state in a dynamically allocated struct. That way the domains become wholly independent.

stkimmer:
We added a new/free pair for keeping an independent object instead of using global variables. The callback list is now implemented by using the fibheap priority queue for each dispatcher. To keep the functionality of being able to enable/disable specific dispatchers, dispatchers are kept in a linked list.  
Note that this changes the public interface and as such other pull request may need to be changed as well.

eboasson:
Thanks @jrebelo and @stkimmer. I remain thoroughly unconvinced by the interface (with the enabling and disabling in this way) despite @MarcelJordense’s comment regarding the use, but (pursuant to policy regarding this particular branch, &c.) I’ll reserve judgment until I have the complete picture. Hopefully it either will start making sense, and otherwise it is hoped that this decision will allow the implementation of DDS Security to proceed more efficiently than it otherwise would, if at the cost of additional rework.

However, I do have a number of review comments regarding various details in this particular bit of code:
* ``struct dds_security_timed_cb_data`` still has an ``init_cnt``, but it is no longer used.
* ``struct dds_security_timed_cb_data`` has a pointer to the thread id, but there always is a thread and so the indirection shouldn’t be there.
* One ought to check/modify ``tcb->terminate`` within the lock, but then obviously checking it inside the inner loop in ``timed_dispatcher_thread`` becomes of rather little value.
* I’m not convinced by the decision to hold the lock across the invocation of the event handler, but that may be acceptable given the use cases.
* ``dds_security_timed_dispatcher_free `` has a really overcomplicated loop for removing the dispatcher from the list: whether ``dispatched_node->next`` is a null pointer or makes no difference.

If it weren’t for the locking issue, I’d be willing to accept it as-is. It doesn’t make too much sense to keep at this little bit of code if it all has to be reworked anyway because of changes to the interface …

jrebelo:
Thank you for the feedback. We agree that there might be a better/simpler way to reach the goal here but, as you say, without a complete picture of the use cases it is indeed hard to rationalize about it. That being said, refactoring later will probably be a good idea anyway.

Regarding your comments:
- We removed the unused init_cnt from the struct
- We removed the indirection
- tcb->terminate is modified within the lock and the (now unnecessary) checks in the loop removed. One of the checks remained to handle the case when the thread is initialised and terminated in quick succession which can cause a deadlock.
- with the current architecture we don't know how to do different. Can be also part of a later refactoring depending on what changes in the architecture are done
- we simplified the loop



On 17 Jan 2020 15:03, eboasson <notifications@github.com> wrote:

Thanks @jrebelo<https://github.com/jrebelo> and @stkimmer<https://github.com/stkimmer>. I remain thoroughly unconvinced by the interface (with the enabling and disabling in this way) despite @MarcelJordense<https://github.com/MarcelJordense>’s comment regarding the use, but (pursuant to policy regarding this particular branch, &c.) I’ll reserve judgment until I have the complete picture. Hopefully it either will start making sense, and otherwise it is hoped that this decision will allow the implementation of DDS Security to proceed more efficiently than it otherwise would, if at the cost of additional rework.

However, I do have a number of review comments regarding various details in this particular bit of code:

  *   struct dds_security_timed_cb_data still has an init_cnt, but it is no longer used.
  *   struct dds_security_timed_cb_data has a pointer to the thread id, but there always is a thread and so the indirection shouldn’t be there.
  *   One ought to check/modify tcb->terminate within the lock, but then obviously checking it inside the inner loop in timed_dispatcher_thread becomes of rather little value.
  *   I’m not convinced by the decision to hold the lock across the invocation of the event handler, but that may be acceptable given the use cases.
  *   dds_security_timed_dispatcher_free has a really overcomplicated loop for removing the dispatcher from the list: whether dispatched_node->next is a null pointer or makes no difference.

If it weren’t for the locking issue, I’d be willing to accept it as-is. It doesn’t make too much sense to keep at this little bit of code if it all has to be reworked anyway because of changes to the interface …

eboasson:
Closed via #374 after rebasing.

PR #345, opened 2019-12-02
Register local entities

MarcelJordense:


MarcelJordense:
This pull request implements the functions which are used to register the entities on creation with the security plugins when security is enabled. Note that the actual integration with the security plugin initialization and loading is not included.

MarcelJordense:
I will first process the review comments made by @martinbremmer and close this pull request. When that is finished I will create a new pull request.

Issue #343, opened 2019-11-29
mpt_basic assert


martinbremmer:
When running mpt_basic, it asserts:
```
=====================================================
Suite: helloworld
Test:  domain_42
=====================================================
--- [Subscriber(24147)] Start(-1) ...
=== [Publisher(24148)] Start(42) ...
--- [Subscriber(24147)] Waiting for 1 sample(s) ...
=== [Publisher(24148)] Send: { 24148, Now domain 42 is used }
--- [Subscriber(24147)] Received: { 24148, Now domain 42 is used }
mpt_basic: /home/martinb/cyclone/cyclonedds/src/core/ddsi/src/q_lease.c:75: void lease_management_term(struct q_globals *): Assertion `ddsrt_fibheap_min (&lease_fhdef, &gv->leaseheap) == NULL' failed.
mpt_basic: /home/martinb/cyclone/cyclonedds/src/core/ddsi/src/q_lease.c:75: void lease_management_term(struct q_globals *): Assertion `ddsrt_fibheap_min (&lease_fhdef, &gv->leaseheap) == NULL' failed.
Process helloworld::domain_42::pub failed (1)
Process helloworld::domain_42::sub kill(24147)
```

The backtrace is
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f5024996801 in __GI_abort () at abort.c:79
#2  0x00007f502498639a in __assert_fail_base (fmt=0x7f5024b0d7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7f5025c261a0 <.str> "ddsrt_fibheap_min (&lease_fhdef, &gv->leaseheap) == NULL", 
    file=file@entry=0x7f5025c26200 <.str.1> "/home/martinb/cyclone/cyclonedds/src/core/ddsi/src/q_lease.c", line=line@entry=75, 
    function=function@entry=0x7f5025c26260 <__PRETTY_FUNCTION__.lease_management_term> "void lease_management_term(struct q_globals *)") at assert.c:92
[...]

PR #346, opened 2019-12-04
Fixes for liveliness qos implementation

dennis-adlink:
This PR fixes a few issues in the liveliness qos implementation:
- Moved de-registration of a lease to a separate function, so that it can be called synchronously when a proxy writer is deleted, to prevent the lease from expiring (before this change, the lease was renewed with `T_NEVER`, but that does not work because of a check with early out in lease_renew)
- Handle `proxypp->owns_lease` correctly: when an OpenSplice instance was running in the same network and participants from OpenSplice were discovered, the lease-renewal fails in case the proxy participant's lease was not registered in minl_auto, which happens when the proxypp depends on its parent (ddsi2) participant. This fixes issue #343 
- Increased lease duration in create_delete_writer stress test to avoid failed tests due to delayed pmd message reception
- Fixed the indenting in liveliness tests source file

### DIFF
--- a/src/core/ddsc/tests/liveliness.c
+++ b/src/core/ddsc/tests/liveliness.c
@@ -44,44 +44,44 @@ static dds_entity_t g_sub_subscriber = 0;
 
 static char *create_topic_name(const char *prefix, uint32_t nr, char *name, size_t size)
 {
-        /* Get unique g_topic name. */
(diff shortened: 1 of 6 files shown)


eboasson:
Thanks @dennis-adlink ☺️

PR #347, opened 2019-12-06
Finite State Machine for Security

kurtuluso:
A generic FSM has been added to DDS Security Core component to realize the authentication handshake process.
The list of the states and the transitions are given in the creation and the FSM is started with a start call.
Passing arguments to transition funstions is possible.
Timeout transitions are possible.

I could not find the equivalent of global init and exit functions:
https://github.com/kurtuluso/cyclonedds/blob/fsm/src/security/core/src/fsm.c#L482
https://github.com/kurtuluso/cyclonedds/blob/fsm/src/security/core/src/fsm.c#L469

And the threads should be clean up gracefully at exit. A call back can be send to DDSI in the future:
https://github.com/kurtuluso/cyclonedds/blob/fsm/src/security/core/src/fsm.c#L487

Signed-off-by: Kurtulus Oksuztepe <kurtulus.oksuztepe@adlinktech.com>

### DIFF
--- a/src/security/core/src/dds_security_fsm.c
+++ b/src/security/core/src/dds_security_fsm.c
@@ -9,91 +9,87 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
-
 #include <string.h>
 #include <assert.h>
-#include "dds/security/core/dds_security_fsm.h"
-#include "dds/ddsrt/threads.h"
+#include <stdbool.h>
+
 #include "dds/ddsrt/sync.h"
 #include "dds/ddsrt/heap.h"
 #include "dds/ddsrt/atomics.h"
 #include "dds/ddsrt/retcode.h"
 #include "dds/ddsrt/time.h"
 #include "dds/ddsrt/timeconv.h"
-#include <stdbool.h>
+#include "dds/ddsrt/fibheap.h"
+#include "dds/ddsi/q_thread.h"
+#include "dds/security/core/dds_security_fsm.h"
 
-typedef struct dds_security_fsm {
-  const dds_security_fsm_transition *transitions;
-  int size;
-  void *arg;
-  const dds_security_fsm_state *current;
-  dds_time_t current_state_endtime;
-  ddsrt_atomic_uint32_t ref_cnt;
-  dds_security_fsm_debug debug_func;
-  struct dds_security_fsm_context *context;
 
-  struct dds_security_fsm *next;
-  struct dds_security_fsm *prev;
-} dds_security_fsm;
-
-typedef struct fsm_event {
+struct fsm_event
+{
   struct dds_security_fsm *fsm;
   int event_id;
   struct fsm_event *next;
   struct fsm_event *prev;
-} fsm_event;
+};
 
-typedef struct fsm_state_timeout {
+typedef enum fsm_timeout_kind {
+  FSM_TIMEOUT_STATE,
+  FSM_TIMEOUT_OVERALL
+} fsm_timeout_kind_t;
+
+struct fsm_timer_event
+{
+  ddsrt_fibheap_node_t heapnode;
   struct dds_security_fsm *fsm;
+  fsm_timeout_kind_t kind;
   dds_time_t endtime;
-} fsm_state_timeout;
+};
 
-typedef struct fsm_overall_timeout {
-  struct dds_security_fsm *fsm;
-  dds_time_t endtime;
-  dds_security_fsm_action func;
-  struct fsm_overall_timeout *next;
-  struct fsm_overall_timeout *prev;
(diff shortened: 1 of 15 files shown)


src/security/core/src/fsm.c
```diff
@@ -0,0 +1,755 @@
...
+
+    if (ut__fsm_context) {
+      /* os_procAtExit( ut__fsm_fini ); ??? */
+      ut_fsm_initialized = true;
+    } else {
+      ret = false;
+    }
+  } else {
+    if (ut_fsm_initialized == false) {
+      dds_duration_t sleep = 100000000; /* 100ms */
```
dennis-adlink: Use DDS_MSECS(100)

src/security/core/src/fsm.c
```diff
@@ -0,0 +1,755 @@
...
+#include "dds/security/core/fsm.h"
+#include "dds/ddsrt/threads.h"
+#include "dds/ddsrt/sync.h"
+#include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/process.h"
+#include "dds/ddsrt/atomics.h"
+#include "dds/ddsrt/retcode.h"
+#include "dds/ddsrt/time.h"
+#include "dds/ddsrt/timeconv.h"
+#include <assert.h>
```
dennis-adlink: Duplicate include

eboasson:
Thanks @kurtuluso and @dennis-adlink for this PR and the review comments so far. I have a few of my own as well, ones that really echo those I just made for #340:
* Setup and teardown need to be invoked from ``rtps_init``/``rtps_fini`` and it would be much better to have the state in dynamically allocated memory.
* The addition of ``ddsrt_gettid_for_thread`` needs a stronger argumen than ``dds_security_fsm_free`` alone, because it is the consequence of what seems to be an overly complicated implementation. Minimising impact to the work (and work-in-progress) that depends on this FSM machinery means avoiding changes to the interface, but the way I see it, that is doable with reasonable overhead if one were to replace the list of FSMs by an index (keyed on the address) and checking whether the FSM has been freed or not after the callback got invoked. (Alternatively, one could just create a ``FREE`` state ...)
* Two threads seems like total overkill. Perhaps I overlooked something but I'd think that tracking pending timeouts is easy enough using a priority queue, which then means the ``fsm_thread`` can simply block until an event arrives or the first timeout expires. So why another thread?
* Thread creation would best be integrated more tightly with the rest of the DDSI stack code, so that thread attributes can be set like for all other threads.

src/security/core/src/dds_security_fsm.c
```diff
@@ -0,0 +1,520 @@
...
+  for (i = 0; i < size; i++)
+  {
+    /* It needs to have a start. */
+    if (transitions[i].begin && transitions[i].event_id == DDS_SECURITY_FSM_EVENT_AUTO)
+      return true;
+  }
+  return true;
+}
+
+static void add_fsm_to_list (struct dds_security_fsm_control *control, struct dds_security_fsm *fsm)
```
eboasson: I foresee trouble here, it looks like some editing mishap occurred.

src/security/core/src/dds_security_fsm.c
```diff
@@ -0,0 +1,520 @@
...
+  ddsrt_mutex_lock (&fsm->control->lock);
+  if (fsm->active)
+  {
+    fsm_dispatch (fsm, event_id, prio);
+    ddsrt_cond_signal (&fsm->control->cond);
+  }
+  ddsrt_mutex_unlock (&fsm->control->lock);
+}
+
+const dds_security_fsm_state* dds_security_fsm_current_state (struct dds_security_fsm *fsm)
```
eboasson: Is this not missing a mutex lock/unlock pair?

src/security/core/src/dds_security_fsm.c
```diff
@@ -0,0 +1,520 @@
...
+{
+  assert(fsm);
+  assert(fsm->active);
+
+  return fsm->current;
+}
+
+void dds_security_fsm_set_debug (struct dds_security_fsm *fsm, dds_security_fsm_debug func)
```
eboasson: Is this not missing a mutex lock/unlock pair?

src/security/core/src/dds_security_fsm.c
```diff
@@ -0,0 +1,520 @@
...
+      }
+      else
+      {
+        dds_time_t now = dds_time ();
+        dur_to_wait = timeout - now;
+      }
+      if (dur_to_wait > 0)
+      {
+        thread_state_asleep (ts1);
+        (void)ddsrt_cond_waitfor( &control->cond, &control->lock, dur_to_wait);
```
eboasson: There is also a ``ddsrt_cond_waituntil``

PR #348, opened 2019-12-07
Define NOMINMAX to suppress windows.h min/max macros

rotu:

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/types/windows.h
+++ b/src/ddsrt/include/dds/ddsrt/types/windows.h
@@ -16,6 +16,10 @@
 #define WIN32_LEAN_AND_MEAN
 #endif
 
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
 #include <windows.h>
 #include <VersionHelpers.h>
 #include <stdint.h>


rotu:
@eboasson could you take a gander at this?

nuclearsandwich:
ROS 2 CI building up to rmw_cyclonedds_cpp and testing rmw_cyclonedds_cpp

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4640/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7241/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8755/)

eboasson:
From what I can find on the web this is a sensible change to make, but it doesn't seem to make much sense to do rush this change if it doesn't solve the issue for reasons not yet understood …

k0ekk0ek:
Hi @rotu. Can you describe the problem that you're running into a bit more? I'd like to understand why this change should be included. Maybe some build output? If I build ROS2 and the RMW layer for Eclipse Cyclone DDS on Windows will I get the same error?

rotu:
Yeah. min/max are very common identifiers and are used in the C++ standard library, e.g. in ` std::numeric_limits<uint32_t>::max()` and in `std::max`. Since this is a public facing header, exposing a macro with those names kills the build.

Here's an example of these kinds of weird results (in this case, it appears that this is not the only place the header is pulled in, so suppressing the macro here does not fully fix the build issue - I need to do a build on Windows myself in order to track down the other places).

https://ci.ros2.org/job/ci_windows/8749/consoleText
```
C:\J\workspace\ci_windows\ws\src\ros2\rmw_cyclonedds\rmw_cyclonedds_cpp\src\Serialization.cpp(208,5): warning C4003: not enough arguments for function-like macro invocation 'max' [C:\J\workspace\ci_windows\ws\build\rmw_cyclonedds_cpp\rmw_cyclonedds_cpp.vcxproj]
```

eboasson:
I've managed to get a Windows box in the hope I will eventually manage to eliminate the problem. The signs aren't good, but I have not yet given up hope ...

As far as this PR goes: I must have misunderstood why the ROS2 CI build badge above was/is still "failing" and so my original reason for not merging it (yet) turns out to have been incorrect.

Looking at the mess that is Windows header files, I do wonder what the consequences of including ``Windows.h`` (in any way) are for application code. If we include it with ``WIN32_LEAN_AND_MEAN`` and ``NOMINMAX`` defined, can an application source file then later add the missing definitions by re-including it? I'm not sure ... if it can, consider this PR approved.

Anyway, the only real solution seems to be to not include it anywhere except in the C files that implement the platform interface for Windows, but that means we must then use our own compatible definitions for various types (handles, sockets, IO vectors, mutexes, condition variables, ...) — and then somehow check they are compatible in the implementation part where we have the "real" ones available. And all that is a real problem!

What should definitely be possible as a first step is to clean up the includes a bit to avoid pulling in more than we need to.

k0ekk0ek:
@eboasson, @rotu, while these headers can be included by external applications, the original intent was simply to provide a decent enough abstraction layer for _this_ project. I'd prefer to not go out of our way and implement yet another abstraction on top of our own (We've had that in the past and we know what that's going to look like, let's say it's not exactly LEAN_AND_MEAN :sweat_smile:).

@rotu, if you add a comment to the source that describes why the define is there. i.e. c++ and the min/max defines I think this patch is a pragmatic short-term solution. Judging by `windows.h` you can't include it twice though...

@eboasson, I think you're on the right track. `windows.h` is too much and I think we can do better indeed. I had a quick look and we should be able to just include a subset of the headers that `windows.h` includes (`basetsd.h` gets you `SSIZE_T`). Then (for example) in the sync header include just enough for the types used there (`synchapi.h`?).

eboasson:
@k0ekk0ek if you don't mind, I'll merge this PR as an improvement over the current state, because I don't see the proper, long-term solution happening in the very near future.

k0ekk0ek:
@eboasson, I don't mind. It seems to be the most pragmatic short-term solution :+1:

PR #349, opened 2019-12-07
Match getopt signature to posix standard

rotu:
This came up when I was trying to cross-compile on Mac:
```
colcon build --cmake-args "-DCMAKE_TOOLCHAIN_FILE=$PWD/cmake/Win.cmake" --packages-select cyclonedds --cmake-clean-cache --cmake-args " -DBUILD_TESTING=OFF"

Starting >>> cyclonedds
--- stderr: cyclonedds                              
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/tools/pubsub/pubsub.c:33:
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/tools/pubsub/common.h:15:
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/include/dds/dds.h:38:
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/ddsrt/include/dds/ddsrt/time.h:26:
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/ddsrt/include/dds/ddsrt/types.h:23:
In file included from /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/ddsrt/include/dds/ddsrt/types/posix.h:20:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/unistd.h:509:6: error: conflicting types for 'getopt'
int      getopt(int, char * const [], const char *) __DARWIN_ALIAS(getopt);
         ^
/opt/ros/master/build/cyclonedds/src/ddsrt/include/getopt.h:22:16: note: previous declaration is here
DDS_EXPORT int getopt(int argc, char **argv, const char *opts);
               ^
1 error generated.
gmake[2]: *** [src/tools/pubsub/CMakeFiles/pubsub.dir/build.make:63: src/tools/pubsub/CMakeFiles/pubsub.dir/pubsub.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:625: src/tools/pubsub/CMakeFiles/pubsub.dir/all] Error 2
gmake: *** [Makefile:152: all] Error 2
---
Failed   <<< cyclonedds	[ Exited with code 2 ]

Summary: 0 packages finished [17.3s]
  1 package failed: cyclonedds
  1 package had stderr output: cyclonedds
```

### DIFF
--- a/src/ddsrt/include/getopt.h.in
+++ b/src/ddsrt/include/getopt.h.in
@@ -19,6 +19,6 @@ DDS_EXPORT extern int optind;
 DDS_EXPORT extern int optopt;
 DDS_EXPORT extern char *optarg;
 
-DDS_EXPORT int getopt(int argc, char **argv, const char *opts);
+DDS_EXPORT int getopt(int argc, char *const argv[], const char *opts);
 
 #endif /* GETOPT_H */
--- a/src/ddsrt/src/getopt.c
+++ b/src/ddsrt/src/getopt.c
@@ -37,7 +37,7 @@ char *optarg;
 int
 getopt(
     int argc,
-    char **argv,
+    char *const argv[],
     const char *opts)
 {
     static int sp = 1;


k0ekk0ek:
Hi @rotu. I think the problem is not really the signature of the function here(?). As you stated, you're trying to cross-compile for Windows, but the macOS version for `getopt.h` is still used. Probably best to have another look at the Toolchain file. Maybe you can post it here?

rotu:
@k0ekk0ek I don't disagree. I'm sure that my cross-compilation setup is not correct, and I don't expect to get it working. I still think it's a code smell, and possibly `#include <getopt.h>` should be changed to `#include "getopt.h"` if the intent is to override instead of use the system header.

k0ekk0ek:
The `<include.h>` vs `"include.h"` means the compiler first searches the system include directories for the first one and switches to local directories if it's not found, the latter means check local include directories before system include directories. On platforms where `getopt.h` is available, that's exactly what we want. Note that the `getopt.h` in Eclipse Cyclone DDS also shouldn't be "exported" on those platforms. On Windows the file is not available on the system, but if it were, we'd like to use that one instead. In your cross-compilation setup the file is available in `/usr/include` (probably), but that directory shouldn't even be checked by the compiler.

So on Windows, yes. Using `"getopt.h"` is more correct, on other supported platforms it is not. I'm not against changing the signature perse, but it'll not make a difference in your setup. Maybe we should bundle the necessary changes once you get cross-compilation to work for Windows?

Just out of curiosity, which compiler are you using? Are you trying to get Visual Studio going on Wine or are you using MinGW?

eboasson:
> The <include.h> vs "include.h" means the compiler first searches the system include directories for the first one and switches to local directories if it's not found, the latter means check local include directories before system include directories.

I'm pretty sure the "and switches to local directories if it's not found" is not true in practice (although the spec gives enough leeway to implementors to pretty much do whatever they like ...  as usual ...)

What surprises me is that the macOS include files get used by a cross-compiler. Anyway, there is also this in src/ddsrt/CMakeLists.txt:
```cmake
if(WIN32)
  configure_file("include/getopt.h.in" "include/getopt.h" COPYONLY)
  list(APPEND headers "${CMAKE_CURRENT_BINARY_DIR}/include/getopt.h")
  list(APPEND sources "${CMAKE_CURRENT_SOURCE_DIR}/src/getopt.c")
endif()
```
So the cross environment is not only dependent on the compiler but also on what is defined when running CMake.

> Just out of curiosity, which compiler are you using? Are you trying to get Visual Studio going on Wine or are you using MinGW?

I am guessing clang 😄

rotu:
I'm building on a Mac, and I was trying to use some combination of clang/gcc/MinGW.

k0ekk0ek:
Correct approach here is probably to make a `dds/ddsrt/getopt.h` that includes the system `getopt` on platforms where it's available and exports it's own signature otherwise. The C file is conditionally compiled already. @rotu, I'm happy to pull the signature change if the other changes are included too. Also wondering if this is still current and if you ever managed to get the cross compilation to work :slightly_smiling_face:

Checking if the platform supports it is something best done in CMake (like autotools). I've been contemplating introducing an autotools like configure header, but that's something for the future.

rotu:
I'm closing this as I had no idea what I'm doing and am not working on a cross-compiling solution  right now.

PR #353, opened 2019-12-09
Refer to roundtrip example

wilcobonestroo:
Text refers to roundtrip example and code to helloworld.

Signed-off-by: Wilco Bonestroo <w.j.bonestroo@saxion.nl>


k0ekk0ek:
I think the `Signed-off-by` footer has the wrong format. Please try `Signed-off-by: First Last <local-part@domain.name>`. You can then simply force-push and the check should pass.

PR #338, opened 2019-11-23
Update README.md

wilcobonestroo:
I started with a clean Ubuntu 18.04.3 LTS installation and GIT must also be installed manually.

PR #354, opened 2019-12-09
Update README.md

wilcobonestroo:
Signed-off-by: Wilco Bonestroo <w.j.bonestroo@saxion.nl>


k0ekk0ek:
Can you recreate the commit? I think the ECA check is confused because #338 used the same commit id. Fyi, committers can hit an ECA revalidate button if you force-push, so no need to create a new PR every time :wink:

Issue #356, opened 2019-12-10
About dds_takecdr() (and is missing twin brother dds_readcdr())


TheFixer:
Hi,

I noticed a few issues regarding the implementation of dds_takecdr().

1. There is no documentation about dds_takecdr(), and its twin brother dds_readcdr() is missing
2. The signature of dds_takecdr() seems to deviate from the interface of other take-implementation as far as the position of the dds_sample_info_t parameter is concerned. In most other take() implementations it is after the maxs parameter, in dds_takecdr() it is before the maxs parameter. To me this is unnecessarily confusing.

I am happy to address issue 1 and provide a PR for it if you think this makes sense. I doubt whether you think it is worthwhile to address 2 since this would involve an API change.

Can you let me know if it makes sense to provide a PR for issue 1? Can you also let me know what your position is regarding issue 2?

eboasson:
I'd be happy to accept a PR that addresses 1, and in principle I would be happy with one addressing 2 as well. I doubt there are many users of ``dds_takecdr`` but there is at least one (https://github.com/ros2/rmw_cyclonedds).

I have another suggestion for dealing with point 2. I remember ``takecdr`` got introduced as a quick hack to test performance for invoking operations via a unix stream socket and it sort of managed to escape that old experiment and made its way into the API. Back then, Cyclone really only dealt with CDR serialised data, but today the only remnant of that is the message validation code that requires the "encoding" field to be set to CDR.

So, it would arguably make sense to switch to a new name, say ``read_ref`` and ``take_ref``, because what they really do is hand out references to the internal abstracted samples. (This is something different from the "loaned samples" in the spec, because those are in application representation.) In those new functions, the parameters can have the correct order and ``dds_takecdr`` can then become a deprecated wrapper.

Does that make sense?

TheFixer:
@eboasson Yes, that makes sense. I'll try to address these issues one by one in different PRs.

PR #357, opened 2019-12-10
Add missing documentation for dds_takecdr()

TheFixer:
This pull request adds missing documentation for dds_takecdr(). The related issue is #356.

I decided NOT to implement dds_readcdr() because it would become deprecated implemented once the pull request for item 2 of #356 appears.

Signed-off-by: TheFixer <thefixer@iteazz.com>

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -1712,12 +1712,28 @@ DDS_EXPORT void
 dds_write_flush(dds_entity_t writer);
 
 /**
- * @brief Write a CDR serialized value of a data instance
+ * @brief Write a serialized value of a data instance
+ *
+ * This call causes the writer to write the serialized value that is provided
+ * in the serdata argument.
  *
  * @param[in]  writer The writer entity.
- * @param[in]  serdata CDR serialized value to be written.
+ * @param[in]  serdata Serialized value to be written.
  *
  * @returns A dds_return_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK
+ *             The writer successfully wrote the serialized value.
+ * @retval DDS_RETCODE_ERROR
+ *             An internal error has occurred.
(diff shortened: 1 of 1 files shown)


src/core/ddsc/include/dds/dds.h
```diff
@@ -2668,6 +2668,42 @@ dds_take_mask_wl(
...
+ *
+ * Return value provides information about number of samples read, which will
+ * be <= maxs. Based on the count, the buffer will contain data to be read only
+ * when valid_data bit in sample info structure is set.
+ * The buffer required for data values, could be allocated explicitly or can
+ * use the memory from data reader to prevent copy. In the latter case, buffer and
+ * sample_info should be returned back, once it is no longer using the data.
+ *
+ * @param[in]  reader_or_condition Reader, readcondition or querycondition entity.
+ * @param[out] buf An array of pointers to samples in CDR format into which data is read (pointers can be NULL).
```
martinbremmer: As far as I can tell, ```struct ddsi_serdata **buf``` is not an array of pointers, but a pointer to an array of ```ddsi_serdata``` structures in which I don't spot any pointers to actual cdr data. Am I missing something?

TheFixer: @martinbremmer You are correct, the suggestion that this is restricted to cdr data is incorrect.
The `buf` is a pointer to an array containing pointers to `struct ddsi_serdata`. This structure contains (references to) serialised data and functions to access the data, but not necessarily in cdr format.  So I agree with your comment and will change the description. Unfortunately I cannot change the rename of the API function called `take_cdr` because that would potentially break backward compatibility.

eboasson:
Thanks @TheFixer, the documentation of Cyclone can really use some work and so this PR is very welcome.

While the reality is slightly more complicated than the documentation comment suggests (it is an abstract sample from which you can extract a serialised representation) I suspect the distinction would actually make it less clear what it does to someone not very familiar with the internals.

PR #361, opened 2019-12-13
Force md5 keyhash when needed.

martinbremmer:
When using certain Security settings, the keyhash has to be md5.
This ticket introduces the possibility to force the keyhash on the wire to be md5. It does not introduce the actual 'decision' depending on Security settings because those settings are not available yet.

Simple test patch: 
[md5_test.patch.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/3961108/md5_test.patch.txt)

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
@@ -132,6 +132,11 @@ typedef bool (*ddsi_serdata_eqkey_t) (const struct ddsi_serdata *a, const struct
      returning bufsize-1) if it had to truncate) */
 typedef size_t (*ddsi_serdata_print_t) (const struct ddsi_sertopic *topic, const struct ddsi_serdata *d, char *buf, size_t size);
 
+/* Add keyhash (from serdata) to buffer (forcing md5 when necessary).
+   - key needs to be set within serdata (can already be md5)
+   - buf needs to be at least 16 bytes large */
+typedef void (*ddsi_serdata_get_keyhash_t) (const struct ddsi_serdata *d, struct nn_keyhash *buf, bool force_md5);
+
 struct ddsi_serdata_ops {
   ddsi_serdata_eqkey_t eqkey;
   ddsi_serdata_size_t get_size;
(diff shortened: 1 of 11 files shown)


src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
```diff
@@ -132,6 +132,11 @@ typedef bool (*ddsi_serdata_eqkey_t) (const struct ddsi_serdata *a, const struct
      returning bufsize-1) if it had to truncate) */
 typedef size_t (*ddsi_serdata_print_t) (const struct ddsi_sertopic *topic, const struct ddsi_serdata *d, char *buf, size_t size);
 
+/* Add keyhash (from serdata) to buffer (forcing md5 when necessary).
+   - key needs to be set within serdata (can already be md5)
+   - buf needs to be at least 16 bytes large */
+typedef void (*ddsi_serdata_add_keyhash_t) (const struct ddsi_serdata *d, char *buf, bool force_md5);
```
eboasson: I'd rather see ``struct nn_keyhash *buf`` here (the name is crappy and it is defined in the wrong place, but as it is used already for ``ddsi_serdata_from_keyhash_t``, using it here without fixing up those definitional matters doesn't introduce any new problems). That type is simply a struct with a single member, an array of 16 bytes.

Secondly, I don't like the name ``add_keyhash``: it doesn't "add" the keyhash as much as it stores it in the provided buffer. In my view the behaviour of function would be more obvious if it were called ``get_keyhash``.

src/core/ddsi/src/ddsi_serdata_default.c
```diff
@@ -656,6 +656,24 @@ static size_t serdata_default_print_raw (const struct ddsi_sertopic *sertopic_co
...
+static void serdata_default_add_keyhash (const struct ddsi_serdata *serdata_common, char *buf, bool force_md5)
+{
+  const struct ddsi_serdata_default *d = (const struct ddsi_serdata_default *)serdata_common;
+  assert(buf);
+  assert(d->keyhash.m_set);
+  if (force_md5 && d->keyhash.m_iskey /* m_iskey == !md5 */)
+  {
+    ddsrt_md5_state_t md5st;
+    ddsrt_md5_init  (&md5st);
+    ddsrt_md5_append(&md5st, (ddsrt_md5_byte_t*)(d->keyhash.m_hash), 16);
```
eboasson: From the DDS Security 1.1 spec (page 42): "The KeyHash shall be computed as the 128 bit MD5 Digest (IETF RFC 1321) applied to the CDR Big-Endian encapsulation of all the Key fields in sequence. Unlike the rule stated in sub clause 9.6.3.3 of the DDS specification, the MD5 hash shall be used regardless of the maximum-size of the serialized key."

I read that as calculating the MD5 over the bytes actually used for representing the key value, not simply all 16 bytes. That information currently gets discarded along the way ... It looks like it would be pretty trivial to let ``dds_stream_extract_keyhash`` and ``dds_stream_write_keyBE`` return the number of bytes: they have available internally anyway as a position in the output buffer and they currently don't return a value anyway.

As to storing it ... it's guaranteed to be ≤ 16 bytes; I'd just use an additional 5 bits large field in ``struct dds_keyhash``.

martinbremmer: I don't really have an counter argument. So, I think your interpretation is correct.

PR #362, opened 2019-12-16
Lifespan QoS implementation

dennis-adlink:
This PR enables specifying a duration for data to be valid when writing samples. After this duration, samples are dropped from the reader and writer history cache. See section 2.2.3.16 of the DDS specification for more details on this QoS. 

The expiration of samples in the reader history cache is calculated based on the reception timestamp of the sample and uses the monotonic clock. As a result, the current implementation does not rely on clock synchronisation between reader and writer. There may be reasons to change this behavior in future and use the source timestamp instead.

### DIFF
--- a/src/core/ddsc/src/dds_rhc_default.c
+++ b/src/core/ddsc/src/dds_rhc_default.c
@@ -39,6 +39,9 @@
 #include "dds/ddsi/q_entity.h" /* proxy_writer_info */
 #include "dds/ddsi/ddsi_serdata.h"
 #include "dds/ddsi/ddsi_serdata_default.h"
+#ifdef DDSI_INCLUDE_LIFESPAN
+#include "dds/ddsi/ddsi_lifespan.h"
+#endif
 #include "dds/ddsi/sysdeps.h"
 
 /* INSTANCE MANAGEMENT
@@ -122,9 +125,8 @@
    DDSI implementation, but still) will be done by also reseting "wr_iid"
    when an exclusive ownership writer lowers its strength.
 
-   Lifespan, time base filter and deadline, are based on the instance
-   timestamp ("tstamp").  This time stamp needs to be changed to either source
-   or reception timestamp, depending on the ordering chosen.
+   Lifespan is based on the reception timestamp, and the monotonic time is
+   used for sample expiry if this QoS is set to something else than infinite.
 
    READ CONDITIONS
    ===============
@@ -244,6 +246,10 @@ struct rhc_sample {
   bool isread;                 /* READ or NOT_READ sample state */
   uint32_t disposed_gen;       /* snapshot of instance counter at time of insertion */
   uint32_t no_writers_gen;     /* __/ */
+#ifdef DDSI_INCLUDE_LIFESPAN
+  struct lifespan_fhnode lifespan;  /* fibheap node for lifespan */
+  struct rhc_instance *inst;   /* reference to rhc instance */
+#endif
 };
 
 struct rhc_instance {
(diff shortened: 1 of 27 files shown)


src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -708,13 +816,18 @@ static void free_instance_rhc_free_wrap (void *vnode, void *varg)
...
+  ddsrt_mutex_lock (&rhc->lock);
   assert (rhc_check_counts_locked (rhc, true, true));
   ddsrt_hh_enum (rhc->instances, free_instance_rhc_free_wrap, rhc);
   assert (rhc->nonempty_instances == NULL);
   ddsrt_hh_free (rhc->instances);
   lwregs_fini (&rhc->registrations);
   if (rhc->qcond_eval_samplebuf != NULL)
     ddsi_sertopic_free_sample (rhc->topic, rhc->qcond_eval_samplebuf, DDS_FREE_ALL);
+#ifdef DDSI_INCLUDE_LIFESPAN
+  lifespan_fini (&rhc->lifespan);
```
eboasson: Lifespan expiry events (from the tev thread) can still fire until this point, so you can have the tev thread blocked until you release the lock, then end up in the dds_rhc_default_sample_expired_cb, at which point stuff will break.

Calling ``dds_rhc_default_sample_expired_cb(rhc,NN_MTIME_NEVER); lifespan_fini (&rhc->lifespan);`` at the beginning should ensure that all samples with lifespan attached to them are gone and that the callback event no longer executes/will execute. (I think.)

src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -708,13 +816,18 @@ static void free_instance_rhc_free_wrap (void *vnode, void *varg)
 
 static void dds_rhc_default_free (struct dds_rhc_default *rhc)
 {
+  ddsrt_mutex_lock (&rhc->lock);
```
eboasson: Locking & unlocking a mutex protecting an object prior to unconditionally freeing it and without any intervening synchronisation to ensure all threads that might have been blocked on the mutex have left their critical regions prior to that freeing is pretty much always a problem.

src/core/ddsc/src/dds_whc.c
```diff
@@ -440,6 +468,10 @@ void whc_default_free (struct whc *whc_generic)
     }
   }
 
+#ifdef DDSI_INCLUDE_LIFESPAN
```
eboasson: I think the same considerations as for the freeing of the RHC hold here.

src/core/ddsc/src/dds_write.c
```diff
@@ -99,7 +99,7 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
     {
       dds_duration_t max_block_ms = wr->xqos->reliability.max_blocking_time;
       struct ddsi_writer_info pwr_info;
-      ddsi_make_writer_info (&pwr_info, &wr->e, wr->xqos);
+      ddsi_make_writer_info (&pwr_info, &wr->e, wr->xqos, now_mt(), payload->statusinfo);
```
eboasson: Considering that all paths simply call ``now_mt()`` to get the new 4th parameter and that it is entirely superfluous when ``DDSI_INCLUDE_LIFESPAN`` is not set, it would probably make sense to make that call inside ``ddsi_make_writer_info`` instead.

(The refactoring suggested at the beginning of ``whc_default_insert_seq`` might well affect this.)

src/core/ddsi/src/q_entity.c
```diff
@@ -311,12 +311,22 @@ nn_vendorid_t get_entity_vendorid (const struct entity_common *e)
...
+#ifndef DDSI_INCLUDE_LIFESPAN
+  DDSRT_UNUSED_ARG (trecv);
+  DDSRT_UNUSED_ARG (statusinfo);
+#endif
   wrinfo->guid = e->guid;
   wrinfo->ownership_strength = xqos->ownership_strength.value;
   wrinfo->auto_dispose = xqos->writer_data_lifecycle.autodispose_unregistered_instances;
   wrinfo->iid = e->iid;
+#ifdef DDSI_INCLUDE_LIFESPAN
+  if ((statusinfo & (NN_STATUSINFO_UNREGISTER | NN_STATUSINFO_DISPOSE)) == 0 && xqos->lifespan.duration != DDS_INFINITY)
```
eboasson: I'd flip the two conjuncts: the primary consideration is whether lifespan is enabled, and only if it is is it of interest whether the thing is a dispose or unregister. (It should be marginally faster on average as well: the vast majority of the stuff is regular data with a lifespan of ∞, so flipping them means you know after one test, rather than after two. That never hurts!)

src/core/ddsi/src/q_transmit.c
```diff
@@ -910,10 +910,18 @@ static int insert_sample_in_whc (struct writer *wr, seqno_t seq, struct nn_plist
...
-    res = insres;
   else
-    res = 1;
+  {
+    nn_mtime_t exp = NN_MTIME_NEVER;
+#ifdef DDSI_INCLUDE_LIFESPAN
+    /* Don't set expiry for samples with flags unregister or dispose, because these are required
+     * for sample lifecycle and should always be delivered to the reader so that is can clean up
+     * its history cache. */
+    if ((serdata->statusinfo & (NN_STATUSINFO_UNREGISTER | NN_STATUSINFO_DISPOSE)) == 0 && wr->xqos->lifespan.duration != DDS_INFINITY)
```
eboasson: See ``ddsi_make_writer_info``

eboasson:
Thank you @dennis-adlink! I don't see any relationship to the changes for the two build failures, so that's not a problem. Two details remain:

* an indentation slip up in ``whc_default_free``
* I would like to squash the commits that implement the lifespan support. I think it makes sense to keep the "xevents" one as well as the one that updates the travis configuration.

If you're happy to do it, please go ahead and force-push them, otherwise I'll do it myself.

dennis-adlink:
@eboasson Thanks for your review comments! I've squashed the lifespan commits into one, and kept the other 2 commits for xevents and travis config. The livespan commit includes fixes for all issues you've addressed.

PR #365, opened 2019-12-27
Hook-up preprocessor to IDL compiler

k0ekk0ek:
This PR ties the preprocessor to the IDL compiler created by @FransFaase to make it at least a somewhat usable tool (it actually works :wink:).

To do so I added `ddsrt_vasprintf` and added the `DDS_RETCODE_BAD_SYNTAX` return code to `ddsrt`. Apart from that I fixed handling of `@key` and annotations in general and moved handling of compiler directives (just `#pragma keylist` for now) to the right location.

While there's still much ground to cover, all the basics are now in place and all-in-all this marks the first actual working version of `idlc`.

@eboasson, if you can find the time, please skim the changes before I merge them into the `xtypes` branch.

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/io.h
+++ b/src/ddsrt/include/dds/ddsrt/io.h
@@ -25,6 +25,15 @@ extern "C" {
  * @brief Write a formatted string to a newly allocated buffer.
  */
 DDS_EXPORT int
+ddsrt_vasprintf(
+  char **strp,
+  const char *fmt,
+  va_list ap);
+
+/**
+ * @brief Write a formatted string to a newly allocated buffer.
+ */
+DDS_EXPORT int
 ddsrt_asprintf(
   char **strp,
   const char *fmt,
--- a/src/ddsrt/include/dds/ddsrt/retcode.h
+++ b/src/ddsrt/include/dds/ddsrt/retcode.h
@@ -77,6 +77,8 @@ typedef int32_t dds_return_t;
 #define DDS_RETCODE_OUT_OF_RANGE        DDS_XRETCODE(9)
 /** Not found */
 #define DDS_RETCODE_NOT_FOUND           DDS_XRETCODE(10)
+/** Syntax error */
+#define DDS_RETCODE_BAD_SYNTAX          DDS_XRETCODE(11)
 /**
  * @}
  */
--- a/src/ddsrt/src/io.c
+++ b/src/ddsrt/src/io.c
@@ -52,6 +52,40 @@ ddsrt_asprintf(
   return ret;
 }
 
+int
+ddsrt_vasprintf(
+  char **strp,
+  const char *fmt,
+  va_list ap)
+{
+  int ret;
+  unsigned int len;
+  char buf[1] = { '\0' };
+  char *str = NULL;
+  va_list ap2;
+
+  assert(strp != NULL);
+  assert(fmt != NULL);
+
+  va_copy(ap2, ap); /* va_list cannot be reused */
+
+  if ((ret = vsnprintf(buf, sizeof(buf), fmt, ap)) >= 0) {
+    len = (unsigned int)ret;
+    if ((str = ddsrt_malloc(len + 1)) == NULL) {
+      ret = -1;
+    } else if ((ret = vsnprintf(str, len + 1, fmt, ap2)) >= 0) {
+      assert(((unsigned int)ret) == len);
+      *strp = str;
+    } else {
+      ddsrt_free(str);
+    }
+  }
+
+  va_end(ap2);
+
+  return ret;
+}
+
 #if defined(_MSC_VER) && (_MSC_VER < 1900)
 int
 snprintf(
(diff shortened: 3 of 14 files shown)


eboasson:
@k0ekk0ek it generally looks ok to me but there are a few build failures — the one I have looked at are nuisance ones. I have one comment though: at first reading it looks like calling ``idl_scan`` actually does the parsing. Given that "scanning" is typically used for no more than tokenization, that name seems less than ideal. (Or perhaps I am mistaken.)

k0ekk0ek:
Thanks @eboasson! Didn't expect you to take a look so quickly.

Yes, `idl_scan` indeed does the parsing. I think I'll rename `idl_scan_token` to `idl_scan` and the current `idl_scan` to `idl_parse`.

While writing tests though, it turned out that the scanner does not handle preprocessor directives all that well. At least, not all variants. To illustrate:
```
/* comment */ # line
# line 1 /* comment */ "filename"
# line 1 /* comment
  */ "filename"
# line 1 "filename" foo bar baz
```
The same is of true for other directives (e.g. `#pragma keylist`) as well. I'm in the process of fixing it. The changes also make the scanner much easier on the eye, which is a nice bonus.

k0ekk0ek:
Quick update.

I've tackled the preprocessor problem by implementing a niftly little parser for subset of directives that must be supported (`#line` and `#pragma`), but ran into the limitations of flex. Specifically handling of line continuations.

Normally I would have opted to simply not support line continuation as `idlpp` (mcpp) takes care of it, but since Eclipse Cyclone DDS uses compiler directives to denote keys and generating type support from strings is useful for tests and possibly runtime type generation, it makes sense to implement support.

The subset of preprocessor functionality that will be supported for type generations from strings:
 * `#line` directives (maybe without filename if generated from a string)
 * `#pragma keylist` directives
 * line continuation

The compiler itself (`idlc`) of course supports things like includes, macro expansion, conditionals, etc because `idlpp` is used as a preprocessor.

I opted to swap out [flex](https://github.com/westes/flex) for [re2c](https://re2c.org/), which fits much better.

Conceptionally the stack will look like:

    mcpp > re2c > Bison

With some glue code to make it all work.

rvloon:
I would suggest looking at Boost::Wave. This is a complete integratable C/C++ preprocessor, with callbacks for #pragma's and the like. It also offers lexer-support in general as well as semantical parsing. It can also output stuff according to a grammar. (Karma).

Advantage is that it is flexible, well-known (part of Boost) and as such does not require a dependency on something no longer maintained.


Ronald

On 01/01/2020 19:34:55, Jeroen Koekkoek <notifications@github.com> wrote:
Quick update.
I've tackled the preprocessor problem by implementing a niftly little parser for subset of directives that must be supported (#line and #pragma), but ran into the limitations of flex. Specifically handling of line continuations.
Normally I would have opted to simply not support line continuation as idlpp (mcpp) takes care of it, but since Eclipse Cyclone DDS uses compiler directives to denote keys and generating type support from strings is useful for tests and possibly runtime type generation, it makes sense to implement support.
The subset of preprocessor functionality that will be supported for type generations from strings:
* #line directives (maybe without filename if generated from a string)
* #pragma keylist directives
* line continuation
The compiler itself (idlc) of course supports things like includes, macro expansion, conditionals, etc because idlpp is used as a preprocessor.
I opted to swap out flex [https://github.com/westes/flex] for re2c [https://re2c.org/], which fits much better.
Conceptionally the stack will look like:
mcpp > re2c > Bison
With some glue code to make it all work.

FransFaase:
@rvloon, please go ahead, show us your coding skills, clone the repository, port Boost::Wave to C99, make all the tests work, and submit a pull request.

rvloon:
Internal discussions at ADLINK have led us to believe there to be no need for IDL-related activities to be implemented in C99. Sone initial experiments with Boost show that the included IDL sample might be a good starting point.RonaldOp 2 jan. 2020 09:20 schreef Frans Faase <notifications@github.com>:@rvloon, please go ahead, show us your coding skills, clone the repository, port Boost::Wave to C99, make all the tests work, and submit a pull request.

k0ekk0ek:
Thanks @rvloon. Boost::Wave looks like a very useful tool. For now I'll stay on the `mcpp` + `re2c` route though. The amount of glue code required to make everything work is rather small and introducing C++ to the mix introduces some other challenges. For example, having a pure-C stack probably makes integration into other languages easier (I'm envisioning a dynamic language based tool that allows you to paste IDL and start publishing/subscribing).

As for mcpp: while it is certainly true that it is not the most actively maintained project, it is very mature and the likelihood of new features being introduced to the C99 language specification is rather small :sweat_smile:, therefore it makes (some) sense that it does not receive many updates. That being said, there are some forks out there that have patches that I would like to incorporate.

You are absolutely right btw. in the sense that the preprocessor is only a small component and rather irrelevant to IDL. It's actually the reason why I'm just trying to get it out of the way... I'm hoping future contributions/discussions will focus on implementing IDL features rather than the preprocessor.

@rvloon: I removed the quotes from your responses for readability. You're probably replying via email, if so, can you try not to quote the original message(?)

eboasson:
@k0ekk0ek yes, please just go ahead. Moving the `xtypes` branch into the modern era while also consolidating and the bits and pieces that were floating around in old PRs is a very good thing. A few minor things here and there can easily be corrected.

k0ekk0ek:
Thanks @eboasson. Much appreciated.

PR #366, opened 2020-01-07
Deadline QoS implementation

dennis-adlink:
This PR adds the implementation of the deadline QoS for readers and writers. Details of this QoS can be found in the DDS specification section 2.2.3.7. I've also included a commit (thanks @eboasson for your contribution to these changes) with refactoring of the (cyclic) linked list implementation that is used in the reader history cache, to make it reusable for the deadline qos administration in both rhc and whc.

### DIFF
--- a/src/core/ddsc/src/dds_rhc_default.c
+++ b/src/core/ddsc/src/dds_rhc_default.c
@@ -43,6 +43,9 @@
 #ifdef DDSI_INCLUDE_LIFESPAN
 #include "dds/ddsi/ddsi_lifespan.h"
 #endif
+#ifdef DDSI_INCLUDE_DEADLINE_MISSED
+#include "dds/ddsi/ddsi_deadline.h"
+#endif
 #include "dds/ddsi/sysdeps.h"
 
 /* INSTANCE MANAGEMENT
@@ -273,6 +276,9 @@ struct rhc_instance {
   ddsi_guid_t wr_guid;           /* guid of last writer (if wr_iid != 0 then wr_guid is the corresponding guid, else undef) */
   nn_wctime_t tstamp;          /* source time stamp of last update */
   struct ddsrt_circlist_elem nonempty_list; /* links non-empty instances in arbitrary ordering */
+#ifdef DDSI_INCLUDE_DEADLINE_MISSED
+  struct deadline_elem deadline; /* element in deadline missed administration */
+#endif
   struct ddsi_tkmap_instance *tk;   /* backref into TK for unref'ing */
   struct rhc_sample a_sample;  /* pre-allocated storage for 1 sample */
 };
@@ -324,6 +330,9 @@ struct dds_rhc_default {
   void *qcond_eval_samplebuf;        /* Temporary storage for evaluating query conditions, NULL if no qconds */
 #ifdef DDSI_INCLUDE_LIFESPAN
   struct lifespan_adm lifespan;      /* Lifespan administration */
+#endif
+#ifdef DDSI_INCLUDE_DEADLINE_MISSED
+  struct deadline_adm deadline; /* Deadline missed administration */
 #endif
 };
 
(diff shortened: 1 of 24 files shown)


src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -855,6 +917,14 @@ static bool add_sample (struct dds_rhc_default *rhc, struct rhc_instance *inst,
      counts - are included in the sample).  While this would be place to do it,
      we do it later to avoid having to roll back on allocation failure */
 
+#ifdef DDSI_INCLUDE_DEADLINE_MISSED
+  /* receiving a sample from a writer always results in renewing the
```
eboasson: I wonder whether considering the arrival of a sample that doesn't get stored as sufficient reason to do this makes sense. The reader-side deadline is for monitoring whether the reader's instance got updated; if it doesn't get stored (for whatever reason), arguably it hasn't been updated. To me it makes more sense to do it if and only if the instance really has been updated.

src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -1531,16 +1604,25 @@ static bool dds_rhc_default_store (struct dds_rhc_default * __restrict rhc, cons
          WD     D;gen++     D
          Simplest way is to toggle istate when it is currently DISPOSED
          and the operation is WD. */
-      if (has_data && inst->isdisposed)
+      if (has_data && inst->isdisposed && !is_dispose)
```
eboasson: Why this change? It changes the behaviour of the disposed generations for the case of a write-dispose operation on an already disposed topic. It used to treat as a write followed by a dispose, and hence consider the instance alive, even if that was observable only by the change in the generation; now it simply remains in the disposed state. I might well be inclined to agree with it, just want to know 😄 

(And if the change stands, the comment preceding it needs to be updated)

src/core/ddsc/src/dds_whc.c
```diff
@@ -403,7 +464,7 @@ struct whc *whc_new (struct q_globals *gv, int is_transient_local, uint32_t hdep
   whc->seq_hash = ddsrt_hh_new (1, whc_node_hash, whc_node_eq);
 #endif
 
-  if (whc->idxdepth > 0)
+  if (wrinfo->idxdepth > 0)
```
eboasson: Not having an index for a KEEP_ALL topic is attractive for performance, but the downside is that there is then no instance at all to use in a deadline administration. Correctly implementing register/unregister appears to require one as well, so it appears one should be created always.

So as things stand now, I suspect "deadline" won't work for a KEEP_ALL volatile writer and for a KEEP_ALL transient-local writer with a KEEP_ALL durability_service.history setting.

src/core/ddsi/include/dds/ddsi/ddsi_deadline.h
```diff
@@ -0,0 +1,83 @@
...
+
+typedef nn_mtime_t (*deadline_missed_cb_t)(void *hc, nn_mtime_t tnow);
+
+struct deadline_adm {
+  struct ddsrt_circlist list;               /* linked list for deadline missed */
+  struct xevent *evt;                       /* xevent that triggers when deadline expires for an instance */
+  deadline_missed_cb_t deadline_missed_cb;  /* callback for deadline missed; this cb can use deadline_next_missed_locked to get next instance that has a missed deadline */
+  size_t list_offset;                       /* offset of deadline_adm element in whc or rhc */
+  size_t elem_offset;                       /* offset of deadline_elem element in whc or rhc instance */
+  ddsi_duration_t dur;                      /* deadline duration */
```
eboasson: A ``ddsi_duration_t`` matches the wire format (fixed-point with a resolution of 2^-32s) and forces conversions left and right. It'd be better to represent it as nanoseconds throughout.

src/core/ddsi/src/ddsi_deadline.c
```diff
@@ -0,0 +1,108 @@
...
+
+/* Gets the instance from the list in deadline admin that has the earliest missed deadline and
+ * removes the instance element from the list. If no more instances with missed deadline exist
+ * in the list, the deadline (nn_mtime_t) for the first instance to 'expire' is returned. If the
+ * list is empty, NN_MTIME_NEVER is returned */
+nn_mtime_t deadline_next_missed_locked (struct deadline_adm *deadline_adm, nn_mtime_t tnow, void **instance)
+{
+  struct ddsrt_circlist_elem *list_elem = NULL;
+  struct deadline_elem *elem = NULL;
+  if (!ddsrt_circlist_isempty (&deadline_adm->list) && (list_elem = ddsrt_circlist_oldest (&deadline_adm->list)) != NULL)
```
eboasson: Surely ``list_elem`` can't be a null pointer if the list isn't empty?

src/core/ddsi/src/ddsi_deadline.c
```diff
@@ -0,0 +1,108 @@
...
+  {
+    elem = DDSRT_FROM_CIRCLIST (struct deadline_elem, e, list_elem);
+    if (elem->t_deadline.v <= tnow.v)
+    {
+      ddsrt_circlist_remove (&deadline_adm->list, &elem->e);
```
eboasson: In the normal course of events, one would reregister the instance immediately afterward, right? At which point it will be the new latest and so you're really just rotating the circular list by one position. Then again, missing deadlines would be an indication of trouble anyway, so it is good enough for now 😄 .

dennis-adlink:
@eboasson thanks for the review. In my latest 2 commits I added fixes for the issues you found:
- Refactored `whc_default_insert` so that an index node is also created in case of index depth 0, which fixes deadline for KEEP_ALL writers
- Introduced `deadline_clear` function and use this in whc; assume deadline administration is clean when rhc is freed
- Changed duration type to dds_duration_t in deadline_adm
- Removed unneeded check in `deadline_next_missed_locked`
- Moved condition `!is_dispose` in rhc store so that `disposed_gen` counting works correctly again

PR #367, opened 2020-01-09
Remove non-standard "CM" discovery topics

eboasson:
There was a race condition in updating a proxy participant's "plist" on receipt of the non-standard "CMParticipant" discovery topic, which at least once caused the address sanitizer to report a use-after-free.

It could have been fixed by locking/unlocking the proxy participant's ``qos_lock`` mutex around the update of its ``plist``, but as these topics are non-standard and not actually used anywhere in Cyclone, it makes rather more sense to remove them than to fix this.


k0ekk0ek:
@eboasson, judging by the changes only the `CMParticipant` topic was properly supported, which means that OpenSplice nodes won't get the process name and identifier when they discover an Eclipse Cyclone DDS node(?) I'd say not communicating that over the network is probably a good thing from a security perspective too. Changes look fine to me, mostly removal of stub code, but it does make the implementation simpler in some places. Feel free to merge if you're confident the changes are correct (I did spot a failing test on the Visual Studio 15 target).

eboasson:
@k0ekk0ek I am pretty sure the failing test is not related to these changes: @dennis-adlink and myself have spotted a crash like this on different PRs and I suspect one of the more recent commits on master introduced it. I've been trying to figure out the cause, but so far without success.

PR #368, opened 2020-01-09
Liveliness QoS compatibility with OpenSplice for lease duration zero

dennis-adlink:
OpenSplice uses liveliness QoS lease duration zero as default, which results in the entity immediately getting not-alive in Cyclone. So we consider lease duration zero as infinite lease duration when communicating with an OpenSplice entity.

### DIFF
--- a/src/core/ddsi/src/q_plist.c
+++ b/src/core/ddsi/src/q_plist.c
@@ -1075,6 +1075,14 @@ static dds_return_t dvx_resource_limits (void * __restrict dst, const struct dd 
   return validate_resource_limits_qospolicy (dst);
 }
 
+static dds_return_t dvx_liveliness (void * __restrict dst, const struct dd * __restrict dd)
+{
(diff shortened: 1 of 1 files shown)


eboasson:
@dennis-adlink the way you have chosen to rewrite the lease_duration setting is entirely correct, but on consideration, I believe it is the wrong approach to the problem: basically, OpenSplice bugs should be dealt with in OpenSplice unless there is a pressing need to work around them in Cyclone.

As far as I have been able to determine, it really concerns only two writers used in an OpenSplice-specific protocol for dealing with transient data. Cyclone supports proxy writers with lease duration 0, so all the consequence is simply that those writers would considered not-alive by Cyclone, except for very brief periods when a sample arrives. Therefore, as far as I can tell, there is no problem in most cases, and probably no significant one even for anyone implementing that particular protocol.

In short: thanks, but no thanks 😄

Issue #74, opened 2018-12-12
dds_read_wl/dds_take_wl should allow to read/take all samples


kydos:
`dds_read_wl`/`dds_take_wl` allows to read/take samples while loaning the memory that holds the samples themselves. However these APIs currently assume that the `buf` passed il pre-allocated to provide, say **n** slots. Why don't we allow this call to actually allocate `buf` when what is passed is a null pointer. We would have to change the signature, tu be able to return the newly allocated buffer, but at least we would make it possible to read all samples in one shot w/o having to issue multiple read/take and while still minimising the memory allocation. 

@eboasson what are your thoughts?

A+

eboasson:
Absolutely.

I have to say I find the current read/take interface quite illogical even for preallocated use: I'd rather provide a buffer of some size and have it use the equivalent of a "placement new" to subdivide that buffer. So perhaps a little TLC for the read/take interface is needed in general.

martinbremmer:
I'm working on a C++ language binding (IsoCpp) on top of cyclonedds. This C++ binding will return all available samples with a single read/take. So, it would be very helpful if this is supported.

There is a un-documented kludgy 'feature' that works around the problem, so I can continue for now. But it would be nice if we could replace this kludge with a proper call.

PR #369, opened 2020-01-14
Added extra header file needed to work around an issue #74

ThijsSassen:
This is needed for reading samples with the C++API

### DIFF
--- a/src/core/ddsc/CMakeLists.txt
+++ b/src/core/ddsc/CMakeLists.txt
@@ -51,6 +51,7 @@ PREPEND(hdrs_public_ddsc "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include/dd
     ddsc/dds_public_qosdefs.h
     ddsc/dds_public_status.h
     ddsc/dds_rhc.h
+    ddsc/dds_internal_api.h
 )
 
 PREPEND(hdrs_private_ddsc "${CMAKE_CURRENT_LIST_DIR}/src"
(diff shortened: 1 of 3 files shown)


k0ekk0ek:
Hi @ThijsSassen! Thanks! Good to hear the C++ api is being worked on. Not sure whether it makes sense to use this function as I haven't seen the actual code that uses it, but the function is currently unused. Judging from what it's meant to do, I'd say it should just be defined in the public api. Defining it again, and thus twice, might cause problems in the future. The filename might also send the wrong message as the function would (possibly) also be required by other language bindings(?) I'd name it dds_reader.h and mark the function `@private` or something in the documentation. When/if you'd follow that route, which would be much appreciated, please keep an eye on the include guard. `DDS_CXX_H` is a little too generic I think. Most others seem to follow the filename. You may also want to define `DDS_READ_WITHOUT_LOCK` in the header as the documentation with the original function states it must be used with `dds_read` if `dds_reader_lock_samples` was previously called.

ThijsSassen:
@k0ekk0ek I processed your comments and added more information on why this function is needed, as soon as issue #74 is solved this extra header should be removed as its only needed for the work around to get unlimited samples in the c++ api.

eboasson:
@k0ekk0ek as far as I am concerned you can merge it.

k0ekk0ek:
Well, I was going to say the `@private` marker will not be picked up because it's in a `/* */` block instead of a `/** */` block. Only the latter is picked up by doxygen, which is the only reason the private mark is there. And having two definitions still kind of bugs me. The only reason the function is there is to be used outside of the current core and thus should be public. On the other hand, they're only minor nuisances, so here goes.

PR #370, opened 2020-01-15
Register local entities with security

MarcelJordense:
The main purpose of this pull request is to register and de-register the local entities with the security plugins and to maintain the security information related to the matching of local and remote entities. For that purpose also some of the corresponding functions related to remote entities are also implemented but this has still to be completed with another pull request.


MarcelJordense:
Part of this implementation is relying on code that is still to be implemented and therefor it is better to combine that first before considering this pull request. Therefor I will close this pull request.

PR #371, opened 2020-01-15
Make really short liveliness settings work

eboasson:
This PR was caused by pondering the wisdom of #368: is it wise to rewrite a lease duration of 0 received via discovery in some specific case? There are solid arguments against it:
* Endpoint matching takes lease durations into account, so rewriting it can result in two endpoints no longer matching;
* There is a valid argument in favour of supporting (and even optimising) lease_duration = 0, which is that it effectively eliminates the distinction between the "alive" and "not-alive-no-writers" states in an instance. That in turn affects eventual consistency-after-reconnect.

The problem intended to be addressed #368 being a very narrow one and the wisdom being doubtful, a wiser approach is to make sure a writer lease duration of 0 is handled correctly. Testing a automatic one is next to impossible (because it forces infinitely fast updates of the participants liveliness) so instead some checks in manual mode seemed to be in order.

This PR thus introduces such a test and some more-or-less related changes:

* A proxy-writer lease expiry during the creation of the proxy writer object could result in the proxy-writer not being marked "not alive" because the object might not yet be accessible via the GUID. Using the entity pointer that's embedded in the lease instead of looking it up via the GUID (retrieved using that pointer!) solves the problem.
* A manual-by-topic proxy writer did transition correctly to not-alive after expiry of the deadline, and did correctly transition back to alive on the receipt of a sample. However, it did not restart the deadline and so would then stay alive indefinitely.
* I found it sensible to use ``dds_wait_for_acks`` in the test code, so it now works for writers. The intent is that it can also be called on a publisher, where it would mean to wait on all the publishers writers in parallel, but that I haven't implemented yet. (Also, the testing is rather minimal ...)
* I suddenly realised that the warning ``selected interface is not multicast-capable: disabling multicast`` would be much improved if the name of the selected interface were to be included ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -215,7 +215,7 @@ script:
         ${SCAN_BUILD} cmake --build . --config ${BUILD_TYPE} --target install
         ;;
     esac
-  - CYCLONEDDS_URI='<CycloneDDS><Domain><Internal><EnableExpensiveChecks>all</EnableExpensiveChecks></Internal><Tracing><Verbosity>config</Verbosity><OutputFile>stderr</OutputFile></Tracing></Domain></CycloneDDS>' ctest -j 4 --output-on-failure -T test -E '^CUnit_ddsrt_random_default_random$' -C ${BUILD_TYPE}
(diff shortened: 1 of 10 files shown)


dennis-adlink:
@eboasson thanks for fixing these 2 bugs in the liveliness implementation. I left out the case of lease duration 0 in the original set of liveliness tests in PR #274 (#342), as I considered this to be not relevant, however PR #368 showed that was a wrong assumption. So its nice that we support this case now as well. 
Another (unlikely) case, using a very short lease duration for a participants lease, may also need a similar fix in `new_proxy_participant`, by moving `builtintopic_write` to just before `lease_register` (and it can be moved to outside the proxypp lock).

eboasson:
Thanks for catching that one @dennis-adlink !

PR #376, opened 2020-01-22
Default to a single receive thread on Windows

eboasson:
This works around a termination issue on Windows caused by the process
sometimes being unable to send a packet to itself to wake up a thread
stuck in a blocking read on a socket.


eboasson:
Thanks for taking a look @dennis-adlink 😄

PR #378, opened 2020-01-28
Liveliness local readers

dennis-adlink:
This PR adds support for liveliness QoS when using local readers. The implementation for (liveliness) expiration of writers used here is similar to that used with proxy writers, and it also supports the three liveliness kinds (1) automatic, which is trivial when using a local reader and writer, (2) manual-by-participant and (3) manual-by-topic.

In addition, these changes and fixes are included:
- Fixed a bug in heartbeat handling in the reader: for manual-by-participant writers the lease was not updated on reception of a heartbeat message with liveliness flag set. This is fixed and a test-case is added.
- Include the liveliness flag in a heartbeat message to the trace 
- Trace all lease renewals, including liveliness leases
- Replaced liveliness changed state 'twitch' by 2 subsequent calls to the status callback
- Added a test for liveliness duration 0 and 1ns (for both local and remote readers)


eboasson:
Thanks @dennis-adlink, there's really only one comment I have and that is that the need for a bit of refactoring to reduce the code duplication for local vs. proxy entities is really rather obvious. Which is not to say that I disagree with your choice here — indeed, I might well have made the same choice 😄 .

There have been a few maven problems in the builds and a switchover from travis-ci.org to travis-ci.com screwed up my ability to restart them. That should be over now, but I need to trigger a new build just to be sure — so I'll squash them by hand and do a force-push. But that's just administrative hassle after having decided to merge it.

PR #379, opened 2020-01-28
Update authentication plugin with respect to the remote identity handle

MarcelJordense:
The handle returned by the function validate_remote_identity of the authentication plugin should be based on the match of the remote participant and the local participant. For each matching pair of participants different key material is allocated which is identified by the remote identity handle.

### DIFF
--- a/src/security/builtin_plugins/authentication/src/authentication.c
+++ b/src/security/builtin_plugins/authentication/src/authentication.c
@@ -205,7 +205,6 @@ typedef struct HandshakeInfo {
 
 typedef struct dds_security_authentication_impl {
     dds_security_authentication base;
-    int id; //sample internal member
     ddsrt_mutex_t lock;
     struct ddsrt_hh *objectHash;
     struct ddsrt_hh *remoteGuidHash;
@@ -225,7 +224,7 @@ typedef struct {
 static int compare_relation (const void *va, const void *vb);
 
 const ddsrt_avl_treedef_t relations_treedef =
-  DDSRT_AVL_TREEDEF_INITIALIZER (offsetof (struct IdentityRelation, avlnode), offsetof (struct IdentityRelation, remoteIdentity), compare_relation, 0);
+    DDSRT_AVL_TREEDEF_INITIALIZER_INDKEY (offsetof (struct IdentityRelation, avlnode), offsetof (struct IdentityRelation, remoteIdentity), compare_relation, 0);
 
 static bool
 security_object_valid(
@@ -234,7 +233,7 @@ security_object_valid(
 {
     if (!obj) return false;
     if (obj->kind != kind) return false;
-    if ((ddsrt_address)obj->handle != (ddsrt_address)obj) return false;
+    if ((uintptr_t)obj->handle != (uintptr_t)obj) return false;
     return true;
 }
 
@@ -275,12 +274,12 @@ security_object_find(
 
 static int compare_relation (const void *va, const void *vb)
 {
-    const struct RemoteIdentityInfo * const *ha = va;
-    const struct RemoteIdentityInfo * const *hb = vb;
+    const struct RemoteIdentityInfo * ha = va;
+    const struct RemoteIdentityInfo * hb = vb;
 
-    if (*ha < *hb) {
+    if (IDENTITY_HANDLE(ha) < IDENTITY_HANDLE(hb)) {
         return -1;
-    } else if (*ha > *hb) {
+    } else if (IDENTITY_HANDLE(ha) > IDENTITY_HANDLE(hb)) {
         return 1;
     }
    return 0;
@@ -294,7 +293,7 @@ security_object_init(
     assert(obj);
 
     obj->kind = kind;
-    obj->handle = (int64_t)(ddsrt_address)obj;
+    obj->handle = (int64_t)(uintptr_t)obj;
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -199,14 +204,13 @@ typedef struct HandshakeInfo {
 
 typedef struct dds_security_authentication_impl {
     dds_security_authentication base;
+    int id; //sample internal member
```
eboasson: What is this?

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -216,21 +220,19 @@ typedef struct {
...
     if (obj->kind != kind) return false;
-    if (kind == SECURITY_OBJECT_KIND_IDENTITY_RELATION) {
-        IdentityRelation *relation = (IdentityRelation *)obj;
-        if (!relation->localIdentity || !relation->remoteIdentity || (ddsrt_address)obj->handle != (ddsrt_address)relation->localIdentity) {
-            return false;
-        }
-    } else if ((ddsrt_address)obj->handle != (ddsrt_address)obj) {
-        return false;
-    }
+    if ((ddsrt_address)obj->handle != (ddsrt_address)obj) return false;
```
eboasson: My apologies for overlooking the ``ddsrt_address`` in some previous review, but there really is no need for it to exist. This should just use ``uintptr_t``.

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -216,21 +220,19 @@ typedef struct {
 } validity_cb_info;
 
 
+static int compare_relation (const void *va, const void *vb);
+
+const ddsrt_avl_treedef_t relations_treedef =
+  DDSRT_AVL_TREEDEF_INITIALIZER (offsetof (struct IdentityRelation, avlnode), offsetof (struct IdentityRelation, remoteIdentity), compare_relation, 0);
```
eboasson: I had naively expected that identies would be compared on the content of them, but this, combined with the definition of ``compare_relation``, rather says identies are compared by the address at which they are stored. Is that really the intent? It could be ...

(And if it is not the intent, I would suggest using ``DDSRT_AVL_TREEDEF_INITIALIZER_INDKEY``.)

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -269,17 +271,28 @@ security_object_find(
     return (SecurityObject *) ddsrt_hh_lookup(hh, &template);;
 }
 
+static int compare_relation (const void *va, const void *vb)
+{
+    const struct RemoteIdentityInfo * const *ha = va;
+    const struct RemoteIdentityInfo * const *hb = vb;
+
+    if (*ha < *hb) {
```
eboasson: I think this should be comparing ``*ha`` and ``*hb`` cast to ``uintptr_t``, though it is possible I am mistaken in thinking that pointer comparison is among the operations causing undefined behaviour when the addresses come from unrelated objects.

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -1473,9 +1400,9 @@ validate_remote_identity(
         DDS_Security_set_token_nil(local_auth_request_token);
     }
 
-    *remote_identity_handle = IDENTITY_HANDLE(remoteIdent);;
+    *remote_identity_handle = IDENTITY_HANDLE(relation);;
```
eboasson: double semi-colon (and since there are some comments anyway, I might as well point it out)

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -1429,38 +1360,34 @@ validate_remote_identity(
...
            check if the remote identity token matches with the existing one
          */
-        if (!DDS_Security_DataHolder_equal(remoteIdent->remoteIdentityToken, remote_identity_token)) {
+        if (!DDS_Security_DataHolder_equal(rinfo->remoteIdentityToken, remote_identity_token)) {
             result = DDS_SECURITY_VALIDATION_FAILED;
             DDS_Security_Exception_set(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, (int)result,
                     "validate_remote_identity: remote_identity_token does not match with previously received one");
             goto err_inv_duplicate;
         }
+        relation = ddsrt_avl_lookup(&relations_treedef, &linfo->relations, rinfo);
```
eboasson: While it is fine to use ``ddsrt_avl_lookup`` followed by ``ddsrt_avl_insert`` if it isn't found, it does effectively require a second lookup inside the ``ddsrt_avl_insert``. There is a version suffixed with ``_ipath`` that allows you to avoid the second lookup by caching information in the lookup.

src/security/builtin_plugins/authentication/src/authentication.c
MarcelJordense: I do not know. The id is not used therefor I will remove the it.

src/security/builtin_plugins/authentication/src/authentication.c
MarcelJordense: No this was not the intend. Will be changed as suggested.

eboasson:
Thanks @MarcelJordense — the only problem is that the "Signed-off-by" footer is missing ... Well, and that my merging another PR now causes a conflict, but one that looks pretty trivial to me.

PR #380, opened 2020-01-28
Register and validate entities w.r.t. security

MarcelJordense:
Add the functions that register the local and remote entities and validate their permissions with the security plugins. This includes:
- validate local and remote identities with the authentication plugin
- check local and remote permissions with the access_control plugin
- register local and remote entities with the crypto plugin


MarcelJordense:
Ongoing implementations of the remaining security functionality based on this pull request has made this pull request obsolete. This pull request will therefor be closed.

PR #382, opened 2020-02-05
Allow a topic to have multiple definitions at once

eboasson:
This PR has a primary purpose and some fall-out. The PR looks massive (and it is ...) but it is not nearly as bad as the first impression because of two global renames. Looking at the commits individually should make it a little more manageable.

The primary purpose is to make it possible to have multiple topic entities in a single participant with topic name, but all (or some) with different underlying definitions. The main benefit of this is that it allows multiple languages (with different mappings of the types used in the topics) to coexist in a single application, using a single participant. It also makes the ros1-to-2 bridge work.

As a side-effect, and hopefully this pleases @martinbremmer, a consequence of this is that each call to ``dds_create_topic`` now creates a new topic object. This is something he tried to contribute in #304 that I rejected because it would allow the QoS of the different entities describing the same topic to diverge. This would, for example, result in built-in topics describing readers for the same topic but with different values for the ``TOPIC_DATA`` setting. In this PR, there is only one set of QoS settings per topic name, and topic entities refer to these.

The fall-out is:
* A change in the interface for custom topics: a few new functions are now required. On a sertopic, equality testing and hashing (for eliminating duplicate definitions); on a serdata, an alternative form to construct a sample from serialised form, taking an "I/O vector" instead of the very complicated fragchain. The need to define these can be determined using preprocessor macros. In the longer term, I am not really happy with duplicate interface (even though it is trivial to implement both) and would like to eliminate the ``nn_fragchain`` from this interface, but that's yet another project.
* A "clean up" of the tracing of parameter lists and QoS, now being able to convert them automatically to text, with added bonus of being able to trace the contents of discovery data just like the contents of ordinary samples can be traced.
* A few global renames, because some names started leaking out via the interface and I really don't want the old messed-up names polluting the interface any more than necessary — in particular: ``q_globals`` is now ``ddsi_domaingv``, ``nn_plist`` is now ``ddsi_plist`` and ``nn_xqos`` is now ``ddsi_xqos``. The alternative would've been to use, e.g., a void pointer, but that is painful for other reasons. Better to type it properly.
* The CDR serdes code moved from ``ddsc`` to ``ddsi``: the "default" topic implementation is relied upon by the lower layer, so this removes a (highly undesirable) depenency from the lower layer on the upper layer. Some elimination of unnecessary required a few more changes to leave the externally visible (and ``idlc``-generated) type descriptors unchanged. A nice side-effect of those is that it is now possible to create a topic with a dynamically allocated ``dds_topic_descriptor_t``.

With respect to the fall-out: I have tried to clean it up a bit, but the global renames require an amount of rework that is of such a magnitude that I think it is not really worth it. Those renames are necessary just to reduce the mess that some of the ancient names in the code base are.


eboasson:
Thank you for the comments, @dennis-adlink 😄 I have pushed two commits to cover them that I'll squash into the original commits before merging — provided they are themselves correct and no further issues pop up anywhere, of course.

dennis-adlink:
Lgtm, I couldn't find any other issues.

martinbremmer:
> As a side-effect, and hopefully this pleases @martinbremmer
I'm easily pleased and thanks :)

PR #385, opened 2020-02-06
Directory operations for reading trusted CA

stkimmer:
This PR depends on PR#377 Triggering of certificate and permission expires.


src/ddsrt/include/dds/ddsrt/filesystem.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * Precondition:
+ *   none
+ *
+ * Possible results:
+ * - return DDS_RETCODE_OK if directory 'name' is opened
+ * - DDS_RETCODE_ERROR if 'name' could not
+ *     be found or is not a directory.
+ */
+DDS_EXPORT dds_return_t ddsrt_opendir(const char *name, ddsrt_dirHandle *dir);
```
eboasson: Most of the code — and ddsrt decidedly also — uses lower-case-with-underscores for names and (most) typedefs and on ``_t``, and so ``ddsrt_dirHandle`` ought to be ``ddsrt_dir_handle_t``.

Naturally the same is true for other names, e.g., ``ddsrt_fileNormalize`` would have to be ``ddsrt_file_normalize``, etc.

src/ddsrt/src/filesystem/posix/filesystem.c
```diff
@@ -0,0 +1,121 @@
...
+#include "dds/ddsrt/filesystem.h"
+#include "dds/ddsrt/string.h"
+#include "dds/ddsrt/heap.h"
+
+dds_return_t ddsrt_opendir(const char *name, ddsrt_dirHandle *dir)
+{
+    dds_return_t result;
+    DIR *d;
+
+    result = DDS_RETCODE_ERROR;
```
eboasson: It shouldn't come as a big surprise that I comment on the structure here: this could be almost a one-liner ... I wonder why not something like
```C
dds_return_t ddsrt_opendir(const char *name, ddsrt_dirHandle *dir)
{
  *dir = opendir (name);
  return *dir ? DDS_RETCODE_OK : DDS_RETCODE_OK;
}
```
(although I admit that it is ever so slightly different in that it overwrites ``dir`` on failure). Anyway, I would appreciate it if you went over this cleaning it up a bit.

And as ``ddsrt_dirHandle`` is typedef to a ``DIR *``, the type casts between the two are superfluous (and I would argue, therefore dangerous).

src/ddsrt/src/filesystem/posix/filesystem.c
```diff
@@ -0,0 +1,121 @@
...
+char * ddsrt_fileNormalize(const char *filepath)
+{
+    char *norm;
+    const char *fpPtr;
+    char *normPtr;
+
+    norm = NULL;
+    if ((filepath != NULL) && (*filepath != '\0')) {
+        norm = ddsrt_malloc(strlen(filepath) + 1);
+        /* replace any / or \ by OS_FILESEPCHAR */
```
eboasson: The world has sadly moved on beyond Acorn ADFS, Classic MacOS, VMS and some others with entirely different ideas on the notation of path names. Don't worry: I just had to get it off my chest, it is not a request to actually change this.

src/ddsrt/include/dds/ddsrt/filesystem.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * Possible results:
+ * - return DDS_RETCODE_OK if stat is successful
+ * - return DDS_RETCODE_ERROR if stat fails.
+ */
+DDS_EXPORT dds_return_t ddsrt_stat(const char *path, struct ddsrt_stat *buf);
+
+/** \brief Transforms the given filepath into a platform specific filepath.
+ *
+ * This translation function will replace any platform file seperator into
```
eboasson: This is not what the function does (because of the way it handles quotes)

src/ddsrt/src/filesystem/windows/filesystem.c
```diff
@@ -0,0 +1,122 @@
...
+    WIN32_FIND_DATA FileData;
+    HANDLE hList;
+
+    result = DDS_RETCODE_ERROR;
+    if (dir) {
+        snprintf(szDir, DDSRT_PATH_MAX, "%s\\*", name);
+        hList = FindFirstFile(szDir, &FileData);
+
+        if (hList != INVALID_HANDLE_VALUE) {
+            *dir = (ddsrt_dirHandle)hList;
```
eboasson: This cast (and all like it) are also superfluous and had better be removed. (Or, alternatively, one could wrap turn ddsrt_dirHandle into a struct wrapping the underlying object, &c. Your choice.)

src/ddsrt/include/dds/ddsrt/filesystem.h
```diff
@@ -0,0 +1,122 @@
...
+ * Possible results:
+ * - returns normalized filepath conform current platform
+ * - return NULL if out of memory.
+ */
+DDS_EXPORT char* ddsrt_fileNormalize(const char *filepath);
+
+/** \brief Get file seperator
+ *
+ * Possible Results:
+ * - "<file-seperator-character>"
```
eboasson: If it is guaranteed to be a single character, a return type of ``char`` would be better; considering that the one call site assumes it is a string, it'd probably be better to document it as returning a string that is possibly longer than 1.

src/security/builtin_plugins/authentication/src/auth_utils.c
```diff
@@ -1190,40 +1192,35 @@ get_trusted_ca_list ( const char* trusted_ca_dir,
...
-            ddsrt_strcat(full_file_path, os_fileSep());
-            ddsrt_strcat(full_file_path, d_entry.d_name);
-
-            if (os_stat (full_file_path, &status) == os_resultSuccess) { /* accessable */
+    if (r == DDS_RETCODE_OK && ca_buffer_array_size < MAX_TRUSTED_CA) { /* accessable */
+        r = ddsrt_readdir(d_descr, &d_entry);
+        while (r == DDS_RETCODE_OK) {
+            full_file_path_size = strlen(trusted_ca_dir) + strlen(ddsrt_fileSep()) + strlen(d_entry.d_name) + strlen(ddsrt_fileSep()) + 1 ;
+            full_file_path = (char*) ddsrt_malloc(full_file_path_size);
+            ddsrt_strlcpy(full_file_path, trusted_ca_dir, full_file_path_size);
```
eboasson: ``ddsrt_asprintf`` would be a better choice here, I'd say.

eboasson:
@stkimmer, thank you for making those changes, all of them work for me. I don't see any changes to the test code that generated that ``Test_Permissions_listener.p7s`` file, but perhaps you are still looking at why it gets generated there?

In any case, I can't push to your branch, so for me to merge this pull request you will really need to rebase on the current security branch — which, as I mentioned before, really means dropping a commit.

stkimmer:
@eboasson, I removed the the Test_Permissions_listener.p7s file from the repo. It needs to be generated from the test (along with Example_Alice_Cert.pem and Example_Bob_Cert.pem).

eboasson:
> @eboasson, I removed the the Test_Permissions_listener.p7s file from the repo. It needs to be generated from the test (along with Example_Alice_Cert.pem and Example_Bob_Cert.pem).

I did see that you removed it, however, building/testing Cyclone shouldn't generate any files outside the build directory. Writing a previously non-existent file in the source tree is not so much dangerous as it is unexpected and risks accidentally adding it to the repository (which is the way it was introduced in the first place).

So I really want to see the test changed to write that file in a different location.

stkimmer:
> I did see that you removed it, however, building/testing Cyclone shouldn't generate any files outside the build directory. Writing a previously non-existent file in the source tree is not so much dangerous as it is unexpected and risks accidentally adding it to the repository (which is the way it was introduced in the first place).
> 
> So I really want to see the test changed to write that file in a different location.

Very good point. The files that the tests generate are now in the build directory.

eboasson:
Thanks @stkimmer!

PR #389, opened 2020-02-10
Unable to run CycloneDDS because of locale issue in Windows.

evshary:
The problem is about CycloneDDS running on Windows.
`ddsrt_asprintf` will not malloc memory for `ifa->name` while `iface->FriendlyName` is not English.
(In my case, `iface->FriendlyName` is traditional Chinese.)
This will cause CycloneDDS unable to work because `ifa->name == NULL`.
Therefore, I allocate the memory first and then call `wcstombs` to transform wide char into char.

### DIFF
--- a/src/ddsrt/src/ifaddrs/windows/ifaddrs.c
+++ b/src/ddsrt/src/ifaddrs/windows/ifaddrs.c
@@ -189,10 +189,14 @@ copyaddr(
   if ((ifa = ddsrt_calloc_s(1, sizeof(*ifa))) == NULL) {
     err = DDS_RETCODE_OUT_OF_RESOURCES;
   } else {
+    size_t namelen = wcslen(iface->FriendlyName);
     ifa->flags = getflags(iface);
     ifa->type = guess_iftype(iface);
     ifa->addr = ddsrt_memdup(sa, sz);
-    (void)ddsrt_asprintf(&ifa->name, "%wS", iface->FriendlyName);
+    ifa->name = ddsrt_malloc(namelen + 1);
+    if (ifa->name != NULL) {
+      wcstombs(ifa->name, iface->FriendlyName, namelen);
+    }
     if (ifa->addr == NULL || ifa->name == NULL) {
       err = DDS_RETCODE_OUT_OF_RESOURCES;
     } else if (ifa->addr->sa_family == AF_INET6) {
--- a/src/ddsrt/src/ifaddrs/windows/ifaddrs.c
+++ b/src/ddsrt/src/ifaddrs/windows/ifaddrs.c
@@ -9,8 +9,10 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


src/ddsrt/src/ifaddrs/windows/ifaddrs.c
```diff
@@ -189,10 +189,14 @@ copyaddr(
...
     err = DDS_RETCODE_OUT_OF_RESOURCES;
   } else {
+    size_t namelen = wcslen(iface->FriendlyName);
     ifa->flags = getflags(iface);
     ifa->type = guess_iftype(iface);
     ifa->addr = ddsrt_memdup(sa, sz);
-    (void)ddsrt_asprintf(&ifa->name, "%wS", iface->FriendlyName);
+    ifa->name = ddsrt_malloc(namelen + 1);
+    if (ifa->name != NULL) {
+      wcstombs(ifa->name, iface->FriendlyName, namelen);
```
k0ekk0ek: `wcstombs` is locale dependent itself. Most of the time it'll probably be alright, but I think it would be better to use `_mbstowcs_s_l` and force the encoding to be UTF-8. `_mbstowcs_s_l` can also be used to determine the number of bytes required in the destination judging by the [documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mbstowcs-s-mbstowcs-s-l).

evshary:
Hi, @k0ekk0ek Thanks for your suggestions! They're really helpful.
The PR is updated. I get the user local environment and then use `_wcstombs_s_l` to get length and transform into char.
BTW, it' fine with me if you find some mistakes and want to modify directly.

k0ekk0ek:
Thanks @evshary! I haven't tested your branch. It'll probably work, but it's still somewhat dependent on the locale. It probably would've been better to use something like English+UTF-8 for a locale(?). The important thing is that the interface name is encoded in UTF-8, so instead of using locale dependent conversion I opted for `WideCharToMultiByte` icw `CP_UTF8`, which hopefully works. I based a patch on your comments here: https://github.com/k0ekk0ek/cyclonedds/tree/bugfix_iface_name. Can you give it a try?

evshary:
Thanks @k0ekk0ek!  Actually, I'm not really familiar with the encoding, but maybe using UTF-8 is a better idea if `ifa->name` is for internal use and not display to users.
Your patch works well in my environment. It's fine to use your patch.

k0ekk0ek:
@evshary, I did a rebase and opened PR #394 for another committer to review. You should be able to use master relatively soonish. Thanks again for reporting!

Basically the locale controls things like number formatting, time conversion etc. The real problem (I hope :slightly_smiling_face:) was the encoding used. Your system uses wide characters (UTF-16) to encode Unicode code points while Eclipse Cyclone DDS expects a regular character string. So the locale doesn't actually matter that much, the interface even has the same name, but that name is now encoded in [UTF-8](https://nl.wikipedia.org/wiki/UTF-8) (US-ASCII is a subset) instead of UTF-16.

PR #392, opened 2020-02-12
Retention of samples in keep-last WHC + running ddsperf on travis

eboasson:
This PR fixes an accidental change in the retention of samples in a keep-last WHC introduced in commit 231cb8c. It also adds running a simple combined throughput and roundtrip sanity check using ``ddsperf`` on CI builds. This sanity check runs for 10s and fails if:
* the two processes don't discover each other in time;
* RSS grows by more than a set amount (10% + 0.5MB from initial to final, determined by experimentation) — better tests are needed, but at the data rates involved this should catch some real memory growth;
* sample rate is less than 30kS/s (because timing on Travis is really problematic, I've seen its once-per-seconds time stamps jump from for 2s straight to 9s ...);
* number of round trips is less than 3k.

Trying the "MonitorPort" setting I chanced upon two small independent issues, those are fixes as well.

### DIFF
--- a/src/core/ddsc/tests/whc.c
+++ b/src/core/ddsc/tests/whc.c
@@ -110,26 +110,40 @@ static dds_entity_t create_and_sync_reader(dds_entity_t subscriber, dds_entity_t
   return reader;
 }
 
-static void check_whc_state(dds_entity_t writer, seqno_t exp_min, seqno_t exp_max)
+static void get_writer_whc_state (dds_entity_t writer, struct whc_state *whcst)
 {
   struct dds_entity *wr_entity;
   struct writer *wr;
-  struct whc_state whcst;
   CU_ASSERT_EQUAL_FATAL(dds_entity_pin(writer, &wr_entity), 0);
   thread_state_awake(lookup_thread_state(), &wr_entity->m_domain->gv);
   wr = entidx_lookup_writer_guid(wr_entity->m_domain->gv.entity_index, &wr_entity->m_guid);
   CU_ASSERT_FATAL(wr != NULL);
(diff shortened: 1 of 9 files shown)


src/core/ddsc/tests/whc.c
```diff
@@ -191,6 +211,23 @@ static void test_whc_end_state(dds_durability_kind_t d, dds_reliability_kind_t r
       ret = dds_write (writer, &sample_keyless);
       CU_ASSERT_FATAL (ret == DDS_RETCODE_OK);
     }
+
+    /* if history is truly keep all, there may never be more data present than the max of the
+       history depth(s) */
```
dennis-adlink: I think this comment should be 'if history is truly keep _last_, ...'.

src/core/ddsc/tests/whc.c
```diff
@@ -110,6 +110,27 @@ static dds_entity_t create_and_sync_reader(dds_entity_t subscriber, dds_entity_t
   return reader;
 }
 
+static void check_intermediate_whc_state(dds_entity_t writer, seqno_t exp_min, seqno_t exp_max)
```
dennis-adlink: Nit-picking: maybe parameterize `check_whc_state` so that there is less code duplication?

src/core/ddsc/tests/whc.c
eboasson: Oops - fixing

src/core/ddsc/tests/whc.c
eboasson: Good point

PR #397, opened 2020-02-13
Fixed build error on FreeRTOS

ThijsSassen:
recent change in ddsperf causes a build error on freeRTOS.


k0ekk0ek:
It's a syntax error on more platforms if they don't support rusage and/or thread list. Waiting for Travis to says it's ok and then I'll merge. Thanks @ThijsSassen!

PR #399, opened 2020-02-18
Fix some typos.

evshary:
As title, just some typos.


eboasson:
Hi @evshary all the builds failed, as I am sure you have noticed. I suspect you simply didn't know about the little bit of nastiness with changing the options and/or descriptions. I'll try to explain :)

The settings with their descriptions in ``q_config.c`` are processed by a perl script that [generates a markdown file with the options](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/manual/options.md) and a ["Relax NG Compact form" specification](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/etc/cyclonedds.rnc) for editor support. There is also an [XSD file](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/etc/cyclonedds.xsd) but I haven't (yet) done the work to make that perl script generate the XSD directly as well. Instead, I've used an open source converter tool (trang), but that final step doesn't run on Travis (see https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/CMakeLists.txt).

While these files are in principle generated, they are still committed in the source tree because I think it is useful to have them available for reference at any time. And so the build process compares the ones it can generate with the ones in the repository and fails if there is a discrepancy.

It is quite silly to take the C source and transform it in this manner, rather than using a decent format for the option specification and generate the C code for the decoder from that. On the other hand it works well enough in practice — and so it just doesn't get priority.

If you'd update those options.md, cyclonedds.rnc and cyclonedds.xsd files by copying the generated version over the one in the source tree and adding those to the commit, it'll be ok again. I can do it, but I figured it's better to at least try to explain why and how. If you'd rather not do this right now, let me know.

evshary:
Oops, you're right. I didn't know the relationship between the document and source code.
I'll try to correct it later. Thank you for your kind explanation.

evshary:
Hi @eboasson , it seems OK now.
However, I'm not sure why cdds.md and options.md both exist at the same time.
They are almost the same file.
While I used trang, it only generated options.md.

eboasson:
@evshary Yes it is, and thank you for mentioning cdds.md, that one shouldn't exist and we should delete. Would you like the honour of doing that?

evshary:
@eboasson Sure.

eboasson:
Thank you!

PR #403, opened 2020-02-19
Add config options to decrease ACK rates

trickapm:
A number of config options have been added for tweaking the protocol between reliable readers/writers to decrease ACK rates.
- Disable ParticipantMessageData readers (which trigger creation of matching writers) to disable periodical liveliness messages and associated ACK traffic. Liveliness is still derived indirectly from other traffic.
- Option for ACK suppression that makes reliable writers send heartbeats with ACK requests only if neccesary eg. to clean up writer-history cache (suppressing ACKs when 'all is well').
- Option for (randomized) ACK delay to prevent all relevant readers from responding to ACK requests at the same time causing spikes on the network.

All options are by default turned off so unless explicitly configured the default behaviour is unchanged.


k0ekk0ek:
@eboasson, is this something we're interested in? If so, @trickapm, maybe you can update after #487 has been merged? Sorry nobody looked into this yet.

PR #404, opened 2020-02-20
Not receiving multicast data on Windows

dennis-adlink:
Fixed issue that Cyclone does not receive multicast data on Windows when the destination cache of the network stack is in a certain state. The issue is fixed by binding unicast sockets (incoming unicast and all outgoing traffic) to the address of the interface instead of `inaddr_any` (0.0.0.0).
Set the new configuration option `internal/BindUnicastToInterfaceAddr` to false to get the old behavior.
@eboasson, thanks for your help in analysing and solving this issue!


eboasson:
@dennis-adlink thanks for simply adding the final fix rather than force-pushing stuff, but I'd prefer having two commits here, not three: one with the change in the configuration naming (thanks for that, by the way, it was long overdue!) and one with the complete fix.

Do you mind combining the 1st and 3rd commits (so that there are two commits, one for the multicast issue, and one for the rename)? Thanks!

dennis-adlink:
Sure, no problem, I've force-pushed the squashed commit for the multicast fix

eboasson:
Thanks!

PR #410, opened 2020-02-26
WIP: Switch atomic operations to __atomic functions

### LINKED ISSUE #409: Atomic functions don't look atomic
`ddsrt_atomic...` functions don't look atomic.
1. Their arguments are marked volatile which doesn't seem to serve a purpose
2. Loading and storing is definitely *not atomic*.
3. The gcc implementation uses the "legacy" `__sync_...` functions which have been superceded by `__atomic_...` functions since gcc 4.7.4

https://github.com/eclipse-cyclonedds/cyclonedds/blob/6f33dc0e6d48110fa96fce001e4a6477cecc4f9c/src/ddsrt/include/dds/ddsrt/atomics/gcc.h#L29-L33

see https://stackoverflow.com/questions/25647710/atomic-get-load-and-set-store-with-older-gcc-sync-builtins for discussion of what synchronization looks like for plain loads and stores (which we don't do)

rotu:
1. Make atomic load and store operations actually atomic
2. Switch definitions to use `__atomic_...` functions instead of legacy `__sync_...` functions.
3. Clean up function signatures.
 
Fix #409

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/atomics.h
+++ b/src/ddsrt/include/dds/ddsrt/atomics.h
@@ -65,23 +65,23 @@ typedef ddsrt_atomic_uintptr_t ddsrt_atomic_voidp_t;
    the form of a different implementation where it is used, or as an emulation using a mutex in
    ddsrt.  It seems that the places where they'd be used end up adding a mutex, so an emulation in
    ddsrt while being able to check whether it is supported by hardware is a sensible approach.  */
-DDS_EXPORT uint64_t ddsrt_atomic_ld64 (const volatile ddsrt_atomic_uint64_t *x);
-DDS_EXPORT void ddsrt_atomic_st64 (volatile ddsrt_atomic_uint64_t *x, uint64_t v);
-DDS_EXPORT void ddsrt_atomic_inc64 (volatile ddsrt_atomic_uint64_t *x);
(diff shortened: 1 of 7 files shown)


k0ekk0ek:
Hi @rotu. Thanks so much for you efforts. I'm reading this and I (may) have some problems with it. My memory is a bit rusty on atomic access, compiler ordering, etc, but just removing the `volatile` keyword everywhere isn't the best idea. In my recollection plain loads and stores of word sizes is atomic if the address is aligned. That's very likely the reason why there's no __sync_fetch builtin(? :sweat_smile: ) The compiler doesn't optimize with regards to ordering if the `volatile` keyword is used. The builtins combine the two and that seems to be exactly what the `ddsrt_atomic_*` functions do as well(?). Removing `volatile`, combined with the functions being `inline` seems very likely to break compiler ordering. When using the GCC builtins, that is solved for platforms using GCC, but what about other platforms? Apart from that, there might be platforms/ports out there that depend on older versions of GCC that do not have the `__atomic` family of builtins. @eboasson's Solaris x.x port comes to mind... As always, when changing things with regards to locking, atomic access, etc taking great care that one's absolutely right is the wise thing to do because these things being incorrect can lead to very nasty problems.

With that in mind, can you attest that the current implementation is incorrect and that yours is not? Apart from that, is there a very strong reason to swap out the `__sync_` family of functions for the `__atomic_` family of functions?

PR #411, opened 2020-02-27
Avoid race causing thread-state pointer aliasing

eboasson:
The thread_states array resets the "state" to ZERO on thread termination
to indicate that the slot was unused, but it leaves the thread id
unchanged because some platforms don't have a defined value that will
never be used as a thread id.  A consequence is that a newly created
thread may result in multiple slots containing their own thread id, but
generally there will only be one that is not in state ZERO.

However, the code for create_thread used to set the state to ALIVE prior
to creating the thread, and so if the events get scheduled like:

1. thread A: X.state = ALIVE
2. create new thread B, storing tid in X.tid
3. thread A: Y.state = ALIVE
4. new thread B: lookup self (and cache pointer)
5. create new thread C, storing tid in Y.tid
6. new thread C: lookup self (and cache pointer)

Then B will observe two slots in the ALIVE state, with X.tid certain to
match and Y.tid undefined (and hence possibly matching).  It may
therefore pick Y.  C will (in this schedule) of course always choose Y.
They cache the pointer and never look at X and Y again, except for
updating their virtual clocks.

These virtual clocks are updated non-atomically (by design it is private
to the thread) and so if both B & C use Y they can end up racing each
other in updating the virtual clock and cause the nesting level of the
"awake" state controlling garbage collection to get stuck (or wrap
around, or do other horrible things).  The consequence can be anything,
from a somewhat benign variant where GC effectively stops and some
operations (deleting readers and writers and shutting down) block
forever, to use-after-free and the undefined behaviour that implies.

This commit avoids looking up the slot in the newly created threads,
instead passing the correct address in the argument.  It also adds an
intermediate state INIT that serves to reserve the slot until the new
thread is actually running.  It does make the look-up safe (if one were
to do it), and as it is essentially free and gives more insight in the
state of the system when viewed from a debugger, it appears a useful
addition.

(https://github.com/ros2/rmw_cyclonedds/issues/104)

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_thread.h
+++ b/src/core/ddsi/include/dds/ddsi/q_thread.h
@@ -50,6 +50,8 @@ typedef int32_t svtime_t; /* signed version */
 
 enum thread_state {
   THREAD_STATE_ZERO, /* known to be dead */
+  THREAD_STATE_STOPPED, /* internal thread, stopped-but-not-reaped */
+  THREAD_STATE_INIT, /* internal thread, initializing */
   THREAD_STATE_LAZILY_CREATED, /* lazily created in response because an application used it. Reclaimed if the thread terminates, but not considered an error if all of Cyclone is shutdown while this thread hasn't terminated yet */
   THREAD_STATE_ALIVE /* known to be alive - for Cyclone internal threads */
 };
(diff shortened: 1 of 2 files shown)


rotu:
I just ran this against test_tf2 a bunch of times and it seems to be a pretty solid fix.

PR #402, opened 2020-02-19
Add security handshake and encryption

MarcelJordense:
The following security functionality is added:
- authentication handshake
- encoding and decoding
The added functionality is tested with a single publisher and subscriber using the following security configuration:
allow_unauthenticated_participants =false
enable_join_access_control = true
discovery_protection_kind = SIGN_WITH_ORIGIN_AUTHENTICATION
liveliness_protection_kind = ENCRYPT
rtps_protection_kind = ENCRYPT_WITH_ORIGIN_AUTHENTICATION
enable_liveliness_protection = true
enable_discovery_protection = true
enable_read_access_control = false
enable_write_access_control = false
metadata_protection_kind = ENCRYPT_WITH_ORIGIN_AUTHENTICATION
data_protection_kind =ENCRYPT

PR #412, opened 2020-02-28
Add and refactor security core tests

dennis-adlink:
In this PR I've refactored tests for security core: introduced generic wrappers for security plugins and updated the existing plugin loading tests to use these instead of specific wrappers per test. I've also added tests for securing communication (all combinations of protection kinds in governance configuration) and handshake tests (including a test that checks there is no communication when using different identity CAs for two nodes). Note that some of these security core tests will fail until PR #402 is in. 

In the two other commits I've included some changes related to the tests:
- the way definitions in CMake files are set, so that all definitions are consistent for all executables and libraries under /src. 
- expand_vars so that a custom lookup function can be used, which makes it usable for replacing tags by values other than environment variables


dennis-adlink:
@eboasson thanks for your review comments! I've addressed the issues in a new commit. I left the buffer matching as-is for now, because as you already mentioned, the size difference for the buffers that are compared is not significant so this won't really harm in current tests. The buffer checking approach needs more rework/improvement anyway, which I will do in a separate PR (then I'll replace the matching with `memmem` on linux and mac and a Boyer-Moore implementation on windows)

PR #414, opened 2020-02-28
Fix undefined behavior when hash function given null pointer

rotu:
```
[test_subscriber-12] /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/ddsrt/src/mh3.c:28:53: runtime error: applying zero offset to null pointer
[test_subscriber-12] SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /opt/ros/master/src/eclipse-cyclonedds/cyclonedds/src/ddsrt/src/mh3.c:28:53 in
```


eboasson:
Merged by hand (had to drop the "don't pass null to memcmp" commit first).

PR #420, opened 2020-03-03
deaf/mute mode and UDPv6 support in decode-trace

eboasson:
This PR originally simply aimed to add UDPv6 support to the ``decode-trace`` script. It also improves the heuristics of what might be called an "asymmetric disconnection", a situation where in a connected pair of nodes, A and B, a situation arises where A temporarily considers B's lease expired, but not vice-versa.

The "deaf/mute" mode part refers to the enabling of some test code to allow simulating the process to be either deaf, mute or both, by throwing away packets right where they arrive from the network and/or where they are handed off to the network. I've simply added a function to control this (``dds_domain_set_deafmute``).

Comments welcome, as always.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -3358,6 +3358,40 @@ DDS_EXPORT dds_return_t
 dds_assert_liveliness (
   dds_entity_t entity);
 
+/**
+ * @brief This operation allows making the domain's network stack
+ * temporarily deaf and/or mute. It is a support function for testing and,
+ * other special uses and is subject to change.
+ *
+ * @param[in] entity  A domain entity or an entity bound to a domain, such
+ *                    as a participant, reader or writer.
+ * @param[in] deaf    Whether to network stack should pretend to be deaf and
+ *                    ignore any incoming packets.
+ * @param[in] mute    Whether to network stack should pretend to be mute and
+ *                    discard any outgoing packets where it normally would.
+ *                    pass them to the operating system kernel for transmission.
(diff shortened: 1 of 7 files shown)


TheFixer:
@eboasson You asked me to look in a bit more detail to the deaf/mute implementation, and I finally had some time to do that. The implementation looks OK to me. The ability to deaf/mute a node is a handy feature for testing the behaviour of a node in the context of (asymmetric) disconnects/reconnects. I do have one remark related to this feature (although I don't think it should be part of this pull request). 

The current deaf/mute implementation allows a node to make itself deaf or mute for others. This basically isolates a node from every other node. As a possible extension I believe it would be handy to be able to isolate other nodes from yourself, e.g., in case these other nodes show malicious behaviour. In that case I can image that you want to blacklist such malicious behaving nodes, so that they cannot harm your system. I think at some point in time the ability to blacklist malicious nodes will be handy, and its implementation may use similar techniques as deaf/mute. But as said before, this sounds like a different feature, and its urgency is yet to be determined.

eboasson:
Thanks @TheFixer for taking the time to have a look at it. I agree with you that the ability to ignore certain nodes is valuable (for a variety of reasons) and the most obvious path to doing so is to implement the "ignore participant" operation that has been part of the DDS specification from the very beginning.

Dealing with malicious nodes is a much more complicated affair. DDS Security provides a reasonable set of mechanisms for trying to keep them out, but if one ever were to make it in, the problem is no longer merely stopping communication with that node but determining the damage already done by it. It is an interesting but complicated subject – and definitely not for this PR!

PR #426, opened 2020-03-06
WIP: Code style - convert some enum-like #defines to actual enums

rotu:


rotu:
oops. looks like I may have changed the size of some struct in a platform-specific way.

eboasson:
> oops. looks like I may have changed the size of some struct in a platform-specific way. 

Hi @rotu, that could well be, but I'm pretty sure that addressing the issues will be rather straightforward. All it takes is changing the structs to use the correctly-sized integer and casting to/from the enum.

The problem here is that these structs are used in interpreting data coming from the network, and therefore there is no way of knowing whether an input value really converts to a defined enum symbol or not (what Ada's ``Valid`` attribute does), and so one has to be careful in handling the enums. But that is a manageable issue.

eboasson:
The old considerations still stand: it is a good idea to use `enum`s where possible, but for some their use on the wire means it doesn't seem like a good idea in all cases. Besides which, so much has changed since this PR was originally created that I'd rather re-do the work then try updating this PR.

PR #428, opened 2020-03-09
Don't link winsock1

rotu:
wsock32.lib is only needed for the legacy version of Winsock and is not needed with Winsock2 (the current version).
This appears to be a root cause of the multicast issue on Win10 and may allow us to reverse #404

### DIFF
--- a/src/ddsrt/CMakeLists.txt
+++ b/src/ddsrt/CMakeLists.txt
@@ -236,7 +236,7 @@ if(NOT WITH_FREERTOS)
 endif()
 
 if(WIN32)
-  target_link_libraries(ddsrt INTERFACE wsock32 ws2_32 iphlpapi bcrypt)
+  target_link_libraries(ddsrt INTERFACE ws2_32 iphlpapi bcrypt)
 elseif(UNIX)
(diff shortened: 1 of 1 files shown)


rotu:
@eboasson here's the build with #404 backed out and this PR added: https://ci.ros2.org/job/nightly_win_extra_rmw_rel/693/

eboasson:
Thanks @rotu! I haven't been able to find out when or why it got added — it's part of the initial contribution and I presume someone simply made a mistake.

It's too late to simply reverse #404 because #427 effectively reversed it and solved "The Problem" by adding an extra socket (something that seems to match how most other DDS vendors do it, incidentally). One extra socket is not too terrible a burden but it does complicate firewalling machines and eliminating it in the future is definitely desirable.

As there is plenty of cleaning up to be done in the creating of sockets in Cyclone, sorting out this detail is probably most practically done in that context.

rotu:
I mean there’s no need for #427 either, with this fix, afaict.

eboasson:
@rotu I agree, but it was in already ... removing the extra socket will not be complicated.

joespeed:
@eboasson @rotu did the temporary workaround's extra socket get removed?

PR #429, opened 2020-03-09
Install security plugins

SidFaber:
Security plugins are built but not installed. This PR adds target to CMakeLists.txt for three security plugins.


eboasson:
Thanks @SidFaber, looks good to me. However, I can't merge it yet because it fails the Eclipse ECA check. All commits in Eclipse projects must have a "Signed-off-by:" in the footer matching the commit author, and there has to be a valid Eclipse Contributor Agreement on file for the author. Please see https://www.eclipse.org/legal/ecafaq.php

PR #431, opened 2020-03-10
A bit of spring cleaning

eboasson:
Remove semi-duplication of functions for getting the current time, i.e., ``now`` vs ``dds_time``, ``now_mt`` vs ``ddsrt_time_monotonic``, &c. The return types of these also differed (``mtime_t`` vs ``dds_time_t``, &c.) and this is fixed by renaming ``mtime_t`` to ``ddsrt_mtime_t`` and changing the return type of ``ddsrt_time_monotonic`` to the latter. The other clocks have been adjusted as well. This obviously affects quite a few files ...

There were also a lot of references to PrismTech in constants (and some types) for vendor-specific extensions, but the name "PrismTech" no longer exists and these have been renamed to Adlink, as that's the current name. A few unused types and fields related to these existed in the code (as opposed to defines providing names for various numeric constants) and have been removed.

The third bit of spring cleaning is moving some code from a generic "network support" file (q_nwif.c) to the transports and deleting a no-longer used function for hashing an IPv6 address. That means the file only deals with choosing an interface now. (Which, too, should move to the transport code, but I can't do everything at once.)


TheFixer:
I decided to do a bit of reviewing of part of this large commit. I did not focus on renamings since they are checked by the compiler. I did focus on the changes related to times.

In src/ddsrt/include/dds/ddsrt/time.h
- Can ddsrt_time_add_duration can handle corner cases such as INVALID + INFINITE? I am aware that the current code doesn't use these cases, but that does not prevent that somebody in future (accidentally) will do this. If such cases can potentially occur  I would suggest to either explicitly indicate that such corner cases are excluded by making clear that this is a precondition, or handle these cases.
- I am a bit surprises to see DDS_NEVER being used as reltime. I was under the impression that it is used for time, not durations

eboasson:
@TheFixer thanks for diving into this rather large and monotonous PR. Regarding your questions:

> Can ddsrt_time_add_duration can handle corner cases such as INVALID + INFINITE? I am aware that the current code doesn't use these cases, but that does not prevent that somebody in future (accidentally) will do this. If such cases can potentially occur I would suggest to either explicitly indicate that such corner cases are excluded by making clear that this is a precondition, or handle these cases.

I'd say this is covered by asserting that the time and the interval are both non-negative (which eliminates the "invalid" case because that is mapped to -2^63) and by representing infinity/"The Future" as the maximum representable number (2^63-1) so that ``reltime >= NEVER - abstime`` test is a correct way of determining whether the addition would overflow.

> I am a bit surprises to see DDS_NEVER being used as reltime. I was under the impression that it is used for time, not durations

Your impression is correct. Technically here it is used as such (NEVER - abstime is a reltime), but the equation could just as easily have been rewritten to use (NEVER - reltime). The two are simple equivalent in this particular instance.

PR #432, opened 2020-03-10
Install security plugins

SidFaber:
Security plugins are built but not installed.  This PR adds targets to CMakeLists.txt for three security plugins.

Signed-off-by: Sid Faber <sid.faber@canonical.com>


eboasson:
Thanks @SidFaber !

PR #436, opened 2020-03-13
Update encode decode

MarcelJordense:
Set volatile secure reader initially out_of_sync
Correct handling of identity certificate with EC key


eboasson:
Thanks @MarcelJordense! I've decided to create a PR (#437) based on this one that is just a rebased version of this and merged that one. Closing this one because it is no longer relevant.

PR #438, opened 2020-03-14
permissions checks

eboasson:
This PR fixes two issues with the permissions checks:
* passing of the ``relay_only`` parameter to the access control ``check_remote_datareader`` operation;
* correct an inverted test for the availability of a permissions handle prior to invoking the access control plugin
It furthermore now forwards the ``relay_only`` setting to the crypto plugin.


MarcelJordense:
The changes are ok. I have tested this with these changes an the tests succeeds.

eboasson:
Thanks for checking @MarcelJordense

PR #439, opened 2020-03-16
Code formatting fixes and clean-up authentication plugin

dennis-adlink:
This PR fixes code formatting in the authentication plugin, fixes a memory leak in `validate_handshake_reply_token` and makes error handling and return values more consistent with the other two security plugins.

Some additional refactoring could be done to remove code duplication in the authentication plugin (same applies to the other two security plugins), but this will be done in a future PR.

### DIFF
--- a/src/security/builtin_plugins/authentication/src/auth_utils.c
+++ b/src/security/builtin_plugins/authentication/src/auth_utils.c
@@ -151,13 +151,12 @@ static DDS_Security_ValidationResult_t check_key_type_and_size(EVP_PKEY *key, in
     }
     if (isPrivate)
     {
-      RSA *rsaKey = EVP_PKEY_get1_RSA(key);
+      RSA *rsaKey = EVP_PKEY_get0_RSA(key);
       if (rsaKey && RSA_check_key(rsaKey) != 1)
       {
         DDS_Security_Exception_set_with_openssl_error(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, DDS_SECURITY_VALIDATION_FAILED, "RSA key not correct : ");
         return DDS_SECURITY_VALIDATION_FAILED;
       }
-      RSA_free(rsaKey);
     }
     return DDS_SECURITY_VALIDATION_OK;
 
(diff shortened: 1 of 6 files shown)


src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -474,1812 +395,1382 @@ identityRelationFree(
...
-    if (memcmp(&localIdent->adjustedGUID, &remoteIdent->guid, sizeof(DDS_Security_GUID_t)) < 0) {
-        result = DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_REQUEST;
-    } else {
-        result = DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_MESSAGE;
-    }
-
-    return result;
+  *remote_identity_handle = IDENTITY_HANDLE(remoteIdent);
+  return memcmp(&localIdent->adjustedGUID, &remoteIdent->guid, sizeof(DDS_Security_GUID_t)) < 0 ?
+    DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_REQUEST : DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_MES
```
eboasson: This leaks ``pdhkey``, I think — it was already wrong before

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -2290,1203 +1781,871 @@ validate_handshake_reply_token(
...
+      add_validity_end_trigger(impl, IDENTITY_HANDLE(handshake->relation->remoteIdentity), cert_exp);
+  }
+  ddsrt_mutex_unlock(&impl->lock);
+  return result;
 
 err_invalid_expiry:
-    ddsrt_free( handshake->shared_secret_handle_impl->shared_secret );
-    ddsrt_free( handshake->shared_secret_handle_impl );
+  ddsrt_free(handshake->shared_secret_handle_impl->shared_secret);
+  ddsrt_free(handshake->shared_secret_handle_impl);
```
eboasson: This leaves a dangling pointer in ``handshake->shared_secret_handle_impl``, and IINM can lead to a double free

dennis-adlink:
@eboasson thanks for the review! I have pushed a commit that fixes these issues. I'll add another commit to remove some of the code duplication in this plugin. As discussed, we'll fix the serialization (use ddsi's serializer) in a future PR and not as part of this PR.

src/security/builtin_plugins/authentication/src/auth_utils.c
```diff
@@ -905,60 +903,53 @@ void free_ca_list_contents(X509Seq *ca_list)
...
-        }
-      }
+      if (load_X509_certificate_from_file(fname, &ca, ex) == DDS_SECURITY_VALIDATION_OK)
+        ca_buf[ca_cnt++] = ca;
+      else
+        DDS_Security_Exception_reset(ex);
+      ddsrt_free(fname);
+    }
+    ddsrt_free(fpath);
+    if (ca_cnt >= MAX_TRUSTED_CA)
```
eboasson: This fails if ``MAX_TRUSTED_CA`` have been read, even though one would expect a directory with ``MAX_TRUSTED_CA`` entries to be ok. 100 is arbitrary and 99 is equally arbitrary, but it has the potential of confusing the reader. It would be better to test before adding it to the buffer, then bailing out if the maximum has already been reached.

eboasson:
Thanks!

PR #440, opened 2020-03-16
Security core tests: new tests and refactoring

dennis-adlink:
Refactoring security core tests and adding new tests:
- Dynamically generate ca and identity certificates in authentication tests, so that certificate expiry is tested
   - The memory leak fix from PR #439 must be in for these tests to pass
   - Certificate expiry during a session is not yet handled correctly in ddsi, this test case is disabled until that issue is fixed in a future PR)
   - Added writing/reading samples in these tests to ensure that nodes can (or cannot) communicate  in a specific test case
- Secure communication tests: improved the validation of encryption in wrapper
- Test access control plugin settings. This only adds a first very basic test for access control; additional access control tests will be added in a future PR. 
- Replaced the (in code) test identities, added test ca private keys (to be able to change default test configuration or add test identities in a later stage)


src/security/core/tests/authentication.c
```diff
@@ -95,57 +141,276 @@ static void authentication_init(bool different_ca, const char * trusted_ca_dir,
...
+static void validate_hs(struct Handshake *hs, bool exp_fail_hs_req, const char * fail_hs_req_msg, bool exp_fail_hs_reply, const char * fail_hs_reply_msg)
 {
-  authentication_init (true, NULL, false);
-  validate_handshake (DDS_DOMAINID1, true, NULL, true, "error: unable to get local issuer certificate");
-  validate_handshake (DDS_DOMAINID2, true, NULL, true, "error: unable to get local issuer certificate");
-  authentication_fini (true);
+  DDS_Security_ValidationResult_t exp_result = hs->node_type == HSN_REQUESTER ? DDS_SECURITY_VALIDATION_OK_FINAL_MESSAGE : DDS_SECURITY_VALIDATION_OK;
+  if (hs->node_type == HSN_REQUESTER && exp_fail_hs_req)
+  {
+    CU_ASSERT_EQUAL_FATAL (hs->finalResult, exp_fail_hs_req ? DDS_SECURITY_VALIDATION_FAILED : exp_result);
```
eboasson: Execution never gets here if ¬exp_fail_hs_req — I do think one would want to check the result in the case of an expected successful handshake, and that therefore this needs a bit of rework.

src/security/core/tests/common/handshake_test_utils.c
```diff
@@ -22,6 +22,9 @@
 
 #define TIMEOUT DDS_SECS(2)
 
+static const char * validatation_result_str[] = { "OK", "FAILED", "PENDING_RETRY", "PENDING_HANDSHAKE_REQUEST", "PENDING_HANDSHAKE_MESSAGE", "OK_FINAL_MESSAGE" };
```
eboasson: There's a downside to doing the enum-to-string mapping this way (and there's a typo in the name besides) and that is that there's no compile-time check that all enum symbols are covered correctly. The old switch statement comes a lot closer, although in my view that one has a default case that shouldn't be there. If there's no need for an array, I much rather see a switch, but then reformatted and tweaked a bit, e.g.,
```C
static char * get_validation_result_str(DDS_Security_ValidationResult_t result)
{
  switch (result)
  {
    case DDS_SECURITY_VALIDATION_OK: return "ok";
    case DDS_SECURITY_VALIDATION_PENDING_RETRY: return "pending retry";
    case DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_REQUEST: return "handshake request";
    case DDS_SECURITY_VALIDATION_PENDING_HANDSHAKE_MESSAGE: return "handshake message";
    case DDS_SECURITY_VALIDATION_OK_FINAL_MESSAGE: return "ok final";
    case DDS_SECURITY_VALIDATION_FAILED: return "failed";
  }
  abort ();
  return "";
}
```
that'll cause a compiler warning if a case has been missed; the risk is that some compilers may warn about the abort/return at the end being unreachable ...

(And if you do need to use an array, designated initializers are nice, like ``[DDS_SECURITY_VALIDATION_OK] = "ok"``, etc.)

dennis-adlink:
Closing this PR, because it requires the fix from #439 and conflicts with changes in #442. I will open a new PR with the commits from this PR rebased on security branch after 442 (which needs some rework) is merged.

PR #441, opened 2020-03-16
Fix and simplify building docs and schemas

rotu:
1. Do an in-source build of derived schemas instead of building into build and checking that the files match.
2. Fix paths of installed docs, which were failing on me
3. Consolidate docs into same cmakelists that generates schema files
4. Trang is usually available as an executable, so find it with find_program.
5. Add doc dependencies to package.xml


eboasson:
Thanks @rotu for improving this. There are a few things that seem important to me but missing right now, perhaps you could see to what extent they can fit in (or explain why I am mistaken)?

* Not generating the schemata and ``options.md`` files if ``BUILD_DOCS`` isn't set means that most people aren't ever going to re-generate them. I can see the argument for making this part optional, but perhaps it would be better to have two options, one for running doxygen/sphinx (defaulting to OFF), and one for generating these 3 files (defaulting to ON)?

* For Travis builds, I would like to either retain the "old" behaviour or a fully automated update. The "old" means an error if committed versions do not match the source, which I think is sensible because ultimately the XSD &c. changes should be part of the same commit that changes those settings. A fully automated update could be an alternative, but I have my doubts about the feasibility of that.
  As to the mechanics of it, I guess a build option would be a sensible way.

rotu:
1. I kinda agree with you, but right now it's rather unfortunate that doing so would require Perl and Trang or break the build. I don't like the idea that the build could "succeed" but not generate these files because of a missing dependency.

2. I made a further change that *should* retain the old behavior

CMakeLists.txt
```diff
@@ -216,7 +216,11 @@ if(BUILD_IDLC)
 endif()
 
 option(BUILD_DOCS "Build documentation." OFF)
-add_subdirectory(docs)
+option(BUILD_SCHEMA "Build generated schema for configuration options." OFF)
+
+if(BUILD_DOCS)
+  add_subdirectory(docs)
```
eboasson: Shouldn't this be unconditional again?

eboasson: Answer to self: yes, otherwise it doesn't try to build it ... and when I try it fails on trang.

Ubuntu presumably has a package, homebrew has "jing-trang", chocolatey doesn't seem to have one. Arguably it isn't necessary to check this on each platform as it is platform independent.

But that little detail clearly needs some further investigation.

docs/CMakeLists.txt
```diff
@@ -8,43 +8,51 @@
...
-      COMMAND
-        ${PERL_EXECUTABLE} -w "${CMAKE_CURRENT_SOURCE_DIR}/compare.pl" cyclonedds.xsd "${CMAKE_CURRENT_SOURCE_DIR}/../etc/cyclonedds.xsd"
-      DEPENDS
-        "cyclonedds.rnc")
-    add_custom_target(options_xsd ALL DEPENDS "cyclonedds.xsd")
-  else()
-    message(STATUS "Java or not trang not found: not converting/checking RNC to XSD")
+  find_program(TRANG_EXECUTABLE trang)
+  if(NOT TRANG_EXECUTABLE)
+    message(FATAL_ERROR "Trang not found - needed for generating xsd files")
```
ThijsSassen: @rotu you could also download trang when needed by doing

```
  set(TRANG_ZIP ${CMAKE_CURRENT_BINARY_DIR}/trang.zip)
  if(NOT EXISTS ${TRANG_ZIP})
    message(STATUS "Trang not detected: Downloading..")
    file(DOWNLOAD https://github.com/relaxng/jing-trang/releases/download/V20181222/trang-20181222.zip ${TRANG_ZIP})
    execute_process(COMMAND ${CMAKE_COMMAND} -E tar -xf ${TRANG_ZIP}
                    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
  endif()
  set(TRANG_EXECUTABLE ${CMAKE_CURRENT_BINARY_DIR}/trang-20181222/trang.jar)
```
However i think you want to make a cmake option if you want to download trang or not in the case that it is missing.

CMakeLists.txt
rotu: Good catch

docs/CMakeLists.txt
rotu: 1. A jar is not directly executable, at least on my Mac, so this would break:
```
$ cat `which trang`
#!/bin/bash
exec java  -jar /usr/local/Cellar/jing-trang/20181222/libexec/build/trang.jar "$@"

$ /usr/local/Cellar/jing-trang/20181222/libexec/build/trang.jar
zsh: permission denied: /usr/local/Cellar/jing-trang/20181222/libexec/build/trang.jar
```

2. Java is not necessarily installed either.

3. I think it's quite reasonable to expect build dependencies to be installed by the user or their build dependency management solution (e.g. `rosdep`).

eboasson: Would it not be an option to search for ``trang`` and if it doesn't exist, download the jar run it using ``java -jar``?

I rather like your changes (and dislike the thought of them getting abandoned because of a triviality like this). I guess I could also simply generate the XSD from my perl script: that has its advantages over being dependent on an external tool. I quickly abandoned that route in the past after seeing what an XSD looks like, but it is (self-evidently) just a matter of generating the right syntax.

k0ekk0ek: Downloading files and executing them from somewhere is a bad idea IMHO. If something like that is necessary, I would suggest doing that in the `.travis.yml`. It's really just like any other dependency(?)

rotu: I don't love it, but I've implemented a cmake module to conditionally download Trang. Seems like a lot of code to save the user from a single `apt install trang`.

k0ekk0ek: Definitely like the CMake module. I think we should disregard @ThijsSassen's comment on downloading trang if it's missing, but stick with the module. Setting up a system should be done before building the software. At least, that's what I think. @ThijsSassen, is there a specific reason why we should download trang during the build and not when installing the system itself?

rotu: Well the advantage of doing it in CMake is *all* builds of this project use CMake. On the other hand, some builds use `rosdep` (which doesn't even support `<doc_depend>` yet though it's in the package xml spec) some use `conan` (which does not have a listed version of `trang`), IDLC uses `Maven`.

It would be nice to have some saner dependency management, but ... https://xkcd.com/927/

k0ekk0ek: When I created the build system, I set it up with the basic rule that Conan is used for all C dependencies (libraries). Tools, like the compiler, Conan, CMake, Sphinx, etc must be available on the system beforehand. CMake is of course free to require a certain version, like you do in the module. It's just that when you mix building software and installing the system together, things will get messy. `rosdep` isn't used by Cyclone DDS (I believe) and Maven will eventually be removed when I get around to replace IDLC (Soonish, probably June/July)

k0ekk0ek: For the record, I'm sorry to only take part in this discussion now. It slipped past me. Other than the download bit the changes are a welcomed and very worthwhile addition. Thanks.

rotu: Right now, Conan is *not* used for ROS2 builds. Only rosdep is, which uses the package.xml file. That may be the wrong call, but it's how things are expected for ROS2 where the workflow is generally:

`git clone`, `rosdep install`, `colcon build`

Rosdep is a frontend that dispatches to apt / pip / homebrew / chocolatey, etc. based on the platform and package. We could certainly use conan there, b

I agree that the download is non-ideal practice, but I just want to get these changes merged in :-(

k0ekk0ek: I'll let @eboasson, be the judge here as getting it in might be important enough to take this one and clean it up afterwards. If trang is required for the ROS2 build, I suggest specifying it in `package.xml` (can even be just a `depend`) and remove the download code from the CMake module. If it's not, list it as a `doc_depend` in `package.xml` and remove the download code from the CMake module. If it's required in the Eclipse Cyclone DDS builds on Travis, add it to `.travis.yml` too. That would be my preference.

ThijsSassen: @k0ekk0ek i agree that when setting up the System would be the time to also install trang i.e. add it to the travis.yml as you suggested. As we currently have maven for java dependencies but as already mentioned maven does not provide trang i suggested to download it in cmake, but its also not my preferred method.

k0ekk0ek: @rotu, saw [this PR](https://github.com/ros/rosdistro/pull/24231). Much appreciated! Again, if this is required short-term, with a Chocolatey package not being available, I'm fine with it. We can worry about the proper solution later(?)

k0ekk0ek: Doesn't seem like rosdep is supported on Windows yet? Also, don't see a Chocolatey package installer implementation for rosdep, so Windows may not be a problem(?) That being said, I do see a SourceInstaller Python module there. @rotu, can that installer module be used to download the archive and install it for ROS2 or would we need an actual Chocolatey package? Anyway, just curious. If you have some information, let me know, I might be able to help out.

k0ekk0ek: Took the liberty to create a [patch for `.travis.yml`](https://gist.github.com/k0ekk0ek/8a53f708eb098cd9c4006bbb674e6c24).

rotu: `rosdep` [does not yet support `doc_depend`](https://www.ros.org/reps/rep-0149.html#doc-depend-multiple), so even if Trang is available for all target platforms and rosdep works perfectly on Windows, it's not going to work yet. I added `trang` to the package.xml more for documentation than for functionality at this point in time.

I don't want to throw any more time down this hole right now. Please come to an agreement with @ThijsSassen and @eboasson and I'll gladly code it up later whichever way you guys like.

k0ekk0ek: We (@eboasson, @ThijsSassen and I) are in agreement that downloading is definitely not the way to go. Once that's removed, I'm happy to merge this in. Thanks.

rotu: @k0ekk0ek Done.

rotu:
Why didn't this rerun ci after recent changes?

docs/CMakeLists.txt
k0ekk0ek: Didn't go unnoticed @rotu. Actually had a quick look, but lost myself in trying to find a way to ditch Perl and Java. Not important for this PR, I'll have another look later today.

rotu: I support that!

k0ekk0ek:
I'm accepting the PR once the Travis check succeeds (I'll look into that). `compare.pl` got removed, in favor of using `git diff`(?) This makes Git a dependency. Not a problem when building from the repository, but maybe a problem when building from a release archive. Though there are likely more non-UNIX users that have Git installed than there are users that have Perl available (the 90's have been over for some time and Perl is not as widespread as it used to be. Still, it's hard to beat for this sort of use...), on the other hand, if Perl is not available `makernc.pl` also won't run... Nevermind. (talking about the 90's :wink:). I'd rather had the files were copied to the build directory (or generated) instead of generating them in the source tree, but that too is something we can worry about later. Thanks @rotu!

rotu:
That compare check is really just there to make sure that, after updating the config parser, the schemas get regenerated. It's a check for CI that the user didn't forget to regenerate the schemas and check them in.

Yes, the files should be built - I'm in favor of building them and hosting them somewhere web-accessible so that they can be used easily for schema validation.

k0ekk0ek:
@rotu, `.travis.yml` is corrupt. If you remove the added comment it'll build.
```
diff --git a/.travis.yml b/.travis.yml
index 4803479d..bc890677 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -227,8 +227,9 @@ script:
       CMAKE_C_FLAGS="-DCMAKE_C_FLAGS=-fsanitize=${USE_SANITIZER}";
     fi
   - cd ..
-  - if [ ${BUILD_SCHEMA} ]; then
-      git diff --exit-code; # check that no files (e.g. generated schemas) are changed
+  - |
+    if [ ${BUILD_SCHEMA} ] then
+      git diff --exit-code
     fi
   - mkdir helloworld_build
   - cd helloworld_build
```

k0ekk0ek:
@rotu, I tried branching, then fixing and issueing a PR, but the ECA check then doesn't run. If you would be so kind to fix the `.travis.yml`, I'll merge the changes. Thanks. (sorry that the process is taking this long).

rotu:
@k0ekk0ek very good catch!

PR #443, opened 2020-03-18
Set LateAckMode default to True

rotu:
This seems to fix issue with spotty service discovery


rotu:
seeing how many things fail, I'm guessing this is a bad idea to change across the board

PR #448, opened 2020-03-24
Improvements to building against & using master/security-enabled/security-but-disabled versions

eboasson:
This PR adds a two key features (+ a test):

* Define a ``DDS_HAS_PROPERTY_LIST_QOS`` macro (as discussed in https://github.com/ros2/rmw_cyclonedds/pull/123) to allow users of Cyclone to conditionally include support for setting the QoS properties needed for configuring security. This'll allow, e.g., the RMW the layer to cover both master & security.
* Refuse to create a participant if some "dds.sec." properties are set but support security is not included in the build.
* Add a test to check the behaviour of the above. It relies on the CI doing builds with/without security included (we do both), but I don't see how it can be verified otherwise.

The diff's a bit bigger than strictly necessary merely for the above:

* Some of the property checking code was an over-parametrised mess where two simple functions do the trick. That's bad in general but even more so in the context of checking whether security options are set.
* ``new_participant_guid`` had grown unwieldy over time and which gave a memory leak (if a rare setting was enabled) a chance to hide. A big reason was the huge gob of code for creating the built-in endpoints. A bit of refactoring helps.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -739,8 +739,22 @@ dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
  *
  * @retval >0
  *             A valid participant handle.
+ * @retval DDS_RETCODE_NOT_ALLOWED_BY_SECURITY
+ *             An invalid DDS Security configuration was specified (whether
+ *             that be missing or incorrect entries, expired certificates,
+ *             or anything else related to the security settings and
+ *             implementation).
+ * @retval DDS_RETCODE_PRECONDITION_NOT_MET
+ *             Some security properties specified in the QoS, but the Cyclone
+ *             build does not include support for DDS Security.
+ * @retval DDS_RETCODE_OUT_OF_RESOURCES
+ *             Some resource limit (maximum participants, memory, handles,
+ *             &c.) prevented creation of the participant.
  * @retval DDS_RETCODE_ERROR
- *             An internal error has occurred.
+ *             The "CYCLONEDDS_URI" environment variable lists non-existent
+ *             or invalid configuration files, or contains invalid embedded
+ *             configuration items; or an unspecified internal error has
+ *             occurred.
  */
 DDS_EXPORT dds_entity_t
 dds_create_participant(
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -25,6 +25,7 @@
 #endif
 
 #include "dds/export.h"
+#include "dds/features.h"
 
 /**
  * Handle to an entity. A valid entity handle will always have a positive
--- a/src/core/ddsc/include/dds/ddsc/dds_public_qos.h
+++ b/src/core/ddsc/include/dds/ddsc/dds_public_qos.h
@@ -23,6 +23,11 @@
 #include "dds/export.h"
 #include "dds/ddsc/dds_public_qosdefs.h"
 
+/* Whether or not the "property list" QoS setting is supported in this version.  If it is,
+   the "dds.sec." properties are treated specially, preventing the accidental creation of
+   an non-secure participant by an implementation built without support for DDS Security. */
+#define DDS_HAS_PROPERTY_LIST_QOS 1
+
 #if defined (__cplusplus)
 extern "C" {
 #endif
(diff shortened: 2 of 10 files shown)


src/core/ddsi/src/q_entity.c
```diff
@@ -828,77 +956,18 @@ dds_return_t new_participant_guid (ddsi_guid_t *ppguid, struct ddsi_domaingv *gv
...
+    goto not_allowed;
+  if ((ret = q_omg_security_check_create_participant (pp, gv->config.domainId)) != DDS_RETCODE_OK)
+    goto not_allowed;
+  *ppguid = pp->e.guid;
+#else
+  if (ddsi_xqos_has_prop_prefix (&pp->plist->qos, "dds.sec."))
+  {
+    /* disallow creating a participant with a security configuration if there is support for security
+       has been left out */
+    ret = DDS_RETCODE_NOT_ALLOWED_BY_SECURITY;
```
dennis-adlink: Maybe replace this error code by DDS_RETCODE_PRECONDITION_NOT_MET?

eboasson: I had a debate about it with myself before pushing it to GitHub, and I just don't know what is best. My reasoning was that "not allowed by security" at least makes it immediately obvious it has something to do with the security configuration, whereas "precondition not met" is one of those overly generic ones. But at the same time, it might well make one go looking in exactly the wrong place: trying to find out what is wrong with the configuration one has specified ...

I'll mull it over a bit in the hope of getting a few votes either way before having to make the call 😀

src/core/ddsc/include/dds/ddsc/dds_public_qos.h
```diff
@@ -23,6 +23,11 @@
 #include "dds/export.h"
 #include "dds/ddsc/dds_public_qosdefs.h"
 
+/* Whether or not the "property list" QoS setting is supported in this version.  If it is,
+   the "dds.sec." properties are treated specially, preventing the accidental creation of
+   an non-secure participant by an implementation built without support for DDS Security. */
+#define DDS_HAS_PROPERTY_LIST_QOS 1
```
kyrofa: Can you expose a `DDS_HAS_SECURITY_ENABLED` (or similar) as well? Otherwise we will need to assume security is enabled if `DDS_HAS_PROPERTY_LIST_QOS` is defined, which doesn't seem to be an accurate assumption to make unless I'm missing something.

eboasson: Yes, I can — and I have just pushed a commit that provides a ``DDS_HAS_SECURITY`` you can check at compile time.

It wasn't there at first because I thought it wasn't strictly necessary, because the non-security-enabled builds will (with this PR) outright refuse the creation of any DDS participant that requests security — and secondarily because it saved me from having to deal with CMake. But I think you're right that having it is indeed useful and the only downside (having to deal with CMake) is rather silly!

src/core/ddsi/src/q_entity.c
eboasson: Mulled it over in the absence of votes one way or the other. ``DDS_PRECONDITION_NOT_MET`` it will be. Thanks!

src/core/ddsc/include/dds/ddsc/dds_public_qos.h
kyrofa: > because it saved me from having to deal with CMake

Haha, been there man. Thanks for adding!

src/core/ddsc/include/dds/dds.h
```diff
@@ -738,8 +739,22 @@ dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
...
  *             A valid participant handle.
+ * @retval DDS_RETCODE_NOT_ALLOWED_BY_SECURITY
+ *             An invalid DDS Security configuration was specified (whether
+ *             that be missing or incorrect entries, expired certificates,
+ *             or anything else related to the security settings and
+ *             implementation).
+ * @retval DDS_RETCODE_PRECONDITION_NOT_MET
+ *             Some security properties specified in the QoS, but the Cyclone
+ *             build does not include support for DDS Security.
+ * @retval DDS_OUT_OF_RESOURCES
```
kyrofa:
```suggestion
 * @retval DDS_RETCODE_OUT_OF_RESOURCES
```

eboasson:
@kyrofa and @dennis-adlink thanks for the review comments

PR #450, opened 2020-03-25
Fix read/return_loan memory leak

eboasson:
This PR addresses an edge case that caused a one-off memory leak on Travis, thanks to some fortunate timing. It improves the tests to no longer be dependent on timing for detecting this particular kind of problem.

* read/take failed to restore the null pointer in the first entry of the sample pointer array it gets passed, in the case no "loan" had been allocated yet and it returned an empty set.  The consequence is that on a subsequence read it will reuse the address without marking at as in use, so that a *second* read using with a null pointer in that first entry will overwrite the first result.  (Introduced by d16264f.)
* return_loan failed to free all memory if its argument wasn't actually a loan.  There are many good arguments why the read/take/return_loan interface is messed up, but in the context of the existing interface this is a perfectly reasonable case: there is at most one "loan" for each reader, but one can keep calling read/take and return_loan as if there's an infinite number of "loans".  It's just that the first gets cached and the others don't.

In a separate commit it moves a function duplicated in nearly every test case to a common file: the updated tests for ``dds_return_loan`` require independence from other tests running simultaneously and I could not bring myself to copy that function yet another time.


dennis-adlink:
LGTM. When improving the interface, we may consider that calling read/take with loan requires a previously provided loan for this reader to be returned. I agree that this change should not be part of this fix.

eboasson:
Thanks @dennis-adlink

PR #453, opened 2020-03-27
Update crypto plugin w.r.t. origin authentication.

MarcelJordense:
When origin authentication is configured the specific reader sign's present in the received message should be checked if the sign corresponds with the targeted reader. For this purpose the cryptographic plugin should associate the key-id present in the message with the correct key material. When there are more than one participant per application it could occur that the cryptographic plugin did not find the key material by the provide key-id. For this purpose a separate index is implemented which provides the relation between the specific reader key material and the corresponding key-id. 
Another problem that is solved is that when discovery information about remote entities is received before the authentication handshake with the remote participant is finished that the sending of the corresponding cryptographic tokens is delayed until the handshake is finished and the shared secret is determined which is needed for the exchange of the cryptographic tokens.

### DIFF
--- a/src/core/ddsi/src/q_entity.c
+++ b/src/core/ddsi/src/q_entity.c
@@ -3620,7 +3620,7 @@ static void new_writer_guid_common_init (struct writer *wr, const struct ddsi_se
     assert ((wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL) ||
             (wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_STATELESS_MESSAGE_WRITER));
   }
-  wr->handle_as_transient_local = (wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL);
+  wr->handle_as_transient_local = (wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL || wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER);
   wr->include_keyhash =
     wr->e.gv->config.generate_keyhash &&
     ((wr->e.guid.entityid.u & NN_ENTITYID_KIND_MASK) == NN_ENTITYID_KIND_WRITER_WITH_KEY);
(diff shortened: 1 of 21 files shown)


src/core/ddsi/src/q_entity.c
```diff
@@ -3601,7 +3601,7 @@ static void new_writer_guid_common_init (struct writer *wr, const struct ddsi_se
     assert ((wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL) ||
             (wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_STATELESS_MESSAGE_WRITER));
   }
-  wr->handle_as_transient_local = (wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL);
+  wr->handle_as_transient_local = (wr->xqos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL || wr->e.guid.entityid.u == NN_ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER);
```
eboasson: This bothers me: there is almost a guarantee that a volatile reader won't retrieve historical data, and it is also not clear to me how this avoids being a resource leak. Maybe I'm overlooking something.

MarcelJordense: You are correct, this change is incorrect. I will remove this change.

eboasson:
Thanks @MarcelJordense — with this set of commits, the "multiple_readers" test runs pretty reliable on, my laptop, out of approximately twenty runs or so, I have (so far) twice seen:
```
Testing: 1 subscriber domains, 1 pp per domain, 1 rd per pp; 1 publishing domains, 1 pp per domain, 1 wr per pp
Governance configuration: DISCOVERY_PROTECTION_KIND=NONE; LIVELINESS_PROTECTION_KIND=NONE; RTPS_PROTECTION_KIND=ENCRYPT; METADATA_PROTECTION_KIND=SIGN; DATA_PROTECTION_KIND=ENCRYPT; 

Suite ddssec_secure_communication, Test protection_kinds had failures:
    1. /Users/erik/C/cdds/src/security/core/tests/secure_communication.c:283  - ret >= 1
```
That run was with simulated packet loss, by the way. No proper logs that I can provide you with ...

An interesting tidbit: with a small amount of (simulated) packet loss (1%) the security handshakes consistently go faster! Clearly we have some work to do still ...

I think the above observations simply mean there is more work to be done yet (and I look forward to a future PR 🙂). I don't see a reason to not merge this PR because of them.

PR #455, opened 2020-03-27
Stop threads doing handshake processing earlier

eboasson:
In particular before the state they depend on gets torn down.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_handshake.h
@@ -169,6 +169,13 @@ struct ddsi_handshake * ddsi_handshake_find(struct participant *pp, struct proxy
 void ddsi_handshake_admin_init(struct ddsi_domaingv *gv);
 
 /**
+* @brief Stop handshake background processing.
+*
+* @param[in] gv         The global parameters
+*/
+void ddsi_handshake_admin_stop(struct ddsi_domaingv *gv);
+
+/**
  * @brief Deinitialze the handshake administration.
  *
  * @param[in] gv         The global parameters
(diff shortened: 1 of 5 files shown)


dennis-adlink:
LGTM! I've seen issues with fsm de-init a couple of times when running tests locally (not easily reproducible), so this PR will probably fix this. 

Some nit-picking: `gv` is passed to `ddsi_handshake_admin_stop` and `ddsi_handshake_admin_deinit`, but only `gv->hsadmin` is used here. And in line with that, `ddsi_handshake_admin_init` could also return `struct ddsi_hsadmin *hsadmin` so it does not require gv.

eboasson:
> LGTM! I've seen issues with fsm de-init a couple of times when running tests locally (not easily reproducible), so this PR will probably fix this.
> 
> Some nit-picking: `gv` is passed to `ddsi_handshake_admin_stop` and `ddsi_handshake_admin_deinit`, but only `gv->hsadmin` is used here. And in line with that, `ddsi_handshake_admin_init` could also return `struct ddsi_hsadmin *hsadmin` so it does not require gv.

Good point, thanks for noticing it!

eboasson:
> > LGTM! I've seen issues with fsm de-init a couple of times when running tests locally (not easily reproducible), so this PR will probably fix this.
> > Some nit-picking: `gv` is passed to `ddsi_handshake_admin_stop` and `ddsi_handshake_admin_deinit`, but only `gv->hsadmin` is used here. And in line with that, `ddsi_handshake_admin_init` could also return `struct ddsi_hsadmin *hsadmin` so it does not require gv.
> 
> Good point, thanks for noticing it!

On consideration, making that change has the downside of having a new/free pair of functions for instantiating a "handshake administration", where the remainder of the handshake implementation assumes that the underlying administration is pointed to by ``gv->hsadmin``. That discrepancy doesn't exist right now.

That means the scope of the change needs to be a fair bit bigger and is in my view best done in a separate PR. This one can then be merged as-is as the minimal change needed to address the immediate problem.

PR #456, opened 2020-03-29
Do not rewrite secure messages in retransmit queue

eboasson:
Messages to be retransmitted spend some time on a transmit queue, and are subject to the rewriting of the destination information to reduce the number of outgoing copies.   -evidently, altering the message header does not sit well with encryption and/or authentication of messages.

The way the rewriting works is that the offset of the "reader entity id" in the DATA submessage is saved on message construction (the GUID prefix is at a fixed location), so that it can be read and possibly zero'd out later. The crypto transformations move the message around and it so happens that it can end up pointing to the key id in the encoded message.  Zeroing that one out leads to uninterpretable messages.

This commit adds a message/event kind to distinguish between retransmit that may and retransmit that may not be merged (and thus rewritten) and gets used when the crypto plugin is invoked to transform a message.

### DIFF
--- a/src/core/ddsi/src/q_xmsg.c
+++ b/src/core/ddsi/src/q_xmsg.c
@@ -553,7 +553,9 @@ void nn_xmsg_submsg_remove(struct nn_xmsg *msg, struct nn_xmsg_marker sm_marker)
   /* Just reset the message size to the start of the current sub-message. */
   msg->sz = sm_marker.offset;
 
-  /* if it is a DATA_REXMIT one, reset the readerId offset because it may no longer be valid */
+  /* Deleting the submessage means the readerId offset in a DATA_REXMIT message is no
+     longer valid.  Converting the message kind to a _NOMERGE one ensures no subsequent
+     operation will assume its validity. */
   if (msg->kind == NN_XMSG_KIND_DATA_REXMIT)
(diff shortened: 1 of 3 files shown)


src/core/ddsi/src/q_xmsg.c
```diff
@@ -573,6 +578,11 @@ void nn_xmsg_submsg_replace(struct nn_xmsg *msg, struct nn_xmsg_marker sm_marker
 
   /* Replace the sub-message. */
   memcpy(msg->data->payload + sm_marker.offset, new_submsg, new_len);
+
+  /* if it is a DATA_REXMIT one, reset the readerId offset because it may no longer be valid,
+     or patching it might break stuff if the replaced data is encrypted or authenticated */
```
dennis-adlink: This comment is not completely clear to me. I see that `lookup_msg` doesn't return the msg when it is of kind `NN_XMSG_KIND_DATA_REXMIT_NOMERGE` and therefore security-encoded messages won't be merged. But maybe you could add in this comment at what point 'the readerId offset' is reset?

eboasson: Ouch. I changed my mind: first I thought I'd reset the readerId and use that as an indication, then later decided to use a different kind and failed to update the comment.

PR #459, opened 2020-03-30
Improve security-core tests and implement identity and permissions revocation

dennis-adlink:
This PR adds tests to the security-core access control and authentication test suites, and contains some refactoring to reduce duplicate test code. It also contains the implementation for the permissions and identity certificate revocation, which is required for these tests to pass.

A summary of the changes in this PR (most of this description is also in the commit messages)
- Added a `struct domaingv` pointer to the 3 security plugins context, required by the revocation handlers to enumerate (proxy)participants
- Implement handler for access control on_revoke_permissions. This callback function disconnects and deletes all proxy participant that are using the revoked permissions handle (in case of remote permissions expire) and proxy participant that are connected with a participant for which the permissions expire (local permissions expire).
- Impement the revoke identity callback in DDSI, that is called when the identity certificate of a participant expires. In case the identity handle that expires is from a local participant, all proxy pp connections will be dropped for this participant. In case the identity that expires is from a remote participant, the corresponding proxy participant is deleted.
- Added tests for access control, wrt plugin configuration and permission expiry (more tests will be added in another PR)
- Some minor code style (whitespace) updates in the security plugins


src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -686,7 +695,112 @@ static void deinit_plugin_suite_config (dds_security_plugin_suite_config *suite_
...
+  {
+    struct dds_security_context *sc = q_omg_security_get_secure_context(pp);
+    ddsrt_mutex_lock (&pp->sec_attr->lock);
+    if (pp->sec_attr->permissions_handle == handle)
+    {
+      uint32_t i = 0;
+      ddsrt_avl_citer_t it;
+      local = true;
+      for (struct pp_proxypp_match *ppm = ddsrt_avl_citer_first (&pp_proxypp_treedef, &pp->sec_attr->proxy_participants, &it); ppm; ppm = ddsrt_avl_citer_next (&it), i++)
+        pp_proxypp_unrelate_locked (sc, pp, &ppm->proxypp_guid);
```
MarcelJordense: I do not think that only removing the security relation between the participant and the proxy_participant will be sufficient. The application will not be aware of this event and this will probably cause that the associated reader and writer entities still have valid crypto handles and still will try to communicate.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -686,7 +695,112 @@ static void deinit_plugin_suite_config (dds_security_plugin_suite_config *suite_
...
+    while ((proxypp = entidx_enum_proxy_participant_next (&eproxypp)) != NULL)
+    {
+      ddsrt_mutex_lock (&proxypp->sec_attr->lock);
+      uint32_t i = 0;
+      ddsrt_avl_iter_t it;
+      bool del_proxypp = true;
+      for (struct proxypp_pp_match *ppm = ddsrt_avl_iter_first (&proxypp_pp_treedef, &proxypp->sec_attr->participants, &it); ppm; ppm = ddsrt_avl_iter_next (&it), i++)
+      {
+        if (ppm->permissions_handle == handle)
+          proxypp_pp_unrelate_locked (proxypp->sec_attr->sc, proxypp, &ppm->pp_guid, ppm->pp_crypto_handle);
```
MarcelJordense: Is it not sufficient to delete the proxy_participant. This will also cleanup the relations.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -686,7 +695,112 @@ static void deinit_plugin_suite_config (dds_security_plugin_suite_config *suite_
   deinit_plugin_config (&suite_config->cryptography);
 }
 
-dds_return_t q_omg_security_load (dds_security_context *sc, const dds_qos_t *qos)
+static DDS_Security_boolean on_revoke_permissions_cb(const dds_security_access_control *plugin, const DDS_Security_PermissionsHandle handle)
+{
+  struct ddsi_domaingv *gv = plugin->gv;
```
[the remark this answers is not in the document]
dennis-adlink: I think this is not a problem and will also work when using multiple domains (which have their own `gv` and own `gv->security_context`), because the plugin context (e.g. `gv->security_context->authentication_context`) is allocated in the plugin's init functions for each domain.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -686,7 +695,112 @@ static void deinit_plugin_suite_config (dds_security_plugin_suite_config *suite_
...
+  struct proxy_participant *proxypp;
+  bool local = false;
+  thread_state_awake (lookup_thread_state (), gv);
+
+  /* Find participants using this permissions handle */
+  entidx_enum_participant_init (&epp, gv->entity_index);
+  while ((pp = entidx_enum_participant_next (&epp)) != NULL)
+  {
+    struct dds_security_context *sc = q_omg_security_get_secure_context(pp);
+    ddsrt_mutex_lock (&pp->sec_attr->lock);
```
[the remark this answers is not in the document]
dennis-adlink: Thanks for catching this, I'll fix it by adding a null check on `pp->sec_attr`

src/core/ddsi/src/ddsi_security_omg.c
dennis-adlink: As each pp-proxypp match has its own permissions_handle and only one of these handles will be expired in a call to this callback, we need to disconnect the pp-proxypp pair. As you pointed out [here](https://github.com/eclipse-cyclonedds/cyclonedds/pull/459/files/c99af434341029c9176d621bd80fe10b3e8fcca1#r400793611) this requires more than just calling `proxypp_pp_unrelate`, so I will fix this code in the same way.

dennis-adlink: As each pp-proxypp match has its own `permissions_handle` and only one of these handles will be expired in a call to this callback, we need to disconnect the pp-proxypp pair. As you pointed out [here](https://github.com/eclipse-cyclonedds/cyclonedds/pull/459#discussion_r400793611), this requires more than just calling `proxypp_pp_unrelate`, so I will fix this code in the same way.

src/core/ddsi/src/ddsi_security_omg.c
dennis-adlink: With some new insights and after discussing this with @eboasson we've decided to change the approach for this, and delete the (proxy) participant in case of expired identity certificate or expired permissions. See the code comment for `on_revoke_permissions_cb` and `on_revoke_identity_cb` functions in the latest commit in this PR for more details.

src/core/ddsi/src/ddsi_security_omg.c
dennis-adlink: see [this comment](https://github.com/eclipse-cyclonedds/cyclonedds/pull/459#discussion_r401698870) above

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -686,7 +695,123 @@ static void deinit_plugin_suite_config (dds_security_plugin_suite_config *suite_
...
+{
+  struct proxy_participant *proxypp;
+  struct entidx_enum_proxy_participant eproxypp;
+  bool result = false;
+  entidx_enum_proxy_participant_init (&eproxypp, gv->entity_index);
+  while ((proxypp = entidx_enum_proxy_participant_next (&eproxypp)) != NULL)
+  {
+    if (q_omg_proxy_participant_is_secure(proxypp) && expired_proxypp_check_fn (proxypp, handle))
+    {
+      delete_proxy_participant_by_guid (gv, &proxypp->e.guid, ddsrt_time_wallclock (), false);
```
eboasson, line 729: The last parameter is whether the proxy sent a message explicitly stating that it was deleted by the application, or whether it was (I suppose) "implicit" because of lease expiry. I think the case of an expired certificate can be argued either way, but "implicit" and hence `true` seems slightly better to me.

dennis-adlink: You're right, this should be an implicit delete, although the only result of setting this parameter to `true` is that the `lease_expired` flag on the proxy participant is set, which is currently not used anywhere I think. I've changed this in 23de64a

dennis-adlink:
@eboasson thanks for reviewing! And yes, these commits contain, besides the implementation of revocation of identity and permissions, mostly code cleanup (I couldn't resist removing all the unnecessary whitespace) and refactoring to better support the test cases I'm currently working on.

eboasson:
@dennis-adlink I'd be happy to merge it but there are some conflicts. Can you please address those? Thanks.

dennis-adlink:
@eboasson I've rebased the commits in this PR on the security branch

eboasson:
Thanks @dennis-adlink !

PR #463, opened 2020-03-31
Crypto endpoint relation compare routines cleanup

eboasson:
* Trying not to assume an int is at least 32 bits.

* Technically speaking, comparing "unrelated" addresses is undefined behaviour which can be avoided by a cast to uintptr_t.

* The early out if either local_crypto == 0 does work in context, provided the nodes in tree never have local_crypto == 0.  That implies crypto_insert_endpoint_relation must never have a 0 in there, which I think the callers do respect.  Still I think it is better to not hide these assumptions in the compare function and address the problem in the lookup function instead.

These changes likely make the code fractionally slower, but I do think they improve clarity.

@MarcelJordense could you please have a look and see whether it makes any sense at all?


eboasson:
Thanks for reviewing @MarcelJordense

PR #467, opened 2020-04-01
Correct length of encrypted submessage body

MarcelJordense:
When a sub-message is encrypted the length of the resulting sec_body sub-message is set to 4 bytes higher than the actual encrypted payload. Although this does not provide a problem because the length field octet sequence is correct it is better to correct this.
Further for the proxy reader and proxy writer the security_info element present in the entity common is used.

### DIFF
--- a/src/core/ddsi/src/q_entity.c
+++ b/src/core/ddsi/src/q_entity.c
@@ -4475,9 +4475,6 @@ dds_return_t delete_reader (struct ddsi_domaingv *gv, const struct ddsi_guid *gu
   GVLOGDISC ("delete_reader_guid(guid "PGUIDFMT") ...\n", PGUID (*guid));
   builtintopic_write (rd->e.gv->builtin_topic_interface, &rd->e, ddsrt_time_wallclock(), false);
   entidx_remove_reader_guid (gv->entity_index, rd);
-#ifdef DDSI_INCLUDE_SECURITY
-  q_omg_security_deregister_reader(rd);
-#endif
   gcreq_reader (rd);
   return 0;
 }
@@ -5369,8 +5366,7 @@ static int proxy_endpoint_common_init (struct entity_common *e, struct proxy_end
     memset (&c->group_guid, 0, sizeof (c->group_guid));
 
 #ifdef DDSI_INCLUDE_SECURITY
-  c->security_info.security_attributes = 0;
-  c->security_info.plugin_security_attributes = 0;
+  q_omg_get_proxy_endpoint_security_info(e, &proxypp->security_info, plist, &c->security_info);
 #endif
 
   if ((ret = ref_proxy_participant (proxypp, c)) != DDS_RETCODE_OK)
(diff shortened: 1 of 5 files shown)


src/core/ddsi/src/q_entity.c
```diff
@@ -5518,7 +5515,7 @@ int new_proxy_writer (struct ddsi_domaingv *gv, const struct ddsi_guid *ppguid,
 
 #ifdef DDSI_INCLUDE_SECURITY
   set_proxy_writer_security_info(pwr, plist);
-  q_omg_get_proxy_writer_security_info(pwr, plist, &(pwr->security_info));
+  q_omg_get_proxy_writer_security_info(pwr, plist, &(pwr->c.security_info));
```
eboasson, line 5518: I'd rather see this call moved to ``proxy_endpoint_common_init`` — and then using ``q_omg_get_proxy_endpoint_security_info`` which is how ``q_omg_get_proxy_writer_security_info`` and ``q_omg_get_proxy_reader_security_info`` are implemented anyway.

src/core/ddsi/src/q_entity.c
```diff
@@ -5805,7 +5802,7 @@ int new_proxy_reader (struct ddsi_domaingv *gv, const struct ddsi_guid *ppguid,
   ddsrt_avl_init (&prd_writers_treedef, &prd->writers);
 
 #ifdef DDSI_INCLUDE_SECURITY
-  q_omg_get_proxy_reader_security_info(prd, plist, &(prd->security_info));
+  q_omg_get_proxy_reader_security_info(prd, plist, &(prd->c.security_info));
```
eboasson, line 5805: Same as for ``new_proxy_writer``

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -1192,9 +1194,8 @@ encode_datawriter_submessage_encrypt (
     }
   }
   else
-  {
```
eboasson, line 1195: I wouldn't remark on this if I had no other comments, but: as rule I don't like ``if(P) X else Y`` where X is significantly longer than Y, and I really don't like Y to then be a single statement without braces. Just so you know :)

src/core/ddsi/src/q_entity.c
```diff
@@ -4475,9 +4475,6 @@ dds_return_t delete_reader (struct ddsi_domaingv *gv, const struct ddsi_guid *gu
   GVLOGDISC ("delete_reader_guid(guid "PGUIDFMT") ...\n", PGUID (*guid));
   builtintopic_write (rd->e.gv->builtin_topic_interface, &rd->e, ddsrt_time_wallclock(), false);
   entidx_remove_reader_guid (gv->entity_index, rd);
-#ifdef DDSI_INCLUDE_SECURITY
```
eboasson, line 4478: I don't understand why this should disappear

MarcelJordense: The call to q_omg_security_deregister_reader at this location could cause a crash because the sec_attr field could still be accessed at that moment. The function gc_delete_reader also calls q_omg_security_deregister_reader where it is safe.

eboasson: Ah, ok, that makes sense. Doesn't the same argument apply to ``q_omg_security_deregister_writer``?

MarcelJordense: Yes you are right I missed that one will correct it.

eboasson:
Thanks, @MarcelJordense !

PR #470, opened 2020-04-03
Remove temporally stored crypto handles and tokens after entities are…

MarcelJordense:
Before local and remote entities can be matched the entities have to be registered which the cryptographic plugin which returns a corresponding handle and the corresponding cryptographic tokens have to be received from the remote entity. It can be that the cryptographic tokens have been received from the remote entity before the corresponding discovery information arrives in that case the received tokens must be cached. At moment the match is made either when receiving the token or the discovery information the temporally cached security information should be released.

### DIFF
--- a/src/core/ddsi/src/ddsi_security_omg.c
+++ b/src/core/ddsi/src/ddsi_security_omg.c
@@ -415,20 +415,6 @@ static void pending_match_expiry_cb(struct xevent *xev, void *varg, ddsrt_mtime_
   ddsrt_mutex_unlock(&index->lock);
 }
 
-static void inc_guid(ddsi_guid_t *guid)
-{
-  guid->entityid.u++;
-  if (guid->entityid.u == 0)
-  {
-    for (int i = 2; i >= 0; i++)
-    {
-      guid->prefix.u[i]++;
-      if (guid->prefix.u[i] != 0)
-        break;
-    }
-  }
-}
-
 static void clear_pending_matches_by_local_guid(dds_security_context *sc, struct pending_match_index *index, const ddsi_guid_t *local_guid)
 {
   struct pending_match *match;
@@ -443,8 +429,7 @@ static void clear_pending_matches_by_local_guid(dds_security_context *sc, struct
       ddsrt_avl_delete(&pending_match_index_treedef, &index->pending_matches, match);
       if (match->expiry.v != DDS_NEVER)
         ddsrt_fibheap_delete(&pending_match_expiry_fhdef, &index->expiry_timers, match);
-      inc_guid(&match->guids.remote_guid);
-      next = ddsrt_avl_lookup_succ_eq(&pending_match_index_treedef, &index->pending_matches, &match->guids);
+      next = ddsrt_avl_lookup_succ(&pending_match_index_treedef, &index->pending_matches, &match->guids);
       unregister_and_free_pending_match(sc, match);
     }
     match = next;
--- a/src/core/ddsi/src/ddsi_security_omg.c
+++ b/src/core/ddsi/src/ddsi_security_omg.c
@@ -48,6 +48,9 @@
 #define AC_NAME "Access Control"
 #define CRYPTO_NAME "Cryptographic"
 
+/* TODO: This constant which determines the time pending matches are maintained
+ * and not used should be made a configurable parameter,
+ */
 #define PENDING_MATCH_EXPIRY_TIME 300
 
 #define EXCEPTION_LOG(sc,e,cat, ...) \
@@ -159,6 +162,20 @@ struct participant_sec_index {
   ddsrt_avl_ctree_t participants;
 };
 
+/* The pending _match_index uses an avl tree to store pending_match's where the
+ * guid_pair is used as the key. The remote_guid is the primary key and
+ * the local_guid is the secondary key. The use of the remote_guid as the primary key
+ * is used in the function clear_pending_matches_by_remote_guid to clear the
+ * pending matches associated with a remote entity.
+ *
+ * The table containing the pending matches is protected by the pending_match_index:lock.
+ * It is allowed to access the fields (crypto_handle and tokens) of a pending_match outside
+ * the pending_match_index:lock provided that the pending_match is protected by the
+ * lock of the entity corresponding to the local_guid.
+ * A pending_match is either created when registering and matching an remote entity and
+ * the corresponding crypto tokens are not available or when the crypto tokens associated
+ * with a remote entity are received but it has not yet been discovered.
+ */
 struct guid_pair {
   ddsi_guid_t remote_guid;
   ddsi_guid_t local_guid;
@@ -321,6 +338,7 @@ static struct pending_match * find_or_create_pending_entity_match(struct pending
     match->tokens = NULL;
     match->guids = guids;
     match->kind = kind;
+    match->expiry = DDSRT_MTIME_NEVER;
     ddsrt_avl_insert_ipath(&pending_match_index_treedef, &index->pending_matches, match, &ipath);
   }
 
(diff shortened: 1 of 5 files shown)


src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -258,97 +296,162 @@ void q_omg_log_exception(const struct ddsrt_log_cfg *lc, uint32_t cat, DDS_Secur
...
-  if (match)
-    ddsrt_avl_delete_dpath(&entity_match_treedef, &list->matches, match, &path);
-  ddsrt_mutex_unlock(&list->lock);
-
-  return match;
+  ddsrt_mutex_lock(&index->lock);
+  match = ddsrt_avl_find_min(&pending_match_index_treedef, &index->pending_matches);
+  while (match)
+  {
+    struct pending_match *next = ddsrt_avl_find_succ(&pending_match_index_treedef, &index->pending_matches, match);
```
eboasson, line 414: I'm not sure whether the benefit would be worth the trouble, but you could do better than scanning all entries, you can always skip forward to the next entry that with a greater remote guid and a local guid greater-than-or-equal to.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -258,97 +296,162 @@ void q_omg_log_exception(const struct ddsrt_log_cfg *lc, uint32_t cat, DDS_Secur
...
-  struct guid_pair guids;
-
-  guids.src = *src;
-  guids.dst = *dst;
+  struct pending_match *found;
+  ddsrt_avl_dpath_t dpath;
 
-  return ddsrt_avl_lookup(&entity_match_treedef, &list->matches, &guids);
+  ddsrt_mutex_lock(&index->lock);
+  if ((found = ddsrt_avl_lookup_dpath(&pending_match_index_treedef, &index->pending_matches, &match->guids, &dpath)) != NULL)
```
eboasson, line 380: This lookup bothers me. If the locking and allocation/freeing strategies are sound it appears there should be no possibility of it not being in the index. I can see ``pending_match_expiry_cb`` possibly occurring just before locking ``index->lock``, but then that really means the caller's activities immediately preceding this function are a potential use-after-free.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -258,97 +296,162 @@ void q_omg_log_exception(const struct ddsrt_log_cfg *lc, uint32_t cat, DDS_Secur
...
+    ddsrt_avl_insert_ipath(&pending_match_index_treedef, &index->pending_matches, match, &ipath);
+  }
+
+  if (crypto_handle)
+    match->crypto_handle = crypto_handle;
+
+  if (tokens)
+  {
+    match->tokens = tokens;
+    match->expiry = ddsrt_mtime_add_duration(ddsrt_time_monotonic(), DDS_SECS(PENDING_MATCH_EXPIRY_TIME));
```
eboasson, line 333: I'm missing a ``ddsrt_fibheap_insert`` call here ...

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -2334,73 +2441,83 @@ static bool q_omg_security_register_remote_reader_match(struct proxy_reader *prd
...
   bool send_tokens = false;
+  bool allowed = false;
 
-  *crypto_handle = 0;
-
-  if ((pm = get_pp_proxypp_match_if_authenticated(pp, proxypp, prd->e.guid.entityid)) == NULL)
+  if ((proxypp_match = get_pp_proxypp_match_if_authenticated(pp, proxypp, prd->e.guid.entityid)) == NULL)
     return false;
 
   /* TODO: the security_entity_match should be removed after the the received tokens are stored in the plugin.
```
eboasson, line 2452: Given that this "TODO" disappeared in ``q_omg_security_register_remote_writer_match`` I'm inclined to think it was meant to be removed here as well.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -2334,73 +2441,83 @@ static bool q_omg_security_register_remote_reader_match(struct proxy_reader *prd
...
-          match->tokens = NULL;;
-          GVTRACE("match_remote_reader "PGUIDFMT" with writer "PGUIDFMT": tokens available\n", PGUID(prd->e.guid), PGUID(wr->e.guid));
+          if (sc->crypto_context->crypto_key_exchange->set_remote_datareader_crypto_tokens(
+              sc->crypto_context->crypto_key_exchange, wr->sec_attr->crypto_handle, *crypto_handle, pending_match->tokens, &exception))
+          {
+            GVTRACE(" match_remote_reader "PGUIDFMT" with writer "PGUIDFMT": tokens available\n", PGUID(prd->e.guid), PGUID(wr->e.guid));
+            allowed = true;
+          }
+          else
+            EXCEPTION_ERROR(sc, &exception, "Failed to set remote reader crypto tokens "PGUIDFMT" --> "PGUIDFMT, PGUID(prd->e.guid), PGUID(wr->e.guid));
```
eboasson, line 2500: Is ``remove_pending_match`` really the correct error handling if this error occurs? I'm just asking: it might be, but it isn't obvious.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -48,6 +48,8 @@
 #define AC_NAME "Access Control"
 #define CRYPTO_NAME "Cryptographic"
 
+#define PENDING_MATCH_EXPIRY_TIME 300
```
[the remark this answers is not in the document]
MarcelJordense: For the moment left this as is. Added a comment to make it configurable later.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -258,97 +296,162 @@ void q_omg_log_exception(const struct ddsrt_log_cfg *lc, uint32_t cat, DDS_Secur
...
-  entity_match_free(match);
+  ddsrt_mutex_init(&index->lock);
+  ddsrt_avl_init(&pending_match_index_treedef, &index->pending_matches);
+  ddsrt_fibheap_init(&pending_match_expiry_fhdef, &index->expiry_timers);
+  index->gv = gv;
+  index->evt = NULL;
 }
 
-static void security_match_index_free(struct dds_security_match_index *list)
+static void pending_match_index_deinit(struct pending_match_index *index)
```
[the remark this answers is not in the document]
MarcelJordense: Had to add an extra deinit function of the security context to implement this. This because the security context can only be freed when the xevent queue has been freed before. Callbacks scheduled on the xevent queue could still access the security context.

src/core/ddsi/src/ddsi_security_omg.c
MarcelJordense: Implemented you suggestion. Although I would expect that the pending_match_index is mostly empty.

src/core/ddsi/src/ddsi_security_omg.c
MarcelJordense: The problem is what to do when setting the crypto tokens fails. When that happens there is no way to recover from this. The situation will not change. Thus for the moment only a warning could be issued and the pending tokens be freed.

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -415,7 +441,11 @@ static void clear_pending_matches_by_local_guid(dds_security_context *sc, struct
     if (guid_compare(&match->guids.local_guid, local_guid) == 0)
     {
       ddsrt_avl_delete(&pending_match_index_treedef, &index->pending_matches, match);
-      clear_and_free_pending_match(sc, match);
+      if (match->expiry.v != DDS_NEVER)
+        ddsrt_fibheap_delete(&pending_match_expiry_fhdef, &index->expiry_timers, match);
+      inc_guid(&match->guids.remote_guid);
+      next = ddsrt_avl_lookup_succ_eq(&pending_match_index_treedef, &index->pending_matches, &match->guids);
```
eboasson, line 447: Doing this instead:
```suggestion
      next = ddsrt_avl_lookup_succ(&pending_match_index_treedef, &index->pending_matches, &match->guids);
```
saves you from having to define ``inc_guid``

src/core/ddsi/src/q_init.c
```diff
@@ -1200,7 +1200,7 @@ int rtps_init (struct ddsi_domaingv *gv)
   add_property_to_xqos(&gv->builtin_volatile_xqos_rd, "dds.sec.builtin_endpoint_name", "BuiltinParticipantVolatileMessageSecureReader");
   add_property_to_xqos(&gv->builtin_volatile_xqos_wr, "dds.sec.builtin_endpoint_name", "BuiltinParticipantVolatileMessageSecureWriter");
 
-  q_omg_security_init(gv);
+//  q_omg_security_init(gv);
```
eboasson, line 1203: This line would be best be removed

src/core/ddsi/src/ddsi_security_omg.c
MarcelJordense: Thanks, will use your suggestion.

eboasson:
Merged via #486 after rebasing. Thanks @MarcelJordense!

Issue #396, opened 2020-02-12
Breaks when configured for a network interface setup for DHCP


nikechristo:
Hi,

The config xml allows you to set the network interface address - `NetworkInterfaceAddress`. What I noticed is that if the interface's IP changes, say during a DHCP lease renewal, the DDS layer breaks.

In my case, I had two nodes setup, one with static IP of 172.20.1.1 and another on DHCP with its initial IP being 172.20.1.2. At the time of renewal, the IP on node 2 changed from 172.20.1.2 to 172.20.1.4. If you look at the screenshot below, node 2 was still identifying itself as being on 172.20.1.2 even though its new IP is 172.20.1.4. The multicast traffic was able to make it through so node 1 was still receiving some data.

![wireshark](https://user-images.githubusercontent.com/880324/74384778-0ae41200-4dc0-11ea-9ffd-ea7f6344ea70.png)

Are there any plans to add support for DHCP interfaces?

Thanks.

eboasson:
@nikechristo I would say it is certainly desirable — the difficulty is that many things are desirable and time is limited, so I can't promise much other than saying that it is on the wish list and that protocol and implementation in principle can support it without too much trouble.

Should anyone reading this feel like giving it a shot (PRs are always welcome!) ... at first sight, one would have to:

* Detect the address change (obviously!) and adjust the domain-global "ownip" and list of known interfaces accordingly.
* Publish updated discovery messages. All outgoing SPDP (participant discovery) messages have the unicast address embedded in them. Currently there is no way to override the unicast address for an individual endpoint while using UDP, so I think there would be no need for updating the SEDP (endpoint discovery) messages.
* On receipt of an SPDP message, the corresponding ``proxy_participant``'s address set must be updated.
* The endpoints inherit the unicast locator from the participant if they don't provide one in their own SEDP messages, and as Cyclone stores these explicitly in a ``proxy_reader``/``proxy_writer`` address set, the change needs to be propagated. Presumably it would be necessary to track whether an address is inherited from the participant or not to this propagation, but with that addition, it would appear to be quite straightforward.

I have no idea how much work would be involved in the first point, because I don't actually know how to monitor an interface for address changes (I guess one could always poll). The others seem quite straightforward to me, but might well be caused by my knowing the code fairly well.

stkimmer:
Point one regarding the address change detection, is not handled within #471

PR #471, opened 2020-04-03
Add IP adress change notification

stkimmer:
This add a function that can be used to register
callback functions to monitor the change of an
IP address of a specified network interface.
It is implemented for Linux and Windows 64.
A new test executable was added to allow testing
of the functionallity on dynamically added dummy
network interfaces. This test excecutable must be
run with elevated rights.

Signed-off-by: Stefan Kimmer <skimmer@s2e-systems.com>

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_radmin.h
+++ b/src/core/ddsi/include/dds/ddsi/q_radmin.h
@@ -102,6 +102,13 @@ DDSRT_STATIC_ASSERT (sizeof (struct nn_rmsg) == offsetof (struct nn_rmsg, chunk)
 #define NN_RMSG_PAYLOAD(m) ((unsigned char *) (m + 1))
 #define NN_RMSG_PAYLOADOFF(m, o) (NN_RMSG_PAYLOAD (m) + (o))
 
+/* Align rmsg chunks to the larger of sizeof(void*) or 8.
+
+Ideally, we would use C11's alignof(struct nn_rmsg); however, to avoid dependency on C11,
+we ensure rmsg chunks are at least aligned to sizeof(void *) or 8,
+whichever is larger. */
+#define ALIGNOF_RMSG (sizeof(void *) > 8 ? sizeof(void *) : 8)
+
 struct receiver_state {
   ddsi_guid_prefix_t src_guid_prefix;       /* 12 */
   ddsi_guid_prefix_t dst_guid_prefix;       /* 12 */
--- a/src/core/ddsi/src/q_radmin.c
+++ b/src/core/ddsi/src/q_radmin.c
@@ -303,9 +303,11 @@ static void nn_rbuf_release (struct nn_rbuf *rbuf);
 #define RMSGTRACE(...)              TRACE_CFG (rmsg, rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 #define RDATATRACE(rdata, ...)      TRACE_CFG ((rdata)->rmsg, (rdata)->rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 
-static uint32_t align8uint32 (uint32_t x)
+static uint32_t align_rmsg (uint32_t x)
 {
-  return (x + 7u) & (uint32_t)-8;
+  x += (uint32_t) ALIGNOF_RMSG - 1;
+  x -= x % (uint32_t) ALIGNOF_RMSG;
+  return x;
 }
 
 #ifndef NDEBUG
@@ -553,15 +555,15 @@ struct nn_rmsg *nn_rmsg_new (struct nn_rbufpool *rbp)
 
 void nn_rmsg_setsize (struct nn_rmsg *rmsg, uint32_t size)
 {
-  uint32_t size8 = align8uint32 (size);
-  RMSGTRACE ("rmsg_setsize(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8);
+  uint32_t size8P = align_rmsg (size);
+  RMSGTRACE ("rmsg_setsize(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8P);
   ASSERT_RBUFPOOL_OWNER (rmsg->chunk.rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
   assert (ddsrt_atomic_ld32 (&rmsg->refcount) == RMSG_REFCOUNT_UNCOMMITTED_BIAS);
   assert (rmsg->chunk.u.size == 0);
-  assert (size8 <= rmsg->chunk.rbuf->max_rmsg_size);
+  assert (size8P <= rmsg->chunk.rbuf->max_rmsg_size);
   assert (rmsg->lastchunk == &rmsg->chunk);
-  rmsg->chunk.u.size = size8;
+  rmsg->chunk.u.size = size8P;
 #if USE_VALGRIND
   VALGRIND_MEMPOOL_CHANGE (rmsg->chunk.rbuf->rbufpool, rmsg, rmsg, offsetof (struct nn_rmsg, chunk.u.payload) + rmsg->chunk.size);
 #endif
@@ -620,7 +622,7 @@ void nn_rmsg_commit (struct nn_rmsg *rmsg)
   ASSERT_RBUFPOOL_OWNER (chunk->rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
   assert (chunk->u.size <= chunk->rbuf->max_rmsg_size);
-  assert ((chunk->u.size % 8) == 0);
+  assert ((chunk->u.size % ALIGNOF_RMSG) == 0);
   assert (ddsrt_atomic_ld32 (&rmsg->refcount) >= RMSG_REFCOUNT_UNCOMMITTED_BIAS);
   assert (ddsrt_atomic_ld32 (&rmsg->chunk.rbuf->n_live_rmsg_chunks) > 0);
   assert (ddsrt_atomic_ld32 (&chunk->rbuf->n_live_rmsg_chunks) > 0);
@@ -677,15 +679,15 @@ void *nn_rmsg_alloc (struct nn_rmsg *rmsg, uint32_t size)
 {
   struct nn_rmsg_chunk *chunk = rmsg->lastchunk;
   struct nn_rbuf *rbuf = chunk->rbuf;
-  uint32_t size8 = align8uint32 (size);
+  uint32_t size8P = align_rmsg (size);
   void *ptr;
-  RMSGTRACE ("rmsg_alloc(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8);
+  RMSGTRACE ("rmsg_alloc(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8P);
   ASSERT_RBUFPOOL_OWNER (rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
-  assert ((chunk->u.size % 8) == 0);
-  assert (size8 <= rbuf->max_rmsg_size);
+  assert ((chunk->u.size % ALIGNOF_RMSG) == 0);
+  assert (size8P <= rbuf->max_rmsg_size);
 
-  if (chunk->u.size + size8 > rbuf->max_rmsg_size)
+  if (chunk->u.size + size8P > rbuf->max_rmsg_size)
   {
     struct nn_rbufpool *rbp = rbuf->rbufpool;
     struct nn_rmsg_chunk *newchunk;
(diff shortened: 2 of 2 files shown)


stkimmer:
Note 1; This Pull request skips tests on Windows 32bit version. This is necessary since the tests require the installation of a network interface. This is only possible on native 32bit and 64bit Windows versions. Travis is 32bit is running on 64bit which is prohibited by the API (ERROR_IN_WOW64). 
Note 2: IPv6 is not handled yet.

src/ddsrt/src/ifaddrs/windows/ifaddrs.c
```diff
@@ -242,7 +242,7 @@ copyaddr(
         }
       }
 
-      assert(i < addrtable->dwNumEntries);
+      //assert(i < addrtable->dwNumEntries);
```
stkimmer, line 245: In Windows, this assertion triggers if the interface has no IP address assigned to it. If the for example the following command is executed:
`netsh interface ip set address name="Ethernet 4" static 172.19.5.2/24`
It first deletes the address before it adds the new one. Hence there is a short moment when the network interface has no IP assigned.

k0ekk0ek:
Hi @stkimmer. I had a little peek at this PR (not nearly enough to call it an actual review). I really like the idea of adding tests and appreciate the fact that you did go through the trouble of adding them, but I think adding network interfaces isn't something we should do in unit tests. The Linux version uses `sudo` and while it's a bit uncertain what qualifies as a unit test and what doesn't I think we can agree that having a separate process acquiring root privileges means it's probably better done elsewhere(?) I'm not sure what the best way is in this case, but my opinion would be that it's to tricky. Perhaps we would be better served by just verifying the internal mechanism when the rest of the functionality is done(?)

Another thing I noticed is that only IPv4 is covered. I know IPv6 isn't widely deployed, I'm not even sure if we'll ever get there, but I'm not sure it's ok to not implement for it here. After all, the DDSI stack is able to cope with it.

Lastly, one thing I'm not sure about is spawning another thread. I saw some variables named `osres`, so this might be code that originated in OpenSplice(?). There it would make sense to do it this way. With Eclipse Cyclone DDS there might be a somewhat nicer way. os_sockWaitset comes to mind, but event loops can be found in other places as well. Of course, other people are more qualified to provide input here.

Consider the above ideas that may or may not be of value to this PR.

Again, really like the functionality! I'll try to have a real look, but I'm a little pressed for time at the moment.

stkimmer:
Thanks @k0ekk0ek for the peek at the code. I basically agree on all your points, however some remarks:
- To have the tests that include the creation of a network interface is essential to develop this functionality. So in my opinion the code must be kept somewhere. If its not at the unit tests, where could it be placed?
- IPv6 is indeed not covered. Should be definitely be added, but in another PR
- To spawn another thread is indeed questionable. It could be done by another "maintenance" thread. But I did not know which one.

k0ekk0ek:
Regarding the test, completely agree. There's no one answer. For `rhc_torture` we added a pragmatic `xtests` directory. The questionable bit here is adding interfaces and devices on the fly. I see `25.12.0.1` being used, which is a valid address. One would expect at least a (documentation) [reserved address](https://en.wikipedia.org/wiki/Reserved_IP_addresses) being used. But even then, users might use such a range locally, leading to problems. The real test therefore needs user input(?) I guess I'd make an `xtests` directory in `ddsrt`, place it there and not run it automatically. The part of the code that delivers the trigger can be tested in a unit test perhaps(?)

stkimmer:
> 
> 
> Regarding the test, completely agree. There's no one answer. For `rhc_torture` we added a pragmatic `xtests` directory. The questionable bit here is adding interfaces and devices on the fly. I see `25.12.0.1` being used, which is a valid address. One would expect at least a (documentation) [reserved address](https://en.wikipedia.org/wiki/Reserved_IP_addresses) being used. But even then, users might use such a range locally, leading to problems. The real test therefore needs user input(?) I guess I'd make an `xtests` directory in `ddsrt`, place it there and not run it automatically. The part of the code that delivers the trigger can be tested in a unit test perhaps(?)

Very good point. We will change the used IP addresses. Also to use the xtests directory sounds like a good idea to me. We would change also that, after a review of this PR.

stkimmer:
@eboasson In which thread/monitoring system could the IP address chaage notification be living instead of creating an own thread?

eboasson:
The best way would be to have a good look at the "socket waitset", generalize it a bit and integrate the triggering in there.

* on Linux (and I'm sure also on under Unix variants) it is all events on file descriptors, so that fits well;
* on Windows it is a bit trickier in that it currently uses "select" which can only handle sockets. It has code to use WaitForMultipleEvents that isn't actually in use because in some configurations there may be too many sockets for a call to WFME. The common case of a fixed set of sockets with UDP doesn't have that problem, so in the common case it might well work without requiring an additional theread.

Refactoring that q_sockWaitset.c and putting it in `ddsrt` is something that's long overdue ... it shouldn't be hard, but I'm sure they'll be some little snags here and there ...

stkimmer:
Thanks @eboasson for the reply. 
To be able to finish this PR in a finite time we would suggest:

1. We refactor the code such that it can be integrated in a waitset easily. As such we remove the thread creation from the public interface and use a thread only to test the functionality. We will not refactor the socket waitset in this PR as it would mix up too many things in one go.

2. On top of that we move the tests to a xtests directory and change the IP addresses there to use reserved documentation addresses. 

3. IPv6 need to be implemented in another PR as well

k0ekk0ek:
Hi @stkimmer. Don't know how @eboasson feels about this, but IMHO pulling any of these changes before they work as desired doesn't make much sense. You probably want to finish pieces of work, which is understandable, but please try to factor in the perspective from the project as well. I don't feel comfortable pulling anything into master that isn't finished. Perhaps we can create a separate branch so that we can accept the work in logical pieces and merge it to master once it's finished? Would that work for you?

My proposal would then be:
1. We create a branch once the tests are in order.
2. Add support for IPv6 but keep the separate threads for now.
3. Move `os_sockWaitset` to `ddsrt`. Basically add an event mechanism in `ddsrt` (correct me if I'm wrong @eboasson).
4. Move away from the separate thread and hook it up to mechanism formerly known as `os_sockWaitset`.

Before pulling the finished work into master we'd have to ensure the mechanism works with FreeRTOS too, but I'm fine with you focussing on Linux (epoll?), macOS (kqueue, like FreeBSD) and Windows (WaitForMultipleEvents?). This could potentially give us performance improvements on Linux as well as select is still used there.

stkimmer:
Hi @k0ekk0ek thanks for your quick reply. I can certainly understand your concerns especially given the complicated and sensitive nature of handling network involving multiple OSes. Though, we feel that a slightly different route than you propose might be better for this specific task.

The changes in this PR are working as desired: An IP change can be detected and the system can be notified as desired. The tests do also work and verify its functionality. In the current PR, the implementation of the network change detection is done by registering a callback function and having a thread running the checks (specific for each OS) and executing the callback function where appropriate. On Windows, the detection is done using the NotifyAddrChange and WaitForSingleObject API. On Linux the detection is not based on a file descriptor but rather on reading messages sent by the kernel on a PF_NETLINK socket. The unit tests show that this works for different scenarios (add IP, modify IP, no change, ...). If I understand correctly yours and @eboasson previous messages the main concern here is to have this function running on a separate thread rather than as a waitset.

As such and in the scope of this PR, we would propose modifying the interface to work as a waitset and adjust the unit tests accordingly and, as you suggested before, we would move the unit tests to a different folder (to avoid the permission issues) and adjust the IP adresses to use some meaningful reserved numbers. This would result in a ddsrt API that is functional and tested for detecting IPv4 changes for Windows and Linux and since it is not used in the actual code it should cause no unintended harm.

The rework of the q_sockwaitset seems to be quite a task in itself so we see it as a separate PR as well as the support for detecting IP changes on additional platforms and IPv6. This way things can be done in a more incremental approach which we hope reduces the risk to the project. In this sense we think that a new branch is not required but if that is your prefered route that is of course fine for us.

k0ekk0ek:
Hi @stkimmer.

The changes may very well be "working" and your efforts are appreciated, but I cannot pull this as is.

Firstly, please note that nothing is pulled into master before it's finished. This process is somewhat unofficial, but it has worked very well for Eclipse Cyclone DDS so far. The `security` branch is a good example of this practice.

You're free to develop this in your own repository and submit a PR when everything is done, but like you said, it may be easier for reviewers and yourself if work can be accepted in chunks on a separate branch.

Some prelimary ramblings.

1. The tests shuffle interfaces around on a live system and use `sudo` to acquire root privileges. IMHO it is unacceptable to do so in an automated fashion. I'm not sure whether CTest supports something like conditionally disabling tests, I'm thinking environment variables and soft-failing or something if they're not set. I'd require all variables to be specified. Maybe adding the interface can even be done with a separate executable, that'd be really useful to have while testing and seems like a small task. With requiring environment variables, we can configure something to run in an automated fashion on Travis, but users won't be surprised when running locally. These are mostly just random thoughts... Feel free to come up with something that works convenient for you. That being said, I will require extra checks. e.g. creating an interface may fail because it exists, but the function is `void` and `delete_if` is happily called at the end either way.

2. It's much better to detect whether the system supports adding an interface or not for the Windows case from the test code. For example, `src/ddsrt/tests/ifaddrs.c` does some checks too to see if IPv6 is available at runtime on Travis.

3. A socket is just a file descriptor. I'm guessing they made it a socket specifically so you can use it in your program's event loop(?)

4. If something isn't supported on a specific platform, there should be a feature macro. Pulling this now may "work" for macOS, but breaks FreeRTOS. A macro like DDSRT_HAVE_IPV6, but for this feature is the way to go. That way, we can test if a feature is supported in the upper layers and we can ditch the macOS dummy code. That being said, I think it's rather nice to at least have this for all "tier-1" platforms.

5. I very much dislike the way a thread is simply created and joined and a thread per interface name seems like wasting resources.

That's it for the preliminary comments for now.

A quick glance at the Linux code shows the `PF_NETLINK` socket isn't cleaned up if thread creation fails and the thread that's supposed to be notified loops as opposed to being notified like with WaitForSingleObject on Windows.

Before I'll consider doing a proper review (this is my spare time after all), you should probably decide how everything is going to fit together. That is really the primary concern that I (and probably @eboasson too) have. DDSI will have to be informed that something changed at some point, be it by the kernel or an intermediate thread writing a message to a pipe via a callback. The `os_sockWaitset` is a good starting point, so moving that down to `ddsrt` is the best (and quickest) way to go. It's basically the waitset interface that you describe. Also, it's something that can be submitted to master directly, the address-changed event can be bolted on later (perhaps even as a PR per platform and possibly directly on master if so desired and it makes sense).

I'm sorry for the amount of text and the fact that I cannot pull as is.

stkimmer:
Thanks a lot again for this detailed feedback @k0ekk0ek,This PR is certainly not pull-able at this stage for reasons you and @eboasson mentioned: 
1. the tests are run by default while need super user rights
2. the static IP addresses used are not from the documentation range
3. there is an additional thread running to be able to react on an IP change 

This points will be addressed within this PR in forth coming commits. 
This PR is part of getting issue #396 (DHCP reconfiguration) handled. This is in my point of view a very desired and distinguishing feature of a DDS implementation. To be able to get such feature into CycloneDDS some efforts have to be taken that may not be "complete" but working.
Would you suggest having all of the points mentioned in issue #396 within one PR? We carefully tried to get working functionality into CycloneDDS that is tested and working that does not break the existing system with this PR. 

Regarding your points:
1. (sudo rights) The tests will be moved out of forced test runs (to a xtest). The current tests have enabled the development of the IP change detection and have thereby discovered corner cases and side effects. As such I would rather keep the the tests and provide them to the community.
- To use an extra executable was indeed another option to reach the goal of testing the functionality if an IP address change can be detected. The current implementation did choose do to the testing without an extra executable to avoid keeping track of the the executable in the build system. If there is a need for it, we can of course change it to do so.
- Do the functions in a test need to be that explicit/restrictive to check for interfaces already being existing? It is a test for testing if an IP address change can be detected after all. I feel the tests should not be taken ad absurdum until they get more complex then the software itself. 

2. (iface creation caps) That would be certainly much better (may be in a later PR?)

3. (sockets are fds) To which event loop has it to go? Also: on Windows, the IP address change detection is not bound to a specific socket.

4. (Non supported Macro) It will not work for MacOS and FreeRTOS (since its not implemented for it yet). Certainly its nice to have it for all platforms especially tier-1 platforms. Could you point to more details (or example) on how such Macro should look like?

5. (wasteful thread creation) The own thread will be removed.

The question is how to get an IP address change detection into CycloneDDS while keeping its high performance standards up. This PR is meant to give the first steps to it in Windows and Linux without breaking the code, and that it does provide the opportunity for further development on that topic. If too much stuff ends up in one PR it will be hard to find enough time for a review and for developers to find time to continue to include all possible features (like other OSes).
And as such which changes should we make to the code that this PR touched to get the issue #396 (DHCP configured network interface) solved?

k0ekk0ek:
I'm happy to accept smaller chunks, but they must add something. If I leave out thread-creation, there's little left to pull and what is left is very likely to change because it's unclear how changes will be picked up by DDSI.

Let's forget about all the details above and get this on track.

`os_sockWaitset` is used by the DDSI recv thread to wait for incoming packets (it even does a tiny bit of connection handling already). That's the thread that should notice changes in connections. A possible way to handle changes there is to add an event type as its main-loop iterates over events already (`os_sockWaitsetNextEvent`). On Linux, select is used and thus can accept your socket too, on Windows, I think currently select is used, but in order to use it for this too, you'll need to use WaitForMultipleEvents.

I'd personally start there.

Please read `docs/dev/modules.md` for details on feature discovery.

PR #472, opened 2020-04-05
Prevent time-out in handshake

dennis-adlink:
The security handshake is started when a node receives an SPDP message. The SPDP receiver will reply with an SPDP, followed by a `dds.sec.auth_request`. Because the initial SPDP sender will receive the `auth_request` immediately after (or even before) the SPDP reply message, that node may not have finished (or not even started) matching the remote writers and therefore it drops the `auth_request` message. This results in a time-out in the handshake process, and the `auth_request` has to be re-send. To avoid this, a short (rather arbitrarily chosen, based on local testing) sleep is introduced before the `auth_request` message is sent.

### DIFF
--- a/src/core/ddsi/src/ddsi_handshake.c
+++ b/src/core/ddsi/src/ddsi_handshake.c
@@ -143,18 +143,20 @@ static bool validate_handshake(struct ddsi_handshake *handshake, struct particip
 #define AUTHENTICATION_TIMEOUT  DDS_SECS(100)
 #define INITIAL_DELAY           DDS_MSECS(10)
 
-static void func_validate_remote_identity     (struct dds_security_fsm *fsm, void *arg);
-static void func_handshake_init_message_resend(struct dds_security_fsm *fsm, void *arg);
-static void func_begin_handshake_reply        (struct dds_security_fsm *fsm, void *arg);
-static void func_begin_handshake_request      (struct dds_security_fsm *fsm, void *arg);
-static void func_process_handshake            (struct dds_security_fsm *fsm, void *arg);
-static void func_handshake_message_resend     (struct dds_security_fsm *fsm, void *arg);
-static void func_validation_ok                (struct dds_security_fsm *fsm, void *arg);
-static void func_validation_failed            (struct dds_security_fsm *fsm, void *arg);
-static void func_send_crypto_tokens_final     (struct dds_security_fsm *fsm, void *arg);
-static void func_send_crypto_tokens           (struct dds_security_fsm *fsm, void *arg);
+static void func_validate_remote_and_begin_reply (struct dds_security_fsm *fsm, void *arg);
+static void func_validate_remote_identity        (struct dds_security_fsm *fsm, void *arg);
+static void func_handshake_init_message_resend   (struct dds_security_fsm *fsm, void *arg);
+static void func_begin_handshake_reply           (struct dds_security_fsm *fsm, void *arg);
+static void func_begin_handshake_request         (struct dds_security_fsm *fsm, void *arg);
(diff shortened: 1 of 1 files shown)


src/core/ddsi/src/ddsi_handshake.c
```diff
@@ -553,7 +553,15 @@ static void func_validate_remote_identity(struct dds_security_fsm *fsm, void *ar
    * to be send.
    */
   if (handshake->local_auth_request_token.class_id && strlen(handshake->local_auth_request_token.class_id) != 0)
+  {
+    /* A short sleep to give the remote node some time for matching the
+       BuiltinParticipantStatelessMessageWriter, so that it won't drop the
+       sample we're sending (that would result in a time-out and the
+       handshake taking 1s longer than required) */
+    dds_sleepfor (DDS_MSECS(10));
```
eboasson, line 561: Ouch, an avoidable sleep in a thread that is supposed to do work on behalf of many independent entities. Adding a short delay for to cover this is a pragmatic solution, but surely there's a more elegant way to make this happen? Perhaps using an extra delay state in the FSM? That might be more complicated than it sounds, but it is worth exploring. If that proves to be too hard, then we can let this in as a temporary workaround with a comment.

dennis-adlink: Yes, I agree that this is not elegant at all. Although blocking the FSM thread once per handshake won't have too much performance impact in many use cases I think, I will explore the option of adding the extra state in the FSM.

dennis-adlink:
@eboasson in the additional commit I've changed the approach to what you suggested, which (as expected) also successfully eliminates the 1s time-outs in the handshake. I'm not completely sure about the correctness of the transition from `state_initial_delay` on `EVENT_RECEIVED_MESSAGE_REQUEST`, @MarcelJordense could you please verify this?

src/core/ddsi/src/ddsi_handshake.c
```diff
@@ -367,6 +379,17 @@ VALIDATION_FAILED |                                         |
...
+       BuiltinParticipantStatelessMessageWriter, so that it won't drop the auth_request
+       we're sending (that would result in a time-out and the handshake taking longer
+       than required). For the node that receives the auth_request, the transition for
+       the event EVENT_RECEIVED_MESSAGE_REQUEST is added, because that node may receive the
+       auth_request during this delay and can continue immediately (as the sender already
+       waited for this delay before sending the request) */
+    { &state_initial_delay,                 EVENT_TIMEOUT,                              NULL,
+                                            &state_validate_remote_identity                },
+    { &state_initial_delay,                 EVENT_RECEIVED_MESSAGE_REQUEST,             NULL,
                                             &state_validate_remote_identity                },
```
MarcelJordense, line 393: When receiving a handshake request while waiting for the initial timeout to finish it is may be better to have an additional state which handles both the validate_remote_identity and the received handshake request message, because otherwise the handshake request will not be handled and we have to wait for a resent of this request.

dennis-adlink: You're right, the current approach leads to some overhead. In 1c85b868d95c33df9d3acf89fdc0900a68898be2 I've added the additional state that does both `validate_remote_identity` and `begin_handshake_reply`. 

I think that, in case of a transition from `state_initial_delay` to the newly introduced state `state_validate_remote_and_begin_reply` (which means the node received an `auth_request`) the result from `validate_remote_identity` should always be `PENDING_HANDSHAKE_MESSAGE` (or `FAILED` in case something went wrong). So I added a check on this return value and the handshake fails if this is not the case.

eboasson:
Thanks @dennis-adlink and @MarcelJordense !

PR #477, opened 2020-04-07
Fix in initialization of remote participants key material

dennis-adlink:
The initialization of key material for a remote participant was not protected by the `remote_participant_crypto` lock, which could result in using partially initialized remote key material. This caused intermittent test failures with assert on key_size in `crypto_cipher_decrypt_data`. This PR fixes this issue by adding locking for the remote key material initialization.


dennis-adlink:
Closing this PR as the new fix depends on #459. When that PR is merged to security, I'll open a new PR for this issue.

PR #483, opened 2020-04-09
Using GUIDs in the API

eboasson:
Adds ``dds_guid_t`` as a less confusing alias for ``dds_builtintopic_guid_t``. It is alias to retain backwards compatibility. All uses in the sources of Cyclone have been updated.

Adds ``dds_get_guid`` to retrieve the GUID of an participant, reader or writer (the only entities visible in the discovery). This is *much* more convenient (and more efficient) than something like:
```
dds_entity_t x = dds_create_reader(h, matching_builtin_topic, NULL, NULL);
dds_sample_info_t si;
void *raw = NULL;
dds_instance_handle_t ih;
dds_get_instance_handle(h, &ih);
int32_t n;
dds_read_instance(x, &raw, &si, 1, 1, ih);
(e.g. dds_builtintopic_participant_t *)s = raw;
guid = &s->key;
dds_return_loan(x, &raw, 1);
dds_delete(x);
```

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -186,17 +186,21 @@ typedef struct dds_builtintopic_guid
 }
 dds_builtintopic_guid_t;
 
+/* "dds_builtintopic_guid_t" is a bit of a weird name for what everyone just calls a GUID,
+   so let us try and switch to using the more logical one */
+typedef struct dds_builtintopic_guid dds_guid_t;
+
 typedef struct dds_builtintopic_participant
 {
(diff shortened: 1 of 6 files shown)


dennis-adlink:
LGTM. The only thing I could find is that you're not testing getting a guid from invalid entity kinds in the test case, but as the code in `dds_get_guid` is so trivial thats not really worth the effort.

eboasson:
Thanks @dennis-adlink for checking! I agree without about the test ... both in the sense that it really feels like spending time on this particular detail is a waste and in that spending time on improving the tests in that area in general would be a valuable exercise.

PR #489, opened 2020-04-15
Rejigger UDP buffer config options

rotu:
Previous behavior:
Only had minimum UDP buffer sizes, so you either get small buffers or you get an error.

Now:
We request a certain buffer size and use the largest buffer the system will give us up to that size. The minimum only comes into play if the buffers are smaller, in which case an error is logged.


k0ekk0ek:
I'm a little pressed for time... I'll do my best to find time, definitely have to study things related to lwIP, I'm almost certain that's going to break things. @ThijsSassen can you testdrive the changes? I don't have an embedded build setup atm.

ThijsSassen:
@k0ekk0ek sure

rotu:
Now that I understand this code a bit better, I think it makes sense the buffer logic to another function and share it with ddsi_tcp as well.

rotu:
Also, I made some changes like translating `ENOPROTOOPT` (which is often a platform-dependent failure) to `DDS_RETCODE_UNSUPPORTED` rather than `DDS_RETCODE_BAD_PARAMETER`. I'd have to also change the documentation in `sockets.h`. This change probably belongs in a separate PR and I have to take it out of this one.

src/core/ddsi/src/q_config.c
```diff
@@ -534,11 +534,14 @@ static const struct cfgelem internal_cfgelems[] = {
...
-  { LEAF("MinimumSocketReceiveBufferSize"), 1, "default", ABSOFF(socket_min_rcvbuf_size), 0, uf_maybe_memsize, 0, pf_maybe_memsize,
-    BLURB("<p>This setting controls the minimum size of socket receive buffers. The operating system provides some size receive buffer upon creation of the socket, this option can be used to increase the size of the buffer beyond that initially provided by the operating system. If the buffer size cannot be increased to the specified size, an error is reported.</p>\n\
-<p>The default setting is the word \"default\", which means DDSI2E will attempt to increase the buffer size to 1MB, but will silently accept a smaller buffer should that attempt fail.</p>") },
-  { LEAF("MinimumSocketSendBufferSize"), 1, "64 KiB", ABSOFF(socket_min_sndbuf_size), 0, uf_memsize, 0, pf_memsize,
-    BLURB("<p>This setting controls the minimum size of socket send buffers. This setting can only increase the size of the send buffer, if the operating system by default creates a larger buffer, it is left unchanged.</p>") },
+  { LEAF("MinimumSocketReceiveBufferSize"), 1, "64 KiB", ABSOFF(socket_min_rcvbuf_size), 0, uf_maybe_memsize, 0, pf_maybe_memsize,
```
eboasson, line 537: The "maybe_memsize" function sets the "isdefault" flag if the value is "default" and clears it if a number is given, so all these set "isdefault" to false. That's probably not the intent.

Issue #488, opened 2020-04-14
Failing ROS2 comms on QEMU


dodsonmg:
TL;DR//I am trying to troubleshoot what appear to be communication issues running ROS2 on QEMU.  I tried to build and run the cyclonedds examples, to eliminate at least one layer as my problem source, but the Maven dependency kept me from doing so.

---

Background:

I am trying to cross-compile ROS2 demos on a FreeBSD-based research system running on QEMU.  I am using the cyclonedds middleware.

My QEMU is configured to allow communication from guest -> host, but *not* host -> guest.  If I run a simple publisher on the guest, the host can see the node, but can't query it, which makes sense.  What makes less sense (though this may be a conceptual error on my part), the host can't see the topics to which the publisher is publishing.

If I attempt to run a publisher and subscriber separately, but simultaneously, on the guest (to avoid trying to discover and communicate between guest and host), I get the following error upon starting the second process:

```
1586640423.685765 [42]     100051: UDP make_socket failed for multicast port 17900
[ERROR] [1586640423.781729296] [rmw_cyclonedds_cpp]: rmw_create_node: failed to create domain, error Error

>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'error not set, at /home/broomstick/cheri/ros2_min/src/ros2/rcl/rcl/src/rcl/node.c:336'

with this new error message:

  'rcl node's rmw handle is invalid, at /home/broomstick/cheri/ros2_min/src/ros2/rcl/rcl/src/rcl/node.c:485'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
[ERROR] [1586640423.895518056] [rcl]: Failed to fini publisher for node: 1
Terminating due to uncaught exception 0x48960300 of type rclcpp::exceptions::RCLError
Abort
```

When I try to run a composed publisher and subscriber (two ROS2 nodes in the same process), I get a slightly different error.  As you can see, the subscriber, which should simply echo the publisher's message, echos only an empty string.

```
[INFO] [1586660033.578303216] [publisher_node]: Publisher: 'Hello, world! 0'

>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

[...]

PR #490, opened 2020-04-15
add freebsd support

dodsonmg:
Including FreeBSD in preprocessor conditionals for APPLE fixes issues with UDP make_socket, as described in issue #488.

### DIFF
--- a/src/ddsrt/src/sockets/posix/socket.c
+++ b/src/ddsrt/src/sockets/posix/socket.c
@@ -32,9 +32,9 @@
 #include <fcntl.h>
 #endif
 
-#ifdef __APPLE__
+#if defined(__APPLE__) || defined(__FreeBSD__)
 #include <sys/sockio.h>
-#endif /* __APPLE__ */
+#endif /* __APPLE__ || __FreeBSD__ */
 #endif /* LWIP_SOCKET */
 
 dds_return_t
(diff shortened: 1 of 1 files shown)


src/ddsrt/src/sockets/posix/socket.c
```diff
@@ -32,9 +32,9 @@
 #include <fcntl.h>
 #endif
 
-#ifdef __APPLE__
+#ifdef __APPLE__ || __FreeBSD__
```
k0ekk0ek, line 35: This should be changed to: `#if defined(__APPLE__) || defined(__FreeBSD__)`

dodsonmg: Incorporated.  Thanks!

rotu:
I'd like to see this logic removed from `socket.c`. I think it belongs as two different calls to `ddsrt_setsockopt` to set the two options.

rotu:
Here's what I mean:
https://github.com/dodsonmg/cyclonedds/pull/1

k0ekk0ek:
Thanks @dodsonmg!

@rotu, you're absolutely right in that that the socket options can be managed a bit better than is currently the case. You're encouraged to improve the situation (which could be as simple as opening the same PR here :wink:). I'm merging these changes because it makes Eclipse Cyclone DDS work on FreeBSD and it doesn't worsen the situation.

@dodsonmg, you went through the trouble of testing and opening a PR (which is very much appreciated!). I think asking you to solve a bigger problem is a bit much. That being said, if you want to do additional work on this, I'd welcome any improvements. Feel free to do so, don't feel obligated.

rotu:
You're right, @k0ekk0ek, and you worded that very diplomatically. I appreciate it!

@k0ekk0ek thank you for your contribution and I'm sorry if I made your help feel at all unwelcome. You rock!

dodsonmg:
Thanks @k0ekk0ek for the merge.

@rotu I looked at your proposed changes and they're a bit beyond my experience, so I'll have to dig a bit deeper to understand them before proposing them myself.  That said, if you think they're correct and work on other platforms, I'm happy to give them a run, as is, on my FreeBSD platform and confirm they do the trick.

PR #491, opened 2020-04-15
Support docker build for cyclonedds

prasannabhat:
Dockerfile & dockerbuild scripts added, which enable building & testing cyclonedds in docker containers.
With dockers both IPC & network based use cases can be tested in the same machine

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+
+# It is assumed that this script is called from the root of the project
+WORKSPACE=$(pwd)
+echo "WORKSPACE is $WORKSPACE"
+
+CYCLONEDDS_INSTALL_PREFIX=$WORKSPACE/build/install/
+
+rm -rf build/
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=$CYCLONEDDS_INSTALL_PREFIX -DBUILD_IDLC=ON ..
+cmake --build .
+cmake --build . --target install
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -6,9 +6,46 @@ WORKSPACE=$(pwd)
 echo "WORKSPACE is $WORKSPACE"
 
 CYCLONEDDS_INSTALL_PREFIX=$WORKSPACE/build/install/
+CLEAN_BUILD=false
 
-rm -rf build/
-mkdir build
+usage() {
+	echo ""
+	echo "Helper script to build cyclonedds"
+	echo "Usage : "
+	echo "<workspace>/scripts/build.sh [options]"
+	echo "Supported options are "
+	echo "clean 		: remove build folder (Default : keep build folder)"
+	echo ""
+}
+
+parse_args() {
+	for arg in "$@"
+	do
+		case $arg in
+		    -h | --help)
+			usage
+			exit 0
+			;;
+			clean)
+			CLEAN_BUILD=true
+			;;
+			*)
+			echo "ERROR : unknown parameter $arg"
+			usage
+			exit 1
+			;;
+		esac
+	done
+}
+
+parse_args $@
+
+if [ $CLEAN_BUILD == true ]
+then
+    rm -rf build/
+fi
+
+mkdir -p build
 cd build
 cmake -DCMAKE_INSTALL_PREFIX=$CYCLONEDDS_INSTALL_PREFIX -DBUILD_IDLC=ON ..
 cmake --build .
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,4 +1,15 @@
 #!/bin/bash
+
+# Copyright(c) 2020 Prasanna Bhat
+
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+# v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+
 set -e
 
 # It is assumed that this script is called from the root of the project
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu:bionic
+
+# Dependencies required to build cyclonedds
+RUN apt update && apt install -y \
+        cmake \
+        default-jdk \
+        maven \
+        g++
+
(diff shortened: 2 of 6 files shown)


eboasson:
Thanks @prasannabhat, that's something that I am sure will be appreciated by many people. I've only used docker only rarely and in very limited and the only thing I can say is that it looks fine to me, except for two things:
* I do think it would be good to add copyright/license blurbs to the files that have a comment syntax (like in most other files in the repository);
* I suspect people use a directory named "build" for building, and so running your "scripts/build.sh" would potentially remove a pre-existing directory. There might be some things in there one would like to keep. It might be better to refuse to run if the directory exists unless, e.g., a "force" (-f, I suppose) option is given.

I'll try to find a moment to follow instructions and see what happens 🙂 If there's anyone reading this who has done a lot with docker and feels qualified to review this properly, I'd be very grateful!

k0ekk0ek:
Hi all! Firstly, thanks @prasannabhat, contributions are always greatly appreciated! I had a quick read the other day. Overall, I'm fine with this. I'd like to give it a try and see if it can be used on Windows as well, because Docker is supported there too, but that's something that could potentionally be fixed afterwards. My main concern here is that nobody who contributes to Eclipse Cyclone DDS on a regularly basis has much experience with Docker (to my knowledge). For the foreseeable future that's not an issue, but if at some point things break that would be a serious degradation in user experience, and it would slip by unnoticed. I'm not sure how to solve that issue, so input from others is appreciated here. Maybe we could add a notice in the README or setup a Docker build on Travis or something?

If I find the time, I too, will try the instructions provided and do a bit of reading-up.

eboasson:
@prasannabhat I finally installed ubuntu+docker and gave it a go. The first issue I ran into is that it appears you must build `ubuntu:cyclonedds` in order to be able to build `cyclonedds:latest` which wasn't clear to me from your README (it didn't take long to figure it out because the error message from docker was clear enough).

The second is that:
> - Launch docker container, from the root of the project
> `docker run --name cyclonedds -it -v $(pwd):/cyclonedds --workdir /cyclonedds cyclonedds:latest /bin/bash`
> This will open terminal connected to docker @ root of the project.  
> It will look something like this
> ```
> root@22ff409a33e1:/cyclonedds# 
> ```
> - Navigate to `/build/bin` to run the example applications.

puts me in `/cyclonedds` where the source of cyclone is visible (it looks like it is the actual directory), rather than the built and installed version and there is no `/build`. Removing the `-v $(pwd):/cyclonedds` lands me in the version that includes the build output in a `build` subdirectory. I think it's just a bit of a fiddling with command lines and overlay directories.

I'd appreciate it if you could have a look and update the readme (and add a copyright notice to the scripts, please!) to make sure a copy-paste of the command lines in there works out fine. That's the only major thing stopping me from merging it.

prasannabhat:
@eboasson 
First of all , thanks a lot for trying out the docker stuff and thanks for the detailed feedback.
I updated Read me file with the suggestions.

Earlier you had also mentioned about the default *clean* option in *scripts/build.sh*.
I wrote this file mainly with the intention to use within docker scripts, though it can be used outside as well. I updated this script , now it doesn't remove *build* directory by default. If you want to remove, then pass *clean* as argument.
It can be used to build *cyclonedds* in general. (only systems which support shell script execution)

I am not sure what copyright notice I should add , in case of a contribution from a private individual.
Should it be 
Copyright(c) 2006 to 2018 ADLINK Technology Limited and others 
OR
I mention my name ?
Please guide me here (I am okay with any option, I am unsure which suits here based on licensing etc.)

prasannabhat:
@k0ekk0ek Thanks for the comments.
Docker is supported in Windows , but for Linux container environment.
From their [homepage](https://docs.docker.com/docker-for-windows/install-windows-home/)
*You can now install Docker Desktop on Windows Home machines using the WSL 2 backend. Docker Desktop on Windows Home is a full version of Docker Desktop for Linux container development.*
Technically it should work the same , as it works in Linux host.

I am not sure if docker supports *windows containers* in windows (Multi level virtualization going on in there :-))

Regarding breaking of the docker, as long as cyclonedds doesn't introduce new dependency , it should be fine. (Docker container is built with these dependencies)

At the moment , there is no link to Docker build in the main README file.
It would be good idea to introduce that to let others know of this.

eboasson:
> I am not sure what copyright notice I should add , in case of a contribution from a private individual.
> Should it be
> Copyright(c) 2006 to 2018 ADLINK Technology Limited and others
> OR
> I mention my name ?
> Please guide me here (I am okay with any option, I am unsure which suits here based on licensing etc.)

I'm not a lawyer and the legal details of copyright are intricate and dependent on jurisdiction, so I don't presume to know exactly what to do when. But as I understand it, by default the copyright belongs to the person who authored the file, but employment (and contract work, I suppose) can change that. When you adapt an existing file the adaptations are copyrightable in themselves and so you can then add your name as well. (It might be silly to do that if it is only a trivial adaptation). The Eclipse Handbook gives some more information: https://www.eclipse.org/projects/handbook/#ip-copyright-headers

So for me it is easy (ADLINK); for you it would be your name. The year would simply be 2020.

Actually when I wrote "copyright" I meant "copyright and license reference" but I'm sure you understood that. So basically:

```
#
# Copyright(c) 2020 YOUR NAME
#
# This program and the accompanying materials are made available under the
# terms of the Eclipse Public License v. 2.0 which is available at
# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
# v. 1.0 which is available at
# http://www.eclipse.org/org/documents/edl-v10.php.
#
#
```

In short you're not handing anything over to ADLINK.

prasannabhat:
> Actually when I wrote "copyright" I meant "copyright and license reference" but I'm sure you understood that. So basically:
> 
> ```
> #
> # Copyright(c) 2020 YOUR NAME
> #
> # This program and the accompanying materials are made available under the
> # terms of the Eclipse Public License v. 2.0 which is available at
> # http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
> # v. 1.0 which is available at
> # http://www.eclipse.org/org/documents/edl-v10.php.
> #
> # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
> #
> ```
> 
> In short you're not handing anything over to ADLINK.

Sorry , this slipped from my RADAR for sometime.
Thanks @eboasson 
Copyrights sometimes could get complex as you mentioned, and sometimes take away the joy of s/w development from the developers :neutral_face: (but still necessary)

I added copyright headers, which was standing between this PR's final destination :wink:

eboasson:
> I added copyright headers, which was standing between this PR's final destination 😉

Indeed. Thanks, @prasannabhat!

PR #493, opened 2020-04-15
Respect IEC decimal prefixes

### LINKED ISSUE #492: MiB and MB should not be synonymous.
In the config file, we interpret IEC binary prefixes the same as decimal prefixes. This is somewhat confusing. We should either allow "MB" to refer to the colloquial (though incorrect) meaning of 2<sup>20</sup> bytes and disallow "MiB", or switch "MB" to the correct meaning 10<sup>2</sup> bytes

rotu:
fix #492


eboasson:
There was already some discussion on #492. Those wishing to comment on the wisdom of changing should go there; those who want to comment on the details of the proposed changes probably should be here 🙂.

PR #494, opened 2020-04-16
Use SO_REUSEPORT whenever it's defined instead of sniffing platform

rotu:


k0ekk0ek:
Thanks @rotu. Not really sure about the lwIP part yet, but I'll have a closer look this evening.

rotu:
My thought there is that lwIP should already return ENOPROTOOPT for that case, so it's redundant.

k0ekk0ek:
I recall lwIP having this defined, and the option is recognized, but it maybe disabled... It's all a little confusing because the user can really reduce the implementation to the bare minimum. The exception here is very likely correct, but I can't verify because I currently don't have access to builds with (FreeRTOS+)lwIP. Let's leave this one in(?)

Then, onto `SO_REUSEPORT` and `SO_REUSEADDR`. It just so happens that I had to deal with socket options a lot lately. Surprise, [it's a real mess](https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ). I very much agree with you that calling `setsockopt` is nicer than what's currently there, but in this case it really depends on the platform, or rather, the IP stack, because you can use lwIP on Linux (and Windows, ...).

I would like to propose the following, let's introduce the function like `ddsrt_setreuseaddr` (others may follow, spoiler: there's a SO_SNDBUF and SO_SNDBUFFORCE, like with SO_RCVBUF, don't remember the difference) and call that one from `ddsi_tcp.c` and `ddsi_udp.c`. We describe what the intent of the function is and make the others behave as alike as possible (that's really what it currently does for Apple and FreeBSD, if `SO_REUSEADDR`, then add `SO_REUSEPORT` too). So, the platform check needs to remain, but it may make it a bit more obvious what's going on.

@eboasson, any thoughts/remarks?

k0ekk0ek:
Btw, really sorry for not being able to accept it as-is. I didn't recall if `SO_REUSEPORT` and SO_REUSEADDR` were part of the tricky ones. I'd understand if you don't want to continue work on this...

rotu:
Yes, lwIP has this defined and disabled, which returns ENOPROTOOPT
http://git.savannah.nongnu.org/cgit/lwip.git/tree/src/api/sockets.c#n2958

But you're right it does define SO_REUSEPORT, so I need to change this check:
http://git.savannah.nongnu.org/cgit/lwip.git/tree/src/include/lwip/sockets.h#n219

> Surprise, [it's a real mess](https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ).

It is a mess - totally agreed.

> let's introduce the function

I also agree with factoring this out to a function. The only reason I didn't already is because I didn't know where to put it. I guess ddsrt [is more appropriate](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/modules.md) for smoothing out platform-specific implementation details than ddsi.

> Btw, really sorry for not being able to accept it as-is.

Nothing to apologize for. This does require a little more elbow grease.

eboasson:
Effectively merged as #1029.

PR #495, opened 2020-04-16
Fix in initialization of remote participants key material

dennis-adlink:
The initialization of remote participant's key material was not protected by the `remote_participant_crypto` lock, which could result in using partially initialized remote key material. This caused intermittent test failures with assertions on `key_size` in `crypto_cipher_decrypt_data`. This commit fixes this issue by adding locking for the remote key material.


eboasson:
LGTM! Thanks @dennis-adlink 🙂

PR #496, opened 2020-04-17
Access control tests

dennis-adlink:
Add test cases for the `join_access_control` governance setting and for the access control plugin `check_create_` and `check_remote_` hooks, using a wrapper plugin that simulates failure for each of these, to test the DDSI integration with the access control plugin. This commit also contains:
- replaced assert on DDS_RETCODE_OK in dds_create_reader and dds_create_writer that caused the application to terminate in case creation of a reader or writer is not allowed by security
- fixed not matching a proxy reader that has the `relay_only` set to true, which is currently unsupported
- fixed condition for setting the identity subject in remote rights struct
- consistent return type for plugin init and fini functions

### DIFF
--- a/src/core/ddsi/src/ddsi_security_omg.c
+++ b/src/core/ddsi/src/ddsi_security_omg.c
@@ -2137,7 +2137,6 @@ bool q_omg_security_check_remote_writer_permissions(const struct proxy_writer *p
   DDS_Security_SecurityException exception = DDS_SECURITY_EXCEPTION_INIT;
   DDS_Security_PublicationBuiltinTopicDataSecure publication_data;
   DDS_Security_TopicBuiltinTopicData topic_data;
-  bool result = true;
 
   if (!sc)
     return true;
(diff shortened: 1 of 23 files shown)


dennis-adlink:
@eboasson thanks for the review! I've addressed both issues in 29ac3f0. For the assert on participant in `dds_create_reader` and `dds_create_writer` I agree with your reasoning: the subscriber handle is removed via `delete_children` in `really_delete_pinned_closed_locked`, before `dds_entity_deriver_delete` is called, and from that point (so before the DDSI participant is deleted) it is not possible to pin the publisher/subscriber.

dennis-adlink:
I forgot to rewrite the logic in `q_omg_security_check_remote_reader_permissions` in my previous commit, so that's fixed in 14efd5e

src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -2155,31 +2155,41 @@ bool q_omg_security_check_remote_writer_permissions(const struct proxy_writer *p
...
-    if ((permissions_handle = get_permissions_handle(pp, pwr->c.proxypp)) == 0)
-    {
-      GVTRACE("Secure remote writer "PGUIDFMT" proxypp does not have permissions handle yet\n", PGUID(pwr->e.guid));
-      return false;
-    }
+  q_omg_shallow_copy_PublicationBuiltinTopicDataSecure(&publication_data, &pwr->e.guid, pwr->c.xqos, &pwr->c.security_info);
+  bool result = sc->access_control_context->check_remote_datawriter(sc->access_control_context, permissions_handle, (int)domain_id, &publication_data, &exception);
+  if (!result)
+  {
+    if (!is_topic_discovery_protected(pp->sec_attr->permissions_handle, sc->access_control_context, publication_data.topic_name))
```
eboasson, line 2172: I have one more question: here, as well as a few lines below and in `q_omg_security_check_remote_reader_permissions`, if this function returns true, it silently rejects the reader/writer, else it rejects with an error message. I don't understand why that difference exists.

eboasson:
The failure in the test run (ddsperf failing to complete (mutual) discovery in one run) is a new one to me, but I have checked the changed again insofar as they could have affected this and I don't see a reason why these changes would be the cause of that failure.

src/core/ddsi/src/ddsi_security_omg.c
dennis-adlink: The reason for not logging this is to prevent topic names from being visible in log files, in case discovery protection is enabled for a topic (topic names could be considered as sensitive data in certain cases). In ca2fbe7 I've changed this so that the message is logged in DDS_LC_TRACE when discovery protection is turned on.

eboasson:
Thanks @dennis-adlink !

Issue #497, opened 2020-04-17
Pointer alignment causing traps


dodsonmg:
I'm using a [research system](https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/) with 16 byte pointers.  

The following code snippet traps when `rst` is passed to `memset`.

https://github.com/eclipse-cyclonedds/cyclonedds/blob/f139dbcd5e8bd7d625cc8dfca02aa89e5e2e494e/src/core/ddsi/src/q_receive.c#L2606-L2611

As far as I can tell, `rst` is a pointer to an 8 byte aligned chunk of a buffer, based on the following snippets.

https://github.com/eclipse-cyclonedds/cyclonedds/blob/f139dbcd5e8bd7d625cc8dfca02aa89e5e2e494e/src/core/ddsi/src/q_radmin.c#L306-L309

https://github.com/eclipse-cyclonedds/cyclonedds/blob/f139dbcd5e8bd7d625cc8dfca02aa89e5e2e494e/src/core/ddsi/src/q_radmin.c#L676-L716

Because my system uses 16 byte pointers, and is very picky about alignment, this code combination traps.

I know 16 byte pointers aren't at the top of many people's lists of concerning items, but I was wondering if you would consider a PR to generalise the `align8uint32` function, with sensible renaming of variables that get assigned the return value from this function (they all appear to be named `size8` at the moment).  Something like this:

```
static uint32_t align_literal_uint32 (uint32_t x)
{
  size_t align_size = alignof(void *) > 8 ? alignof(void *) : 8;
  return (x + align_size - 1) & (uint32_t)-align_size;
}
```

This compiles and runs beautifully on my system.

Future proofing?

eboasson:
I really should do a proper review to be sure that it truly is this simple, but if you'll forgive me responding based on my recollection of this part of the code ... there is nothing truly special about the 8-byte alignment: it is simply an approximation of "suitably aligned for any basic type", but it might exploit (or might have exploited) the fact that the lower 3 bits are 0 as well. In any case, 16 byte alignment won't do any harm.

I doubt `alignof` is universally available, in consequence of which I think it would probably be better to take the maximum of `sizeof(void*)` and `8` instead. Anyway, a lovely bit of future proofing!

dodsonmg:
Thanks for the quick response.  I think `sizeof` could work.  I'll verify and submit at PR for more formal review.

dodsonmg:
[...]

PR #498, opened 2020-04-18
align buffer chunks to sizeof(void*) instead of hard coding 8 bytes

dodsonmg:
Currently, buffer chunks are aligned to a hard-coded 8 bytes.  This is causing exceptions on my research system which uses 16 byte pointers.

This PR generalises the alignment to `sizeof(void *)`, which will work on my system and has the side-effect of future-proofing the code segment.

See a longer discussion of where the trap occurs in issue #497.

### DIFF
--- a/src/core/ddsi/src/q_radmin.c
+++ b/src/core/ddsi/src/q_radmin.c
@@ -303,9 +303,11 @@ static void nn_rbuf_release (struct nn_rbuf *rbuf);
 #define RMSGTRACE(...)              TRACE_CFG (rmsg, rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 #define RDATATRACE(rdata, ...)      TRACE_CFG ((rdata)->rmsg, (rdata)->rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 
-static uint32_t align8uint32 (uint32_t x)
+static uint32_t align_rmsg (uint32_t x)
 {
-  return (x + 7u) & (uint32_t)-8;
+  x += (uint32_t) ALIGNOF_RMSG - 1;
+  x -= x % (uint32_t) ALIGNOF_RMSG;
+  return x;
 }
 
 #ifndef NDEBUG
@@ -553,15 +555,15 @@ struct nn_rmsg *nn_rmsg_new (struct nn_rbufpool *rbp)
 
 void nn_rmsg_setsize (struct nn_rmsg *rmsg, uint32_t size)
 {
-  uint32_t size8 = align8uint32 (size);
-  RMSGTRACE ("rmsg_setsize(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8);
+  uint32_t size8P = align_rmsg (size);
+  RMSGTRACE ("rmsg_setsize(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8P);
   ASSERT_RBUFPOOL_OWNER (rmsg->chunk.rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
   assert (ddsrt_atomic_ld32 (&rmsg->refcount) == RMSG_REFCOUNT_UNCOMMITTED_BIAS);
   assert (rmsg->chunk.u.size == 0);
-  assert (size8 <= rmsg->chunk.rbuf->max_rmsg_size);
+  assert (size8P <= rmsg->chunk.rbuf->max_rmsg_size);
   assert (rmsg->lastchunk == &rmsg->chunk);
-  rmsg->chunk.u.size = size8;
+  rmsg->chunk.u.size = size8P;
 #if USE_VALGRIND
   VALGRIND_MEMPOOL_CHANGE (rmsg->chunk.rbuf->rbufpool, rmsg, rmsg, offsetof (struct nn_rmsg, chunk.u.payload) + rmsg->chunk.size);
 #endif
@@ -620,7 +622,7 @@ void nn_rmsg_commit (struct nn_rmsg *rmsg)
   ASSERT_RBUFPOOL_OWNER (chunk->rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
   assert (chunk->u.size <= chunk->rbuf->max_rmsg_size);
-  assert ((chunk->u.size % 8) == 0);
+  assert ((chunk->u.size % ALIGNOF_RMSG) == 0);
   assert (ddsrt_atomic_ld32 (&rmsg->refcount) >= RMSG_REFCOUNT_UNCOMMITTED_BIAS);
   assert (ddsrt_atomic_ld32 (&rmsg->chunk.rbuf->n_live_rmsg_chunks) > 0);
   assert (ddsrt_atomic_ld32 (&chunk->rbuf->n_live_rmsg_chunks) > 0);
@@ -677,15 +679,15 @@ void *nn_rmsg_alloc (struct nn_rmsg *rmsg, uint32_t size)
 {
   struct nn_rmsg_chunk *chunk = rmsg->lastchunk;
   struct nn_rbuf *rbuf = chunk->rbuf;
-  uint32_t size8 = align8uint32 (size);
+  uint32_t size8P = align_rmsg (size);
   void *ptr;
-  RMSGTRACE ("rmsg_alloc(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8);
+  RMSGTRACE ("rmsg_alloc(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8P);
   ASSERT_RBUFPOOL_OWNER (rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
-  assert ((chunk->u.size % 8) == 0);
-  assert (size8 <= rbuf->max_rmsg_size);
+  assert ((chunk->u.size % ALIGNOF_RMSG) == 0);
+  assert (size8P <= rbuf->max_rmsg_size);
 
-  if (chunk->u.size + size8 > rbuf->max_rmsg_size)
+  if (chunk->u.size + size8P > rbuf->max_rmsg_size)
   {
     struct nn_rbufpool *rbp = rbuf->rbufpool;
     struct nn_rmsg_chunk *newchunk;
@@ -703,7 +705,7 @@ void *nn_rmsg_alloc (struct nn_rmsg *rmsg, uint32_t size)
   }
 
   ptr = (unsigned char *) (chunk + 1) + chunk->u.size;
-  chunk->u.size += size8;
+  chunk->u.size += size8P;
   RMSGTRACE ("rmsg_alloc(%p, %"PRIu32") = %p\n", (void *) rmsg, size, ptr);
 #if USE_VALGRIND
   if (chunk == &rmsg->chunk) {
--- a/src/core/ddsi/include/dds/ddsi/q_radmin.h
+++ b/src/core/ddsi/include/dds/ddsi/q_radmin.h
@@ -102,6 +102,13 @@ DDSRT_STATIC_ASSERT (sizeof (struct nn_rmsg) == offsetof (struct nn_rmsg, chunk)
 #define NN_RMSG_PAYLOAD(m) ((unsigned char *) (m + 1))
 #define NN_RMSG_PAYLOADOFF(m, o) (NN_RMSG_PAYLOAD (m) + (o))
 
+/* Align rmsg chunks to the larger of sizeof(void*) or 8.
+
+Ideally, we would use C11's alignof(struct nn_rmsg); however, to avoid dependency on C11,
+we ensure rmsg chunks are at least aligned to sizeof(void *) or 8,
+whichever is larger. */
+#define ALIGNOF_RMSG (sizeof(void *) > 8 ? sizeof(void *) : 8)
+
 struct receiver_state {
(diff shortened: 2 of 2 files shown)


dodsonmg:
This appears to be failing with the following:

```
/home/travis/build/eclipse-cyclonedds/cyclonedds/src/core/ddsi/src/q_radmin.c:310:31: error: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘uint32_t’ {aka ‘unsigned int’} may change value [-Werror=conversion]
   return (x + align_size - 1) & (uint32_t)-align_size;
```

This isn't a practical problem, because `align_size` holds the size of a `void *` (16 bytes in my case, 8 in most others), so the cast to `uint32_t` won't lose anything.

I can make a modification to avoid the warning, but I think it's cleanest as is.  Since the function returns `uint32_t`, there's going to be a conversion where, since `sizeof` returns `size_t`.

What's your preference?

dodsonmg:
One possibility which builds on Ubuntu 18.04 with gcc:

```
static uint32_t align_literal_uint32 (uint32_t x)
{
  uint32_t align_size = sizeof(void *) > 8 ? sizeof(void *) : 8;
  return (x + align_size - 1) & -align_size;
}
```

This changes `align_size` from `size_t` to `uint32_t` and removes the cast in the return statement.  Since the conditional assignment will resolve to `8`, it seems to be quite happy assigning that to `uint32_t align_size`.

This also builds for my target using clang and 16 byte pointers.  I'll revise the PR.

dodsonmg:
I can't see why the Windows checks are failing.  Seems to be script-related, rather than my PR, but I'm not sure.

src/core/ddsi/src/q_radmin.c
```diff
@@ -14,6 +14,7 @@
 #include <stdarg.h>
 #include <stdlib.h>
 #include <limits.h>
+#include <stdalign.h>
```
rotu, line 17: were you going to use this? It seems to be C++ and later.

src/core/ddsi/src/q_radmin.c
```diff
@@ -303,9 +304,10 @@ static void nn_rbuf_release (struct nn_rbuf *rbuf);
 #define RMSGTRACE(...)              TRACE_CFG (rmsg, rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 #define RDATATRACE(rdata, ...)      TRACE_CFG ((rdata)->rmsg, (rdata)->rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 
-static uint32_t align8uint32 (uint32_t x)
+static uint32_t align_literal_uint32 (uint32_t x)
```
rotu, line 307: It would be nice to have a why in here. "align_literal_uint32" doesn't make it clear that this is for aligning a `struct nn_rmsg`.

eboasson:
> This appears to be failing with the following:
> 
> ```
> /home/travis/build/eclipse-cyclonedds/cyclonedds/src/core/ddsi/src/q_radmin.c:310:31: error: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘uint32_t’ {aka ‘unsigned int’} may change value [-Werror=conversion]
>    return (x + align_size - 1) & (uint32_t)-align_size;
> ```
> 
> This isn't a practical problem, because `align_size` holds the size of a `void *` (16 bytes in my case, 8 in most others), so the cast to `uint32_t` won't lose anything.
> 
> I can make a modification to avoid the warning, but I think it's cleanest as is. Since the function returns `uint32_t`, there's going to be a conversion where, since `sizeof` returns `size_t`.
> 
> What's your preference?

I suspect it was GCC ultimately complaining about the conversion of `sizeof(void*)` to an `uint32_t` because it has the friendly habit of (still) not taking into consideration the value of a compile time constant being converted, like clang does. There are also some compilers that warn if you negate an unsigned value, so `-align_size` can cause warnings by itself.

I've found that I get in the least trouble with the various compilers (and compiler versions) if I effectively write it like this:
```C
T align(T x, U a) { const T m = (T)(a-1); return (T) ((x+m) & ~m); }
```
and it all comes down to the same thing anyway ... (That cast of the return value is because a T narrower than an `int` will promote everything to `int`, and with some compilers causes that same warning you got. C's type system makes superficial sense at best.)

src/core/ddsi/src/q_radmin.c
eboasson: @dodsonmg no surprise there, it is very hard to spot the error on Windows, but this here is indeed the reason:
```
C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(17): fatal error C1083: Cannot open include file: 'stdalign.h': No such file or directory [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
```

src/core/ddsi/src/q_radmin.c
eboasson: Fair point, maybe `align8Puint32` would be better? 8 bytes or pointer, whichever is widest.

Anyway, I'm not known for consistently coming up with beautiful names, so I am not going to fuss over the name of some very locally used alignment function ...

rotu: `align8Puint32` is correct but doesn’t make the code any more self-evident. Maybe `align_rmsg` or change it to be a general alignment function that takes as an argument a constant `ALIGNOF_RMSG`.

rotu:
While we're sharing favorite alignment expressions, the version I find most intuitive is this lovely couplet:
```
x += align - 1;
x -= x % align;
```

With power of 2 alignment and unsigned arguments, this simplifies to: `(x + -x % align)`.

There's also the slightly less readable but cute (again, unsigned and powers of 2 alignment):

`( x | (align-1)) + 1`

src/core/ddsi/src/q_radmin.c
dodsonmg: I agree.  This is unnecessary since I'm using `sizeof` instead of `alignof`.

src/core/ddsi/src/q_radmin.c
dodsonmg: This makes sense.  How about:

```
static uint32_t align_rmsg (uint32_t x)
{
  /* Align rmsg chunks to the larger of sizeof(void*) or 8.
  Ideally, we would use C11's alignof(void *), but sizeof(void *) is equivalent
  for primitive types. */
  const uint32_t align_size = sizeof(void *) > 8 ? sizeof(void *) : 8;
  return (x + align_size - 1) & -align_size;
}
```

and then returning to the original naming convention for the lhs of the assignment.  For example:

`uint32_t size8P = align_rmsg (size);`

eboasson:
Bummer ... Microsoft struck again ...
```
  q_radmin.c
C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): error C2220: warning treated as error - no 'object' file generated [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): warning C4146: unary minus operator applied to unsigned type, result still unsigned [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
```

dodsonmg:
I guess I'll try to be very explicit to get through the MS check.  I've taken a simple, one line function and made if four lines.  Blah.

src/core/ddsi/src/q_radmin.c
```diff
@@ -303,9 +303,15 @@ static void nn_rbuf_release (struct nn_rbuf *rbuf);
 #define RMSGTRACE(...)              TRACE_CFG (rmsg, rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 #define RDATATRACE(rdata, ...)      TRACE_CFG ((rdata)->rmsg, (rdata)->rmsg->chunk.rbuf->rbufpool->logcfg, __VA_ARGS__)
 
-static uint32_t align8uint32 (uint32_t x)
+static uint32_t align_rmsg (uint32_t x)
 {
-  return (x + 7u) & (uint32_t)-8;
+  /* Align rmsg chunks to the larger of sizeof(void*) or 8.
+  Ideally, we would use C11's alignof(void *), but sizeof(void *) is equivalent
```
rotu, line 309: `alignof` can be used on aggregate types, so ideally we’d use `alignof(struct nn_rmsg)` rather than assuming “the strictest alignment need of rmsg comes from either uint32 or void*”

src/core/ddsi/src/q_radmin.c
```diff
@@ -620,7 +626,7 @@ void nn_rmsg_commit (struct nn_rmsg *rmsg)
   ASSERT_RBUFPOOL_OWNER (chunk->rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
   assert (chunk->u.size <= chunk->rbuf->max_rmsg_size);
-  assert ((chunk->u.size % 8) == 0);
+  assert ((chunk->u.size % (sizeof(void *) > 8 ? sizeof(void *) : 8)) == 0);
```
rotu, line 629: Maybe extract to a constant `ALIGNOF_RMSG`

src/core/ddsi/src/q_radmin.c
dodsonmg: That makes sense.  I'll revise.

rotu:
> Bummer ... Microsoft struck again ...
> 
> ```
>   q_radmin.c
> C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): error C2220: warning treated as error - no 'object' file generated [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
> C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): warning C4146: unary minus operator applied to unsigned type, result still unsigned [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
> ```

Yeah. I understand why Visual Studio alerts on this, and the warning is sensible. It's not totally intuitive that `-(1u)` is unsigned whereas `-(1)` is signed. It's totally legit code but to make MS happy, `(0u-x)` is a more explicit version of `-x`.

src/core/ddsi/src/q_radmin.c
dodsonmg: just pushed a change.  see if it builds...

dodsonmg:
I'm getting gcc warnings about size_t -> uint32_t again...

Sorry.  I pushed as a shortcut to move code from one place to another for testing and forgot it would rebuild here automatically...

dodsonmg:
That last push includes a cast to avoid the gcc warnings (and also gets rid of trailing whitespace).

eboasson:
> > Bummer ... Microsoft struck again ...
> > ```
> >   q_radmin.c
> > C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): error C2220: warning treated as error - no 'object' file generated [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
> > C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\src\core\ddsi\src\q_radmin.c(312): warning C4146: unary minus operator applied to unsigned type, result still unsigned [C:\Users\travis\build\eclipse-cyclonedds\cyclonedds\build\src\core\ddsc.vcxproj]
> > ```
> 
> Yeah. I understand why Visual Studio alerts on this, and the warning is sensible. It's not totally intuitive that `-(1u)` is unsigned whereas `-(1)` is signed. It's totally legit code but to make MS happy, `(0u-x)` is a more explicit version of `-x`.

Not trying to get into a senseless discussion here, but I'd say this is one of the milder cases of arithmetical counter-intuitiveness in C. For example:
```C
# cat > xx.c
#include <stdio.h>
#include <stdlib.h>
int f (uint8_t x)
{
  return (x + x > (uint8_t)255);
}
int g (uint8_t x)
{
#define x2 x*x
#define x4 x2*x2
#define x8 x4*x4
  return (x8 >= (uint8_t)0);
}
int main (int argc, char **argv)
{
  const uint8_t x = (uint8_t) atoi (argv[1]);
  printf ("%d + %d > 255   %s\n", x, x, f (x) ? "true" : "false");
  printf ("%d ** 8 >= 0   %s\n", x, g (x) ? "true" : "false");
  return 0;
}
# clang -g -fsanitize=undefined -Wall xx.c -o xx
# ./xx 255                                          
255 + 255 > 255   true
/Users/erik/C/xx.c:12:11: runtime error: signed integer overflow: 16581375 * 255 cannot be represented in type 'int'
/Users/erik/C/xx.c:12:11: runtime error: signed integer overflow: -66716671 * 255 cannot be represented in type 'int'
/Users/erik/C/xx.c:12:11: runtime error: signed integer overflow: 167118079 * 255 cannot be represented in type 'int'
/Users/erik/C/xx.c:12:11: runtime error: signed integer overflow: -334562815 * 255 cannot be represented in type 'int'
/Users/erik/C/xx.c:12:11: runtime error: signed integer overflow: 585828095 * 255 cannot be represented in type 'int'
255 ** 8 >= 0   false
```
The first is an expression with *only* unsigned 8-bit numbers, and so I would say the intuitive result would be 254, which is less than 255 ... But whatever. I'm sure Microsoft knows best what's good for me, and that is: negating an unsigned integer is scary and needs a warning but implicit promotion to signed integer (possibly triggering undefined behaviour where that's impossible for unsigned types) is perfectly logical ...

Oh well. Rant over 🙂

@dodsonmg Thank you very much, PR looks good to me!

PR #499, opened 2020-04-20
Improvements to dds_read_instance, the addition of dds_readcdr of nicer printing of samples

eboasson:
* dds_read_instance had an unspeakably bad implementation (I plead innocent despite git blame showing my name):
  * dds_read_impl did an entirely superfluous check to see whether the instance handle was known globally, and as there is no efficient global instance handle lookup that meant a scan of all known instance handles
  * the RHC implementation is built around a hashing instance handles, but instead of using a hash lookup it scanned all instances
  And so what should have been a pretty efficient lookup ended up being *two* linear scans of key values ... This PR fixes that by dropping the superfluous first scan and replacing the second by a hash table lookup.
* The code for reading/taking samples is somewhat complicated. I've refactored it a bit and in the process also removed the duplicate implementation in `dds_takecdr`. That interface is not quite a stable, backwards-compatible interface, but it has proved very useful on multiple occasions and I believe deserves to stay. But then the absence of a `dds_readcdr` companion is starting to grate. With the refactoring, adding the latter is a trivial thing. Once again: it's an evolving interface, so think before you decide to rely on it.
* A layer of wrapper functions that did nothing more than casting the RHC pointer from the generic `dds_rhc` to the specific `dds_rhc_default` got removed as well. Those were just noise.
* The trace optionally contains the sample contents in a readable form, but the IDL preprocessor erased so much type information that it became impossible to tell whether, e.g., a 4 byte object was originally an enumerated type, a 32-bit signed integer, a 32-bit unsigned integer, or a 32-bit floating-point number. The enums coming out as numbers is no big deal, but signed integers, and especially floating-point numbers, being printed as unsigned decimal numbers is quite annoying. So despite my reluctance to touch IDLC, this changes its output to set flags for signed and floating-point numbers. These flags only affect the printing. No requirement to regenerate anything.

### DIFF
--- a/src/core/ddsc/src/dds_rhc_default.c
+++ b/src/core/ddsc/src/dds_rhc_default.c
@@ -337,7 +337,7 @@ struct dds_rhc_default {
 };
 
 struct trigger_info_cmn {
-  unsigned qminst;
+  uint32_t qminst;
   bool has_read;
   bool has_not_read;
 };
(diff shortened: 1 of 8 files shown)


src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -2014,418 +1959,326 @@ static bool take_sample_update_conditions (struct dds_rhc_default *rhc, struct t
...
-          break;
-        }
-      }
+      n += take_w_qminv_inst (rhc, inst, values + n, info_seq + n, max_samples - n, qminv, qcmask, &ntriggers, to_sample, to_invsample);
       inst = inst1;
     }
   }
   TRACE ("take: returning %"PRIu32"\n", n);
   assert (rhc_check_counts_locked (rhc, true, false));
   ddsrt_mutex_unlock (&rhc->lock);
```
dennis-adlink, line 2254: One would expect that this unlock is using the same conditon (`lock`) as used when taking this lock in the beginning of the function, but for some reason (related to the `lock_samples` function used to get the sample count) that's not the case. Although this PR does not change this (not so intuitive) behaviour, it would be good to add a comment here that explains the reasoning behind this. Same applies to the unlock in `read_w_qinv`.

eboasson: Good point, thanks!

eboasson:
Subsumed in #512 (which targets it at security).

PR #505, opened 2020-04-30
Preparatory work for implementing autopurge on RHC

eboasson:
With apologies for the mass of this PR ... and even with that it is still only preparation for I really meant to do because the bit of refactoring I started doing with an aim to reduce the number of points that would be affected by implementing "autopurge" for instances in the no writers and in the disposed states turned up some other things.

There are many commits, I believe I managed to get all of them to work, even though they were mostly reconstructed based on the final form. If anyone feels like reviewing anything, I'd suggest looking at the individual commits.

* The first 4 commits are the same as #499, rebased to current master.
* "Null instance pointer ..." avoids passing what is effectively a dangling pointer. The dangling pointer did not get dereferenced, but I found it quite unwise as it was.
* "Track deadline registrations ..." uses a dedicated flag to track in the instances to track whether it is currently registered in the deadline administration and uses this to perform all updates at the end updating an instance, instead of spread out through the code.
* "Auto-dispose ..." is a slight change in auto-dispose behaviour, instead of treating an unregister as a dispose, it now atomically transitions a NO_WRITERS instance to a DISPOSED one (also fits better with the name: "auto-dispose *unregistered* instances", instead of "auto-dispose on unregister").
* "Always add invalid sample ...": it turns out there were a few cases where an invalid sample wasn't generated when it should have been.
* "Refactor" is some of the refactoring in the title, shouldn't affect the behaviour other than some tiny changes in tracing output (those traces remain unworthy of recommendation as reading material anyway).
* "Standard byte order ...": a small clean up to try not to break expectations.
* "Implement "from_sample" ...": that function didn't exist by mistake, causing `dds_lookup_instance` on a built-in topic to crash.
* "Rework listener tests" manages to leave the number of lines of the test file very nearly unchanged while vastly increasing the number of tests by replacing tons of boilerplate by an inscrutable and hacked-together DSL with an interpreter. I'm not sure it is an improvement, but it did help me find missing data available notifications, missing invalid samples and a missing "from_sample" function, so it certainly has been a valuable exercise.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -2741,6 +2741,53 @@ dds_take_mask_wl(
   uint32_t maxs,
   uint32_t mask);
 
+#define DDS_HAS_READCDR 1
+/**
+ * @brief Access the collection of serialized data values (of same type) and
+ *        sample info from the data reader, readcondition or querycondition.
+ *
+ * This call accesses the serialized data from the data reader, readcondition or
+ * querycondition and makes it available to the application. The serialized data
+ * is made available through \ref ddsi_serdata structures. Returned samples are
+ * marked as READ.
+ *
+ * Return value provides information about the number of samples read, which will
+ * be <= maxs. Based on the count, the buffer will contain serialized data to be
+ * read only when valid_data bit in sample info structure is set.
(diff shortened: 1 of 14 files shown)


eboasson:
Subsumed in #512 (which targets it at master).

PR #507, opened 2020-05-01
Document the versioning policy

rotu:


rotu:
Not sure if this belongs here or the README or CONTRIBUTING files

rotu:
Liveliness, Lifespan, Deadline were all introduced in 0.5.1  / 0.5.0 release (with true semver, these features should have been accompanied by a minor version bump, not just a patch version bump).

Are you sure you want to switch?

eboasson:
> Liveliness, Lifespan, Deadline were all introduced in 0.5.1 / 0.5.0 release (with true semver, these features should have been accompanied by a minor version bump, not just a patch version bump).
> 
> Are you sure you want to switch?

Fair point, but yes, I'm sure. Otherwise one keeps having to explain oneself.

Anyway, I believe this sort of thing to be more about risk management than about a strict distinction between bug fix and feature (as long as backwards compatibility is maintained within a major version. Consider, e.g., `dds_get_guid` that I added recently to master: that's arguably a new feature but I think anyone would accept it in a patch. The QoS settings use interfaces that exist but don't all quite do what they were supposed to do and so could be argued to be pure bug fixes ... In hindsight, treating as a bug fix was probably the wrong decision.

rotu:
> Otherwise one keeps having to explain oneself.

Hence a written policy. It doesn't matter what it is, so long as we adhere consistently. We could adopt [RomVer](https://github.com/romversioning/romver) instead.

If we switch to SemVer, we are committing to:

1. No breaking API changes or removals until the 1.0.0 release
2. No API changes or additions *whatsoever* in a patch release

I want to double check: is that okay?

> Anyway, I believe this sort of thing to be more about risk management than about a strict distinction between bug fix and feature (as long as backwards compatibility is maintained within a major version.

That's the thing. In SemVer:

1. No compatibility guarantees between major versions.
2. Minor versions are backwards-compatible.
3. Patch versions are internal and cannot affect the API.

> Consider, e.g., `dds_get_guid` that I added recently to master: that's arguably a new feature but I think anyone would accept it in a patch.

Adding a new public API function like `dds_get_guid` in semver definitely requires a minor version bump and cannot be done in a patch release.

eboasson:
Thanks for digging a bit more and coming up with RomVer. RomVer has some attraction in that acknowledges reality is messy, but I don't like its distinction between "human" and "major": I think it is contrived because bumping "major" is always sufficient. The "human" looks like the most-significant one of Solaris (or Linux): in the end it just got dropped because it had turned into a constant.

All these formalizations are truly silly, when you think about it. Reality is that bugs are unavoidable in practice, and at times the bug will be in maintaining backwards compatibility. Been there, done that. Consequently, I can't take these documents seriously. I could try to come up with my own formalization, but the more precise I'd try to make it, the more it would suffer from the same issue.

SemVer with a bit of wiggle room is really the most realistic. I'd sketch it as X.Y.Z meaning
* a Z bump is highly unlikely to cause any trouble (for existing programs)
* a Y bump is unlikely to cause any trouble and tends to introduce meaningful improvements that are not bug fixes
* a X bump should be assumed to cause trouble unless otherwise stated

... but I'm sure the world will say that's too sloppy. And yet it is a lie without the wiggle room ...

eboasson:
New proposal:

"Cyclone DDS versioning follows semantic versioning with two exceptions: (1) major version 0 is also considered stable; (2) patch releases may add new features."

rotu:
> New proposal:
> 
> "Cyclone DDS versioning follows semantic versioning with two exceptions: (1) major version 0 is also considered stable; (2) patch releases may add new features."

So what’s the difference between a minor and a patch release if both can have externally visible changes? That’s the only distinction between the two levels in SemVer.

eboasson:
> So what’s the difference between a minor and a patch release if both can have externally visible changes? That’s the only distinction between the two levels in SemVer.

I know ... the only difference is between my "highly unlikely" vs "unlikely" to "cause trouble" and there's no way you can formalize that distinction. But strict semantic versioning means that if you ever get into a situation where you need to add something to the API to deal with a bug [\*] you need to bump the minor and that (in some cases I know of) forces re-qualification of a system, which can be very expensive proposition. Therefore, in my experience you need more freedom for patch releases, where it is understood that one doesn't anything "risky" in a patch.

If you can come up with a good phrase for that, I'd be delighted.

[\*] It could be a workaround, e.g., that `dds_get_guid` function does not add functionality, it only removes overhead. I know of cases where the bug was performance and the only solution was to add a function.

rotu:
> I know ... the only difference is between my "highly unlikely" vs "unlikely" to "cause trouble" and there's no way you can formalize that distinction.

That’s why SemVer chose a bright line: internally changes versus externally visible changes. Bumping minor version signifies that there are new ways to interact with the package, even though everything should still work.

Adding to the public API *doesn’t* resolve bugs. It only makes the downstream packages *able* to resolve bugs.

>  you need to bump the minor and that (in some cases I know of) forces re-qualification of a system

Okay. Let’s start here. What sort of system qualification are we looking at? Because apparently that imposes meaning on our versioning scheme.

It seems there are also functional milestones we’re aiming for listed here: https://projects.eclipse.org/projects/iot.cyclonedds
Are they current and should they inform the versioning scheme?

It seems like maybe minor versions can be for planned feature set milestones and patch versions can be for unplanned but necessary changes that come up in between. It’s still not SemVer, but it does seem like maybe what you have in mind?

eboasson:
> > I know ... the only difference is between my "highly unlikely" vs "unlikely" to "cause trouble" and there's no way you can formalize that distinction.
> 
> That’s why SemVer chose a bright line: internally changes versus externally visible changes. Bumping minor version signifies that there are new ways to interact with the package, even though everything should still work.
> 
> Adding to the public API _doesn’t_ resolve bugs. It only makes the downstream packages _able_ to resolve bugs.

Depends on where you draw the line with bugs. Sometimes the only way to provide a bug fix is to add a function. If the bug is in the specification of the interface, for example, the interface has been defined in a way that gives rise to race condition under certain scenarios, then there is no way to fix the bug without breaking existing uses unless you add a new interface. This is the sort of thing that SemVer completely ignores.

> > you need to bump the minor and that (in some cases I know of) forces re-qualification of a system
> 
> Okay. Let’s start here. What sort of system qualification are we looking at? Because apparently that imposes meaning on our versioning scheme.

Of course it matters: the whole point of being careful about what you do with your releases and how you number is to help users. Release qualification testing for large, highly dependable systems is no small matter and takes weeks (if not months).

> It seems there are also functional milestones we’re aiming for listed here: https://projects.eclipse.org/projects/iot.cyclonedds
> Are they current and should they inform the versioning scheme?

I must be overlooking them, but it doesn't matter much because there are indeed some functional milestones and they do matter. Not necessarily for the versioning *scheme* but certainly for the version itself.

There will be an API change on a few points that I am not sure can be done in an entirely backwards compatible manner (the "read" interface in particular), although I will do my best to provide backwards compatibility as much as possible. That'll probably force the jump to 1.0. Tradition in Eclipse is that one bumps the version number to 1.0 for graduation from the incubation phase, so with a bit of luck the two can be timed to coincide.

It seems likely that preallocation of memory (a.k.a. "static allocation") will require some changes and it may well be that transient data support, group coherency and a number of other major features will do so to. In any case, at some point a simple risk assessment says it is better to bump the major than the minor even though it ought to be backwards compatible. So I also definitely see a 2.x in the future.

> It seems like maybe minor versions can be for planned feature set milestones and patch versions can be for unplanned but necessary changes that come up in between. It’s still not SemVer, but it does seem like maybe what you have in mind?

That's a pretty decent way to put it.

rotu:
> Of course it matters: the whole point of being careful about what you do with your releases and how you number is to help users. Release qualification testing for large, highly dependable systems is no small matter and takes weeks (if not months).

Okay 


> Sometimes the only way to provide a bug fix is to add a function. If the bug is in the specification of the interface, for example,

Again, exposing a new function does not *FIX* a bug. It provides downstream packages a tool to fix the bug. Bumping the minor version signals that such changes may be present. That new APIs will not appear in a patch release is a deliberate choice by SemVer - one that doesn't seem to be a priority to you (I happen to agree with your priority - I expect *hardly any* Cyclone releases to leave the API completely unchanged).

> > > you need to bump the minor and that (in some cases I know of) forces re-qualification of a system
> > 
> > 
> > Okay. Let’s start here. What sort of system qualification are we looking at? Because apparently that imposes meaning on our versioning scheme.
> 
> Of course it matters: the whole point of being careful about what you do with your releases and how you number is to help users. Release qualification testing for large, highly dependable systems is no small matter and takes weeks (if not months).

Maybe I asked this question wrong. Here's what I need to know:

What *release qualification processes* is or might this project subject to? Does calling a release a *patch release* change how the software is qualified versus calling it a *minor release*, assuming the exact same code changes?

eboasson:
> > Sometimes the only way to provide a bug fix is to add a function. If the bug is in the specification of the interface, for example,
> 
> Again, exposing a new function does not _FIX_ a bug. It provides downstream packages a tool to fix the bug.

I think we simply construe things differently, because I don't agree with you despite considering your view a logical and consistent one.

The difference is that I consider an implementation that meets an incorrect specification still a bug because the goal is to build a _system_ — and to me the system is far more interesting and important than any random piece of middleware that happens to be used in it. When changing the specification isn't an option, the only possible solution for the system then becomes an extension. I consider that a bug fix (it allows me to have a working system), but I can see the logic of saying it isn't a fix (the bug that was there is still present).

> Maybe I asked this question wrong. Here's what I need to know:
> 
> What _release qualification processes_ is or might this project subject to? Does calling a release a "_patch release_" change how the software is qualified versus calling it a "_minor release_", assuming the exact same code changes?

If "this project" is Cyclone, then, ideally, the development of a formal specification of DDS jointly with all other stakeholders along with a very high quality testsuite so that the release process becomes formally validating the implementation against that specification and showing that it passes the testsuite 🙂 — in short: "Beware of bugs in the above code; I have only proved it correct and tried it" (free after Knuth). Obviously, that's in jest and it'll likely just be a improving the natural language description of the expected behaviour of Cyclone and improving the test suite to ever more carefully check that it behaves as expected.

For other projects ... in the end the choice is binary: fully requalify or decide to requalify the parts that are affected by the changes, which becomes infeasible really quickly. (That incidentally means that _adding_ something has less impact than fixing something.) So on the technical side, people will likely inspect the actual changes and then make the final call, regardless of what it is called; on the management side you have a significant risk that they will look at the version number and stated policies and think it meaningful. (Not that that distinction is anything new.)

The trouble is determining the impact if the changes are more than tiny. So imagine you have such a system using version 3.1.4 and discover a problem and the only way to solve the problem with provably low-impact on the qualification is a small addition. If 3.2 has not been released, you can just bump it to 3.2.0 and I'm sure all will be well. But what do you do when 3.2.x exists and the changes from 3.1.4 to 3.2.x are already such that the impact is major?

Of course no-one is going to build such a system on an open source project without some formalised form of support — either in-house or contracted out — and therefore you can always have additional variants. You just take 3.1.4, add a patch, call it something else, and hopefully upstream the changes.

So

> That new APIs will not appear in a patch release is a deliberate choice by SemVer - one that doesn't seem to be a priority to you (I happen to agree with your priority - I expect _hardly any_ Cyclone releases to leave the API completely unchanged).

it looks like we actually agree that the "patch" of SemVer is pretty meaningless in practice, at least for Cyclone, which is really all that matters here. SemVer just rubs me the wrong way by having these very strict rules on "patch" releases that evidently ignore a large part of reality.

Having blown off some steam I'll go full circle and say we should just adopt SemVer with the tacit understanding that patches are not likely to happen. And should the above scenario ever happen, we'll just deal with it outside the SemVer numbering scheme.

rotu:
That was quite the wordy response, but here are my takeaways:

1. Compatible changes to the public API can be considered bug fixes if you consider a "design flaw" a "bug".

I agree, though that still doesn't qualify them as appropriate for patch releases under SemVer. Incrementing the minor version signifies to downstream projects "there is something new to check out and maybe use.

2. It is the *scope* of a change, not the version number that says whether a change needs a system to be recertified.

So the versioning system *does not* affect certification. But it makes sense to number things such that "small changes look small" in the version number. I think that's always a goal.

3. SemVer is impractical

I sort of agree with you, though I think my attitude is a bit more charitable toward SemVer. It's all about communicating how much the public API has changed. The three levels of change in SemVer are meaningful:

MAJOR: the API has changed in an incompatible way so that downstream projects *may need* to make changes.
MINOR: the API has changed in a compatible way so that downstream project *can* make changes to take advantage of these changes.
PATCH: the API has not changed. Downstream projects may be functionally affected but only in non-controversial ways.

> Having blown off some steam I'll go full circle and say we should just adopt SemVer with the tacit understanding that patches are not likely to happen. And should the above scenario ever happen, we'll just deal with it outside the SemVer numbering scheme.

I don't understand what you're saying here.

I don't understand what you're saying about "outside semver" Are we SemVer-compatible or not?

Take for example our release 0.5.0 -> 0.5.1, which is clearly illegal in SemVer.

If we don't care about strict SemVer, we could have called it 0.5.1. We would need to redefine what a "PATCH" is.

With a SemVer-compatible system we could have
* Called it 0.6.0 instead of 0.5.1 (this sort of thing would be incompatible with predeclared feature sets like: https://projects.eclipse.org/projects/iot.cyclonedds/releases/0.6.0-florestan)
* Made it a prerelease like 0.6.0-rc.1

How do you propose to define MINOR versus PATCH level changes?

eboasson:
> That was quite the wordy response, but here are my takeaways:

Sorry ... This is longish, too, but not to worry: you can safely skip to the end.

> 1. Compatible changes to the public API can be considered bug fixes if you consider a "design flaw" a "bug".
> 
> I agree, though that still doesn't qualify them as appropriate for patch releases under SemVer.

Indeed

> Incrementing the minor version signifies to downstream projects "there is something new to check out and maybe use.

I agree

> 1. It is the _scope_ of a change, not the version number that says whether a change needs a system to be recertified.

It *should be* thus, and if it weren't for the existence of powers that decide on things they don't understand, it would be. But as I mentioned, it is not something I foresee being an issue for the Cyclone project itself.

> 1. SemVer is impractical
> 
> I sort of agree with you, though I think my attitude is a bit more charitable toward SemVer. 

🙂

> It's all about communicating how much the public API has changed. The three levels of change in SemVer are meaningful:
> 
> MAJOR: the API has changed in an incompatible way so that downstream projects _may need_ to make changes.
> MINOR: the API has changed in a compatible way so that downstream project _can_ make changes to take advantage of these changes.
> PATCH: the API has not changed. Downstream projects may be functionally affected but only in non-controversial ways.

There is another traditional split: binary compatible, source compatible, anything goes. That one is arguably more valuable because it avoids trying to define what changes fit in "binary compatible".

For C, "binary compatible" leaves a lot of room for changes. However, with C++'s templates and/or virtual functions, you have far less room and the "no additions" in a PATCH makes more sense. Modern languages are generally not easy-going either.

So an uncharitable opinion about SemVer for a C project is arguably fair; an uncharitable opinion about SemVer in general, not so much. Cyclone is a C based library, but there'll certainly be a C++ binding once some paperwork is sorted, and likely others as well.

So my C-based perspective is not actually the right one, even though today it fits.

> > And should the above scenario ever happen, we'll just deal with it outside the SemVer numbering scheme.
> 
> I don't understand what you're saying about "outside semver" Are we SemVer-compatible or not?

Don't worry about that phrase: all it was meant to say is that if one ever must do a special-purpose fix to an old version and there isn't room in the versioning scheme to make the change, one can simply consider it not part of the normal releases. And we don't have the manpower to deal with providing support on old versions anyway ...

> Take for example our release 0.5.0 -> 0.5.1, which is clearly illegal in SemVer.

I agree. Binary compatibility is not an issue, but it is does not fit the definition of a patch version.

> If we don't care about strict SemVer, we could have called it 0.5.1. We would need to redefine what a "PATCH" is.
> 
> With a SemVer-compatible system we could have
> 
> * Called it 0.6.0 instead of 0.5.1 (this sort of thing would be incompatible with predeclared feature sets like: https://projects.eclipse.org/projects/iot.cyclonedds/releases/0.6.0-florestan)
> * Made it a prerelease like 0.6.0-rc.1

Actually 0.5.1 was really meant to just be a tag for the purposes of blooming without knowing how to get around giving it an X.Y.Z identifier ... but that's water under the bridge.

The releases https://projects.eclipse.org/projects/iot.cyclonedds/releases are maintained by me and up to the point that the formal release review process starts, anything can be changed: dates, features, you name it. The system knows of "milestone" builds and that would've been the right thing to do. But again, too late for that. (I added one for 0.7.0.)

> How do you propose to define MINOR versus PATCH level changes?

I'd define it as:

* MINOR: source compatible
* PATCH: binary compatible

and then it all fits. As soon as C++ comes into the mix it will effectively be congruent with SemVer. By the time we get to 1.0 that point will definitely have been reached. Hence:

"Cyclone DDS follows Semantic Versioning. Additionally, major version 0 strives to provide source compatibility for minor releases and binary compatibility for patch releases for application code not directly including any header files other than `dds.h`."

The final bit really needs properly documenting stability levels for the various interface ... but I think this does match the grown practice and is a simple enough rule.

PR #508, opened 2020-05-04
Update security handshake when optional fields of handshake tokens are missing

MarcelJordense:
A number of fields of the handshake tokens that are exchanged are optional. In case these optional fields are missing these field should be constructed at the receiving side to correctly verify the signature present in the handshake token. This is handled by this PR.
Further the (de)serialization of the PropertyQoS parameter present in the p.data field of the handshake tokens is corrrected.
Also the generation of the shared key used for the exchange of the crypto tokens is updated to comply to the specification.

### DIFF
--- a/src/security/builtin_plugins/authentication/src/authentication.c
+++ b/src/security/builtin_plugins/authentication/src/authentication.c
@@ -200,6 +200,7 @@ typedef struct dds_security_authentication_impl
   struct dds_security_timed_cb_data *timed_callbacks;
   struct dds_security_timed_dispatcher_t *dispatcher;
   X509Seq trustedCAList;
+  bool include_optional;
 } dds_security_authentication_impl;
 
 /* data type for timer dispatcher */
@@ -1123,9 +1124,9 @@ DDS_Security_ValidationResult_t begin_handshake_request(dds_security_authenticat
   LocalIdentityInfo *localIdent;
   RemoteIdentityInfo *remoteIdent;
   EVP_PKEY *dhkey;
-  DDS_Security_BinaryProperty_t *tokens, *c_id, *c_perm, *c_pdata, *c_dsign_algo, *c_kagree_algo, *hash_c1, *dh1, *challenge;
   unsigned char *certData, *dhPubKeyData = NULL;
   uint32_t certDataSize, dhPubKeyDataSize;
+  uint32_t tsz = impl->include_optional ? 8 : 7;
   int created = 0;
 
   if (!instance || !handshake_handle || !handshake_message || !serialized_local_participant_data)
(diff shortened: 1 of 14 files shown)


src/security/core/src/dds_security_serialize.c
```diff
@@ -742,7 +751,7 @@ DDS_Security_Deserialize_ParticipantBuiltinTopicData(
                 r = DDS_Security_Deserialize_DataHolder(dser, &pdata->permissions_token);
                 break;
             case PID_PROPERTY_LIST:
-                r = DDS_Security_Deserialize_PropertyQosPolicy(dser, &pdata->property);
+                r = DDS_Security_Deserialize_PropertyQosPolicy(dser, &pdata->property, len);
```
eboasson, line 754: Each policy should be deserialized based on the number of bytes remaining in the parameter, not just the properties. We have a (de)serializer pair for this stuff that works (ddsi_plist.c) and it really should build on that. I realise that takes a bit of effort because of those annoying DDS_Security_... types, but that effort is worth it just to get rid of having to maintain two implementations.

Having said that, that's something for a different PR.

src/security/builtin_plugins/cryptographic/src/crypto_key_factory.c
```diff
@@ -89,8 +89,10 @@ calculate_kx_keys(
   unsigned char *kx_master_salt, *kx_master_sender_key;
   size_t shared_secret_size = get_secret_size_from_secret_handle(shared_secret);
   unsigned char hash[SHA256_DIGEST_LENGTH];
-  size_t concatenated_bytes1_size = DDS_SECURITY_AUTHENTICATION_CHALLENGE_SIZE * 2 + sizeof(KXSALTCOOKIE);
-  size_t concatenated_bytes2_size = DDS_SECURITY_AUTHENTICATION_CHALLENGE_SIZE * 2 + sizeof(KXKEYCOOKIE);
+  size_t KXKEYCOOKIE_SIZE = strlen(KXKEYCOOKIE);
```
eboasson, line 92: While correct, I think it makes more sense to add `#define KXKEYCOOKIE_SIZE (sizeof(KXKEYCOOKIE)-1)` immediately after the definition of `KXKEYCOOKIE` (and same for `KXSALTCOOKIE`).

src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -1221,7 +1216,7 @@ DDS_Security_ValidationResult_t begin_handshake_request(dds_security_authenticat
   handshake_message->class_id = ddsrt_strdup(AUTH_HANDSHAKE_REQUEST_TOKEN_ID);
   handshake_message->properties._length = 0;
   handshake_message->properties._buffer = NULL;
-  handshake_message->binary_properties._length = 8;
+  handshake_message->binary_properties._length = tsz;
```
eboasson, line 1219: `tsz == idx` at this point, is it not? One should at least add an assertion that this is the case. Perhaps just use `idx`. Furthermore, `tsz` is a bit cryptic and `idx` a bit generic. I'd rather seem them renamed to `tokcount` and `tokidx`, or something.

PR #460, opened 2020-03-30
Remove directories from IDLC install list

rotu:
These are unnecessary and can cause build weirdness on Windows (possibly due to directory symlinks being a little weird on the platform (?)).

It also seems to work not as intended - `FOLLOW_SYMLINKS` includes all traversed directories, even if no file inside matches the globbing expression!

```
2020-03-30T15:56:56.1457007Z InitializeBuildStatus:
2020-03-30T15:56:56.1457137Z   Creating "x64\Release\idlc\idlc.tlog\unsuccessfulbuild" because "AlwaysCreate" was specified.
2020-03-30T15:56:56.1457260Z CustomBuild:
2020-03-30T15:56:56.1457492Z   "The build of 'D:\a\rmw_cyclonedds\rmw_cyclonedds\ros_ws\build\cyclonedds\CMakeFiles\304562001b460fcc02dd05a4699f82d1\idlc.rule' depends on 'D:\A\RMW_CYCLONEDDS\RMW_CYCLONEDDS\ROS_WS\BUILD\CYCLONEDDS\SRC\IDLC\TARGET\IDLC-JAR-WITH-DEPENDENCIES.JAR' which is produced by the build of 'D:\a\rmw_cyclonedds\rmw_cyclonedds\ros_ws\build\cyclonedds\CMakeFiles\b5ce88a7e533379faaf5d87b828c151a\idlc-jar-with-dependencies.jar.rule'. The items cannot be built in parallel."
2020-03-30T15:56:56.1457719Z   Building JAR file D:/a/rmw_cyclonedds/rmw_cyclonedds/ros_ws/build/cyclonedds/src/idlc/target/idlc-jar-with-dependencies.jar
[...]

PR #511, opened 2020-05-11
Merge master into security

eboasson:
This proposes merging current master (i.e., 0.6.0rc1) into the security branch, which will then allow merging some of the PRs currently waiting into the security branch rather than into master, consolidate things and then merge the security support into master in preparation for a 0.7.0 release.

This is because of the pending 0.6.0 release as well as ROS2 Dashing still referencing the master branch for a couple of more days.


eboasson:
The commits flagged by the ECA check are all from PR #441, #460 and #478, where they passed the ECA check.

PR #513, opened 2020-05-13
Add security tests and fix some bugs in security implementation

dennis-adlink:
This PR adds a set of security tests and fixes a few bugs that were found when introducing these tests. 

The bugfixes in this PR:
- bb501ea The `decode_payload` function should return false in case encoded data is expected according to the security attributes and decoding fails
- 546a007 In `new_proxy_participant`, communication with a proxy participant should be set-up when there is any non-secured local participant. In case the remote participant allows unauthenticated participants in its security settings and advertises a non-secured topic, a reader of the local non-secure participant can read this data. This is fixed by adding the condition `!q_omg_participant_is_secure (pp)`. 
- b8a28dd Fix reader-writer synchronization in access control expiry_multiple test

New tests added (the test descriptions are also included as code comments):
- 93657cd / ad8dc89 Test discovery and liveliness protection by checking that `encode_datawriter_submessage` is called for `SEDP_BUILTIN_PUBLICATIONS_SECURE_WRITER` and `P2P_BUILTIN_PARTICIPANT_MESSAGE_SECURE_WRITER` depending on the discovery and liveliness protection settings in security configuration. 
- 269aaa8 Testing handshake result for any combination of protection kind values for rtps, discovery, liveliness, metadata (submsg) and payload protection settings. In all cases that have an encoding mismatch, the security handshake is expect to fail. 
- dc2071f Test read/write access control by running test cases with different combinations of allow and deny rules for publishing and subscribing a topic, and check correct working of the default policy.
- 4708b76 Test that all attributes and tokens retrieved from the access control plugin are correctly returned.
- 12f2e9a Check that communication for a topic that is allowed in the permissions config keeps working in case the publisher also creates a writer for a non-allowed topic
- d67436c Test communication for a non-secure participant with a secure participant that allows unauthenticated nodes in its governance configuration. Samples for a secured topic should not be received by a reader in the non-secure participant; samples for a non-secure topic should.
- 9080154 Validate that non-encrypted data will not be received by a reader that has protection enabled for rtps/submsg/payload. The test uses a crypto plugin wrapper mode that force the plugin to write plain data in the encoded output buffer to DDSI, ignoring the security attributes for the reader and writer.
- ba25f98 Check that all tokens that are sent to a remote participant are received correctly by that participant and the token-data stored in the remote participant is equal to the data in the token that was sent.

### DIFF
--- a/src/security/core/tests/common/access_control_wrapper.c
+++ b/src/security/core/tests/common/access_control_wrapper.c
@@ -12,6 +12,7 @@
 #include <string.h>
 #include <stdio.h>
 #include "dds/dds.h"
+#include "dds/ddsrt/circlist.h"
 #include "dds/ddsrt/heap.h"
 #include "dds/ddsrt/sync.h"
 #include "dds/ddsrt/string.h"
@@ -27,7 +28,8 @@ enum ac_plugin_mode {
   PLUGIN_MODE_ALL_OK,
   PLUGIN_MODE_WRAPPED,
   PLUGIN_MODE_NOT_ALLOWED,
-  PLUGIN_MODE_MISSING_FUNC
+  PLUGIN_MODE_MISSING_FUNC,
+  PLUGIN_MODE_CHECK_RETURNS,
 };
 
 enum ac_plugin_not_allowed {
@@ -56,6 +58,12 @@ enum ac_plugin_not_allowed {
 #define NOT_ALLOWED_REMOTE_READER_RELAY_ONLY      (1u << NOT_ALLOWED_ID_REMOTE_READER_RELAY_ONLY)
 #define NOT_ALLOWED_REMOTE_PERM                   (1u << NOT_ALLOWED_ID_REMOTE_PERM)
 
+struct returns_log_data {
+  struct ddsrt_circlist_elem e;
+  void * obj;
+};
+
+
 /**
  * Implementation structure for storing encapsulated members of the instance
  * while giving only the interface definition to user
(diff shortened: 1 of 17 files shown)


src/security/core/tests/common/security_config_test_utils.c
```diff
@@ -46,38 +58,43 @@ static const char *governance_xml =
...
+
+static const char *permissions_xml_sub =
+    "        <subscribe>"
+    "          <topics><topic>${TOPIC_NAME}</topic></topics>"
+    "          <partitions><partition>*</partition></partitions>"
+    "        </subscribe>";
+
+static const char *permissions_xml_allow_rule =
+    "      <allow_rule>"
+    "        <domains>${DOMAIN_ID:+<id>}${DOMAIN_ID:-<id_range><min>0</min><max>230</max></id_range>}${DOMAIN_ID:+</id>}</domains>"
```
eboasson, line 81: Clever :)

src/security/core/tests/common/cryptography_wrapper.c
```diff
@@ -109,6 +125,219 @@ void set_entity_data_secret(struct dds_security_cryptography_impl * impl, const
...
+
+  ddsrt_mutex_lock (&impl->token_data_lock);
+  if (ddsrt_circlist_isempty (&impl->encode_decode_log))
+  {
+    ddsrt_mutex_unlock (&impl->token_data_lock);
+    return tokens;
+  }
+
+  struct ddsrt_circlist_elem *elem0 = ddsrt_circlist_oldest (&impl->token_data_list), *elem = elem0;
+  while (elem != NULL)
```
eboasson, line 238: `elem` can't be a null pointer here, I'd just a `do { ... } while (elem != elem0);`

src/security/core/tests/common/cryptography_wrapper.c
```diff
@@ -109,6 +125,219 @@ void set_entity_data_secret(struct dds_security_cryptography_impl * impl, const
...
+{
+  assert (data_len <= CRYPTO_TOKEN_SIZE);
+  ddsrt_mutex_lock (&impl->token_data_lock);
+  if (ddsrt_circlist_isempty (&impl->encode_decode_log))
+  {
+    ddsrt_mutex_unlock (&impl->token_data_lock);
+    return NULL;
+  }
+  struct ddsrt_circlist_elem *elem0 = ddsrt_circlist_oldest (&impl->token_data_list), *elem = elem0;
+  while (elem != NULL)
```
eboasson, line 263: See `get_crypto_tokens`

src/security/core/tests/common/cryptography_wrapper.c
```diff
@@ -109,6 +125,219 @@ void set_entity_data_secret(struct dds_security_cryptography_impl * impl, const
...
+  return NULL;
+}
+
+static void log_encode_decode (struct dds_security_cryptography_impl * impl, enum crypto_encode_decode_fn function, DDS_Security_long_long handle)
+{
+  ddsrt_mutex_lock (&impl->encode_decode_log_lock);
+  if (!ddsrt_circlist_isempty (&impl->encode_decode_log))
+  {
+    struct ddsrt_circlist_elem *elem0 = ddsrt_circlist_oldest (&impl->encode_decode_log), *elem = elem0;
+    while (elem != NULL)
```
eboasson, line 293: See `get_crypto_tokens`

src/security/core/tests/common/cryptography_wrapper.c
```diff
@@ -109,6 +125,219 @@ void set_entity_data_secret(struct dds_security_cryptography_impl * impl, const
...
+  ddsrt_mutex_unlock (&impl->encode_decode_log_lock);
+}
+
+struct crypto_encode_decode_data * get_encode_decode_log (struct dds_security_cryptography_impl * impl, enum crypto_encode_decode_fn function, DDS_Security_long_long handle)
+{
+  ddsrt_mutex_lock (&impl->encode_decode_log_lock);
+  if (!ddsrt_circlist_isempty (&impl->encode_decode_log))
+  {
+    struct ddsrt_circlist_elem *elem0 = ddsrt_circlist_oldest (&impl->encode_decode_log), *elem = elem0;
+    while (elem != NULL)
```
eboasson, line 322: See `get_crypto_tokens`

src/security/core/tests/common/access_control_wrapper.c
```diff
@@ -63,8 +73,75 @@ struct dds_security_access_control_impl {
...
+  printf("log obj %p\n", obj);
+  ddsrt_mutex_unlock (&impl->returns_log_lock);
+}
+
+static void unregister_return_obj (struct dds_security_access_control_impl * impl, void * obj)
+{
+  assert(impl->mode == PLUGIN_MODE_CHECK_RETURNS);
+  ddsrt_mutex_lock (&impl->returns_log_lock);
+  struct ddsrt_circlist_elem *elem0 = ddsrt_circlist_oldest (&impl->returns_log), *elem = elem0;
+  while (elem != NULL)
```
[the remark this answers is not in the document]
dennis-adlink: I've fixed this, and the other loops over circular lists in crypto and access control plugins, in 7da139a.

src/security/core/tests/common/access_control_wrapper.c
```diff
@@ -63,8 +73,75 @@ struct dds_security_access_control_impl {
...
   uint32_t not_allowed_mask;
+  ddsrt_mutex_t returns_log_lock;
+  struct ddsrt_circlist returns_log;
+  bool invalid_return;
 };
 
+
+static void init_returns_log(struct dds_security_access_control_impl *impl)
+{
+  ddsrt_mutex_init (&impl->returns_log_lock);
```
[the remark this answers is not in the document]
dennis-adlink: `impl` is memset to 0 after allocation, but it's more clear to set this field to `false` explicitly, so I've added that in 7da139a.

src/security/core/tests/common/access_control_wrapper.c
```diff
@@ -63,8 +73,75 @@ struct dds_security_access_control_impl {
...
+    }
+    elem = elem->next;
+    if (elem == elem0)
+      break;
+  }
+  impl->invalid_return = true;
+  ddsrt_mutex_unlock (&impl->returns_log_lock);
+}
+
+bool check_returns (struct dds_security_access_control_impl * impl)
```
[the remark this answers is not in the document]
dennis-adlink: You're right, it's confusing, I changed it in 7da139a.

src/security/core/tests/common/access_control_wrapper.c
```diff
@@ -63,8 +73,75 @@ struct dds_security_access_control_impl {
   dds_security_access_control *instance;
   enum ac_plugin_mode mode;
   uint32_t not_allowed_mask;
+  ddsrt_mutex_t returns_log_lock;
+  struct ddsrt_circlist returns_log;
+  bool invalid_return;
 };
 
+
+static void init_returns_log(struct dds_security_access_control_impl *impl)
```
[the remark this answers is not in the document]
dennis-adlink: Done in 7da139a, and I've removed the specific test case for returns-checking as this is now checked in other tests that use the access control test wrapper in wrapper mode. This also revealed a bug in `is_topic_discovery_protected` where the return function call was missing for the topic attributes (which was not really an issue because the return implementation is currently a no-op).

dennis-adlink:
@eboasson Thanks for reviewing! I've force-pushed 221c8fd to remove the printfs that I added in the returns-logging in the access control plugin wrapper (these produced too much output and as a result it became hard to read other test output). I've also added 6ae0d30, which contains the fix (I think) for a race condition that could occur when permissions expire which causes the participant to be deleted, at the same the participant is deleted in the application (this race condition caused test failures in the access control test `permissions_expiry_multiple`)

eboasson:
Thanks @dennis-adlink, I agree with the changes. Indeed, before the introduction of permissions expiry there would never be multiple threads competing to delete the participant (because of how the API deals with entity deletion), but now that has changed.

PR #514, opened 2020-05-15
Add EDL 1.0 license to package.xml

rotu:
This package is dual-licensed under both EPL 2.0 and EDL 1.0. The latter was omitted from package.xml by mistake.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -6,6 +6,7 @@
     <description>Eclipse Cyclone DDS is a very performant and robust open-source DDS implementation. Cyclone DDS is developed completely in the open as an Eclipse IoT project.</description>
     <maintainer email="cyclonedds-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
     <license>Eclipse Public License 2.0</license>
+    <license>Eclipse Distribution License 1.0</license>
 
     <url type="website">https://projects.eclipse.org/projects/iot.cyclonedds</url>
     <url type="bugtracker">https://github.com/eclipse-cyclonedds/cyclonedds/issues</url>


eboasson:
Good catch! Unfortunately the ECA check fails and the system now tells me there's no valid ECA on file. Something must have changed in the recent past because that there definitely was one a little while ago. Perhaps you can check, @rotu ?

rotu:
Yeah. So the DCO check on various repos checks my real email (dan at digilabs dot up) but the ECA check is looking for my github masked email (rotu@noreply.github.com).

It’s all very silly.

eboasson:
> Yeah. So the DCO check on various repos checks my real email (dan at digilabs dot io) but the ECA check is looking for my github masked email ([rotu@noreply.github.com](mailto:rotu@noreply.github.com)).
> 
> It’s all very silly.

It says "Anonymous (dan@****.io)" is not covered by the necessary legal agreements in order to proceed. So I think there must be something else going on. When I put in the address by hand, it says there isn't any. I have seen the github masked ones confuse it, but this doesn't seem to be one of those days.

Can you please double check the eclipse website tells you you have a valid ECA when you log in? If it tells you all is well and it tells me it is not, then I need to file a bug against the ECA checkers.

rotu:
I figured it out and it *is* a bug in the ECA checker. My Eclipse email has a tag in it (`dan+eclipse@`) but the ECA email does not have that tag (`dan@`). Despite the below notice, it is in fact already checking against the address in my Account Settings, *not* the email I used when signing the ECA.

image

rotu:
submitted https://bugs.eclipse.org/bugs/show_bug.cgi?id=563237

eboasson:
Thanks for figuring *that* out, @rotu !

PR #522, opened 2020-05-20
Link with pthread

mauropasse:
This PR fixes the issue of undefined references when trying to cross-compile for RaspberryPi

```
--- stderr: cyclonedds                                                                                                                                                                                             
You have called ADD_LIBRARY for library ddsc without any source files. This typically indicates a problem with your CMakeLists.txt file
CMake Warning:
  Manually-specified variables were not used by the project:

    THIRDPARTY


/root/sysroot/lib/arm-linux-gnueabihf/libpthread.so.0: undefined reference to `__libc_dl_error_tsd@GLIBC_PRIVATE'
make[2]: *** [bin/ddsls] Error 1
make[1]: *** [src/tools/ddsls/CMakeFiles/ddsls.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/root/sysroot/lib/arm-linux-gnueabihf/libpthread.so.0: undefined reference to `__libc_dl_error_tsd@GLIBC_PRIVATE'
make[2]: *** [bin/pubsub] Error 1
make[1]: *** [src/tools/pubsub/CMakeFiles/pubsub.dir/all] Error 2
make: *** [all] Error 2
```

### DIFF
--- a/src/tools/ddsls/CMakeLists.txt
+++ b/src/tools/ddsls/CMakeLists.txt
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 add_executable(ddsls ddsls.c)
-target_link_libraries(ddsls ddsc)
+target_link_libraries(ddsls ddsc pthread)
 
 install(
   TARGETS ddsls
--- a/src/tools/ddsls/CMakeLists.txt
+++ b/src/tools/ddsls/CMakeLists.txt
@@ -10,7 +10,14 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 add_executable(ddsls ddsls.c)
-target_link_libraries(ddsls ddsc pthread)
+
+if(NOT WITH_FREERTOS)
+  set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+  find_package(Threads REQUIRED)
+  target_link_libraries(ddsls PUBLIC ddsc Threads::Threads)
+elseif()
+  target_link_libraries(ddsls ddsc)
+endif()
 
 install(
   TARGETS ddsls
--- a/src/tools/ddsperf/CMakeLists.txt
+++ b/src/tools/ddsperf/CMakeLists.txt
@@ -12,7 +12,7 @@
 
 idlc_generate(ddsperf_types ddsperf_types.idl)
 add_executable(ddsperf ddsperf.c cputime.c cputime.h netload.c netload.h)
-target_link_libraries(ddsperf ddsperf_types ddsc)
+target_link_libraries(ddsperf ddsperf_types ddsc pthread)
 
 if(WIN32)
   target_compile_definitions(ddsperf PRIVATE _CRT_SECURE_NO_WARNINGS)
--- a/src/tools/ddsperf/CMakeLists.txt
+++ b/src/tools/ddsperf/CMakeLists.txt
@@ -12,7 +12,14 @@
 
 idlc_generate(ddsperf_types ddsperf_types.idl)
 add_executable(ddsperf ddsperf.c cputime.c cputime.h netload.c netload.h)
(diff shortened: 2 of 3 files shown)


eboasson:
Thanks @mauropasse for looking into this and opening a PR for it. As it stands, with this change it can no longer build on Windows (nor on some other non-tier-1 platforms like FreeRTOS) because those don't use pthreads. @k0ekk0ek might have some ideas on a good solution to the platform-specific side of it.

Secondly, I'm rather surprised that adding `-lpthread` would solve the problem that `libpthread.so.0` relies on an undefined symbol: one would think that linking with it twice won't magically make the symbol defined ... but perhaps it is not simply a single shared object and/or there are linker scripts involved. I don't have a RPi with Raspbian at hand, or I would try.

mauropasse:
Hi @eboasson 
You don't need a RPi to test this, as it's in a compilation stage. You can use the iRobot performance framework with cross-compile support for raspbian: https://github.com/irobot-ros/ros2-performance/tree/master/cross-compiling
Regarding adding -lpthread, that's how I found out it worked. I got the raw compilation command failing, added -lpthread and didn't complain any more. Strange, but works.

dgoel:
The correct way to link with threads library in a platform agnostic way is this:

```cmake
find_package(Threads REQUIRED)
target_link_libraries(ddsls PUBLIC Threads::Threads)
```

For linux, this would link against pthread library.

k0ekk0ek:
What @dgoel proposes is indeed the correct way to handle it (`CMakeLists.txt` in ddsrt target line 233). It's a problem with FreeRTOS though. ddsrt (and ddsc`) should propagate the `Threads::Threads` target, if they don't, that's the problem we need to fix.

@mauropasse, let me know if you need help.

mauropasse:
I did the changes to link threads in an agnostic way, not sure if the `elsif`s were necessary.

eboasson:
@mauropasse my apologies for failing to follow up on this. I think @k0ekk0ek would know of a way to eliminate the FreeRTOS check (I imagine one could make `find_package(Threads REQUIRED)` work on on all platforms by wrapping it, and that would also avoid having to do `set(CMAKE_THREAD_PREFER_PTHREAD TRUE)` wherever we use it.

There is also a paperwork issue, the Eclipse Contributor Agreement check fails. So before we would be able to accept it, you'd have to "sign" that agreement as well. See https://www.eclipse.org/legal/ECA.php for the details.

k0ekk0ek:
Completely looked over this. My apologies. No "module", except for `ddsrt` should have to check whether or not it is being built for FreeRTOS. That doesn't scale well. The idea is that `ddsrt` propagates the necessary flags. The problem might be that `FindThreads.cmake` doesn't work as it should when cross compiling and we're fixing it here at the very last step when the executables are linked. Might be worth checking why the first warning is thrown by CMake (the library without source files one). If possible, maybe you can provide some instructions/details on your cross compile environment? That would allow me to give it a go myself.

mauropasse:
Hi @k0ekk0ek , you can reproduce the issue with the following commands:
```
git clone https://github.com/irobot-ros/ros2-performance  ~/ros2-performance
cd ~/ros2-performance/cross-compiling
source env.sh raspbian
bash get_sysroot.sh
bash build.sh
bash get_ros2_sources.sh --distro=master --ros2-path=${HOME}/ros2_cc_ws
bash ignore_pkgs.sh ~/ros2_cc_ws/ master
bash cc_workspace.sh ~/ros2_cc_ws/
```

k0ekk0ek:
Looks easy enough. I'll see if I can run some tests in the coming days.

LanderU:
Hi! I tested the changes proposed on this PR solve the issue for me as well.

alsora:
Hi, as of today this issue is still present with the current Galactic branch of CycloneDDS.

thijsmie:
As far as I can tell, this has since been solved. If someone still runs into new issues, please open a new issue.

PR #527, opened 2020-05-25
Merge security into master

### LINKED ISSUE #132: OMG DDS Security
ADLink has decided to donate their Vortex OpenSplice DDS Security implementation to the Cyclone DDS project. However, that will not be a simple code drop.

The first preliminary steps have been taken and a number of tasks have been identified. For now, it's still on a different fork. But it is the planning to do regular pull requests to Cyclone DDS. Possibly on a separate 'security' branch to not interfere with the implementations on the master branch too much.

For more detailed information, please see https://github.com/eclipse-cyclonedds/cyclonedds/blob/security/docs/dev/dds_security_effort.md
(peviously https://github.com/martinbremmer/cyclonedds/blob/mpt/docs/dev/dds_security_effort.md)

eboasson:
I think it is time, waiting until it is perfect makes no sense. This closes #132.


kyrofa:
Missing a ref to #132?

eboasson:
> Missing a ref to #132?

Yes, definitely!

eboasson:
Thanks @nuclearsandwich ! The error seems to be related to using OpenSSL from a DLL on Windows: https://stackoverflow.com/questions/38933943/openssl-code-works-natively-but-as-dll-causes-openssl-uplink-no-openssl-applink . Our own CI statically links OpenSSL and we never thought of it.

Supposedly one should include "openssl/applink.h" in any process using OpenSSL in a DLL, and do so exactly once in the program. I'm sure we can manage that 🙂 It would be good to know if you that somewhere in the ROS2 CI tests, because that would explain why those tests do pass.

I should a start movement for the eradication of Windows from this universe. I can't guarantee it will bring peace to all (and stop world hunger and pandemics and whatnot) but it surely would eliminate one factor preventing it. Unless all the time and energy wasted on satisfying its demands while tearing one's hair out and inventing new curses is time not spent getting mad at other people 🤔😁

PR #530, opened 2020-05-29
Entity enable

TheFixer:
This PR implements delayed enabling of participants, publishers, subscribers, readers and writers using `dds_enable()` call. Delayed enabling allows qos changes to entities that before they are enabled, and that would result in DDS_RETCODE_IMMUTABLE_POLICY after the entity has been enabled.

In particular, this PR contains the following

1. The `entity_factory.autoenable_created_entities` QoS is implemented for participants, publishers and subscribers. This QoS indicates whether child entities for these must be automatically enabled (default) or disabled when created.
2. Implementation of the API call `dds_enable()` to enable entities that have been created disabled. This operation is a noop when applied to an already enabled entity
3. Delayed creation of readers and writers in ddsi so that disabled readers and writers are not discovered
4. A bunch of tests cases to validate the implementation of this feature

The following issues are still present:

1. Since there is no factory for participants in cyclone that has the autoenable qos it is currently impossible to create a disabled participant. Ofcourse this functionality can be built, but I just did not do it (yet)
2. And perhaps worse, there is one situation were I still encounter problems. This has to do with deleting a participant while there are disabled entities, this gives a timeout. There are test cases added to demonstrate this problem.


eboasson:
@TheFixer I'm sorry it is taking so long to review. It hasn't slipped through the cracks, it is just that it needs and deserves a careful look.

TheFixer:
No problem, @eboasson . I am still struggling a bit with properly cleaning up entities that have not been enabled yet. I suggest I keep struggling  for a while while you spent your valuable time on more urgent issues.

TheFixer:
@eboasson  This PR implements delayed enabling of entities.  A more detailed description of what has been done can be found in https://github.com/TheFixer/cyclonedds/commit/18f39fca458f365981b34e5928e6768148e0fe64. Would it be possible to review this PR?

TheFixer:
It has been a few months ago that I spent a considerable amount of time to attempt to implement the dds_enable() and get it running on the (then current) master. So far I have not received any comments. Therefore, I am about to close this pull request.

eboasson:
@TheFixer I am very sorry that it got ignored for so long. I don't know why I missed this:

> @eboasson This PR implements delayed enabling of entities. A more detailed description of what has been done can be found in TheFixer@18f39fc. Would it be possible to review this PR?

but it sure hasn't helped that it got lost somewhere in either my mailbox or my memory. As to the relevance of the PR, it is a rather important bit considering that content filtering is something that should happen at the reader level (rather than via "content filtered topics") and that the "enable" functionality is a key part of making it possible to set a content a filter on a reader before it starts receiving data. (The alternative would be setting it via a QoS, but that has some other downsides, at least at this time.)

So allow me to reopen the PR and make a priority of doing the review it deserves. We'll see how large the impact is of the changes that have been merged into master since then, but as they don't involve major changes to the basics of creating or deleting entities, I have good hopes it won't be too bad.

And next time, if you don't hear anything and are wondering what's going on, please don't wait so long and just ping me!

eboasson:
@TheFixer the set of conflicts appeared small enough to make it worthwhile rebasing it before doing the review (the advantage is that I have keep track of fewer versions in my head), and so I did just that. I've made my rebased version available at https://github.com/eboasson/cyclonedds/tree/thefixer-entity-enable. The tests still pass locally.

A quick look at the changes I made suggests that it won't affect the review in a major way, so I can give review comments on the current version of this PR without complicating things. However, I think it would make the most sense to update the PR first, provided you agree with the way I resolved the conflicts in the rebased version.

TheFixer:
Thanks @eboasson for taking the effort. I'll have a look at the rebase  but don't expect any issues. I let you know when the updated pull request is available.

TheFixer:
@eboasson I inspected the changes caused by the rebase and did not see any problems. The PR is now updated.

eboasson:
@TheFixer thanks, that makes doing the review a bit easier. (And leaves me no excuse for procrastination!)

src/core/ddsc/src/dds_reader.c
```diff
@@ -497,70 +571,46 @@ static dds_entity_t dds_create_reader_int (dds_entity_t participant_or_subscribe
...
+  if (!dds_entity_creation_allowed (rc))
+    goto err_not_allowed;
   return reader;
 
-#ifdef DDS_HAS_SECURITY
 err_not_allowed:
-  thread_state_asleep (lookup_thread_state ());
-#endif
+  (void) dds_delete (reader);
+  goto del_implicit_sub;
```
eboasson, line 605: I'm not against `goto` for error handling, but this moves just a little too far in the direction of "unstructured" programming to my taste!

(Also for the writer.)

src/core/ddsc/src/dds_publisher.c
```diff
@@ -38,14 +38,30 @@ static dds_return_t dds_publisher_status_validate (uint32_t mask)
...
+{
+  struct dds_publisher *pub = (struct dds_publisher *) e;
+
+  assert (dds_entity_kind (e) == DDS_KIND_PUBLISHER);
+
+  /* calling enable on an entity whose factory is not enabled
+   * returns PRECONDITION_NOT_MET */
+  if ((pub->m_entity.m_parent != NULL) && !dds_entity_is_enabled(pub->m_entity.m_parent))
+    return DDS_RETCODE_PRECONDITION_NOT_MET;
+  pub->m_entity.m_flags |= DDS_ENTITY_ENABLED;
```
eboasson, line 51: These tests, setting the `ENABLED` flag and propagating to children, are a fundamental part of enabling entities. So in my view, these particular parts of enabling an entity should be part of the generic `dds_enable`, and not be part of the entity kind-specific functions.

src/core/ddsc/src/dds_writer.c
```diff
@@ -193,26 +199,33 @@ static void dds_writer_close (dds_entity *e)
...
 
 static dds_return_t dds_writer_delete (dds_entity *e) ddsrt_nonnull_all;
 
 static dds_return_t dds_writer_delete (dds_entity *e)
 {
   dds_writer * const wr = (dds_writer *) e;
-  /* FIXME: not freeing WHC here because it is owned by the DDSI entity */
+
+  /* not freeing WHC here because it is owned by the DDSI entity */
+#if 0
```
eboasson, line 225: I don't see any value in leaving this #if 0/#endif block in.

src/security/core/tests/common/test_utils.c
```diff
@@ -501,6 +501,9 @@ void rd_wr_init_fail(
       goto fail;
     dds_set_status_mask (*rd, DDS_SUBSCRIPTION_MATCHED_STATUS);
   }
+  dds_delete_qos (qos);
```
eboasson, line 504: These three lines don't add anything of value, it means the same code is executed under all circumstances. I prefer the simpler variant.

src/security/core/tests/secure_communication.c
```diff
@@ -448,7 +448,7 @@ CU_TheoryDataPoints(ddssec_secure_communication, multiple_readers) = {
     CU_DataPoints(size_t, 1, 3, 1, 3), /* number of participants per domain */
     CU_DataPoints(size_t, 3, 1, 3, 3), /* number of readers per participant */
 };
-CU_Theory((size_t n_dom, size_t n_pp, size_t n_rd), ddssec_secure_communication, multiple_readers, .timeout = 90, .disabled = false)
+CU_Theory((size_t n_dom, size_t n_pp, size_t n_rd), ddssec_secure_communication, multiple_readers, .timeout = 120, .disabled = false)
```
eboasson, line 451: I don't understand why implementing `dds_enable` would affect the duration of this test. I'd prefer not to increase the timeout if there is no need, but I'm open to the possibility that 90s is just a bit short.

eboasson (commented):
@TheFixer as was to be expected, the review results in some comments. Some are really minor matters, some are more substantial.

What I haven't checked is whether indeed all functions that should fail if the entity they are invoked isn't enabled yet correctly return that error. I'll do that in a second round, I suppose.

Regarding the tests, I cannot help but wonder whether it would not be better to collect all "enable" tests in a single file and structure them a bit differently because a large overlap in the set of checks for the different entity kinds, and a bit of generic code along with some small functions to construct entities would likely result in significantly less code covering exactly same the ground.

Finally, I think that especially for the reader/writer tests it would make sense to also check that the built-in topics come out as they should (generated at the correct time, correct instance handles, stuff like that) and that remote nodes indeed observe the discovery events as a consequence of enabling the entity. (The latter is pretty obvious looking at the code, but the former not so much.)

TheFixer:
@eboasson Thanks for the review. I'll rework the the PR and let you you when ready.

PR #532, opened 2020-06-04
Some plist handling things

eboasson:
* This fixes the use of an uninitialized “durability service” setting in the SPDP writer QoS (and only that one writer). The updated code adds a few assertions to verify this doesn’t happen again by accident, and it memsets empty ddsi_plist_t and dds_qos_t (as well as finalized ones) to garbage to reduce the likelihood further.
* This also adds an early-out to ddsi_plist_fini and ddsi_xqos_fini: these are called for every data sample that arrives and so are meant to be very cheap in the common case where there is nothing to be done. This tiny change has a small but real effect on the CPU load when receiving many samples per second. (Actually, this change is where it all started ...)
* For good measure, I’ve added some documentation comments to all plist/xqos functions. At which point I noticed that there was an unused parameter in ddsi_plist_quickscan, so I removed it.


eboasson:
The build failure is definitely unrelated

PR #533, opened 2020-06-04
Drop macOS 10.12 build from Travis CI cron job

eboasson:
The macOS 10.12 build was put in because of ROS2 "Dashing" specified
10.12 as the supported version, but Eloquent and later specify
10.14. The relevance of this is no longer there because of Foxy. The
build itself took an inordinate amount of time with lots of warnings
about the platform being deprecated.


k0ekk0ek:
:+1:

PR #540, opened 2020-06-08
Remove from ddsrt dynlib and filesystem when building FreeRTOS

MarcelJordense:
The dynlib and filesystem functions are currently not supported on FreeRTOS. The FreeRTOS dependencies are removed from ddsrt.


k0ekk0ek:
Looks good to me @MarcelJordense. Thanks!

PR #541, opened 2020-06-08
Allow building without a C++ compiler

robinlinden:
If you don't explicitly specify which languages you're using in the CMake project call, it defaults to C and CXX and will fail if it can't find a valid C++ compiler.

See the LANGUAGES section under https://cmake.org/cmake/help/latest/command/project.html#options.

Before:
https://travis-ci.com/github/eclipse-cyclonedds/cyclonedds/jobs/346028302#L612

After:
https://travis-ci.com/github/robinlinden/cyclonedds/jobs/346172916#L616

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 cmake_minimum_required(VERSION 3.7)
-project(CycloneDDS VERSION 0.7.0)
+project(CycloneDDS VERSION 0.7.0 LANGUAGES C)
 
 # Set a default build type if none was specified
 set(default_build_type "RelWithDebInfo")
--- a/ports/freertos-posix/CMakeLists.txt
+++ b/ports/freertos-posix/CMakeLists.txt
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 cmake_minimum_required(VERSION 3.5)
-project(FreeRTOS-Sim VERSION 10.0.2.0)
+project(FreeRTOS-Sim VERSION 10.0.2.0 LANGUAGES C)
 
 include(GNUInstallDirs)
 


k0ekk0ek:
Hi @robinlinden! This is much appreciated. We'll have to see if this works everywhere. I remember I added this too at some point, but it failed on one (or more) platforms (probably). This was some time ago, so the makefile situation may have improved since then, but let me check on AppVeyor tomorrow with VS2015. If it passes, this certainly has my blessing. Might be that some ROS builds fail, but I'm sure I can find someone to give it a try. Thanks again.

robinlinden:
Ah, I didn't realise you ran tests on Appveyor too. I ran them all on Travis from my fork to make sure I wasn't completely wasting your time. :P No big deal if it's not possible to merge it; very few systems lack C++-compilers. Thank you for your time, and thank you for the fantastic project! Very happy with it. :)

k0ekk0ek:
[Windows builds](https://ci.appveyor.com/project/k0ekk0ek/cyclonedds/builds/33417578) seem to work for 2015 too. I'm fine with merging these changes and reverting them should other builds fail. Only thing I need to verify is if this doesn't break any ROS builds, I'll do so asap, but I'll have to setup an environment first.

k0ekk0ek:
Now that I think about it, I think some CMake checks needed it or something. We should check if it doesn't break anything with ROS. @rotu or @eboasson, perhaps one of you can make the modification manually and do a build Cyclone rmw build? I currently don't have a ROS2 build in place. Thanks in advance.

eboasson:
Other than a bunch of `extern “C”`s, there is no C++ whatsoever in there and I don’t think the ROS2 builds do anything strange. Stranger things have happened, though. I’ll try it, probably tomorrow.

eboasson:
It doesn't cause any problems for me with ROS2 (macOS, Linux and Windows, although I was lazy and the Linux and Windows builds were using sources of about a month ago). And before you ask, I did remember to throw away the Cyclone build directories to avoid contamination of the results by CMake's cache.

@k0ekk0ek feel free to merge it.

PR #543, opened 2020-06-11
Fix MSYS2 mingw compilation

robinlinden:
A bunch of tests pass, but the ones trying to load plugins are failing (maybe others too, but a vast majority of the test failures I got were `(plugins = load_plugins( NULL , NULL , &crypto )) != NULL`. I'm still looking into the test failures, but I figured I should open this early to check if it's even something you're interested in merging. :P

I replaced the ws2ipdef header with ws2tcpip as specified here:
https://docs.microsoft.com/en-us/windows/win32/api/ws2ipdef/ns-ws2ipdef-sockaddr_gen#remarks

Including it directly causes this error:
```
In file included from ../src/ddsrt/src/netstat/windows/netstat.c:12:
../../mingw64/x86_64-w64-mingw32/include/ws2ipdef.h:36:19: error: field 'Address' has incomplete type
   36 |   struct sockaddr Address;
```

The ifdefs were put in place since mingw doesn't implement __try and the exception hack has a nicer method nowadays anyway.
```
../src/ddsrt/src/threads/windows/threads.c:314:5: error: '__try' undeclared (first use in this function)
  314 |     __try
      |     ^~~~~
```

The MSG_TRUNC flag seems to live in mswsock.h for mingw.
```
../src/ddsrt/src/sockets/windows/socket.c:528:23: error: 'MSG_TRUNC' undeclared (first use in this function)
  528 |     msg->msg_flags |= MSG_TRUNC;
      |                       ^~~~~~~~~
```

And these last 2 simply don't exist in the mingw headers, so I defined them to the same value they are in the real Windows headers:
```
../src/core/ddsi/src/ddsi_udp.c: In function 'joinleave_ssm_mcgroup':
../src/core/ddsi/src/ddsi_udp.c:597:57: error: 'MCAST_JOIN_SOURCE_GROUP' undeclared (first use in this function)
  597 |     rc = ddsrt_setsockopt (socket, IPPROTO_IPV6, join ? MCAST_JOIN_SOURCE_GROUP : MCAST_LEAVE_SOURCE_GROUP, &gsr, sizeof (gsr));
      |                                                         ^~~~~~~~~~~~~~~~~~~~~~~
../src/core/ddsi/src/ddsi_udp.c:597:57: note: each undeclared identifier is reported only once for each function it appears in
../src/core/ddsi/src/ddsi_udp.c:597:83: error: 'MCAST_LEAVE_SOURCE_GROUP' undeclared (first use in this function)
  597 |     rc = ddsrt_setsockopt (socket, IPPROTO_IPV6, join ? MCAST_JOIN_SOURCE_GROUP : MCAST_LEAVE_SOURCE_GROUP, &gsr, sizeof (gsr));
      |                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~
```


eboasson:
Hi @robinlinden, I think it'd be really nice to have support for mingw, and these changes look entirely reasonable to me. I also notice that Travis makes it easy to integrate (https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2) so if you want to go adventurous, you could add it to the set of builds as well.

You might want to try building with `-DENABLE_SECURITY=NO`: then nothing that depends on loading DLLs is included and would be a pretty easy way of confirming your suspicion that it all has to do with that.

Finally, don't worry about the one build failure. There are a few tests that fail somewhat regularly, this is one of them. Most of these failures seem to be caused by really adverse timing on Travis causing some timeouts inside the tests to trigger, but I'm still hoping to get a better understanding of the failures before I modify the tests.

robinlinden:
Yeah, sure thing. I'll have a look at adding it to Travis. I had no idea they supported MSYS2, so that's cool. :) My ultimate goal is cross-compilation from Linux to Windows using mingw, but I got a bunch more errors doing that, so MSYS2 seemed like a good stepping stone.

k0ekk0ek:
@robinlinden, out of curiosity, did you manage to get cross compilation working?

@eboasson, with you approval I'd like to merge these changes. Probably best to get them in before we have to rebase :slightly_smiling_face:

robinlinden:
Sorry, I got sidetracked by other things during the summer and forgot about this. 😁 I will have a look at adding an msys2 job to Travis and fixing the cross-compilation.

k0ekk0ek:
@robinlinden, one of the fixes that went in this week for `cyclonedds-cxx` caused a conflict. I updated your PR to resolve it and then squashed the commits. If you find time to fix tests failures (don't feel obligated), I hope you will consider submitting them in a separate PR. Thanks!

PR #544, opened 2020-06-12
Improve security encryption and decryption functions

MarcelJordense:
A number of unnecessary memory allocations when splitting a received message into it's components are removed. Further the encoding and decoding of the rtps message contained a number of unnecessary memory copies which are removed. The InfoSource submessage which is contained in encoded rtps messages is now passed through.
Further the input checks of the main functions are replaced by asserts.


src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -1630,93 +1271,79 @@ static DDS_Security_boolean encode_rtps_message_encrypt (
...
-    body = add_submessage(&data, SMID_SEC_BODY_KIND, flags, submsg_len);
-    encrypted = (struct encrypted_data *)(body + 1);
-    contents = encrypted->data;
+    encrypted_data.base = body->content.data;
+    encrypted_data.length = secure_body_plain_size;
 
     /* encrypt message */
-    /* FIXME: improve performance by not copying plain_rtps_message to a new buffer (crypto_cipher_encrypt_data should allow encrypting parts of a message) */
-    if (!crypto_cipher_encrypt_data(&session->key, session->key_size, header->session_id, secure_body_plain._buffer, secure_body_plain_size, NULL, 0, contents, &payload_len, &hmac, ex))
+    if (!crypto_cipher_encrypt_data(&session->key, session->key_size, &header->prefix.iv, 2, plain_data, &encrypted_data, &hmac, ex))
```
eboasson, line 1264: There's something to be said for doing `sizeof(plain_data)/sizeof(plain_data[0])`, it didn't take me long to figure out what that `2` means and why it is correct, but I do like how a `sizeof` here would be somewhat self-documenting. (Same holds for some similar places.)

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -1630,93 +1271,79 @@ static DDS_Security_boolean encode_rtps_message_encrypt (
...
     /* the transformation_kind indicates only indicates authentication the determine HMAC */
-    if (!crypto_cipher_encrypt_data(&session->key, session->key_size, header->session_id, NULL, 0, secure_body_plain._buffer, secure_body_plain_size, NULL, NULL, &hmac, ex))
+    if (!crypto_cipher_encrypt_data(&session->key, session->key_size, &header->prefix.iv, 2, plain_data, NULL, &hmac, ex))
       goto enc_rtps_fail_data;
 
     /* copy submessage */
-    memcpy(contents, secure_body_plain._buffer, secure_body_plain_size);
-    payload_len = secure_body_plain_size;
-    data._length += payload_len;
+    memcpy(ptr, plain_data[0].base, plain_data[0].length);
```
eboasson, line 1283: But then, if one decides to not hard-code `2`, this ought to get a static assertion ...

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -2293,93 +1780,75 @@ decode_serialized_payload(
...
-  transform_id = CRYPTO_TRANSFORM_ID(header.transform_identifier.transformation_key_id);
-
-  if (!crypto_factory_get_remote_writer_key_material(factory, reader_id, writer_id, transform_id, &writer_master_key, NULL, &basic_protection_kind, ex))
+  if (!crypto_factory_get_remote_writer_key_material(factory, reader_id, writer_id, estate.prefix.transform_id, &writer_master_key, NULL, &basic_protection_kind, ex))
     goto fail_prepare;
 
   /* calculate the session key */
-  if (!initialize_remote_session_info(&remote_session, &header, writer_master_key->master_salt, writer_master_key->master_sender_key, writer_master_key->transformation_kind, ex))
+  if (!initialize_remote_session_info(&remote_session, &estate.prefix, writer_master_key->master_salt, writer_master_key->master_sender_key, writer_master_key->transformation_kind, ex))
     goto fail_decrypt;
```
eboasson, line 1725: `plain_data.base` is uninitialised at this point

src/security/builtin_plugins/cryptographic/src/crypto_cipher.c
```diff
@@ -18,105 +18,83 @@
...
-  {
-    if (data_len > INT_MAX)
-    {
-      DDS_Security_Exception_set(ex, DDS_CRYPTO_PLUGIN_CONTEXT, DDS_SECURITY_ERR_CIPHER_ERROR, 0, "EVP_EncryptUpdate to update data failed: data_len exceeds INT_MAX");
-      goto fail_encrypt;
-    }
-
-    /* encrypt the message */
-    if (!EVP_EncryptUpdate(ctx, encrypted, &len, data, (int) data_len))
+    if (!EVP_EncryptUpdate(ctx, ptr, &len, inpdata[i].base, (int) inpdata[i].length))
```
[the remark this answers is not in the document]
MarcelJordense: The OpenSSL documentation is unclear about the EVP_EncryptUpdate function.  We use two modes AES-GCM (encrypt and message authentication) or AES-GMAC (message authentication only). When used for GMAC only the message authentication code is computed and in that case the EVP_EncryptUpdate function is called with a NULL pointer for the output argument. Note that the outpdata argument is now used to determining if GCM or GMAC is requested.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -103,656 +108,502 @@ typedef struct dds_security_crypto_transform_impl
...
-  sz = CRYPTO_HMAC_SIZE + sizeof(uint32_t) + len * sizeof(struct receiver_specific_mac);
-  ft = ddsrt_malloc(sz);
+  Header_t *dst = crypto_buffer_append(buffer, RTPS_MESSAGE_HEADER_SIZE);
+  *dst = *src;
+  return dst;
+}
 
-  memcpy(ft->common_mac.data, common_mac.data, CRYPTO_HMAC_SIZE);
-  ft->receiver_specific_macs._length = len;
+static void * crypto_buffer_find_submessage(crypto_buffer_t *buffer, uint8_t submessageId, size_t offset)
```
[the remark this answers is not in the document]
MarcelJordense: The crypto_buffer_t structure is currently only used in the context of encrypting and is thus only used for data sent and not for data received from the network.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -1320,187 +1126,72 @@ encode_datareader_submessage(
...
+    struct secure_prefix *prefix,
+    struct secure_postfix *postfix,
     CryptoObjectKind_t kind,
     DDS_Security_Handle rmt_handle,
     const char *context,
     DDS_Security_SecurityException *ex)
 {
   bool result = false;
   master_key_material *keymat = NULL;
+  crypto_data_t data = { .base = postfix->common_mac.data, .length = CRYPTO_HMAC_SIZE };
```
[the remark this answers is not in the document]
MarcelJordense: In this function only the authentication tag associated with the receiving entity is checked. Thus the function first searches in the list of received authentication tags the tag that is associated with the receiving entity and determines the corresponding receiver specific key information which is used to recalculate the authentication code. Thus only one hmac is being computed.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -555,14 +579,14 @@ static bool read_secure_rtps_body(unsigned char **ptr, unsigned char *endp, DDS_
...
     body->data.base = *ptr;
 
-    if (!read_subnsg_header(ptr, endp, SMID_INFO_SRC, &smhdr, &bswap))
+    if (!read_submsg_header(ptr, endp, SMID_INFO_SRC, &smhdr, &bswap))
       return false;
 
     *ptr += smhdr.octetsToNextHeader;
 
-    while ((uintptr_t)*ptr + sizeof(smhdr) < (uintptr_t)endp)
+    while (sizeof(smhdr) < (size_t)(endp - *ptr))
```
eboasson, line 587: Shouldn't this be ≤?

I also wonder which submessage kinds can be encountered here, and whether there `octetsToNextHeader` fields are always straightforward. In particular, I'm thinking of `PAD` and `SENTINEL`, and about the special case of `octetsToNextHeader` being 0 (for submessages other than `PAD`).

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -353,11 +346,11 @@ static bool initialize_remote_session_info(remote_session_info *info, struct sec
   return crypto_calculate_session_key(&info->key, info->id, master_salt, master_key, transformation_kind, ex);
 }
 
-static bool read_subnsg_header(unsigned char **ptr, unsigned char *endp, uint8_t smid, SubmessageHeader_t *hdr, bool *bswap)
+static bool read_submsg_header(unsigned char **ptr, unsigned char *endp, uint8_t smid, SubmessageHeader_t *hdr, bool *bswap)
 {
   SubmessageHeader_t*smhdr = (SubmessageHeader_t *)*ptr;
 
-  if ((uintptr_t)*ptr + RTPS_SUBMESSAGE_HEADER_SIZE > (uintptr_t)endp)
+  if (RTPS_SUBMESSAGE_HEADER_SIZE > (size_t) (endp - *ptr))
```
eboasson, line 353: Something I didn't spot the first time round: `octetsToNextHeader` can be garbage, but some of the callers trust it. I think it would be best to check it here (and that unfortunately also means dealing with the case of it being 0), and remove some of then-superfluous checks at some of the call sites.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -1191,9 +1229,17 @@ static DDS_Security_boolean encode_rtps_message_encrypt (
   size_t size;
   uint32_t transform_kind, transform_id;
 
+  DDSRT_STATIC_ASSERT(num_segs == 3);
```
eboasson, line 1232: Is this really accepted by all compilers? It could be, but I am not sure and the CI wasn't run for this commit (I suppose because it needs to be rebased first)

eboasson (commented):
Thanks @MarcelJordense, I noticed a few oddities around the `octetsToNextHeader` handling (apologies for not spotting them earlier) and the CI run is missing, presumably because it needs to be rebased to current master first. Could you please do that and have a look at the comments I added?

eboasson (commented):
One of the GCC-based builds failed because c26105318 accidentally got dropped in the rewrite. I'm certain it is a gcc bug to give that warning, but it needs to be addressed nonetheless. I'm willing to believe that the one test failing with a timeout on one windows build is unrelated, that could even be an environmental problem.

(Do check out `git range-diff`: I won't claim to really fully understand its behaviour, but
```
git range-diff `git merge-base master update_crypto_old`..update_crypto_old `git merge-base master FETCH_HEAD`..FETCH_HEAD
```
where those "old" things are local versions of the commits I kept around and FETCH_HEAD is the updated one.

eboasson:
@MarcelJordense I am sure everybody by now knows that sometimes I have an absolutely horrific latency in dealing with PRs, one so long it could even make one wonder whether one could ever trust the result of my work to have acceptable latency. Anyway, I have had another look at this PR, rebased it, and think I found three small imperfections.

- what I believe to be an incorrect assertion,
- what I believe to be an off-by-4 error in input validation;
- what I believe to be an incorrect way of dealing with the prefix submessage size: the code assumes that the submessage is no longer than necessary because it doesn't always skip forward using `octetsToNextHeader`.

I also wanted to be absolutely certain that some code only ever processed trusted inputs, for example `crypto_buffer_find_submessage`. Rather than going over all code paths, I decided to try and make the compiler do the checking for me by splitting buffer types into "trusted" and "tainted" buffers. No issues uncovered, but I would like to invite you to see if you think the changes I made in the process make sense.

https://github.com/eboasson/cyclonedds/tree/marcel-update-crypto-pr544 is the same as this PR after rebasing (which was trivial) and with the above points addressed. The first two are separate commits to make it easy to spot them. The submessage-skipping logic is hidden in the refactoring commit at the end. Also gcc 10 gave me a bit of a fight again, hence the final commit.

If you agree with the changes, I'll squash that final commit into the preceding one and ask you update this PR to that version so that your efforts are correctly reflected in GitHub's statistics. If you tell me not to worry about all that and don't have the time or inclination to look at it, that's fine, too. But please let me know!

MarcelJordense:
@eboasson Thank you for the review and implementing the changes. I agree with the changes you have made.

eboasson:
@MarcelJordense, thanks! I have squashed the trivial fixes and improved my commit message for the non-trivial changes I did, and update my branch (the diff between what it was and what it now is, is empty). If you update your branch/this PR to that version (i.e., 8f4b0a0b), then I can merge it into master.

eboasson:
Thanks @MarcelJordense!

PR #547, opened 2020-06-15
Allow building without a C++ compiler on CMake versions less than 3.12

### LINKED ISSUE #546: Unable to build Cyclone DDS on Ubuntu 18.04
I try to build Cyclone DDS on Ubuntu 18.04.
However, I find that there is the error message while building.

```bash
$ git clone https://github.com/eclipse-cyclonedds/cyclonedds.git
$ cd cyclonedds
$ mkdir build && cd build
$ cmake ..

......(ignore some message)

-- Found Maven: /usr/bin/mvn (found suitable version "3.6.0", minimum required is "3.0") 
-- Found Java: /usr/bin/java (found suitable version "11.0.7", minimum required is "1.8") 
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY
CMake Error at /usr/share/cmake-3.10/Modules/CheckCXXSourceCompiles.cmake:97 (try_compile):
  Unknown extension ".cxx" for file

    /home/chenying/workspace/cyclonedds/build/CMakeFiles/CMakeTmp/src.cxx

  try_compile() works only for enabled languages.  Currently these are:

    C

  See project() command to enable other languages.
  /usr/share/cmake-3.10/Modules/CheckCXXCompilerFlag.cmake:49 (CHECK_CXX_SOURCE_COMPILES)
  /usr/share/cmake-3.10/Modules/GenerateExportHeader.cmake:215 (check_cxx_compiler_flag)
  /usr/share/cmake-3.10/Modules/GenerateExportHeader.cmake:374 (_test_compiler_hidden_visibility)
  src/security/builtin_plugins/access_control/CMakeLists.txt:23 (generate_export_header)


-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY - Failed
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY

......(ignore some message)
```

I guess the error is related to [the issue](https://gitlab.kitware.com/cmake/cmake/-/issues/16967) in CMAKE project. It's fixed in [the PR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/1799/diffs).
Maybe CMAKE in Ubuntu 18.04 haven't had the bugfix yet.

k0ekk0ek:
This would be my proposal to fix #546. It simply uses an updated version of `GenerateExportHeader.cmake` with versions less than 3.12. It may be that we need to do a CQ for this perhaps?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,9 @@ ENDFUNCTION(PREPEND)
 
 # Set module path before defining project so platform files will work.
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/Modules")
+if(CMAKE_VERSION VERSION_LESS_EQUAL 3.12)
+  set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_CURRENT_LIST_DIR}/cmake/Modules/Override")
+endif()
 set(PROJECT_NAME_FULL "Eclipse Cyclone DDS")
 
 # Set some convenience variants of the project-name
--- a/cmake/Modules/Override/GenerateExportHeader.cmake
+++ b/cmake/Modules/Override/GenerateExportHeader.cmake
@@ -1,0 +1,443 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+GenerateExportHeader
+--------------------
+
+Function for generation of export macros for libraries
+
+This module provides the function ``GENERATE_EXPORT_HEADER()``.
+
+The ``GENERATE_EXPORT_HEADER`` function can be used to generate a file
+suitable for preprocessor inclusion which contains EXPORT macros to be
+used in library classes::
+
+   GENERATE_EXPORT_HEADER( LIBRARY_TARGET
+             [BASE_NAME <base_name>]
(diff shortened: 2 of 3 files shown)


k0ekk0ek:
It BSD-3-clause, so shouldn't be a problem license wise...

k0ekk0ek:
Updated to commit message to include the CQ

eboasson:
It would nonetheless still require a CQ. More importantly, I'm not convinced it is a good idea to add so much CMake bloat for a bit of backwards compatibility, the only real advantage over my hack is that the `project` line remains as expected and that can affect other project's tooling.

k0ekk0ek:
The CQ is requested. As for the fix. I agree that it's quite a bit of code for something so trivial, but they differ quite a bit. Your solution adds C++ for CMake versions less than 3.12, mine adds an extra include directory for versions less than 3.12. Note that the added module isn't used for newer CMake versions

eboasson:
Of the four options I see, after a bit of thought, I think this PR is the most appropriate one:
* tell Ubuntu 18.04 users to upgrade their CMake: that's not going to make us any friends;
* my hack (use a variable for languages): I'm concerned that it might confuse tooling;
* revert the change and advertise C++: that's just wrong;
* your proposal: the only part of it I don't like is the amount of CMake stuff that needs to be incorporated, but it is it means _users_ of Cyclone get the best behaviour.

k0ekk0ek:
Dropped in favor of #550.

PR #551, opened 2020-06-19
 Modified flags in typetree to include unsupported types, and to use separate flags to indicate signedness and wideness.

e-hndrks:


k0ekk0ek:
Thanks @e-hndrks. This is more like how imagined the flags field originally. If the checks pass, I'll happily merge.

k0ekk0ek:
The macOS builds fail because the image is too old. It'll be resolved once master is merged into xtypes.

PR #552, opened 2020-06-25
Deal with OpenSSL in a DLL on Windows

eboasson:
OpenSSL doesn't support using BIOs of the "fd" or "file" type when it is built as a DLL and the executable didn't provide it with access to the executable's CRT. Requiring all applications that wish to use security to worry about this "applink.c" thing is too onerous a requirement.

* Check for the existence of "applink.c" in the OpenSSL include directory, adding it to the security tests if it exists.  This way, all of OpenSSL can be used by the tests.
* Include it in the security core and built-in plugin tests.  This way, the test code can use the entirety of OpenSSL.
* In the authentication and access-control plugins, load X509 and private keys from files by first reading them into a "mem" type BIO, then reading them from that BIO.
* Take care not to call ddsrt_free on OpenSSL-allocated memory, either by calling OPENSSL_free, or by allocating the memory using ddsrt_malloc and letting OpenSSL fill that buffer.

On my local setup this fixes https://github.com/ros2/rmw_cyclonedds/issues/201 and https://github.com/eclipse-cyclonedds/cyclonedds/pull/527#issuecomment-642210391 but it needs confirmation from a ROS2 CI run first.

### DIFF
--- a/cmake/Modules/FindOpenSSL.cmake
+++ b/cmake/Modules/FindOpenSSL.cmake
@@ -32,3 +32,11 @@ else()
   endif()
 endif()
 
+# OpenSSL DLL on Windows: use of BIO_s_fd and BIO_s_file (directly or indirectly) requires
+# the executable to incorporate OpenSSL applink.c.  CMake 18 adds support for handling
+# this as part of the OpenSSL package, but we can't require such a new CMake version.
+if(OPENSSL_FOUND AND EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/applink.c")
+  set(CYCLONEDDS_OPENSSL_APPLINK "${OPENSSL_INCLUDE_DIR}/openssl/applink.c")
+else()
+  set(CYCLONEDDS_OPENSSL_APPLINK "")
+endif()
--- a/src/security/builtin_plugins/tests/CMakeLists.txt
+++ b/src/security/builtin_plugins/tests/CMakeLists.txt
@@ -59,7 +59,11 @@ set(security_crypto_test_sources
     "set_remote_participant_crypto_tokens/src/set_remote_participant_crypto_tokens_utests.c"
 )
 
-add_cunit_executable(cunit_security_plugins ${security_auth_test_sources} ${security_ac_test_sources} ${security_crypto_test_sources})
+add_cunit_executable(cunit_security_plugins
+  ${security_auth_test_sources}
+  ${security_ac_test_sources}
+  ${security_crypto_test_sources}
+  ${CYCLONEDDS_OPENSSL_APPLINK})
 
 target_include_directories(
   cunit_security_plugins PRIVATE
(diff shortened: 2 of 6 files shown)


eboasson:
ROS 2 CI is clean

cmake/Modules/FindOpenSSL.cmake
```diff
@@ -32,3 +32,11 @@ else()
   endif()
 endif()
 
+# OpenSSL DLL on Windows: use of BIO_s_fd and BIO_s_file (directly or indirectly) requires
+# the executable to incorporate OpenSSL applink.c.  CMake 18 adds support for handling
```
k0ekk0ek, line 36: Maybe post a link to that in the PR? If it's a module, we could opt to copy that module in and override. Just an idea.

src/security/builtin_plugins/tests/CMakeLists.txt
```diff
@@ -59,7 +59,16 @@ set(security_crypto_test_sources
     "set_remote_participant_crypto_tokens/src/set_remote_participant_crypto_tokens_utests.c"
 )
 
-add_cunit_executable(cunit_security_plugins ${security_auth_test_sources} ${security_ac_test_sources} ${security_crypto_test_sources})
+add_cunit_executable(cunit_security_plugins
+  ${security_auth_test_sources}
+  ${security_ac_test_sources}
+  ${security_crypto_test_sources}
+  ${CYCLONEDDS_OPENSSL_APPLINK})
```
k0ekk0ek, line 66: With this it will "parse" the source for CUnit test signatures. It'll not be a problem in practice, but you could opt to add the source file to the target afterwards.

cmake/Modules/FindOpenSSL.cmake
```diff
@@ -32,3 +32,11 @@ else()
...
+# this as part of the OpenSSL package, but we can't require such a new CMake version.
+if(OPENSSL_FOUND AND EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/applink.c")
+  set(CYCLONEDDS_OPENSSL_APPLINK "${OPENSSL_INCLUDE_DIR}/openssl/applink.c")
+else()
+  set(CYCLONEDDS_OPENSSL_APPLINK "")
```
k0ekk0ek, line 41: I think that if you don't set `CYCLONEDDS_OPENSSL_APPLINK` you can check with `if (NOT CYCLONEDDS_OPENSSL_APPLINK` etc. (little "did you know...", you're certainly not required to verify or change)

cmake/Modules/FindOpenSSL.cmake
eboasson: This is what I saw: https://cmake.org/cmake/help/v3.18/module/FindOpenSSL.html

Given that we have some hacks to deal with Conan-provided OpenSSL already and that this works reliably on the platforms we support, I think it makes the most sense to merge it as-is and deal with those hacks later. That way the ROS 2 CI problems should be addressed.

Since you approved it already, I'm sure you're ok with that :)

PR #555, opened 2020-06-26
Fix long latency for large messages

### LINKED ISSUE #484: Reduce how eager CycloneDDS is in retransmits
We've been testing out CycloneDDS with ROS 2 Foxy prerelease (commit hash 45c0f432a9c72fddf1ae0cce4761e8a92cb567e4 of this repository).  Our particular test is an stereo image to pointcloud pipeline.   There is a trigger at one end; when the trigger topic is published, it causes two images to be generated from simulated Gazebo cameras.  Each of those images is 640x480x3 (921600 bytes).  Those two images are published to a node to do disparity processing, and then that output is published to a node to generate a pointcloud.  Finally, the pointcloud is published and we subscribe to print out parts of it.  That final pointcloud is 9830400 bytes.  There's an example of this pipeline at https://github.com/jacobperron/pc_pipe (along with a README on how to run).

When we run this pipeline with CycloneDDS (all on the same machine), we see that most of the time, the pipeline finishes fairly quickly (< 300 ms).  However, sometimes it takes the pipeline a *very* long time to complete; I've had to wait up to 100 seconds to get a final pointcloud.  When I instrumented, I found that most of that time seems to be spent in the publication of the large final pointcloud.

I discussed this with @eboasson offline, and he recommended that I try some tuning to narrow down the problem.  Towards that end, I increased my /proc/sys/net/core/rmem_max to 30MB, and then instructed CycloneDDS to ask for very large receive buffers with the following XML:

```
<?xml version="1.0" encoding="UTF-8" ?>
<CycloneDDS xmlns="https://cdds.io/config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://cdds.io/config https://raw.githubusercontent.com/eclipse-cyclonedds/cyclonedds/master/etc/cyclonedds.xsd">
    <Domain id="any">
        <Internal>
            <MinimumSocketReceiveBufferSize>20MB</MinimumSocketReceiveBufferSize>
        </Internal>
    </Domain>
</CycloneDDS>
```

With that in place, the long delays are absolutely gone; in 2000 tries, I never had to wait more than 400ms from the initial trigger to the final pointcloud.

Now, the above is a workaround; it requires both root (something not everyone has), and also requires that you know a priori what your largest message size is.  After further discussion with Erik, it seems like Cyclone is being a bit too eager in its retransmissions.  I believe he mentioned that someone is looking into this, but this issue is just to open the problem for visibility.  Please let me know if you need any more information from me about the setup, or additional testing.  I'm happy to provide it.

eboasson:
This fixes #484, which turns out to _not_ be an over-eager retransmission policy, but caused by dropping the heartbeat transmit rate while still retransmitting fragments. It is still not really good enough, but the real crazy long delays are gone. (As an aside, the over-eager policy does exist, but it shows up on, e.g., WiFi. A follow-up PR will bring some improvements there.)

Pipeline latency difference for the https://github.com/jacobperron/pc_pipe demo that triggered the issue (measured on an Ubuntu 20.04 VM under macOS, using Foxy and with a fix in place to use reliable communications in the test):

image

This also addresses some issues in handling retransmit requests of large messages. This is only somewhat related, but whatever ... For example, if the reader temporarily loses the connection with the writer (but not vice versa) — not very likely to happen in well-behaved networks if the writer is sending all the time, but a dirty WiFi is a different story — the resetting of the "count" fields in the retransmit requests cause those to be ignored (as per the spec). This PR adds a timeout after which they are accepted again, bounding the duration of the hiccup.

It also eliminates out some undefined behaviour by changing these counts to unsigned integers (the spec says signed ... but that doesn't mean one has to implement them as signed integers).

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_entity.h
+++ b/src/core/ddsi/include/dds/ddsi/q_entity.h
@@ -131,9 +131,10 @@ struct wr_prd_match {
   seqno_t last_seq; /* highest seq send to this reader used when filter is applied */
   int32_t num_reliable_readers_where_seq_equals_max;
   ddsi_guid_t arbitrary_unacked_reader;
-  nn_count_t next_acknack; /* next acceptable acknack sequence number */
-  nn_count_t next_nackfrag; /* next acceptable nackfrag sequence number */
+  nn_count_t prev_acknack; /* latest accepted acknack sequence number */
+  nn_count_t prev_nackfrag; /* latest accepted nackfrag sequence number */
   ddsrt_etime_t t_acknack_accepted; /* (local) time an acknack was last accepted */
+  ddsrt_etime_t t_nackfrag_accepted; /* (local) time a nackfrag was last accepted */
   struct nn_lat_estim hb_to_ack_latency;
   ddsrt_wctime_t hb_to_ack_latency_tlastlog;
   uint32_t non_responsive_count;
@@ -154,7 +155,7 @@ struct pwr_rd_match {
   ddsi_guid_t rd_guid;
   ddsrt_mtime_t tcreate;
   nn_count_t count; /* most recent acknack sequence number */
-  nn_count_t next_heartbeat; /* next acceptable heartbeat (see also add_proxy_writer_to_reader) */
+  nn_count_t prev_heartbeat; /* latest accepted heartbeat (see also add_proxy_writer_to_reader) */
   ddsrt_wctime_t hb_timestamp; /* time of most recent heartbeat that rescheduled the ack event */
   ddsrt_etime_t t_heartbeat_accepted; /* (local) time a heartbeat was last accepted */
   ddsrt_mtime_t t_last_nack; /* (local) time we last sent a NACK */  /* FIXME: probably elapsed time is better */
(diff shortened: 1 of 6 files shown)


eboasson:
Merged as part of #558.

huntjw:
@eboasson it appears that eboasson/cyclonedds@7269058 wasn't merged with #558.  Is there a reason?  We still see the need to set rmem_max for large messages with release 0.7.0.  I am guessing it is because this fix is missing.

eboasson:
> @eboasson it appears that [eboasson/cyclonedds@7269058](https://github.com/eboasson/cyclonedds/commit/7269058) wasn't merged with #558. Is there a reason? We still see the need to set rmem_max for large messages with release 0.7.0. I am guessing it is because this fix is missing.

It got dropped because I concluded that the change in cd37e9d would ultimately result in `writer_hbcontrol_note_asyncwrite` getting called and resetting `hbs_since_last_write` in the cases where the changes of 7269058 reset it. So basically that it was superfluous.

Your observation makes me now seriously doubt my conclusion ... I can't look into it *right now* but this obviously goes to the top of the list of urgent issues.

I take it you are seeing crazy long latencies at basically 0% network load? That is the telltale symptom for this problem. If you have the time for a small experiment: this should change the heartbeat timing settings and basically disable the rate reduction, and should be a workaround:
```XML
<Internal>
  <HeartbeatInterval max="100ms">100ms</HeartbeatInterval>
<Internal>
```
I checked that it reports the change in maximum, I dodn't have time right now to set up a proper experiment ...

huntjw:
@eboasson<mailto:notifications@github.com> Thanks for the quick response and the explanation.  We have determined that the developer was actually using version 0.5.0 when he experienced this problem.  We will try updating to 0.7.0 to see if the problem is resolved.

eboasson:
Thanks for letting me know, @huntjw. That reduces the urgency a bit. Please do keep me posted.

PR #556, opened 2020-06-26
Replace flex by handcrafted scanner

k0ekk0ek:
I've replaced the flex scanner with a handwritten one to:

* Correctly handle a subset of preprocessor grammer to parse `#line` and `#pragma` directives as well as line continuation sequences; `#line` directives are communicated by mcpp to inform the compiler of the location, `#pragma keylist` directives are required for backwards compatibility with the original IDL compiler (and OpenSplice).

* Treat `cr+lf`, `lf+cr`, `cr` and `lf` sequences as a single newline.

The handwritten scanner also makes it significantly easier to track locations (doing that in flex required all kinds of nasty macro tricks) and it's not even that much larger. There's still some classes missing for IDL, but those will be added when I continue working on the IDL parser. I've also ditched `ddsts` as a module as it'll not be used for type comparison in `ddsi`. I'm planning to change the `ddsts_` prefix to `idl_` as I go along. More tests are required, but those too will be added at a later point in time.

@e-hndrks, @eboasson, I don't expect you to fully review this, but I'd like some feedback on the overall mechanism. Just to know if I overlooked something. If you could give it a quick glance, that would be very much appreciated.

Btw, the handling of line continuation sequences and the newline handling is there so that in theory we could use it as a library to generate types from IDL for dynamic languages at runtime (without support for `#include` and macro's, which aren't needed when you pull the code through some Python, etc function).


src/tools/idlc/src/idl.c
```diff
@@ -18,145 +18,210 @@
...
+      buf, sizeof(buf), "%s:%u:%u: ", loc->first.file, loc->first.line, loc->first.column);
+  else
+    cnt = snprintf(
+      buf, sizeof(buf), "%u:%u: ", loc->first.line, loc->first.column);
+
+  if (cnt == -1)
+    return;
+
+  off = (size_t)cnt;
+  cnt = vsnprintf(buf+off, sizeof(buf)-off, fmt, ap);
```
eboasson, line 86: This potentially gets you into trouble: if `cnt == sizeof(buf)`, the second argument of vsnprintf is 0 and that is, I think, a run-time constraint violation in C11 and undefined behaviour before that.

src/tools/idlc/src/idl.c
```diff
@@ -18,145 +18,210 @@
...
+
+  if (cnt == -1)
+    return;
+
+  off += (size_t)cnt;
+
+  if (off >= sizeof(buf))
```
eboasson, line 93: The proper termination of the string is guaranteed by the `snprintf` family

src/tools/idlc/src/idl.c
```diff
@@ -18,145 +18,210 @@
...
+    buf[sizeof(buf) - 1] = '\0';
+  else
+    buf[off] = '\0';
+
+  fprintf(stderr, "%s\n", buf);
```
eboasson, line 98: Consider `fputs`

src/tools/idlc/src/directive.c
```diff
@@ -0,0 +1,274 @@
...
+        return IDL_PARSE_ERROR;
+      }
+      if (!(dir = ddsrt_malloc(sizeof(*dir)))) {
+        return IDL_MEMORY_EXHAUSTED;
+      }
+      dir->directive.type = idl_line;
+      dir->line = (uint32_t)ullng;
+      dir->file = NULL;
+      proc->directive = (idl_directive_t *)dir;
+      proc->state++;
```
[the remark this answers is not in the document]
k0ekk0ek: Agreed, the state can use some improvement. I'm leaning towards a single enum, I'll have a look. As for the pulling in tokens, that would divert from the way Bison is used. It'd definitely be the way I'd normally do that (Bison does that by default too), but `mcpp` uses print functions to output the preprocessed sources. I could opt to catch all of that in a buffer and then use the "normal" way, but that'd potentially mean having a lot of information in memory if the user has a lot of IDL so I though switching to a push parser would be better. Handling the directives like that allows me to have a simple `if/else` construction to switch between parsing preprocessor directives and IDL code.

src/tools/idlc/src/directive.c
```diff
@@ -0,0 +1,274 @@
...
+  ddsrt_free(dir);
+  proc->directive = NULL;
+  return 0;
+}
+
+static int32_t
+parse_line(idl_processor_t *proc, idl_token_t *tok)
+{
+  idl_line_t *dir = (idl_line_t *)proc->directive;
+  enum { line_number = LINE, filename, extra_tokens, newline } state;
```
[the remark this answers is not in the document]
k0ekk0ek: done

src/tools/idlc/src/directive.c
```diff
@@ -0,0 +1,274 @@
...
+      if (end == tok->value.str || *end != '\0') {
+        idl_error(proc, &tok->location,
+          "invalid line number in #line directive");
+        return IDL_PARSE_ERROR;
+      }
+      if (!(dir = ddsrt_malloc(sizeof(*dir)))) {
+        return IDL_MEMORY_EXHAUSTED;
+      }
+      dir->directive.type = idl_line;
+      dir->line = (uint32_t)ullng;
```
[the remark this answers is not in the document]
k0ekk0ek: done

src/tools/idlc/src/directive.c
```diff
@@ -0,0 +1,274 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "idl.h"
+#include "parser.h"
+
+#include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/strtol.h"
+
+/* 8 bits reserved for directive parsing state.
```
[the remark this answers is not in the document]
k0ekk0ek: done

src/tools/idlc/src/directive.c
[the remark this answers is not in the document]
k0ekk0ek: Pushed a commit that switches state from a bitmask to an enum. I think it's more to you liking, if not, let me know.

src/tools/idlc/src/directive.c
```diff
@@ -0,0 +1,274 @@
...
+    /* expect line or pragma */
+    if (tok->code != IDL_TOKEN_IDENTIFIER) {
+      idl_error(proc, &tok->location, "invalid compiler directive");
+      return IDL_PARSE_ERROR;
+    } else if (strcmp(tok->value.str, "line") == 0) {
+      proc->state |= LINE;
+    } else if (strcmp(tok->value.str, "pragma") == 0) {
+      /* support #pragma keylist for backwards compatibility */
+      proc->state |= PRAGMA;
+    } else {
```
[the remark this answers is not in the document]
k0ekk0ek: Correct. A sole `#` gets ignored. `gcc`'s preprocessor might output extra information to the compiler, but feeding the aforementioned input to `gcc` (manually putting it in a file) will result in an `extra tokens at end of #line directive` warning, the subset of preprocessor login in this PR does that too. `mcpp` throws a `Excessive token sequence ...` error (btw, `idlc -E` only preprocesses the input with `mcpp`).

src/tools/idlc/src/idl.c
k0ekk0ek: done

src/tools/idlc/src/idl.c
k0ekk0ek: done

src/tools/idlc/src/idl.c
k0ekk0ek: Leaving it as is for now as `fputs` doesn't append a `\n`.

src/tools/idlc/src/idl.c
```diff
@@ -171,7 +181,7 @@ int32_t idl_parse(idl_processor_t *proc)
   do {
     if ((code = idl_scan(proc, &tok)) < 0)
       break;
-    if (proc->state & IDL_SCAN_DIRECTIVE)
+    if ((int)proc->state & (int)IDL_SCAN_DIRECTIVE)
```
eboasson, line 184: Bitwise operations are on `unsigned`s so I think I would cast to `unsigned` rather than `int` — it doesn't change the behaviour. I would consider defining an inline function or a macro next to the enum definition that tests whether it is in "directive scanning" mode, but I can live with this, too.

src/tools/idlc/src/scanner.c
```diff
@@ -367,10 +367,10 @@ scan_identifier(idl_processor_t *proc, const char *cur, const char **lim)
   if (cnt < 0)
     return IDL_NEED_REFILL;
   /* detect if scope is attached to identifier if scanning code */
-  if ((proc->state & IDL_SCAN_CODE) && (cnt = have(proc, cur, "::")) < 0)
+  if (((int)proc->state & (int)IDL_SCAN_CODE) && (cnt = have(proc, cur, "::")) < 0)
```
eboasson, line 370: See comment at `idl_parse`

src/tools/idlc/src/scanner.c
```diff
@@ -460,61 +460,58 @@ idl_lex(idl_processor_t *proc, idl_lexeme_t *lex)
...
-          else
-            code = (unsigned char)*cur;
-          lim = cur + 1;
-        } else {
-          lim = cur + 1;
-          code = (unsigned char)*cur;
-        }
-      } else if (chr == '#') {
-        proc->state |= IDL_SCAN_DIRECTIVE;
+    } else if ((int)proc->state & (int)IDL_SCAN_DIRECTIVE) {
```
eboasson, line 463: and again ...

src/tools/idlc/src/scanner.c
```diff
@@ -460,61 +460,58 @@ idl_lex(idl_processor_t *proc, idl_lexeme_t *lex)
...
+        code = scan_pp_number(proc, cur, &lim);
+      } else if ((cnt = have_alpha(proc, cur)) || chr == '_') {
+        code = scan_identifier(proc, cur, &lim);
+      } else if ((cnt = have_digit(proc, cur))) {
+        code = scan_pp_number(proc, cur, &lim);
+      } else {
         lim = cur + 1;
         code = (unsigned char)*cur;
+      }
+    } else if ((int)proc->state & (int)IDL_SCAN_CODE) {
```
eboasson, line 477: and again ...

k0ekk0ek:
@eboasson, I rebased this PR on the current `xtypes` branch (which was rebased on current `master`). I processed your latest comments, but folded everything back into one commit. To make it a bit easier for you to approve the changes, I provided a diff of the diff in the attachment. Basically, I changed the casts from `int` to `unsigned` as you suggested.

[diff.txt](https://github.com/eclipse-cyclonedds/cyclonedds/files/5017678/diff.txt)

PR #557, opened 2020-06-29
Replacement for os_sockWaitSet

reicheratwork:
The current os_sockWaitSet does not support notifying IP address changes.
In order to support this functionality a new struct ddsrt_event_queue is created which will be expanded to support this functionality.
The implementation of the ddsrt_event_queue has mirrored that of the os_sockWaitSet as much as possible in order to make replacing one with the other as effortless as possible.
This draft is created to elicit comments and suggestions regarding the interface in general and the implementation in specific.

* Created new ddsrt_event_queue struct which supports notifying users of data being received on sockets through ddsrt_event structs.
* Implementation for POSIX select type and Darwin kevent type notification.
* Added unittests for checking the newly created functionality.

Signed-off-by:  Martijn Reicher <martijn.reicher@adlinktech.com>

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/events.h
+++ b/src/ddsrt/include/dds/ddsrt/events.h
@@ -1,0 +1,181 @@
+/*
+ * Copyright(c) 2006 to 2020 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef DDSRT_EVENTS_H
+#define DDSRT_EVENTS_H
+
+#include "dds/ddsrt/retcode.h"
+#include "dds/ddsrt/sockets.h"
+#include "dds/ddsrt/atomics.h"
+
+#if __APPLE__
+#include "dds/ddsrt/events/darwin.h"
+#else
+#include "dds/ddsrt/events/posix.h"
+#endif
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+  /**
+  * @name Definitions for event trigger flags.
+  * These flags describe the different type of trigger events which can be watched for.
+  */
+///@{
+  /**< the flag is not set */
+#define DDSRT_EVENT_FLAG_UNSET 0u
+  /**< resource has data available for reading */
+#define DDSRT_EVENT_FLAG_READ (1u<<0) 
+  /**< resource can be written to */
+//#define DDSRT_EVENT_FLAG_WRITE (1u<<1) 
+  /**< resource has been opened */
+//#define DDSRT_EVENT_FLAG_OPEN (1u<<2) 
+  /**< resource has been closed */
+//#define DDSRT_EVENT_FLAG_CLOSE (1u<<3) 
+  /**< timeout has occurred on the resource */
+//#define DDSRT_EVENT_FLAG_TIMEOUT (1u<<4)
+  /**< notification of ip address change on the resource*/
+//#define DDSRT_EVENT_FLAG_IP_CHANGE (1u<<5)  
+///@}
+
+  /**
+  * @brief Describes the type of object being monitored for events.
+  */
+  enum ddsrt_event_type {
+    ddsrt_event_type_unset, /**< unitialized state*/
+    ddsrt_event_type_socket /**< indicating a socket type connection*/
+    /*,
+    ddsrt_event_type_file,
+    ddsrt_event_type_ifaddr*/
+  };
+  typedef enum ddsrt_event_type ddsrt_event_type_t;
+
+  /**
+  * @brief Structure containing the information of which object is being watched.
+  *
+  * Is used by ddsrt_event_queue to keep track of objects being watched.
+  */
+  struct ddsrt_event {
+    uint32_t flags; /**< Type of event being watched for, composited from DDSRT_EVENT_FLAG_$EVENT_TYPE$.*/
+    ddsrt_event_type_t type; /**< Type of object being watched.*/
+    ddsrt_atomic_uint32_t triggered; /**< Trigger status after a call to ddsrt_event_queue_wait.*/
+    union {
+      struct { ddsrt_socket_t sock; } socket;
+      /*struct ddsrt_event_file file;  this is for future expansion to be able to register events on files*/
+      /*struct ddsrt_event_ifaddr ifaddr; this is for future expansion to be able to register IP address changes*/
+    } data; /**< Container for the object being watched.*/
+  };
+
+  typedef struct ddsrt_event ddsrt_event_t;
+
+  typedef struct ddsrt_event_queue ddsrt_event_queue_t;
+  
+  /**
+  * @brief Initializes an existing event to indicate a socket.
+  *
+  * @param ev Pointer to the event to initialize.
+  * @param sock Socket to initialize the event with.
+  * @param flags Flags to set for the event.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags);
+
+  dds_return_t ddsrt_event_fini(ddsrt_event_t* ev);
+
+  /**
+  * @brief Event queue creation function.
+  *
+  * @returns Pointer to the created queue.
+  */
+  ddsrt_event_queue_t* ddsrt_event_queue_create(void);
+
+  /**
+  * @brief Cleans up an event queue.
+  *
+  * Will finish the event queue first, then free the memory of the queue.
+  *
+  * @param queue The queue to clean up.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_destroy(ddsrt_event_queue_t* queue);
+
+  /**
+  * @brief Getter for the number of stored events.
+  *
+  * @param queue The queue to get the number of events of.
+  *
+  * @returns The number of events stored by the queue.
+  */
+  size_t ddsrt_event_queue_nevents(ddsrt_event_queue_t* queue);
+
+  /**
+  * @brief Triggers a wait for events for the queue.
+  *
+  * @param queue The queue to trigger.
+  * @param reltime The maximum amount of time to wait.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t reltime);
+
+  /**
+  * @brief Interrupts a triggered wait of a queue.
+  *
+  * @param queue The queue to interrupt.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_signal(ddsrt_event_queue_t* queue);
+
+  /**
+  * @brief Adds an event to the queue.
+  * 
+  * This event will be stored by the queue, any additional actions such as registration to other services will also be done.
+  *
+  * @param queue The queue to add the event to.
+  * @param evt Pointer to the event to add.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_add(ddsrt_event_queue_t* queue, ddsrt_event_t* evt);
+
+  /**
+  * @brief Removes an event from the queue.
+  *
+  * This event will be removed from the queue, any additional actions such as deregistration from other services will also be done.
+  *
+  * @param queue The queue to remove the event from.
+  * @param evt Pointer to the event to remove.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_remove(ddsrt_event_queue_t* queue, ddsrt_event_t* evt);
+
+  /**
+  * @brief Gets the next triggered events from the queue.
+  *
+  * Will go over the stored events until it reaches one where any trigger flag is set.
+  * Successive calls to this function will start from the previously used point.
+  * Will be reset after a call to ddsrt_event_queue_wait.
+  *
+  * @param queue The queue to retrieve.
+  *
+  * @returns Pointer to the event which has a trigger flag set, NULL if none has this flag set.
+  */
+  ddsrt_event_t* ddsrt_event_queue_next(ddsrt_event_queue_t* queue);
+
+#if defined (__cplusplus)
+}
+#endif
+#endif /* DDSRT_EVENTS_H */
--- a/src/ddsrt/include/dds/ddsrt/events.h
+++ b/src/ddsrt/include/dds/ddsrt/events.h
@@ -82,15 +82,13 @@ extern "C" {
   /**
   * @brief Initializes an existing event to indicate a socket.
   *
-  * @param ev Pointer to the event to initialize.
-  * @param sock Socket to initialize the event with.
-  * @param flags Flags to set for the event.
+  * @param[out] ev Pointer to the event to initialize.
+  * @param[in] sock Socket to initialize the event with.
+  * @param[in] flags Flags to set for the event.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
   dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags);
-
-  dds_return_t ddsrt_event_fini(ddsrt_event_t* ev);
 
   /**
   * @brief Event queue creation function.
@@ -104,7 +102,7 @@ extern "C" {
   *
   * Will finish the event queue first, then free the memory of the queue.
   *
-  * @param queue The queue to clean up.
+  * @param[out] queue The queue to clean up.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
@@ -113,7 +111,7 @@ extern "C" {
   /**
   * @brief Getter for the number of stored events.
   *
-  * @param queue The queue to get the number of events of.
+  * @param[in] queue The queue to get the number of events of.
   *
   * @returns The number of events stored by the queue.
   */
@@ -122,8 +120,8 @@ extern "C" {
   /**
   * @brief Triggers a wait for events for the queue.
   *
-  * @param queue The queue to trigger.
-  * @param reltime The maximum amount of time to wait.
+  * @param[out] queue The queue to trigger.
+  * @param[in] reltime The maximum amount of time to wait.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
@@ -132,7 +130,7 @@ extern "C" {
   /**
   * @brief Interrupts a triggered wait of a queue.
   *
-  * @param queue The queue to interrupt.
+  * @param[out] queue The queue to interrupt.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
@@ -143,8 +141,8 @@ extern "C" {
   * 
   * This event will be stored by the queue, any additional actions such as registration to other services will also be done.
   *
-  * @param queue The queue to add the event to.
-  * @param evt Pointer to the event to add.
+  * @param[out] queue The queue to add the event to.
+  * @param[in] evt Pointer to the event to add.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
@@ -155,8 +153,8 @@ extern "C" {
   *
   * This event will be removed from the queue, any additional actions such as deregistration from other services will also be done.
   *
-  * @param queue The queue to remove the event from.
-  * @param evt Pointer to the event to remove.
+  * @param[out] queue The queue to remove the event from.
+  * @param[in] evt Pointer to the event to remove.
   *
   * @returns DDS_RETCODE_OK if everything went OK.
   */
@@ -169,7 +167,7 @@ extern "C" {
   * Successive calls to this function will start from the previously used point.
   * Will be reset after a call to ddsrt_event_queue_wait.
   *
-  * @param queue The queue to retrieve.
+  * @param[out] queue The queue to retrieve.
   *
   * @returns Pointer to the event which has a trigger flag set, NULL if none has this flag set.
   */
--- a/src/ddsrt/include/dds/ddsrt/events/darwin.h
+++ b/src/ddsrt/include/dds/ddsrt/events/darwin.h
@@ -1,0 +1,23 @@
+/*
+ * Copyright(c) 2006 to 2020 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+#ifndef DDSRT_EVENTS_DARWIN_H
+#define DDSRT_EVENTS_DARWIN_H
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+#if defined (__cplusplus)
+}
+#endif
+#endif /* DDSRT_EVENTS_DARWIN_H */
--- a/src/ddsrt/src/events.c
+++ b/src/ddsrt/src/events.c
@@ -1,0 +1,24 @@
+/*
+ * Copyright(c) 2006 to 2020 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+#include "dds/ddsrt/events.h"
+#include "dds/ddsrt/heap.h"
+
+dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags) {
+  assert(ev);
+  assert(sock);
+  ev->type = ddsrt_event_type_socket;
+  ev->flags = flags;
+  ev->data.socket.sock = sock;
+  ddsrt_atomic_st32(&ev->triggered, DDSRT_EVENT_FLAG_UNSET);
+  return DDS_RETCODE_OK;
+}
--- a/src/ddsrt/src/events/darwin/events_darwin.c
+++ b/src/ddsrt/src/events/darwin/events_darwin.c
@@ -63,7 +63,7 @@ static dds_return_t ddsrt_event_queue_init(ddsrt_event_queue_t* queue) {
   kevent kev;
   EV_SET(&kev, queue->interrupt[0], EVFILT_READ, EV_ADD, 0, 0, 0);
   assert(kevent(kq, &kev, 1, NULL, 0, NULL) != -1);
-#endif
+#endif /* !LWIP_SOCKET */
 
   /*create kevents array*/
   queue->kevents = ddsrt_malloc(sizeof(kevent) * queue->cevents);
@@ -96,7 +96,7 @@ static dds_return_t ddsrt_event_queue_fini(ddsrt_event_queue_t* queue) {
     EV_SET(&kev, interrupt[0], EVFILT_READ, EV_DELETE, 0, 0, 0);
   close(queue->interrupt[0]);
   close(queue->interrupt[1]);
-#endif
+#endif /* !LWIP_SOCKET */
   close(queue->kq);
 
   ddsrt_mutex_destroy(&queue->lock);
@@ -150,7 +150,7 @@ dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t r
 #if !defined(LWIP_SOCKET)
       char buf;
       read(queue->interrupt[0], &buf, 0);
-#endif
+#endif /* !LWIP_SOCKET */
     }
   }
   ddsrt_mutex_unlock(&queue->lock);
@@ -184,8 +184,8 @@ dds_return_t ddsrt_event_queue_signal(ddsrt_event_queue_t* queue) {
   char buf = 0;
   if (-1 == write(queue->interrupt[1], &buf, 1))
     return DDS_RETCODE_ERROR;
-#endif
-  return DDS_RETURN_OK;
+#endif /* !LWIP_SOCKET */
+  return DDS_RETCODE_OK;
 }
 
 dds_return_t ddsrt_event_queue_remove(ddsrt_event_queue_t* queue, ddsrt_event_t* evt) {
--- a/src/ddsrt/src/events/darwin/events_darwin.c
+++ b/src/ddsrt/src/events/darwin/events_darwin.c
@@ -1,0 +1,222 @@
+/*
+ * Copyright(c) 2006 to 2020 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "dds/ddsrt/events/darwin.h"
+#include "dds/ddsrt/events.h"
+#include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/sync.h"
+
+#define EVENTS_CONTAINER_DELTA 8
+
+struct ddsrt_event_queue {
+  ddsrt_event_t**         events;   /**< container for triggered events*/
+  size_t                  nevents;  /**< number of triggered events stored*/
+  size_t                  cevents;  /**< capacity of triggered events stored*/
+  size_t                  ievents;  /**< current iterator for getting the next triggered event*/
+  int                     kq;       /**< kevent polling instance*/
+  kevent*                 kevents;  /**< array which kevent uses to write back to, has identical size as this->events*/
+  ddsrt_mutex_t           lock;     /**< for add/delete */
+#if !defined(LWIP_SOCKET)
+  ddsrt_socket_t          interrupt[2]; /**< pipe for interrupting waits*/
+#endif
+};
+
+/**
+* @brief Initializes an event queue.
+*
+* Will set the counters to 0 and create the containers for triggers and additional necessary ones.
+*
+* @param[in] queue The queue to initialize.
+*
+* @returns DDS_RETCODE_OK if everything went OK.
+*/
+static dds_return_t ddsrt_event_queue_init(ddsrt_event_queue_t* queue) {
+  queue->nevents = 0;
+  queue->cevents = 8;
+  queue->ievents = 0;
+  ddsrt_mutex_init(&queue->lock);
+
+  /*create kevent polling instance */
+  queue->kq = kqueue();
+  assert(queue->kq != -1);
+  assert(fcntl(queue->kq, F_SETFD, fcntl(queue->kq, F_GETFD) | FD_CLOEXEC) != -1);
+
+  queue->events = ddsrt_malloc(sizeof(ddsrt_event_t*) * queue->cevents);
+  assert(queue->events);
+
+#if !defined(LWIP_SOCKET)
+  if (-1 == pipe(queue->interrupt))
+    return DDS_RETCODE_ERROR;
+  /*register interrupt event*/
+  kevent kev;
+  EV_SET(&kev, queue->interrupt[0], EVFILT_READ, EV_ADD, 0, 0, 0);
+  assert(kevent(kq, &kev, 1, NULL, 0, NULL) != -1);
+#endif
+
+  /*create kevents array*/
+  queue->kevents = ddsrt_malloc(sizeof(kevent) * queue->cevents);
+  assert(queue->kevents);
+
+  return DDS_RETCODE_OK;
+}
+
+/**
+* @brief Finishes an event queue.
+*
+* Will free created containers and do any additional cleanup of things created in ddsrt_event_queue_init.
+*
+* @param queue The queue to finish.
+*
+* @returns DDS_RETCODE_OK if everything went OK.
+*/
+static dds_return_t ddsrt_event_queue_fini(ddsrt_event_queue_t* queue) {
+  assert(queue);
+  /*deregister all currently stored triggers from kqueue*/
+  kevent kev;
+  for (unsigned int i = 0; i < queue->nevents; i++) {
+    ddsrt_event_t* evt = queue->events[i];
+    if (evt->flags & DDSRT_EVENT_FLAG_READ)
+      EV_SET(&kev, evt->data.socket, EVFILT_READ, EV_DELETE, 0, 0, evt);
+    assert(kevent(kq, &kev, 1, NULL, 0, NULL) != -1);
+  }
+#if !defined(LWIP_SOCKET)
+  if (evt->flags & DDSRT_EVENT_FLAG_READ)
+    EV_SET(&kev, interrupt[0], EVFILT_READ, EV_DELETE, 0, 0, 0);
+  close(queue->interrupt[0]);
+  close(queue->interrupt[1]);
+#endif
+  close(queue->kq);
+
+  ddsrt_mutex_destroy(&queue->lock);
+  ddsrt_free(queue->events);
+  ddsrt_free(queue->kevents);
+  return DDS_RETCODE_OK;
+}
+
+ddsrt_event_queue_t* ddsrt_event_queue_create(void) {
+  ddsrt_event_queue_t* returnptr = ddsrt_malloc(sizeof(ddsrt_event_queue_t));
+  assert(returnptr);
+  ddsrt_event_queue_init(returnptr);
+  return returnptr;
+}
+
+dds_return_t ddsrt_event_queue_destroy(ddsrt_event_queue_t* queue) {
+  assert(queue);
+  ddsrt_event_queue_fini(queue);
+  ddsrt_free(queue);
+  return DDS_RETCODE_OK;
+}
+
+size_t ddsrt_event_queue_nevents(ddsrt_event_queue_t* queue) {
+  size_t ret;
+  ddsrt_mutex_lock(&queue->lock);
+  ret = queue->nevents;
+  ddsrt_mutex_unlock(&queue->lock);
+  return ret;
+}
+
+dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t reltime) {
+  assert(queue);
+
+  /*reset triggered status*/
+  ddsrt_mutex_lock(&queue->lock);
+  queue->ievents = 0;
+  for (unsigned int i = 0; i < queue->nevents; queue++)
+    ddsrt_atomic_st32(&queue->events[i]->triggered, DDSRT_EVENT_FLAG_UNSET);
+  ddsrt_mutex_unlock(&queue->lock);
+
+  struct timespec tmout = { milliseconds / 1000,     /* waittime (seconds) */
+                           milliseconds * 1000000 }; /* waittime (nanoseconds) */
+  int nevs = kevent(queue->kq, NULL, 0, queue->kevents, queue->nevents, &tmout);
+  ddsrt_mutex_lock(&queue->lock);
+  for (int i = 0; i < nevs; i++) {
+    ddsrt_event_t* evt = queue->kevents[i].udata;
+    if (NULL != evt) {
+      ddsrt_atomic_st32(evt->triggered, DDSRT_EVENT_FLAG_READ);
+    }
+    else {
+#if !defined(LWIP_SOCKET)
+      char buf;
+      read(queue->interrupt[0], &buf, 0);
+#endif
+    }
+  }
+  ddsrt_mutex_unlock(&queue->lock);
+
+  return DDS_RETCODE_OK;
+}
+
+dds_return_t ddsrt_event_queue_add(ddsrt_event_queue_t* queue, ddsrt_event_t* evt) {
+  assert(queue);
+  assert(evt);
+
+  ddsrt_mutex_lock(&queue->lock);
+  if (queue->nevents == queue->cevents) {
+    queue->cevents += EVENTS_CONTAINER_DELTA;
+    ddsrt_realloc(queue->events, queue->events, sizeof(ddsrt_event_t*) * queue->cevents);
+    ddsrt_realloc(queue->kevents, queue->kevents, sizeof(kevent) * queue->cevents);
+  }
+  queue->events[queue->nevents++] = evt;
+  ddsrt_mutex_unlock(&queue->lock);
+
+  /*register to queue->kq*/
+  kevent kev;
+  EV_SET(&kev, evt->data.socket.sock, EVFILT_READ, EV_ADD, 0, 0, evt);
+  assert(kevent(kq, &kev, 1, NULL, 0, NULL) != -1);
+
+  return DDS_RETCODE_OK;
+}
+
+dds_return_t ddsrt_event_queue_signal(ddsrt_event_queue_t* queue) {
+#if !defined(LWIP_SOCKET)
+  char buf = 0;
+  if (-1 == write(queue->interrupt[1], &buf, 1))
+    return DDS_RETCODE_ERROR;
+#endif
+  return DDS_RETURN_OK;
+}
+
+dds_return_t ddsrt_event_queue_remove(ddsrt_event_queue_t* queue, ddsrt_event_t* evt) {
+  ddsrt_mutex_lock(&queue->lock);
+  for (unsigned int i = 0; i < queue->nevents; i++) {
+    if (queue->events[i] == evt) {
+      /*deregister from queue->kq*/
+      kevent kev;
+      EV_SET(&kev, evt->data.socket.sock, EVFILT_READ, EV_DELETE, 0, 0, evt);
+      assert(kevent(kq, &kev, 1, NULL, 0, NULL) != -1);
+
+      memmove(queue->events + i, queue->events + i + 1, (queue->nevents - i - 1) * sizeof(ddsrt_event_t*));
+      if (queue->ievents > i)
+        queue->ievents--;
+      queue->nevents--;
+      break;
+    }
+  }
+  ddsrt_mutex_unlock(&queue->lock);
+
+  return DDS_RETCODE_OK;
+}
+
+ddsrt_event_t* ddsrt_event_queue_next(ddsrt_event_queue_t* queue) {
+  ddsrt_event_t* ptr = NULL;
+  ddsrt_mutex_lock(&queue->lock);
+  while (queue->ievents < queue->nevents) {
+    ddsrt_event_t* evt = queue->events[queue->ievents++];
+    if (DDSRT_EVENT_FLAG_UNSET != evt->triggered)
+      ptr = evt;
+  }
+  ddsrt_mutex_unlock(&queue->lock);
+  return ptr;
+}
--- a/src/ddsrt/src/events/darwin/events_darwin.c
+++ b/src/ddsrt/src/events/darwin/events_darwin.c
@@ -20,6 +20,14 @@
 
 #define EVENTS_CONTAINER_DELTA 8
 
+ /**
+ * @brief Darwin (Apple) implementation of ddsrt_event_queue.
+ *
+ * This implementation uses a kqueue for monitoring a set of filedescriptors for events.
+ * Using the kevent call, the kernel can be told to add/modify fd's on its list for monitoring
+ * or to wait for events on the monitored fd's. Interrupts of waits are done through writes
+ * to an internal pipe.
+ */
 struct ddsrt_event_queue {
   ddsrt_event_t**         events;   /**< container for triggered events*/
   size_t                  nevents;  /**< number of triggered events stored*/
@@ -27,7 +35,7 @@ struct ddsrt_event_queue {
   size_t                  ievents;  /**< current iterator for getting the next triggered event*/
   int                     kq;       /**< kevent polling instance*/
   kevent*                 kevents;  /**< array which kevent uses to write back to, has identical size as this->events*/
-  ddsrt_mutex_t           lock;     /**< for add/delete */
+  ddsrt_mutex_t           lock;     /**< for keeping adds/deletes from occurring simultaneously */
 #if !defined(LWIP_SOCKET)
   ddsrt_socket_t          interrupt[2]; /**< pipe for interrupting waits*/
 #endif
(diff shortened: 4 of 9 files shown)


src/ddsrt/include/dds/ddsrt/events/darwin.h
```diff
@@ -0,0 +1,23 @@
...
+#define DDSRT_EVENTS_DARWIN_H
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+#if defined (__cplusplus)
+}
+#endif
+#endif /* DDSRT_EVENTS_DARWIN_H */
```
eboasson, line 23: Github is unhappy with files not ending on a newline, and it is not the only one:

https://stackoverflow.com/questions/8172246/what-is-the-meaning-of-warning-no-new-line-at-end-of-file?lq=1
> The C language requires that every source file must end with a newline (from C99 5.1. 1.2/1): A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character before any such splicing takes place.

I am not aware of any compilers, other than long-obsolete version of some IDL compiler, that uses the freedom such a rule gives. But as a general rule, it makes sense to have the files end with a newline. You can probably even configure your editor to make sure of it :)

src/ddsrt/src/events.c
```diff
@@ -0,0 +1,24 @@
...
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+#include "dds/ddsrt/events.h"
+#include "dds/ddsrt/heap.h"
+
+dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags) {
```
eboasson, line 16: For consistency with the rest of the code base, I'd prefer the opening brace that is part of function declaration syntax to (generally) sit on a line of its own. The "(generally)" is there because it is such a waste of space for trivial one-line or empty functions.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  /**
+  * @brief Initializes an existing event to indicate a socket.
+  *
+  * @param[out] ev Pointer to the event to initialize.
+  * @param[in] sock Socket to initialize the event with.
+  * @param[in] flags Flags to set for the event.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags);
```
eboasson, line 91: Presumably `ev` must not be a null pointer, adding a `ddsrt_nonnull` or `ddsrt_nonnull_all` is useful documentation — and sometimes it even makes the compilers produce useful warnings at the call sites! See https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/ddsrt/include/dds/ddsrt/attributes.h

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright(c) 2006 to 2020 ADLINK Technology Limited and others
```
eboasson, line 2: I do understand how it came to read "2006 to 2020", but I think for this calendar year a plain "2020" would be a better choice for a new file.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+struct ddsrt_event_queue {
+  ddsrt_event_t**         events;   /**< container for triggered events*/
+  size_t                  nevents;  /**< number of triggered events stored*/
+  size_t                  cevents;  /**< capacity of triggered events stored*/
+  size_t                  ievents;  /**< current iterator for getting the next triggered event*/
+  int                     kq;       /**< kevent polling instance*/
+  struct kevent*          kevents;  /**< array which kevent uses to write back to, has identical size as this->events*/
+  ddsrt_mutex_t           lock;     /**< for keeping adds/deletes from occurring simultaneously */
+#if !defined(LWIP_SOCKET)
+  ddsrt_socket_t          interrupt[2]; /**< pipe for interrupting waits*/
```
eboasson, line 45: I my view this should be typed as an array of `int`: that is the declared element type in the `pipe` call, after all. I also don't think it should be dependent on `LWIP_SOCKET` not being defined. If one wants to use LwIP for network communications, one still has to find ways of implementing the non-socket parts of the interface. On macOS (or Linux, for that matter), pipes are a very basic mechanism and a sensible choice in such a case.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+ * to an internal pipe.
+ */
+struct ddsrt_event_queue {
+  ddsrt_event_t**         events;   /**< container for triggered events*/
+  size_t                  nevents;  /**< number of triggered events stored*/
+  size_t                  cevents;  /**< capacity of triggered events stored*/
+  size_t                  ievents;  /**< current iterator for getting the next triggered event*/
+  int                     kq;       /**< kevent polling instance*/
+  struct kevent*          kevents;  /**< array which kevent uses to write back to, has identical size as this->events*/
+  ddsrt_mutex_t           lock;     /**< for keeping adds/deletes from occurring simultaneously */
```
eboasson, line 43: What about waiting in parallel to adding/deleting entries? 🤔

It happens currently when using TCP, that's an inheritance I've accepted but would like to eliminate.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+static dds_return_t ddsrt_event_queue_init(ddsrt_event_queue_t* queue) {
+  queue->nevents = 0;
+  queue->cevents = 8;
+  queue->ievents = 0;
+  ddsrt_mutex_init(&queue->lock);
+
+  /*create kevent polling instance */
+  queue->kq = kqueue();
+  assert(queue->kq != -1);
+  assert(fcntl(queue->kq, F_SETFD, fcntl(queue->kq, F_GETFD) | FD_CLOEXEC) != -1);
```
eboasson, line 67: I suspect you have carefully taken the kqueue support code from my quick hack in `q_sockwaitset.c`. I have decided to not check the details of the system call usage at this time.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+
+  /*create kevents array*/
+  queue->kevents = ddsrt_malloc(sizeof(struct kevent) * queue->cevents);
+  assert(queue->kevents);
+
+  return DDS_RETCODE_OK;
+}
+
+/**
+* @brief Finishes an event queue.
```
eboasson, line 89: Nitpicking comments ... such a bad attitude, but I mean well. I believe `fini` is an abbreviation of `finalise` or `finalisation`, not of `finish`. Please do check, I might learn something!

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+  assert(queue);
+
+  /*reset triggered status*/
+  ddsrt_mutex_lock(&queue->lock);
+  queue->ievents = 0;
+  for (unsigned int i = 0; i < queue->nevents; queue++)
+    ddsrt_atomic_st32(&queue->events[i]->triggered, DDSRT_EVENT_FLAG_UNSET);
+  ddsrt_mutex_unlock(&queue->lock);
+
+  struct timespec tmout = { reltime / 1000000000,     /* waittime (seconds) */
```
eboasson, line 153: You might want to use `DDS_NSECS_IN_SEC` instead, then the comment also becomes a bit superfluous. (That is, if the timeout doesn't get removed altogether.)

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+*
+* @param[out] queue The queue to finish.
+*
+* @returns DDS_RETCODE_OK if everything went OK.
+*/
+static dds_return_t ddsrt_event_queue_fini(ddsrt_event_queue_t* queue) {
+  assert(queue);
+  /*deregister all currently stored triggers from kqueue*/
+  struct kevent kev;
+  for (unsigned int i = 0; i < queue->nevents; i++) {
```
eboasson, line 101: For correctly typing this, `i` should be `size_t` (or `queue->nevents` an `unsigned`).

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+}
+
+dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t reltime) {
+  assert(queue);
+
+  /*reset triggered status*/
+  ddsrt_mutex_lock(&queue->lock);
+  queue->ievents = 0;
+  for (unsigned int i = 0; i < queue->nevents; queue++)
+    ddsrt_atomic_st32(&queue->events[i]->triggered, DDSRT_EVENT_FLAG_UNSET);
```
eboasson, line 150: I'm not too happy about writing to all events (which are scattered in memory). It is probably unavoidable short of embedding the events in the `ddsrt_event_queue_t`, rather than letting the latter point to events owned elsewhere.

(Not that one can always avoid all cache misses ... but this seems like a place where making an effort to significantly reduce the number of them would pay off.)

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+  ddsrt_mutex_unlock(&queue->lock);
+
+  struct timespec tmout = { reltime / 1000000000,     /* waittime (seconds) */
+                            reltime % 1000000000 }; /* waittime (nanoseconds) */
+  int nevs = kevent(queue->kq, NULL, 0, queue->kevents, (int)queue->nevents, &tmout);
```
eboasson, line 155: There is a locking problem here: one can change the set of events (and possibly even trigger reallocation) between reading `kevents` and `nevents` and the `kevent` call returning.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+  assert(queue->kq != -1);
+  assert(fcntl(queue->kq, F_SETFD, fcntl(queue->kq, F_GETFD) | FD_CLOEXEC) != -1);
+
+  queue->events = ddsrt_malloc(sizeof(ddsrt_event_t*) * queue->cevents);
+  assert(queue->events);
+
+#if !defined(LWIP_SOCKET)
+  if (-1 == pipe(queue->interrupt))
+    return DDS_RETCODE_ERROR;
+  /*register interrupt event*/
```
eboasson, line 75: It would probably be a good idea to make the pipe non-blocking and ignore `EAGAIN`: the point is to trigger it, a full buffer should not block the the trigger operation and spurious triggers are ok.

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+  for (int i = 0; i < nevs; i++) {
+    ddsrt_event_t* evt = queue->kevents[i].udata;
+    if (NULL != evt) {
+      ddsrt_atomic_st32(&evt->triggered, DDSRT_EVENT_FLAG_READ);
+    }
+    else {
+#if !defined(LWIP_SOCKET)
+      char buf;
+      if (1 != read(queue->interrupt[0], &buf, 0))
+        return DDS_RETCODE_ERROR;
```
eboasson, line 166: You mustn't return without unlocking the queue first. (I apologise on C's behalf for it being obsolete since the late '50s.)

src/ddsrt/src/events/darwin/events_darwin.c
```diff
@@ -0,0 +1,236 @@
...
+  return DDS_RETCODE_OK;
+}
+
+ddsrt_event_t* ddsrt_event_queue_next(ddsrt_event_queue_t* queue) {
+  ddsrt_event_t* ptr = NULL;
+  ddsrt_mutex_lock(&queue->lock);
+  while (queue->ievents < queue->nevents) {
+    ddsrt_event_t* evt = queue->events[queue->ievents++];
+    if (DDSRT_EVENT_FLAG_UNSET != ddsrt_atomic_ld32(&evt->triggered))
+      ptr = evt;
```
eboasson, line 232: I think you're missing a `break` here

src/ddsrt/src/events/posix/events_posix.c
```diff
@@ -0,0 +1,301 @@
+/*
```
eboasson, line 1: I am stopping here: I suspect the preceding comments will apply quite naturally in the remainder and so the additional benefit of the additional effort is likely to be small.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef DDSRT_EVENTS_H
+#define DDSRT_EVENTS_H
+
+#include "dds/ddsrt/retcode.h"
+#include "dds/ddsrt/sockets.h"
+#include "dds/ddsrt/atomics.h"
+
+#if __APPLE__
```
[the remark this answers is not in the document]
k0ekk0ek: No objection here. `ddsrt` does the same thing in other places and I've actually been thinking about how to do this feature-thing even better (like autotools). So kqueue, would definitely be a better name :+1:

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+//#define DDSRT_EVENT_FLAG_TIMEOUT (1u<<4)
+  /**< notification of ip address change on the resource*/
+//#define DDSRT_EVENT_FLAG_IP_CHANGE (1u<<5)  
+///@}
+
+  /**
+  * @brief Describes the type of object being monitored for events.
+  */
+  enum ddsrt_event_type {
+    ddsrt_event_type_unset, /**< unitialized state*/
```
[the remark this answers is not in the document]
reicheratwork: You are correct, I will change this.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  /**
+  * @brief Cleans up an event queue.
+  *
+  * Will finish the event queue first, then free the memory of the queue.
+  *
+  * @param[out] queue The queue to clean up.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_destroy(ddsrt_event_queue_t* queue);
```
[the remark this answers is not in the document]
reicheratwork: Will use the destroy syntax from now on.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  * @brief Event queue creation function.
+  *
+  * @returns Pointer to the created queue.
+  */
+  ddsrt_event_queue_t* ddsrt_event_queue_create(void);
+
+  /**
+  * @brief Cleans up an event queue.
+  *
+  * Will finish the event queue first, then free the memory of the queue.
```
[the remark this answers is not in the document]
reicheratwork: I probably should have spent a little more effort in documenting the function.
I will improve the documentation.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  * @param[out] queue The queue to clean up.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_destroy(ddsrt_event_queue_t* queue);
+
+  /**
+  * @brief Getter for the number of stored events.
+  *
+  * @param[in] queue The queue to get the number of events of.
```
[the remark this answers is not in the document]
reicheratwork: True, in C++ you have the whole "mutable" keyword to make this distinction between member variables affecting the state and those that do not.
I personally am of the opinion that for the event code at least, using const causes more problems than it solves:
- using const will lead to some places where we need to cast away constness where it does not "fit", breaking its usefulness

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  * @returns DDS_RETCODE_OK if everything went OK.
+  */
+  dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t reltime);
+
+  /**
+  * @brief Interrupts a triggered wait of a queue.
+  *
+  * @param[out] queue The queue to interrupt.
+  *
+  * @returns DDS_RETCODE_OK if everything went OK.
```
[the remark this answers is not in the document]
reicheratwork: Fully agree with this, I will improve the documentation to describe all returned values and under which conditions (if it can be described within a sentence or two).
About the choice for returning errors or calling an abort, I will go over the code where errors are returned and see whether the state of the program is recoverable, if it is not, i will use some kind of hard stop.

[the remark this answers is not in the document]
k0ekk0ek: When it comes to nitpicking `@retval` should be used for this sort of thing. `@returns` describes what is returned.

src/ddsrt/include/dds/ddsrt/events/darwin.h
reicheratwork: I already saw git complaining about this, though I did not know that this was part of the C language standard.
I will change the files to end on a newline.

src/ddsrt/src/events.c
reicheratwork: I will change the braces on function declarations to match the rest of the code base.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: Maybe I was a little to quick in just copy/paste-ing :)
Will change the date range from 2006-2020 to just 2020.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: You are correct about pipes being always available in macOS or any fullfledged linux, I will remove the dependency on LWIP in the kevent type implementation.
I will let it stay in the posix (select) implementation for the moment.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: You are completely correct, this will be corrected.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: Already saw this as being problematic after running the unittest on a mac.
Will be fixed.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: You are correct, this will be fixed.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  * @param[in] queue The queue to get the number of events of.
+  *
+  * @returns The number of events stored by the queue.
+  */
+  size_t ddsrt_event_queue_nevents(ddsrt_event_queue_t* queue);
+
+  /**
+  * @brief Triggers a wait for events for the queue.
+  *
+  * @param[out] queue The queue to trigger.
```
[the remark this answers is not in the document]
reicheratwork: In this case, since the function was only checking whether the pointer it was supposed to "out" to, in my opinion, the amount of "data" retrieved from the parameter was not enough to have it be an "in".
But in absolute terms, data IS being required from this parameter, so it is to some degree an "in".
I will go over the doxygen documentation and re-check the parameters for in/out/inout status.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  * @param[in] reltime The maximum amount of time to wait.
```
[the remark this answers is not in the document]
reicheratwork: I added the timeout parameter since both the ddsrt_select and kevent functions allow for a timeout to be set on the function call.
I propose to have the timeout to be truly infinite, not just "very long" (9.2 * 10^9 seconds, which is approx. 300 years) if this parameter is set to be DDS_INFINITY to use the explicit infinite wait time for kevent and select methods.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: I will also have a look which functions really need to have a return value, i.e. functions which can only return a single value.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: I will implement this.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: Thanks for the suggestion!
I was using lots of checking whether pointers were not null with assert, which seemed suboptimal in my eyes.
Will examine the use of ddsrt_nonnull and how to use it.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: I will use this definition.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: I was already thinking about the same issue, i will create something that solves this possible conflict.
Maybe something with buffering the changes and only setting them with kevent at a single time?

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,208 @@
...
+  *
+  * This function will set the type to be DDSRT_EVENT_TYPE_SOCKET, the triggered status to DDSRT_EVENT_FLAG_UNSET and
+  * the flags and socket fields to the values supplied.
+  *
+  * @param[in,out] ev Pointer to the event to initialize.
+  * @param[in] sock Socket to initialize the event with.
+  * @param[in] flags Flags to set for the event.
+  *
+  * @retval DDS_RETCODE_OK
+  *             In all cases.
```
k0ekk0ek, line 94: No `DDS_RETOCDE_BAD_PARAMETER` if DDSRT_INVALID_SOCKET is passed?

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: I might have collapsed a number of if checks into a single statement at some locations, but functionally it should work the same.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,208 @@
...
+  * @brief Cleans up an event queue.
+  *
+  * Will close/free any resources managed by the event_queue (if any), then free the memory of the queue.
+  *
+  * @param[in,out] queue The queue to clean up.
+  *
+  * @retval DDS_RETCODE_OK
+  *             In all cases.
+  */
+  DDS_EXPORT dds_return_t ddsrt_event_queue_delete(ddsrt_event_queue_t* queue) ddsrt_nonnull_all;
```
k0ekk0ek, line 120: If it's always `DDS_RETCODE_OK`, wouldn't it be better to just make it `void`?

reicheratwork: This was one of the things I was planning on going over in the next iteration.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,208 @@
...
+
+  /**
+  * @brief Event queue creation function.
+  *
+  * This function will attempt to reserve memory for the event queue and open any necessary additional resources, i.e.:
+  * - a pipe for interrupting its own wait state
+  * - an instance to a kernel event monitor (in the case of BSD operating systems)
+  *
+  * @returns !NULL The event queue that was created.
+  * @returns NULL Event queue could not be created succesfully.
```
k0ekk0ek, line 106: This is a really nitpicking... `@returns` would be better written as `@returns The event queue that was created or NULL on failure`.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This touches on a similar issue you raised before.
I will create a manner in which this is no longer a problem.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events/darwin.h
reicheratwork: This is now implemented.

src/ddsrt/src/events.c
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented, including the support for FREEBSD.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: I will check this.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: The events do not have to be scattered in memory, if the entity using the event queue just has the events which are used in the queue has them in some kind of flat array, and just passes pointers to those entries to the queue.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: This also touches on a similar question made by @eboasson .

src/ddsrt/cmake/events.c
```diff
@@ -0,0 +1,18 @@
...
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#include "dds/ddsrt/events.h"
+
+#if defined __APPLE__
```
k0ekk0ek, line 14: Remember to check for `__FreeBSD__` here as well.

reicheratwork: Already on it.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/cmake/events.c
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: Fixed the hard assert failure for the kqueue, a more clean resolution of this error is done.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: The kevent type event queue is will now allow for modification of the administered while it is waiting for events.
This is done by staging the changes and only modifying the events monitored by kevent when it is triggering a new wait.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
reicheratwork: Removed the return value for this function, it will not validate the input for performance reasons.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+    goto kq_fail;
+  else if (-1 == fcntl(queue->kq, F_SETFD, fcntl(queue->kq, F_GETFD) | FD_CLOEXEC))
+    goto pipe0_fail;
+  /*create interrupt pipe */
+  else if (-1 == pipe(queue->interrupt))
+    goto pipe0_fail;
+  else if (-1 == fcntl(queue->interrupt[0], F_SETFD, fcntl(queue->interrupt[0], F_GETFD) | FD_CLOEXEC) ||
+           -1 == fcntl(queue->interrupt[1], F_SETFD, fcntl(queue->interrupt[1], F_GETFD) | FD_CLOEXEC))
+    goto pipe1_fail;
+
```
eboasson, line 112: I think one should also set "non-blocking" on the interrupt pipe. The risk is minimal, but without making it non-blocking, the `write` call in `signal` could block.

src/ddsrt/src/events/posix/events_posix.c
```diff
@@ -75,102 +77,79 @@ static dds_return_t ddsrt_event_queue_init(ddsrt_event_queue_t* queue) {
...
-    return DDS_RETCODE_ERROR;
-  }
 #elif !defined(LWIP_SOCKET)
   /*simple linux type pipe*/
-  if (pipe(queue->interrupt) == -1) return DDS_RETCODE_ERROR;
-#endif /* _WIN32, __VXWORKS__, !LWIP_SOCKET*/
+  if (pipe(queue->interrupt) == -1)
+    goto alloc_cleanup;
+#endif /* _WIN32 || !LWIP_SOCKET*/
+  ddsrt_mutex_init(&queue->lock);
```
eboasson, line 93: here (like in the `kqueue` variant), making the "interrupt" pipe non-blocking and close-on-exec would make sense.

src/ddsrt/src/events/posix/events_posix.c
```diff
@@ -234,68 +223,99 @@ dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t r
...
+
+dds_return_t ddsrt_event_queue_remove(ddsrt_event_queue_t* queue, ddsrt_event_t* evt)
+{
+  dds_return_t ret = DDS_RETCODE_ALREADY_DELETED;
+  ddsrt_mutex_lock(&queue->lock);
+  for (size_t i = 0; i < queue->nevents; i++)
+  {
+    if (queue->events[i] == evt)
+    {
       memmove(queue->events + i, queue->events + i + 1, (queue->nevents - i - 1)*sizeof(ddsrt_event_t*));
```
eboasson, line 289: Wouldn't it be possible to move the last event into the slot occupied by the removed one (like in the `kqueue` variant)?

src/ddsrt/tests/events.c
```diff
@@ -1,5 +1,5 @@
 /*
```
eboasson, line 1: I will look at the tests some other day ...

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
          only happens when the participant set has changed, so might as well rebuild it */
         rebuild_local_participant_set (ts1, gv, &lps);
-        os_sockWaitsetPurge (waitset, num_fixed);
+        ddsrt_event_queue_trim (waitset, num_fixed);
         for (uint32_t i = 0; i < lps.nps; i++)
         {
           if (lps.ps[i].m_conn)
-            os_sockWaitsetAdd (waitset, lps.ps[i].m_conn);
+          {
+            ddsrt_event_socket_init (&lps.ps[i].m_conn->m_event, ddsi_conn_handle(lps.ps[i].m_conn), DDSRT_EVENT_FLAG_READ);
```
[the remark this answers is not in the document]
reicheratwork: This is a much more elegant solution, i will start work on it immediately.

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
       {
-        int idx;
         ddsi_tran_conn_t conn;
-        while ((idx = os_sockWaitsetNextEvent (ctx, &conn)) >= 0)
+        ddsrt_event_t* evt;
+        while ((evt = ddsrt_event_queue_next (waitset)) != NULL)
         {
+          if (0x0 == (ddsrt_atomic_ld32(&evt->triggered) & DDSRT_EVENT_FLAG_READ) ||
+              DDSRT_EVENT_TYPE_SOCKET != evt->type) continue;
+          uintptr_t tmp = (uintptr_t)evt - offsetof(struct ddsi_tran_conn, m_event);
```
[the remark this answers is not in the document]
reicheratwork: I will move it there.

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
+          conn = (ddsi_tran_conn_t)tmp;
           const ddsi_guid_prefix_t *guid_prefix;
-          if (((unsigned)idx < num_fixed) || gv->config.many_sockets_mode != MSM_MANY_UNICAST)
```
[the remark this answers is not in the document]
reicheratwork: That was indeed a mistake, I was laboring under the impression that the only entries being looped over were the ones left after the trim operation above.
It also played into my preference for not making internal indices in the ddsrt_event_queue externally available.
I will fix this.

src/core/ddsi/src/q_receive.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/core/ddsi/src/q_receive.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/src/events.c
```diff
@@ -13,12 +13,10 @@
...
-dds_return_t ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags) {
-  assert(ev);
-  assert(sock);
-  ev->type = ddsrt_event_type_socket;
+void ddsrt_event_socket_init(ddsrt_event_t* ev, ddsrt_socket_t sock, uint32_t flags) 
+{
+  ev->type = DDSRT_EVENT_TYPE_SOCKET;
   ev->flags = flags;
   ev->data.socket.sock = sock;
   ddsrt_atomic_st32(&ev->triggered, DDSRT_EVENT_FLAG_UNSET);
```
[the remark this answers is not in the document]
reicheratwork: This was done after a suggestion by @k0ekk0ek for making it more future (and multi-threaded) proof.
I agreed with his assessment that simultaneous reads/writes could cause for uncertainty of the state being read out, so I implemented this status flag with an atomic.
I do not think that it hurts the performance in any major way, and it will prevent some possible weird errors, so my vote is for leaving it as it is now.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+dds_return_t ddsrt_event_queue_wait(ddsrt_event_queue_t* queue, dds_duration_t reltime)
+{
+  dds_return_t ret = DDS_RETCODE_OK;
+
+  ddsrt_mutex_lock(&queue->lock);
+
+  /*remove deregistered events*/
+  size_t i = 0;
+  while (i < queue->nevents)
```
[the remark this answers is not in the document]
reicheratwork: I will think of some way that this will not be executed every time a wait is called, maybe a flag indicating that the set of events has been modified?

src/core/ddsi/src/q_receive.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+  {
+    queued_event_t* qe = &(queue->events[i]);
+    if (EVENT_STATUS_DEREGISTERED == qe->status)
+    {
+      /*remove the deregistered event*/
+      qe->internal.flags = EV_DELETE;
+      assert(kevent(queue->kq, &qe->internal, 1, NULL, 0, NULL) != -1);
```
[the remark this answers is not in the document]
reicheratwork: This error demonstrates my inexperience with the whole assert call.
I will fix this.

src/ddsrt/src/events/kqueue/events_kqueue.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
         /* first rebuild local participant set - unless someone's toggling "deafness", this
          only happens when the participant set has changed, so might as well rebuild it */
         rebuild_local_participant_set (ts1, gv, &lps);
-        os_sockWaitsetPurge (waitset, num_fixed);
+        ddsrt_event_queue_trim (waitset, num_fixed);
```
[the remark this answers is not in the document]
reicheratwork: This also overlaps with your other comment starting on events_kqueue.c:199.
I will look into some other manner of filtering the "correct" events from the queue (maybe based on some flag bits in the event class?).

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+      *qe = queue->events[--queue->nevents];
```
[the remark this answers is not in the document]
reicheratwork: This also overlaps with your other comment hitting on the trim/purge function from q_receive.c:3227.
The single solution described there should fix both issues.

src/ddsrt/src/events/posix/events_posix.c
reicheratwork: I was a little more careful in this implementation to conserve the ordering of the events.
I will look into whether that is still necessary.
If the ordering is no longer of importance (if the trim function is moved to a filter type), then yes, we can replace the wholesale copy of large parts of data with just switching the entry to be dropped with one "beyond" the array.

src/core/ddsi/src/q_receive.c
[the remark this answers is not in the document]
reicheratwork: I have moved from a "trimming" of the queue based on index to filtering the queue based on flags set.

src/ddsrt/src/events/kqueue/events_kqueue.c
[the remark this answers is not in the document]
reicheratwork: This is now fixed.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -0,0 +1,179 @@
...
+  */
+  struct ddsrt_event {
+    uint32_t flags; /**< Type of event being watched for, composited from DDSRT_EVENT_FLAG_$EVENT_TYPE$.*/
+    ddsrt_event_type_t type; /**< Type of object being watched.*/
+    ddsrt_atomic_uint32_t triggered; /**< Trigger status after a call to ddsrt_event_queue_wait.*/
+    union {
+      struct { ddsrt_socket_t sock; } socket;
+      /*struct ddsrt_event_file file;  this is for future expansion to be able to register events on files*/
+      /*struct ddsrt_event_ifaddr ifaddr; this is for future expansion to be able to register IP address changes*/
+    } data; /**< Container for the object being watched.*/
```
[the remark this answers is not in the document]
reicheratwork: To be more in line with the rest of the code base, i have changed the name of the struct member to u.

src/ddsrt/src/events/kqueue/events_kqueue.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: This has now been implemented.

src/ddsrt/src/events/kqueue/events_kqueue.c
reicheratwork: This has now been implemented.

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
+      if (DDS_RETCODE_OK == ddsrt_event_queue_wait (waitset,DDS_INFINITY))
       {
-        int idx;
         ddsi_tran_conn_t conn;
-        while ((idx = os_sockWaitsetNextEvent (ctx, &conn)) >= 0)
+        ddsrt_event_t* evt;
+        while ((evt = ddsrt_event_queue_next (waitset)) != NULL)
         {
+          if (0x0 == (ddsrt_atomic_ld32(&evt->triggered) & DDSRT_EVENT_FLAG_READ) ||
+              DDSRT_EVENT_TYPE_SOCKET != evt->type) continue;
```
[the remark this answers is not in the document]
reicheratwork: At the moment SOCKET event type is the only (connection-based) event type which can be set.
I think this discussion should be moved to where we determine which kinds of events we will be supporting.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  /*unset triggered status*/
+  for (i = 0; i < queue->nevents; i++)
+    ddsrt_atomic_st32(&(queue->events[i].external->triggered), DDSRT_EVENT_FLAG_UNSET);
+  ddsrt_mutex_unlock(&queue->lock);
+
+  struct timespec tmout, *ptmout = NULL;
+  if (DDS_INFINITY != reltime &&
+      DDS_NEVER != reltime &&
```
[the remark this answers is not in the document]
reicheratwork: Good catch.
Now there is an assert check for reltime to not be DDS_DURATION_INVALID.

src/ddsrt/src/events/posix/events_posix.c
reicheratwork: I have the pipe non-blocking.
I am not sure what you mean with making the pipe close-on-exec, since there is no child process.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -0,0 +1,388 @@
...
+
+  if (queue->nnewevents == queue->cnewevents)
+  {
+    queue->cnewevents += EVENTS_CONTAINER_DELTA;
+    queue->newevents = ddsrt_realloc(queue->newevents, sizeof(queued_event_t) * queue->cnewevents);
+  }
+  queued_event_t* qe = &(queue->newevents[queue->nnewevents++]);
+  qe->status = EVENT_STATUS_UNREGISTERED;
+  qe->external = evt;
+  EV_SET(&qe->internal, evt->data.socket.sock, EVFILT_READ, EV_ADD, 0, 0, 0);
```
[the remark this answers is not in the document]
reicheratwork: The issue is that the "socket" type also has to support the windows socket handles, which are a much smaller set of entities than fds.
This is why I chose the name "socket" to reflect this limitation.
You are correct in requiring a socket type (or some other, when the spec is extended to support maybe fds) for adding the event to the kqueue.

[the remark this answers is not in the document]
reicheratwork: I have added the additional checks on the type of event.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -309,21 +311,46 @@ int ddsrt_event_queue_add(ddsrt_event_queue_t* queue, ddsrt_event_t* evt)
   queued_event_t* qe = &(queue->newevents[queue->nnewevents++]);
   qe->status = EVENT_STATUS_UNREGISTERED;
   qe->external = evt;
-  EV_SET(&qe->internal, evt->data.socket.sock, EVFILT_READ, EV_ADD, 0, 0, 0);
+  qe->evt_type = evt->type;
```
eboasson, line 314: GitHub won't let me comment on lines that haven't changed ... but I do want to remark that scanning all events in `add` to check for duplicates does mean that adding _n_ entries is an O(_n_^2) operation, and that's not so nice.

I'd be willing to require that the entries must be unique.

src/ddsrt/src/events/posix/events_posix.c
eboasson: The point is that the application may fork and create child processes.

src/core/ddsi/src/q_receive.c
```diff
@@ -3225,25 +3226,37 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
-      if ((ctx = os_sockWaitsetWait (waitset)) != NULL)
+      if (DDS_RETCODE_OK == ddsrt_event_queue_wait (waitset,DDS_INFINITY))
       {
-        int idx;
         ddsi_tran_conn_t conn;
-        while ((idx = os_sockWaitsetNextEvent (ctx, &conn)) >= 0)
+        ddsrt_event_t* evt;
+        while ((evt = ddsrt_event_queue_next (waitset)) != NULL)
         {
+          if (0x0 == (ddsrt_atomic_ld32(&evt->triggered) & DDSRT_EVENT_FLAG_READ) ||
```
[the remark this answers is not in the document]
reicheratwork: I will go over the code and see whether I can improve this, and align all instances to use the same structure.

src/ddsrt/include/dds/ddsrt/events.h
```diff
@@ -50,14 +52,13 @@ extern "C" {
...
-    ddsrt_event_type_unset, /**< unitialized state*/
-    ddsrt_event_type_socket /**< indicating a socket type connection*/
+ typedef enum ddsrt_event_type {
+    DDSRT_EVENT_TYPE_UNSET, /**< uninitialized state*/
+    DDSRT_EVENT_TYPE_SOCKET /**< indicating a socket type connection*/
     /*,
-    ddsrt_event_type_file,
-    ddsrt_event_type_ifaddr*/
-  };
-  typedef enum ddsrt_event_type ddsrt_event_type_t;
```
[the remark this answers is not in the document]
reicheratwork: I will have a look at the rest of the code base and try to align this with the rest.

src/core/ddsi/src/q_receive.c
```diff
@@ -3226,37 +3218,27 @@ uint32_t recv_thread (void *vrecv_thread_arg)
         /* first rebuild local participant set - unless someone's toggling "deafness", this
          only happens when the participant set has changed, so might as well rebuild it */
         rebuild_local_participant_set (ts1, gv, &lps);
-        ddsrt_event_queue_trim (waitset, num_fixed);
+        ddsrt_event_queue_filter (waitset, DDSRT_EVENT_FLAG_USER_0);
```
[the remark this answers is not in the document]
reicheratwork: Implemented this by adding moving the pointer to ddsi_guid_prefix_t from the local participant to ddsi_tran_conn.
For the persistent connections this is not set, for others this uses the address of the participant which is set when rebuilding the participant set.
This also removes the need for filtering the wait set, we can just re-add all remaining connections.

src/core/ddsi/src/q_receive.c
```diff
@@ -3226,37 +3218,27 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
         {
           if (0x0 == (ddsrt_atomic_ld32(&evt->triggered) & DDSRT_EVENT_FLAG_READ) ||
               DDSRT_EVENT_TYPE_SOCKET != evt->type) continue;
-          uintptr_t tmp = (uintptr_t)evt - offsetof(struct ddsi_tran_conn, m_event);
-          conn = (ddsi_tran_conn_t)tmp;
+          ddsi_tran_conn_t conn = ddsi_conn_from_event(evt);
           const ddsi_guid_prefix_t *guid_prefix;
-          if (gv->config.many_sockets_mode != MSM_MANY_UNICAST)
-          {
+          if (evt->flags & DDSRT_EVENT_FLAG_USER_0 || gv->config.many_sockets_mode != MSM_MANY_UNICAST)
```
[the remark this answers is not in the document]
reicheratwork: The attaching of the pointer to ddsi_guid_prefix_t to ddsi_tran_conn removes the need for this check.

src/core/ddsi/src/q_receive.c
```diff
@@ -3226,37 +3218,27 @@ uint32_t recv_thread (void *vrecv_thread_arg)
...
             guid_prefix = NULL;
-          }
           else
-          {
-            tmp = tmp - offsetof(struct local_participant_desc, m_conn) + offsetof(struct local_participant_desc, guid_prefix);
-            guid_prefix = (ddsi_guid_prefix_t*)tmp;
-          }
+            guid_prefix = (ddsi_guid_prefix_t*)((uintptr_t)conn - offsetof(struct local_participant_desc, m_conn) + offsetof(struct local_participant_desc, guid_prefix));
```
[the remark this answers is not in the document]
reicheratwork: I fully agree with the attitude of minimizing the occurance of these kinds of operations, they always give off a dirty hack "odour".
With the way that currently guid-prefix is attached to the connection structure, this calculation is no longer necessary.

src/ddsrt/src/events/kqueue/events_kqueue.c
```diff
@@ -109,7 +111,8 @@ dds_return_t ddsrt_event_queue_init(ddsrt_event_queue_t* queue)
   else if (-1 == pipe(queue->interrupt))
     goto pipe0_fail;
   else if (-1 == fcntl(queue->interrupt[0], F_SETFD, fcntl(queue->interrupt[0], F_GETFD) | FD_CLOEXEC) ||
-           -1 == fcntl(queue->interrupt[1], F_SETFD, fcntl(queue->interrupt[1], F_GETFD) | FD_CLOEXEC))
+           -1 == fcntl(queue->interrupt[1], F_SETFD, fcntl(queue->interrupt[1], F_GETFD) | FD_CLOEXEC) ||
+           0 > fcntl(queue->interrupt[0], F_SETFL, O_NONBLOCK))
```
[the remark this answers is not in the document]
reicheratwork: You are correct in that fcntl only returns -1 for errors, this is now corrected.

src/ddsrt/src/events/kqueue/events_kqueue.c
reicheratwork: If we can drop the requirement for uniqueness checks of events in the event queue, then we can indeed drop the O(N) checks, resulting in O(N^2) complexity.
Removed the uniqueness checks from the code.

src/ddsrt/src/events/darwin/events_darwin.c
reicheratwork: As far as i could determine, this is something that came over from the naming of the initialization (run before main starts) and termination (run after main finishes) functions which can be set through the compiler options initarray and finiarray or the compiler directives #pragma init/fini.
Whether fini is an abbreviation of finish or finalize, i would go for the latter, as it matches more with the init being short for initialize.

src/core/ddsi/src/q_receive.c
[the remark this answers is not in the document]
reicheratwork: This is now implemented.

src/ddsrt/include/dds/ddsrt/events.h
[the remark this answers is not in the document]
reicheratwork: For enums: when looking at src/core/ddsc/include/dds/dds.h and other header files in the same directory for example, the enums do have the _t suffix.

eboasson:
@reicheratwork with apologies for the delay, I have some more comments after reviewing it thoroughly and experimenting a bit with possible ways of solving the problems (see https://github.com/eboasson/cyclonedds/tree/eventqueue — I rebased & squashed your commits). There are still some issues in that, too, though, and not because I limited myself to looking into the `kqueue`-based implementation.

First, the problems I noticed:
1. You use the kqueue's `udata` field to cache pointers to `queued_event_t`, which is the sort of thing that field is meant to be used for. However, when you remove an event, you move the one at the very end of the array into the slot occupied by the one being removed, but that doesn't update the corresponding `udata` pointer. That's not going to be noticeable _until_ you add another event. In a default config, you're not going to see that; with in "ManySocketsMode", that changes, and with TCP even more so.
2. The `events` array can be reallocated to accomodate more entries, at which point the address may change and those `udata` pointers will be pointing to freed memory. It is pretty much just a variant of the previous.
3. The semantics for removing an event aren't clear/are problematic: following a call to `ddsrt_event_queue_remove` the event won't be returned by `ddsrt_event_next` any more, but that doesn't mean the event will not be used any more by the thread using the event queue: e.g., the event could have been returned just before. This is not a new problem, by the way. The semantics were equally unclear in the old code, it's just that there the likelihood of bad things happening was a bit smaller because the operating system will generally check the validity of sockets.
4. There is a bit of confusion regarding the treatment of the `interrupt` pipe, in that the call to wait for data to arrive doesn't reserve room in the result (this is all level triggered, so I doubt one would ever notice), but it also means that at startup it can spin for a short while because the timing can be such that it waits before any events have been added.

The `select` based version doesn't suffer from the first two.

Regarding the 1st and 2nd points, I played with it a bit. I also decided to see if it could be rewritten to reduce `ddsrt_event_wait` to a simple call to `kevent` (it can be, because that particular ways in which the interface is used allow you to restrict which thread can do what). The 4th is just a matter of bookkeeping.

The 3rd is trickier, and I haven't yet solved it to my satisfaction and I don't think it a great idea to replace old code by new code without also addressing this problem. Fortunately, there are not many places where `remove` is called as a prelude to freeing the connection, and one option would be to introduce a callback once there is no possibility of it being used anymore. That callback could then be used as a trigger for freeing the memory.

I figured it would be best to give you an update on the review process. Feel free to comment or to suggest things.

PR #558, opened 2020-07-02
Less aggressive retransmit

eboasson:
This is a follow up to #555 with much more substantial changes to the retransmit behaviour. These changes are triggered by the original assumption on hearing of the problem in #484, that it was caused by an over-eager retransmit policy. The assumption turned out to be wrong in the sense that the real problem was caused by something (and addressed by #555), but the unnecessary retransmits were occuring. This PR is focused on those unnecessary retransmits.

image

The numbers do vary a bit from run to run, but in my testing there is real improvement. Reducing unnecessary retransmits also benefits other scenarios, such as sending large samples from my MBP over a WiFi to an AP and then over Ethernet to an RPi (3B in 64-bit mode, socket buffer of 425984 bytes, which is Linux' default limit):

image

Apologies for it not being easy to read, but especially the 1M and 10M samples are much better behaved, with a higher throughput (red: 10s average; green: 1s average, missing at 10M because sending a sample takes multiple seconds), a dramatic reduction in retransmits (purple) and especially also with less data discarded as duplicate (brown; some is expected as the latency does exceed the default NackDelay setting). The remainder: blue is what macOS reports as "last transmit rate" for the WiFi and ochre what the RPi's Ethernet reports as received network bandwidth.

The improved behaviour allows a tweaking of the defaults settings without risking too much trouble on random networks. It now uses somewhat larger messages, and this tweak changes the out-of-the-box performance over loopback from ~4.5Gb/s (my MBP) to ~11Gb/s for 100kB samples (which seems to be roughly the sweet spot; socket receive buffer of 1MB because macOS is happy to give that much in default config):

image

The pair of machines used for the graph in the README still lives but they have been reconfigured and no longer allow large socket receive buffers. As a consequence, it now sometimes overruns the 425984-byte buffer a bit:

image

(beware of the different colors ...) Nonetheless performance is pretty much indistinguishable from before.

In short, I am tempted to say that this is an improvement.

This PR builds on the work in #555 (and #553) so the only new parts are from 04be1bc8 (Include receive buffer sizes in discovery) onwards:
* It discovers receive buffer sizes using a vendor-specific parameter. In the absence of this parameter in discovery, it assumes the receiver has the same size as it has itself. This size is then used to determine how much data to retransmit in response to a single retransmit request.
* It reduces the number of HEARTBEATs sent in response ACKNACKs and NACKFRAGs to 1 per packet. Before, an ACKNACK + NACKFRAG pair in a single message could result in multiple HEARTBEATs sent in response, which is pretty much a waste of everything.
* It now fully respects the NackDelay setting: if the writer sends a heartbeat requesting a response, it will send an ACK instead of a NACK if sending a NACK would mean sending two NACKs within one NackDelay period. That it didn't respect it was a bug that went unnoticed for a long time, and the fix is the main reason there are fewer retransmits now.
* However, it will send a NACK immediately if there are no outstanding retransmit requests anymore, otherwise the massive number of retransmits requiring for pushing a very large sample to a receiver that can't keep up with the sender would incur an additional NackDelay latency for every retransmit burst. For 10MB sample, that adds up ...


src/core/ddsc/tests/CMakeLists.txt
```diff
@@ -92,3 +93,16 @@ set_tests_properties(
...
+  "oneliner.c"
+  "test_oneliner.c"
+  "test_oneliner.h"
+  "test_util.c"
+  "test_util.h")
+target_include_directories(
+  oneliner PRIVATE
+  "$<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/src/include/>"
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../ddsc/src>"
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../ddsi/include>")
```
k0ekk0ek, line 107: This'll work. It would've been cleaner to use the property from the `ddsc` target though.

eboasson: Hmm, ok, yes, that would be cleaner. This not being the only place (according to a quick grep) fixing it here won't solve the problem of fixing it everywhere. It is probably better to do that in another PR.

And thanks for giving it a quick check!

PR #559, opened 2020-07-03
XTypes type discovery protocol

dennis-adlink:
This PR adds the implementation of the first stage of the XTypes type discovery protocol in Cyclone. This first implementation is not compliant with the DDS XTypes specification in the sense that it currently uses a different `type_information` format (md5 hash of the serialized sertype), and the exchanged types are not a `TypeObject`, but a serialized `ddsi_sertype` object. At a later stage, when the XTypes type system is implemented, we can plug that into this implementation and get a more spec compliant implementation of the type discovery.

This PR includes a partial implementation of the type consistency enforcement QoS policy and a (limited) assignability check for `sertype_default` that is used when matching readers and writers.

A build option `ENABLE_TYPE_DISCOVERY` is added so that type discovery support can be disabled when building Cyclone. By default type discovery is disabled.


src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+  ddsi_tl_meta_unref_impl_locked (gv, tlm, ep_guid);
+  ddsrt_mutex_unlock (&gv->tl_admin_lock);
+}
+
+static struct participant *get_typelookup_pp (const struct ddsi_domaingv *gv, const ddsi_guid_prefix_t *guid_prefix)
+{
+  struct entidx_enum_participant est;
+  struct participant *pp = NULL;
+  bool pp_found = false;
+  entidx_enum_participant_init (&est, gv->entity_index);
```
eboasson, line 192: There is no need to scan the participants:
* if `guid_prefix_empty(guid_prefix)`, one just picks the first (same code, minus the loop).
* otherwise, one can use the fact that all participants have entity id 0x1c1 (there is a name for it, I can't be bothered to look it up right now), and so one can just do a GUID lookup of (effectively) `{guid_prefix,0x1c1}`.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+  if (tlm->state != TL_META_NEW)
+  {
+    GVTRACE ("ddsi_tl_request_type - state not-new (%u) for "PTYPEIDFMT"\n", tlm->state, PTYPEID (*type_id));
+    ddsrt_mutex_unlock (&gv->tl_admin_lock);
+    return false;
+  }
+
+  struct participant *pp = get_typelookup_pp (gv, &tlm->dst_prefix);
+  struct writer *wr = get_builtin_writer (pp, NN_ENTITYID_TL_SVC_BUILTIN_REQUEST_WRITER);
+  if (wr == NULL)
```
eboasson, line 215: So if an arbitrary participant was picked by `get_typelookup_pp` (the first one in the enumeration is a pretty arbitrary one, given how random the ids are) and it doesn't have a writer for requests, does that guarantee no other participant has a writer?

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+  {
+    GVTRACE ("ddsi_tl_request_type("PGUIDFMT") - builtin TypeLookup service request writer not found\n", PGUID (pp->e.guid));
+    ddsrt_mutex_unlock (&gv->tl_admin_lock);
+    return false;
+  }
+
+  type_lookup_request_t request;
+  memcpy (&request.writer_guid, &wr->e.guid, sizeof (wr->e.guid));
+  request.sequence_number = wr->seq;
```
eboasson, line 224: `wr->seq` is an unlocked access. Given that all requests are written by this function and that it always holds `tl_admin_lock` across this access and sending the sample, it obviously is safe as written. But if there's no need to use the writer's sequence number, I'd rather have a request sequence number in the tl admin.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+  request.type_ids.n = 1;
+  request.type_ids.type_ids = tlm->type_id;
+  struct ddsi_serdata *serdata = ddsi_serdata_from_sample (gv->tl_svc_request_type, SDK_DATA, &request);
+  serdata->timestamp = ddsrt_time_wallclock ();
+
+  GVTRACE (" wr-tl-req "PGUIDFMT" typeid "PTYPEIDFMT"\n", PGUID (wr->e.guid), PTYPEID(*type_id));
+  struct ddsi_tkmap_instance *tk = ddsi_tkmap_lookup_instance_ref (gv->m_tkmap, serdata);
+  write_sample_gc (lookup_thread_state (), NULL, wr, serdata, tk);
```
eboasson, line 232: I would prefer not holding any locks while writing this, or to use a guaranteed non-blocking write operation. That all depends on QoS and stuff like that — currently a `KEEP_ALL` writer isn't guaranteed non-blocking. But we could make it such for this case.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+      memcpy (&types.types[types.n - 1].type_identifier, tlm->type_id, sizeof (types.types[types.n - 1].type_identifier));
+      tlm->sertype->ops->serialize (tlm->sertype, &sz, &types.types[types.n - 1].type_object.value);
+      assert (sz <= UINT32_MAX);
+      types.types[types.n - 1].type_object.length = (uint32_t) sz;
+      GVTRACE (" found");
+    }
+  }
+  ddsrt_mutex_unlock (&gv->tl_admin_lock);
+
+  if (types.n > 0)
```
eboasson, line 292: I'd think a reply should be written also if there is no substance in the response: that way the requestor doesn't have to timeout before it can continue.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+
+  GVTRACE (" wr-tl-reply wr "PGUIDFMT"\n", PGUID (wr->e.guid));
+  struct ddsi_tkmap_instance *tk = ddsi_tkmap_lookup_instance_ref (gv->m_tkmap, serdata);
+  write_sample_gc (lookup_thread_state (), NULL, wr, serdata, tk);
+  ddsi_tkmap_instance_unref (gv->m_tkmap, tk);
+}
+
+void ddsi_tl_handle_request (struct ddsi_domaingv *gv, const ddsi_guid_prefix_t *guid_prefix, struct ddsi_serdata *sample_common)
+{
+  assert (!(sample_common->statusinfo & (NN_STATUSINFO_DISPOSE | NN_STATUSINFO_UNREGISTER)));
```
eboasson, line 266: I haven't checked yet, but in case I forget: if this is an assert, the caller should ensure that this precondition is met before calling this function.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+      ddsrt_free (types.types[n].type_object.value);
+    ddsrt_free (types.types);
+  }
+}
+
+void ddsi_tl_handle_reply (struct ddsi_domaingv *gv, struct ddsi_serdata *sample_common)
+{
+  struct generic_proxy_endpoint **gpe_match_upd = NULL;
+  uint32_t n_match_upd = 0;
+  assert (!(sample_common->statusinfo & (NN_STATUSINFO_DISPOSE | NN_STATUSINFO_UNREGISTER)));
```
eboasson, line 307: I haven't checked yet, but in case I forget: if this is an assert, the caller should ensure that this precondition is met before calling this function.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+      st = ddsrt_malloc (sizeof (*st));
+      ddsi_sertype_init_from_ser (gv, &st->c, &ddsi_sertype_ops_default, r.type_object.length, r.type_object.value);
+      ddsrt_mutex_lock (&gv->sertypes_lock);
+      ddsi_sertype_register_locked (&st->c);
+      ddsrt_mutex_unlock (&gv->sertypes_lock);
+
+      tlm->state = TL_META_RESOLVED;
+      tlm->sertype = &st->c; // refcounted by register
+
+      gpe_match_upd = ddsrt_malloc (tlm->n_endpoints * sizeof (*gpe_match_upd));
```
eboasson, line 331: This, inside the outer loop, is not going to work quite so well. And I'm tempted to suggest adding some generic list handling (in `ddsrt`) to make this sort of thing a bit less verbose. Something that constructs a linked list of (short) arrays of entries or so. It could be templatised with macros, too 🙂

src/core/ddsi/src/q_ddsi_discovery.c
```diff
@@ -1545,7 +1565,8 @@ int builtins_dqueue_handler (const struct nn_rsample_info *sampleinfo, const str
     if (pwr) guid = pwr->e.guid; else memset (&guid, 0, sizeof (guid));
     GVTRACE ("data(builtin, vendor %u.%u): "PGUIDFMT" #%"PRId64": ST%x %s/%s:%s%s\n",
              sampleinfo->rst->vendor.id[0], sampleinfo->rst->vendor.id[1],
-             PGUID (guid), sampleinfo->seq, statusinfo, d->topic->name, d->topic->type_name,
+             PGUID (guid), sampleinfo->seq, statusinfo,
+             pwr && (pwr->c.xqos->present & QP_TOPIC_NAME) ? pwr->c.xqos->topic_name : "", d->type->type_name,
```
eboasson, line 1569: It'd be better to guarantee a topic name is present in all endpoints (I actually think it is already, or else it could easily be done because it were only ever the built-in ones that did not have a name).

src/core/ddsi/src/q_qosmatch.c
```diff
@@ -63,73 +66,128 @@ int partitions_match_p (const dds_qos_t *a, const dds_qos_t *b)
   }
 }
 
-bool qos_match_mask_p (const dds_qos_t *rd, const dds_qos_t *wr, uint64_t mask, dds_qos_policy_id_t *reason)
+static bool check_assignability (struct tl_meta *rd_tlm, struct tl_meta *wr_tlm)
 {
+  assert (rd_tlm->sertype != NULL);
+  assert (wr_tlm->sertype != NULL);
+  return rd_tlm->sertype->ops->assignable_from (rd_tlm->sertype, wr_tlm->sertype);
```
eboasson, line 73: `ddsi_sertype_assignable_from`

src/core/ddsi/src/q_qosmatch.c
```diff
@@ -63,73 +66,128 @@ int partitions_match_p (const dds_qos_t *a, const dds_qos_t *b)
...
+    return false;
+
+  if ((rd_typeid == NULL || wr_typeid == NULL) && rd_qos->type_consistency.force_type_validation)
+  {
+    *reason = DDS_TYPE_CONSISTENCY_ENFORCEMENT_QOS_POLICY_ID;
     return false;
-  if ((mask & QP_TYPE_NAME) && strcmp (rd->type_name, wr->type_name) != 0)
+  }
+
+  struct tl_meta *rd_tlm = NULL, *wr_tlm = NULL;
```
eboasson, line 102: I'd make a function out of the type checking.

src/core/ddsi/src/q_entity.c
```diff
@@ -2753,12 +2803,20 @@ static bool topickind_qos_match_p_lock (struct entity_common *rd, const dds_qos_
...
   for (int i = 0; i < 2; i++)
     ddsrt_mutex_lock (locks[i + shift]);
-  bool ret = qos_match_p (rdqos, wrqos, reason);
+  bool ret = qos_match_p (gv, rdqos, rd_typeid, wrqos, wr_typeid, reason, &rd_type_lookup, &wr_type_lookup);
   for (int i = 0; i < 2; i++)
     ddsrt_mutex_unlock (locks[i + shift]);
+
+  if (!ret && (rd_type_lookup || wr_type_lookup))
+  {
+    (void) ddsi_tl_request_type (gv, rd_type_lookup ? rd_typeid : wr_typeid);
```
eboasson, line 2816: I get why you can do `rd_type_lookup ? rd_typeid : wr_typeid`, but it would be worth a comment.

src/core/ddsi/src/q_entity.c
```diff
@@ -4583,12 +4681,24 @@ static void create_proxy_builtin_endpoints(
       assert (is_builtin_entityid (guid1.entityid, proxypp->vendor));
       if (is_writer_entityid (guid1.entityid))
       {
+        // Replace topic name for each pwr
```
eboasson, line 4684: It seems to me that it would make more sense to refactor it a bit and pass topic & type name in as arguments.

src/core/ddsi/src/ddsi_sertype_default.c
```diff
@@ -0,0 +1,213 @@
...
+    return false;
+  if (
+    (a->type.m_nops > 0) &&
+    memcmp (a->type.m_ops, b->type.m_ops, a->type.m_nops * sizeof (*a->type.m_ops)) != 0)
+    return false;
+  assert (a->opt_size == b->opt_size);
+  return true;
+}
+
+static uint32_t sertype_default_hash (const struct ddsi_sertype *tpcmn)
```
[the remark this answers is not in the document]
dennis-adlink: As far as I have found, this is only used as the hash function for `gv->sertypes`. So lookups will use it and I don't think that causes any performance issues, as the resolved `sertype` is stored in the endpoint entities. I'll update this function to use the first 4 bytes of the type id hash.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -0,0 +1,356 @@
...
+
+int ddsi_tl_meta_equal (const struct tl_meta *a, const struct tl_meta *b)
+{
+  return ddsi_typeid_equal (a->type_id, b->type_id);
+}
+
+uint32_t ddsi_tl_meta_hash (const struct tl_meta *tl_meta)
+{
+  uint32_t h = 0;
+  h = ddsrt_mh3 (&tl_meta->type_id->hash, TYPEID_HASH_LENGTH, h);
```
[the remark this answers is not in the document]
dennis-adlink: In the current implementation the type identifier only consists of a hash of the type descriptor, so you're right that we can use this in the hash table for the type lookup administration. In the xtypes spec the type identifier is a union type and only optionally includes a hash, so at some point we'll need to build a hash value in this function. For now I'll use the type id hash.

dennis-adlink:
@eboasson Thanks for your extensive review! In the additional commits on this PR I've addressed most of the issues (I'll give a reply to your comments later). In one of the Travis builds on Windows for my own fork there was a refcounting issue, I'll also fix that.

src/core/ddsc/CMakeLists.txt
```diff
@@ -37,8 +37,8 @@ PREPEND(srcs_ddsc "${CMAKE_CURRENT_LIST_DIR}/src"
     dds_write.c
     dds_whc.c
     dds_whc_builtintopic.c
-    dds_serdata_builtintopic.c
-    dds_sertopic_builtintopic.c
+    dds_serdata_builtintype.c
```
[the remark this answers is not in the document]
dennis-adlink: I've reverted this name change in commit ebf6d2f

src/core/ddsc/include/dds/dds.h
```diff
@@ -206,6 +206,8 @@ typedef struct dds_builtintopic_endpoint
   dds_instance_handle_t participant_instance_handle;
   char *topic_name;
   char *type_name;
+  unsigned char *type_identifier;
+  uint32_t type_identifier_sz;
```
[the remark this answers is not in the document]
dennis-adlink: You're right, I changed it to size_t

src/core/ddsc/include/dds/dds.h
```diff
@@ -1071,21 +1074,23 @@ dds_create_topic(
...
  * @retval DDS_RETCODE_PRECONDITION_NOT_MET
- *             Mismatch between type name in sertopic and pre-existing
+ *             Mismatch between type name in sertype and pre-existing
  *             topic's type name.
  */
 DDS_EXPORT dds_entity_t
 dds_create_topic_generic (
   dds_entity_t participant,
-  struct ddsi_sertopic **sertopic,
+  const char *name,
```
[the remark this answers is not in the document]
dennis-adlink: Added `DDS_HAS_DDSI_SERTYPE` macro

src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
```diff
@@ -230,21 +230,21 @@ DDS_EXPORT inline bool ddsi_serdata_to_sample (const struct ddsi_serdata *d, voi
   return d->ops->to_sample (d, sample, bufptr, buflim);
 }
 
-DDS_EXPORT inline bool ddsi_serdata_topicless_to_sample (const struct ddsi_sertopic *topic, const struct ddsi_serdata *d, void *sample, void **bufptr, void *buflim) {
-  return d->ops->topicless_to_sample (topic, d, sample, bufptr, buflim);
+DDS_EXPORT inline bool ddsi_serdata_topicless_to_sample (const struct ddsi_sertype *type, const struct ddsi_serdata *d, void *sample, void **bufptr, void *buflim) {
```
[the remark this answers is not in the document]
dennis-adlink: Good point, I've changed all 'topicless' function names to 'untyped'

src/core/ddsi/src/ddsi_plist.c
```diff
@@ -1567,6 +1567,7 @@ static const struct piddesc piddesc_omg[] = {
   QP  (PARTITION,                           partition, XQ, XS, XSTOP),
   QP  (TIME_BASED_FILTER,                   time_based_filter, XD),
   QP  (TRANSPORT_PRIORITY,                  transport_priority, Xi),
+  QP  (TYPE_CONSISTENCY_ENFORCEMENT,        type_consistency, XE1, Xb, Xb, Xb, Xb, Xb),
```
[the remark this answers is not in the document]
dennis-adlink: Added Xbx3, Xbx4 and Xbx5

src/core/ddsi/include/dds/ddsi/ddsi_typeid.h
```diff
@@ -0,0 +1,68 @@
...
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+struct ddsi_sertype;
+
+typedef union {
+  unsigned char c[TYPEID_HASH_LENGTH];
+  uint32_t u[4];
```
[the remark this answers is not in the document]
dennis-adlink: Yes, the type identifier is in the current implementation a sequence of `uint8_t`, the `uint32_t` member was added just for convenient access for logging the value, but I rewrote that with some macros

src/core/ddsc/src/dds_domain.c
```diff
@@ -365,3 +365,47 @@ void dds_write_set_batch (bool enable)
...
+    rc = DDS_RETCODE_BAD_PARAMETER;
+    goto failed;
+  }
+
+  type_identifier_t type_id;
+  memcpy (&type_id.hash, type_identifier, TYPEID_HASH_LENGTH);
+  struct tl_meta *tlm = ddsi_tl_meta_lookup (&e->m_domain->gv, &type_id);
+  if (tlm == NULL)
+  {
+    rc = DDS_RETCODE_NOT_FOUND;
```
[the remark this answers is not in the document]
dennis-adlink: I didn't realize that `NOT_FOUND` is not defined in the spec. I've changed it so that `dds_domain_resolve_type` returns `PRECONDITION_NOT_MET` in case the type identifier is not known in the local type lookup administration or no type request writer could be found, and `NO_DATA` in case resolving the type failed within the specified duration.

src/core/ddsc/src/dds_domain.c
```diff
@@ -365,3 +365,47 @@ void dds_write_set_batch (bool enable)
...
+    goto failed;
+  }
+  if (tlm->state == TL_META_RESOLVED)
+    *sertype = tlm->sertype;
+  else
+  {
+    // FIXME: resolve the type information
+    //ddsi_tl_request_type (&e->m_domain->gv, &type_id);
+    rc = DDS_RETCODE_NOT_FOUND;
```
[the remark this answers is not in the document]
dennis-adlink: Yes, it definitely needs a timeout parameter, it wasn't there because I didn't finish the implementation yet. For the return code: I think both `NO_DATA` and `TIMEOUT` would fit here, but I agree with you that `TIMEOUT` is the best option because in most cases one expects the resolving to succeed and return a type.

src/core/ddsc/src/dds_serdata_builtintype.c
```diff
@@ -42,127 +42,127 @@ static uint32_t hash_guid (const ddsi_guid_t *g)
...
 }
 
-static void from_entity_rd (struct ddsi_serdata_builtintopic *d, const struct reader *rd)
+static void from_entity_rd (struct ddsi_serdata_builtintype *d, const struct reader *rd)
 {
   d->pphandle = rd->c.pp->e.iid;
-  ddsi_xqos_copy(&d->xqos, rd->xqos);
-  set_topic_type_from_sertopic(d, rd->topic);
+  ddsi_xqos_copy (&d->xqos, rd->xqos);
+  set_topic_name_type (d, rd->xqos->topic_name, rd->type->type_name);
```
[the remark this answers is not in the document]
dennis-adlink: Makes sense, I changed this so that `ent->xqos` always has the topic name set.

src/core/ddsc/src/dds_topic.c
```diff
@@ -152,7 +152,11 @@ static dds_return_t dds_topic_delete (dds_entity *e)
   struct dds_ktopic * const ktp = tp->m_ktopic;
   assert (dds_entity_kind (e->m_parent) == DDS_KIND_PARTICIPANT);
   dds_participant * const pp = (dds_participant *) e->m_parent;
-  ddsi_sertopic_unref (tp->m_stopic);
+  type_identifier_t *type_id = ddsi_typeid_from_sertype (tp->m_stype);
```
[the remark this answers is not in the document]
dennis-adlink: Yes, type identifiers can have variable size in the Xtypes spec, e.g. when using array or map collection types it will contain one or possibly two nested type identifiers. To be prepared for that, I've used dynamic allocation in the current implementation. The type hash field of the type identifier has a fixed length, currently 16 bytes and in the Xtypes implementation it will be 14 bytes.

src/core/ddsc/src/dds_topic.c
```diff
@@ -152,7 +152,11 @@ static dds_return_t dds_topic_delete (dds_entity *e)
...
+  ddsi_tl_meta_unref (&pp->m_entity.m_domain->gv, type_id, NULL);
```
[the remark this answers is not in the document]
dennis-adlink: Good point, I've changed this so that ref and unref have the same interface.

src/core/ddsi/src/ddsi_sertype.c
```diff
@@ -0,0 +1,183 @@
...
+{
+  assert (sz);
+  assert (buf);
+  uint32_t tnsz = (uint32_t) strlen (tp->type_name);
+  size_t i = 0;
+  *sz = sizeof (tnsz) + tnsz + 1;
+  *buf = ddsrt_malloc (*sz);
+  *(uint32_t *)(*buf + i) = ddsrt_toBE4u (tnsz);
+  i += sizeof (tnsz);
+  memcpy (*buf + i, tp->type_name, tnsz);
```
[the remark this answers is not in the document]
dennis-adlink: Updated this in f29fa4a to use the `pserop` serializer, for both `ddsi_sertype` and `ddsi_sertype_default` (as you mentioned below).

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: Added a `request_seqno` field in `struct tl_meta`

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: I agree, removed the count > 0 condition here, so it'll always reply to an incoming type lookup request (unless there is no typelookup reply writer, which should of course not happen where the is a request reader).

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: I moved this to `q_ddsi_discovery` as a condition for calling `ddsi_tl_handle_request` and `ddsi_tl_handle_request`

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: I've created an abstraction (the `tlm_endpoints_*` functions) for the endpoint list in `struct tl_meta`, which is currently implemented as a simple array, but can easily be replaced by another list implementation if required.

dennis-adlink:
@eboasson the refcounting issue I mentioned in my previous comment is fixed in 9767e54, as well as a bug in locking for the condition variable used by `dds_domain_resolve_type`. The builds on Travis are running stable now for my fork. I'll rebase the PR branch after you've re-reviewed the updates. 
Two things I'll add in short term: the comment on the `dds_domain_resolve_type` functions needs to mention the refcounting for sertype that needs be handled by the application, and I'll add preprocessor conditionals for the type lookup functionality, so that it can be excluded from the build.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -38,56 +38,117 @@ size_t typelookup_service_request_nops = sizeof (typelookup_service_request_ops)
...
+static void tlm_endpoint_add (struct ddsi_domaingv *gv, struct tl_meta *tlm, const ddsi_guid_t *proxy_ep_guid)
 {
-  if (tlm->sertype != NULL)
-    ddsi_sertype_unref ((struct ddsi_sertype *) tlm->sertype);
-  ddsrt_free (tlm->type_id);
-  ddsrt_free (tlm->endpoints);
-  ddsrt_free (tlm);
+  if (proxy_ep_guid != NULL)
+  {
+    if (!tlm_endpoint_exists (tlm, proxy_ep_guid))
```
eboasson, line 60: This data structure doesn't scale. As long as systems are small, it is not an issue, and it doesn't make sense to delay merging the PR (especially considering that it defaults to being disabled) because of it, but I do think it needs to be noted.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -38,56 +38,117 @@ size_t typelookup_service_request_nops = sizeof (typelookup_service_request_ops)
...
 
-static void ddsi_tl_meta_fini (struct tl_meta *tlm)
+static void tlm_endpoint_add (struct ddsi_domaingv *gv, struct tl_meta *tlm, const ddsi_guid_t *proxy_ep_guid)
 {
-  if (tlm->sertype != NULL)
-    ddsi_sertype_unref ((struct ddsi_sertype *) tlm->sertype);
-  ddsrt_free (tlm->type_id);
-  ddsrt_free (tlm->endpoints);
-  ddsrt_free (tlm);
+  if (proxy_ep_guid != NULL)
```
eboasson, line 58: I don't like this test here. It does make sense to represent references from topic entities by using a null pointer in the GUID in add/remove, but I don't think that should be forwarded all the way to `endpoint_add` and `endpoint_remove`.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -138,124 +198,140 @@ void ddsi_tl_meta_ref (struct ddsi_domaingv *gv, const type_identifier_t *type_i
...
-    }
-    if (!found)
-    {
-      tlm->n_endpoints++;
-      tlm->endpoints = ddsrt_realloc (tlm->endpoints, tlm->n_endpoints * sizeof (*tlm->endpoints));
-      memcpy (&tlm->endpoints[tlm->n_endpoints - 1], src_ep_guid, sizeof (*src_ep_guid));
-      GVTRACE (" add ep "PGUIDFMT, PGUID (*src_ep_guid));
-    }
-  }
+  tlm_endpoint_add (gv, tlm, proxy_ep_guid);
```
eboasson, line 206: So this call should be conditional on `proxy_ep_guid` being non-null. A few doc comments for the interface exported by this module would be a good idea, too, by the way.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -138,124 +198,140 @@ void ddsi_tl_meta_ref (struct ddsi_domaingv *gv, const type_identifier_t *type_i
...
+    resolved = true;
   }
   if (dst != NULL && guid_prefix_empty (&tlm->dst_prefix) && !guid_prefix_empty (dst))
     memcpy (&tlm->dst_prefix, dst, sizeof (*dst));
   tlm->refc++;
-  GVTRACE (" state %d refc %u\n", tlm->state, tlm->refc);
+  GVTRACE (" state %d refc %u", tlm->state, tlm->refc);
   ddsrt_mutex_unlock (&gv->tl_admin_lock);
+
+  if (resolved)
```
eboasson, line 220: Why can't this simply be all within `tl_admin_lock`?

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -138,124 +198,140 @@ void ddsi_tl_meta_ref (struct ddsi_domaingv *gv, const type_identifier_t *type_i
...
 {
-  struct entidx_enum_participant est;
-  struct participant *pp = NULL;
-  bool pp_found = false;
-  entidx_enum_participant_init (&est, gv->entity_index);
-  while (!pp_found && (pp = entidx_enum_participant_next (&est)) != NULL)
+  struct participant *pp;
+  struct writer *wr = NULL;
+  thread_state_awake (lookup_thread_state (), gv);
+  if (guid_prefix_empty (guid_prefix))
```
eboasson, line 259: See the comment about `dst_prefix` where the containing struct type is defined. I'm not so sure it's worth having.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -280,7 +356,7 @@ void ddsi_tl_handle_request (struct ddsi_domaingv *gv, const ddsi_guid_prefix_t
       size_t sz;
       types.n++;
       types.types = ddsrt_realloc (types.types, types.n * sizeof (*types.types));
-      memcpy (&types.types[types.n - 1].type_identifier, tlm->type_id, sizeof (types.types[types.n - 1].type_identifier));
+      types.types[types.n - 1].type_identifier = tlm->type_id;
       tlm->sertype->ops->serialize (tlm->sertype, &sz, &types.types[types.n - 1].type_object.value);
```
eboasson, line 360: Error handling? I mean, it is defined as returning success/failure, and one does need to at least allow for types that do not support it.

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -289,68 +365,118 @@ void ddsi_tl_handle_request (struct ddsi_domaingv *gv, const ddsi_guid_prefix_t
...
   {
     ddsrt_mutex_lock (&gv->tl_admin_lock);
     type_identifier_type_object_pair_t r = reply->types.types[n];
     struct tl_meta *tlm = ddsi_tl_meta_lookup_locked (gv, &r.type_identifier);
-    if (tlm != NULL && tlm->state == TL_META_REQUESTED && tlm->n_endpoints > 0)
+    if (tlm != NULL && tlm->state == TL_META_REQUESTED && !tlm_endpoints_empty (tlm))
     {
+      bool sertype_new = false;
       GVTRACE (" type "PTYPEIDFMT, PTYPEID (r.type_identifier));
       st = ddsrt_malloc (sizeof (*st));
```
eboasson, line 430: For the time being I am fine with assuming `ddsi_sertype_default` here

src/core/ddsi/src/ddsi_typelookup.c
```diff
@@ -289,68 +365,118 @@ void ddsi_tl_handle_request (struct ddsi_domaingv *gv, const ddsi_guid_prefix_t
...
+    write_typelookup_reply (wr, req->sequence_number, &types);
+  else
+    GVTRACE (" no tl-reply writer");
+
+  for (uint32_t n = 0; n < types.n; n++)
+    ddsrt_free (types.types[n].type_object.value);
+  ddsrt_free (types.types);
+}
+
+static void ddsi_tl_meta_proxy_endpoint_ref_locked (struct ddsi_domaingv *gv, struct tl_meta *tlm)
```
eboasson, line 379: The name is a bit confusing because the `ref` and `unref` suffix are pretty consistently used for incrementing and decrementing reference counts, and this is not doing that at all.

src/core/ddsi/src/q_entity.c
```diff
@@ -2793,7 +2811,18 @@ static void reader_qos_mismatch (struct reader * rd, dds_qos_policy_id_t reason)
   }
 }
 
-static bool topickind_qos_match_p_lock (struct ddsi_domaingv *gv, struct entity_common *rd, const dds_qos_t *rdqos, const type_identifier_t *rd_typeid, struct entity_common *wr, const dds_qos_t *wrqos, const type_identifier_t *wr_typeid, dds_qos_policy_id_t *reason)
+static bool topickind_qos_match_p_lock (
```
eboasson, line 2814: I have been staring at this (as well as `qos_match_p`) and I don't quite like the set of parameters, but I'm not quite sure much can be done to improve the situation: passing a struct in would require the struct to be initialized, &c. So I'm just mentioning it in case you have a good suggestion 🙂.

src/core/ddsi/include/dds/ddsi/ddsi_typelookup.h
```diff
@@ -76,17 +77,17 @@ extern size_t typelookup_service_reply_nops;
 
 int ddsi_tl_meta_equal (const struct tl_meta *a, const struct tl_meta *b);
 uint32_t ddsi_tl_meta_hash (const struct tl_meta *tl_meta);
-void ddsi_tl_meta_ref (struct ddsi_domaingv *gv, const type_identifier_t *type_id, const struct ddsi_sertype *type, const ddsi_guid_t *src_ep_guid, const ddsi_guid_prefix_t *dst);
+void ddsi_tl_meta_ref (struct ddsi_domaingv *gv, const type_identifier_t *type_id, const struct ddsi_sertype *type, const ddsi_guid_t *proxy_ep_guid, const ddsi_guid_prefix_t *dst);
```
eboasson, line 80: I think it would be advisable to have different versions of `meta_ref` and `meta_unref` so that there are fewer null pointers being passed in. For `ref` it really splits into local entities vs proxies, and for `unref` based on type_id or type.

I am fine with these being wrappers that forward it to the current implementation, it's just that I think it'd be clearer at the call sites.

src/core/ddsc/tests/typelookup.c
```diff
@@ -0,0 +1,341 @@
...
+    else
+      dds_sleepfor (DDS_MSECS (20));
+  }
+  while (result == NULL && dds_time() - t_start <= timeout);
+  CU_ASSERT_FATAL (result != NULL);
+
+  return result;
+}
+
+static void endpoint_info_free (endpoint_info_t *ep_info)
```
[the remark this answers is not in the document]
dennis-adlink: I've used a custom data structure to store only the required data, so I think this free function can stay here.

src/core/ddsi/include/dds/ddsi/q_entity.h
```diff
@@ -419,8 +421,10 @@ struct proxy_endpoint_common
   ddsi_guid_t group_guid; /* 0:0:0:0 if not available */
   nn_vendorid_t vendor; /* cached from proxypp->vendor */
   seqno_t seq; /* sequence number of most recent SEDP message */
+#ifdef DDSI_INCLUDE_TYPE_DISCOVERY
   type_identifier_t type_id; /* type identifier for for type used by this proxy endpoint */
-  const struct ddsi_sertype * type; /* type */
+#endif
+  const struct ddsi_sertype * type; /* sertype for data this endpoint reads/writes */
```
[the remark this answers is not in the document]
dennis-adlink: Yes you're right, I've changed this

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: Moved the line to within the lock

src/core/ddsi/src/q_entity.c
dennis-adlink: I didn't really find a better solution for this, maybe the planned refactoring of fields in (proxy) entities will help in reducing the number of parameters for these function calls.

dennis-adlink (commented):
Thanks for reviewing again @eboasson, in commits ac26478 and b53823e I've addressed the issues you mentioned.

src/core/ddsi/include/dds/ddsi/ddsi_typelookup.h
dennis-adlink: For both `ref` and `unref` I've split up the function into a `proxy` and `local` variant; I think it's more clear to use proxy/local this for unref too, instead of having a `type_id` and `type` variant.

src/core/ddsi/src/ddsi_typelookup.c
dennis-adlink: I agree, I've changed it to `ddsi_tl_meta_register_with_proxy_endpoints`, which better describes what the function does and is less confusing.

src/core/ddsc/src/dds_domain.c
```diff
@@ -378,28 +385,50 @@ dds_return_t dds_domain_resolve_type (dds_entity_t entity, unsigned char *type_i
...
+    ddsrt_mutex_unlock (&gv->tl_admin_lock);
+  }
   else
   {
-    // FIXME: resolve the type information
-    //ddsi_tl_request_type (&e->m_domain->gv, &type_id);
-    rc = DDS_RETCODE_NOT_FOUND;
-    goto failed;
+    ddsrt_mutex_unlock (&gv->tl_admin_lock);
+    ddsrt_mutex_lock (&gv->tl_resolved_lock);
```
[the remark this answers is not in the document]
dennis-adlink: I've refactored this so that it's now only using the `gv->tl_admin_lock` mutex. After requesting the type and re-acquiring the lock, there is a test to check if the type is already resolved, as it may have been resolved after the lock was released in `ddsi_tl_request_type` (which is required because when writing the sample we shouldn't hold the lock). In that case there is no need to wait for `gv->tl_resolved_cond` to be triggered. 

In `ddsi_tl_request_type` there is a test `tlm->state == TL_META_NEW` to avoid duplicate requests for a type, so I think that makes it the 'lookup-and-request-if-unknown' function you mean?

src/core/ddsi/src/ddsi_sertype.c
```diff
@@ -118,40 +118,38 @@ struct sertype_ser
...
 }
 
-bool ddsi_sertype_deserialize (struct ddsi_sertype *tp, size_t src_sz, const unsigned char *src_data, size_t *src_pos)
+bool ddsi_sertype_deserialize (struct ddsi_domaingv *gv, struct ddsi_sertype *tp, const struct ddsi_sertype_ops *sertype_ops, size_t sz, unsigned char *serdata)
 {
-  assert (src_pos);
   struct sertype_ser d;
   size_t srcoff = 0;
-  if (plist_deser_generic_srcoff (&d, src_data, src_sz, &srcoff, false, sertype_ser_ops) < 0)
+  if (plist_deser_generic_srcoff (&d, serdata, sz, &srcoff, DDSRT_ENDIAN != DDSRT_LITTLE_ENDIAN, sertype_ser_ops) < 0)
```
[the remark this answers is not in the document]
dennis-adlink: The last parameter of `plist_ser_generic_embeddable` is a bool that indicates big-endian encoding, not bswap y/n, so that's set to `false` because the xtypes spec (7.3.4.5) requires LE encoding for TypeObject serialization. The deser functions have a `bool bswap` parameter, which make it somewhat confusing. So maybe we could change the ser functions to use `bool bswap` as well?

dennis-adlink:
@eboasson thanks for reviewing again! I've covered most of the remarks from your review in commit 8defdec (see also the commit message), except for the byteswapping, which is not wrong I think, but somewhat confusing because of the parameters of the ser/deser functions.

src/core/ddsi/src/ddsi_sertype.c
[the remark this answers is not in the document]
eboasson: > because the xtypes spec (7.3.4.5) requires LE encoding for TypeObject serialization

Dealing with OMG specs is like dealing with Windows: cruel and unusual punishment — but at least the organisation has a good acronym ... So the security spec mandates big-endian encoding (unnecessarily) while the XTypes spec mandates little-endian encoding (equally unnecessarily) even though the original DDSI specification leaves the decision to the sender except in the one case where a standard endianness is required (the perfidious keyhashes (\*)) and mandates that the receiver can handle either. (I admit to not having read all the fine print in the XTypes spec ...)

The last parameter of `plist_ser_generic_embeddable` is a bool that controls native vs big-endian encoding and was added to accomodate the Security spec's foibles, but setting it to `false` doesn't allow for the XTypes spec's insanity because on a big-endian machine that'll give you big-endian. So it needs to be complexified further and changed into a tri-state.

(\*) The keyhash is perfidious because for keys of which the serialized length may be larger than 16 bytes, it uses the MD5 hash, and therefore creating a topic with an octet sequence as key and writing two instances with keys that happen to produce an MD5 hash collision has "interesting" consequences for implementations that rely on the keyhash. And yet the DDS Security spec mandates the presence of the keyhash for those cases where the readers can't decode the data but need to keep instances separate, effectively forbidding applications from choosing the key values freely, but rather requiring them to avoid MD5 collisions. Not much good can be said about that.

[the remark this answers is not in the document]
eboasson: P.S.: of course a boolean that says whether or not to byteswap while serialising suffices to do any of the modes we need. But that pushes the tests like `DDSRT_ENDIAN != DDSRT_LITTLE_ENDIAN`) to the caller. I think the interface would be easier to use if you could specify the desired result as native/little/big endian and then let the serializer function figure out whether or not it needs to do byteswapping.

dennis-adlink:
Thanks @eboasson! I agree with your suggestion to replace the assert on type info length with a check, that's fixed in c95520e. The refactoring of the endianness parameter in serialization is in c399890, and finally I fixed some null function pointers that should be 0 in commit 6be9031.

eboasson (approved):
@dennis-adlink this PR has spent an embarrassingly long time with me and I am grateful that you have been so patient. Today, I am happy to finally declare it suitable for merging, yay!

Still, the actual merging will have to wait for a little longer: because it breaks source compatibility (which is fine, it is just an evolving interface) some updates need to go into other repositories first. In particular:
* `rmw_cyclonedds_cpp`, for which I have got the changes done and have locally checked that it works with "old" and "new" sources, and also that an old `rmw_cyclonedds_cpp` is binary compatible with a "new" `cyclonedds`. The PR for that should show up today or tomorrow.
* `cyclonedds-cxx`, which needs a similar adjustment to what I did for `rmw_cyclonedds_cpp`. Perhaps @reicheratwork would be willing to have a look?
* the Rust binding on https://github.com/atolab (I suspect it uses a pinned version of `cyclonedds`, it which case that can wait).

PR #564, opened 2020-07-24
Minor improvements to content filter interface

eboasson:
Content filtering is possible in current Cyclone by setting a callback function on a topic, but this is not an interface that is intended to survive.  Still, it can save the day.  This commit improves the interface a bit by allowing an argument to the filter function.  It also adds some tests to verify that it does work.

Also point out in the header file that this really is not an interface that will be around for long and that it needs to be used with care.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -1142,31 +1142,83 @@ dds_get_type_name(dds_entity_t topic, char *name, size_t size);
 
 /** Topic filter function */
 typedef bool (*dds_topic_filter_fn) (const void * sample);
+typedef bool (*dds_topic_filter_arg_fn) (const void * sample, void * arg);
 
 /**
- * @brief Sets a filter on a topic.
+ * @brief Sets a filter on a topic. To be replaced by proper filtering on readers,
+ * no guarantee that this will be maintained for backwards compatibility.
+ *
+ * Not thread-safe with respect to data being read/written using readers/writers
+ * using this topic.  Be sure to create a topic entity specific to the reader you
+ * want to filter, then set the filter function, and only then create the reader.
+ * And don't change it unless you know there are no concurrent writes.
  *
  * @param[in]  topic   The topic on which the content filter is set.
  * @param[in]  filter  The filter function used to filter topic samples.
  */
-DDS_EXPORT void
+DDS_DEPRECATED_EXPORT void
 dds_set_topic_filter(dds_entity_t topic, dds_topic_filter_fn filter);
 
 DDS_DEPRECATED_EXPORT void
 dds_topic_set_filter(dds_entity_t topic, dds_topic_filter_fn filter);
 
 /**
- * @brief Gets the filter for a topic.
+ * @brief Sets a filter and filter argument on a topic. To be replaced by proper
+ * filtering on readers, no guarantee that this will be maintained for backwards
+ * compatibility.
+ *
+ * Not thread-safe with respect to data being read/written using readers/writers
+ * using this topic.  Be sure to create a topic entity specific to the reader you
+ * want to filter, then set the filter function, and only then create the reader.
+ * And don't change it unless you know there are no concurrent writes.
+ *
+ * @param[in]  topic   The topic on which the content filter is set.
+ * @param[in]  filter  The filter function used to filter topic samples.
+ * @param[in]  arg     Argument for the filter function.
+ *
+ * @returns A dds_return_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK  Filter set successfully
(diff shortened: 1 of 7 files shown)


src/core/ddsc/src/dds_topic.c
```diff
@@ -500,71 +501,71 @@ dds_entity_t dds_find_topic (dds_entity_t participant, const char *name)
...
 }
 
-static void dds_topic_mod_filter (dds_entity_t topic, dds_topic_intern_filter_fn *filter, void **ctx, bool set)
+static bool topic_filter_no_arg_wrapper (const void *sample, void *arg)
+{
+  dds_topic_filter_fn f = (dds_topic_filter_fn) arg;
+  return f (sample);
+}
+
+static void dds_set_topic_filter_old (dds_entity_t topic, dds_topic_filter_fn filter)
```
dennis-adlink, line 522: Nit-picking: maybe use `_deprecated` here to make clear that it will be removed in future?

src/core/ddsc/include/dds/dds.h
```diff
@@ -1142,21 +1142,50 @@ dds_get_type_name(dds_entity_t topic, char *name, size_t size);
...
+ * @param[in]  filter  The filter function used to filter topic samples.
+ * @param[in]  arg     Argument for the filter function.
+ *
+ * @returns A dds_return_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK  Filter set successfully
+ * @retval DDS_RETCODE_BAD_PARAMETER  The topic handle is invalid
+*/
+DDS_EXPORT dds_return_t
+dds_set_topic_filter_arg(dds_entity_t topic, dds_topic_filter_arg_fn filter, void *arg);
```
dennis-adlink, line 1184: The name `set_.._filter_arg` may be somewhat confusing, in the sense that it seems like it can be used to set just the argument for the filter function that was set by `set_topic_filter`. Of course the function parameters make clear that it also sets the filter function, and it's a temporary solution, so this is not really an issue.

src/core/ddsc/include/dds/dds.h
```diff
@@ -1142,21 +1142,50 @@ dds_get_type_name(dds_entity_t topic, char *name, size_t size);
...
-DDS_EXPORT void
+DDS_DEPRECATED_EXPORT void
 dds_set_topic_filter(dds_entity_t topic, dds_topic_filter_fn filter);
 
 DDS_DEPRECATED_EXPORT void
 dds_topic_set_filter(dds_entity_t topic, dds_topic_filter_fn filter);
 
 /**
- * @brief Gets the filter for a topic.
+ * @brief Sets a filter on a topic. To be replaced by proper filtering on readers,
```
dennis-adlink, line 1166: Nit-pick: Sets a filter on a topic _with an argument that will be passed to the filter function_

src/core/ddsc/include/dds/dds.h
```diff
@@ -1142,21 +1142,50 @@ dds_get_type_name(dds_entity_t topic, char *name, size_t size);
...
+ * @returns A dds_return_t indicating success or failure.
+ *
+ * @retval DDS_RETCODE_OK  Filter set successfully
+ * @retval DDS_RETCODE_BAD_PARAMETER  The topic handle is invalid
+*/
+DDS_EXPORT dds_return_t
+dds_set_topic_filter_arg(dds_entity_t topic, dds_topic_filter_arg_fn filter, void *arg);
+
+/**
+ * @brief Gets the filter for a topic. To be replaced by proper filtering on readers,
```
dennis-adlink, line 1187: Although it's a temporary situation having these filter functions on a topic, I think it would be better to keep the getters in line with the setters and add a `get_topic_filter_with_arg` that returns a `dds_topic_filter_arg_fn`

src/core/ddsc/tests/filter.c
```diff
@@ -0,0 +1,331 @@
...
+      rd[i][j] = dds_create_reader (dp[i], tp[i][j], qos, NULL);
+      CU_ASSERT_FATAL (rd[i][j] > 0);
+      wr[i][j] = dds_create_writer (dp[i], tp[i][j], qos, NULL);
+      CU_ASSERT_FATAL (wr[i][j] > 0);
+    }
+  }
+  dds_delete_qos (qos);
+
+  dds_set_topic_filter_arg (tp[0][0], filter_long1_eq, (void *) 1);
+  dds_set_topic_filter_arg (tp[1][0], filter_long2_eq, (void *) 1);
```
dennis-adlink, line 145: Maybe add a line to test the get_topic_filter function result?

eboasson:
Thanks @dennis-adlink, those are all sensible comments. I've updated the PR accordingly.

PR #567, opened 2020-07-27
Use-after-free in "security FSM" and test failures

eboasson:
This PR addresses two issues with the FSM used for performing the handshake when security is enabled:
* A potential use-after-free on calling `dds_security_fsm_free `: there was no synchronisation between the thread performing the state transitions and the one freeing the memory, but the former invokes functions associated with the transition and the new state without holding any locks. This takes the easy way out by adding a `busy` flag ...
* A race condition in the FSM tests caused by some of those tests waiting for a certain state to be reached (by polling `dds_security_fsm_current_state`), with the assumption that once that state is reached, all function invocations associated with reaching that state will have occurred. The state transition logic however sets the `current` state to the state after the transition prior to invoking these functions. This PR solves this by no longer relying on the "current" state, but instead on the callback functions themselves and on whether or not the end state has been reached.

There's been a bit of a reshuffle and reformat in the test code. The test code was (and remains) a mess, and, frankly, the FSM itself doesn't make me very happy either. But for now solving these bugs will have to do.

The various "fsm" test failures should be fixed by this, I may also fix a rare segmentation fault in some of the security tests, but I can't be sure at this point.

### DIFF
--- a/src/security/core/include/dds/security/core/dds_security_fsm.h
+++ b/src/security/core/include/dds/security/core/dds_security_fsm.h
@@ -142,10 +142,10 @@ dds_security_fsm_dispatch(struct dds_security_fsm *fsm, int32_t event_id, bool p
  *
  * @param fsm       The state machine
  *
- * @return          The current state of the given state machine
+ * @return          true iff fsm not in initial or final state
  */
-DDS_EXPORT const dds_security_fsm_state*
-dds_security_fsm_current_state(struct dds_security_fsm *fsm);
+DDS_EXPORT bool
+dds_security_fsm_running(struct dds_security_fsm *fsm);
 
 /**
  * Stops the state machine.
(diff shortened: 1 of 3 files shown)


k0ekk0ek:
@eboasson, I trust this solves the segmentation fault you mentioned before?

eboasson:
> @eboasson, I'm guessing this solves the segmentation fault you mentioned before?

I suspect it may. I accidentally pressed a wrong key and turned out to be a shortcut for submitting the PR 🙂

eboasson:
> > @eboasson, I'm guessing this solves the segmentation fault you mentioned before?
> 
> I suspect it may. I accidentally pressed a wrong key and turned out to be a shortcut for submitting the PR 🙂

It doesn't ...

PR #570, opened 2020-07-30
Delayed spring cleaning

k0ekk0ek:
This PR updates the platforms and compilers used to build the software (except Windows) and tidies up `.travis.yml` a little. It'll introduce a little overhead because there are no [Bintray](https://bintray.com/) packages for the newer compilers yet, but I think it's worth the switch. Using GCC 10 will introduce warnings in the CUnit dependency, but I fixed those yesterday and opened a [PR](https://github.com/bincrafters/conan-cunit/pull/3) for the Bincrafters repository. Other than that I added a workaround for a `stringop-overflow` warning introduced by switching to GCC 10.


.travis.yml
```diff
@@ -37,54 +37,60 @@ submit_to_coverity_scan: &submit_to_coverity_scan
...
+      packages: [ gcc-10 g++-10 ]
   before_install:
-    - eval "export CC=gcc-8"
-    - eval "export CXX=g++-8"
+    - eval "export CC=gcc-10 CXX=g++-10"
     - eval "export COV_COMPTYPE=gcc COV_PLATFORM=linux64"
+    - eval "export BUILD_TOOL_OPTIONS='-j 4'"
+    - eval "export GENERATOR='Unix Makefiles'"
   install:
-    - *install_coverity
```
eboasson, line 54: Why is this no longer needed?

.travis.yml
```diff
@@ -37,54 +37,60 @@ submit_to_coverity_scan: &submit_to_coverity_scan
...
-osx_xcode: &osx_xcode
+macos1015_xcode11_6: &macos1015_xcode11_6
   os: osx
-  osx_image: xcode11.1
+  osx_image: xcode11.6
   compiler: clang
   addons:
     homebrew:
-      packages: [ python3 ]
+      packages: [ pyenv-virtualenv ]
```
k0ekk0ek, line 83: This may also not exactly what I intended. `pyenv-virtualenv` probably has `python3` as a dependency... will fix this one too.

.travis.yml
k0ekk0ek: This is a typical example of why reviews are always a good idea... :relaxed: Thanks @eboasson!

k0ekk0ek:
@eboasson, I've processed the comments. Is it ok if I work the last one into the actual change, force-push and then merge?

eboasson:
> @eboasson, I've processed the comments. Is it ok if I work the last one into the actual change, force-push and then merge?

Yes, please do.

PR #571, opened 2020-08-06
Fix for locater test and Coverity issues

dennis-adlink:
Commit d9b0d1c fixes failing locator tests by disabling some of the tests that timeout when calling `getaddrinfo`. The default timeout for name resolving on linux is 5s with 2 retries, so keeping this tests would require an undesirable long timeout for these test cases. Commit bf2a506 contains fixes for some of the Coverity issues.

### DIFF
--- a/src/core/ddsi/tests/locators.c
+++ b/src/core/ddsi/tests/locators.c
@@ -77,7 +77,7 @@ CU_Test (ddsi_locator_from_string, bogusproto)
   CU_ASSERT_FATAL (res == AFSR_UNKNOWN);
   res = ddsi_locator_from_string (&gv, &loc, "bogusproto/xyz:1234", fact);
   CU_ASSERT_FATAL (res == AFSR_UNKNOWN);
-  res = ddsi_locator_from_string (&gv, &loc, "bogusproto/1.2.3.4:1234", fact);
+  res = ddsi_locator_from_string (&gv, &loc, "bogusproto/192.0.2.0:1234", fact);
   CU_ASSERT_FATAL (res == AFSR_UNKNOWN);
   fini (&gv);
 }
@@ -110,10 +110,10 @@ CU_Theory ((enum transport_selector tr), ddsi_locator_from_string, ipv4_invalid)
   snprintf (astr, sizeof (astr), "%s/:1234", fact->m_typename);
   res = ddsi_locator_from_string (&gv, &loc, astr, fact);
   CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
-  snprintf (astr, sizeof (astr), "%s/[1.2.3.4]", fact->m_typename);
+  snprintf (astr, sizeof (astr), "%s/[192.0.2.0]", fact->m_typename);
   res = ddsi_locator_from_string (&gv, &loc, astr, fact);
   CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
-  snprintf (astr, sizeof (astr), "%s/[1.2.3.4]:1234", fact->m_typename);
+  snprintf (astr, sizeof (astr), "%s/[192.0.2.0]:1234", fact->m_typename);
   res = ddsi_locator_from_string (&gv, &loc, astr, fact);
   CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
   fini (&gv);
@@ -166,28 +166,28 @@ CU_Theory ((enum transport_selector tr), ddsi_locator_from_string, ipv4)
   }
 #endif
 
-  res = ddsi_locator_from_string (&gv, &loc, "1.2.3.4", fact);
+  res = ddsi_locator_from_string (&gv, &loc, "192.0.2.0", fact);
   CU_ASSERT_FATAL (res == AFSR_OK);
   CU_ASSERT_FATAL (loc.kind == fact->m_kind);
   CU_ASSERT_FATAL (loc.port == NN_LOCATOR_PORT_INVALID);
   CU_ASSERT_FATAL (loc.tran == fact);
-  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){1,2,3,4}));
+  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){192,0,2,0}));
 
-  snprintf (astr, sizeof (astr), "%s/1.2.3.4", fact->m_typename);
+  snprintf (astr, sizeof (astr), "%s/192.0.2.0", fact->m_typename);
   res = ddsi_locator_from_string (&gv, &loc, astr, fact);
   CU_ASSERT_FATAL (res == AFSR_OK);
   CU_ASSERT_FATAL (loc.kind == fact->m_kind);
   CU_ASSERT_FATAL (loc.port == NN_LOCATOR_PORT_INVALID);
   CU_ASSERT_FATAL (loc.tran == fact);
-  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){1,2,3,4}));
+  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){192,0,2,0}));
 
-  snprintf (astr, sizeof (astr), "%s/1.2.3.4:1234", fact->m_typename);
+  snprintf (astr, sizeof (astr), "%s/192.0.2.0:1234", fact->m_typename);
   res = ddsi_locator_from_string (&gv, &loc, astr, fact);
   CU_ASSERT_FATAL (res == AFSR_OK);
   CU_ASSERT_FATAL (loc.kind == fact->m_kind);
   CU_ASSERT_FATAL (loc.port == 1234);
   CU_ASSERT_FATAL (loc.tran == fact);
-  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){1,2,3,4}));
+  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){192,0,2,0}));
   fini (&gv);
 }
 
@@ -197,12 +197,12 @@ CU_Test (ddsi_locator_from_string, ipv4_cross1)
   struct ddsi_tran_factory * const fact = init (&gv, TRANS_UDP);
   nn_locator_t loc;
   enum ddsi_locator_from_string_result res;
-  res = ddsi_locator_from_string (&gv, &loc, "tcp/1.2.3.4:1234", fact);
+  res = ddsi_locator_from_string (&gv, &loc, "tcp/192.0.2.0:1234", fact);
   CU_ASSERT_FATAL (res == AFSR_OK);
   CU_ASSERT_FATAL (loc.kind == NN_LOCATOR_KIND_TCPv4);
   CU_ASSERT_FATAL (loc.port == 1234);
   CU_ASSERT_FATAL (loc.tran != fact);
-  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){1,2,3,4}));
+  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){192,0,2,0}));
   fini (&gv);
 }
 
@@ -212,12 +212,12 @@ CU_Test (ddsi_locator_from_string, ipv4_cross2)
   struct ddsi_tran_factory * const fact = init (&gv, TRANS_TCP);
   nn_locator_t loc;
   enum ddsi_locator_from_string_result res;
-  res = ddsi_locator_from_string (&gv, &loc, "udp/1.2.3.4:1234", fact);
+  res = ddsi_locator_from_string (&gv, &loc, "udp/192.0.2.0:1234", fact);
   CU_ASSERT_FATAL (res == AFSR_OK);
   CU_ASSERT_FATAL (loc.kind == NN_LOCATOR_KIND_UDPv4);
   CU_ASSERT_FATAL (loc.port == 1234);
   CU_ASSERT_FATAL (loc.tran != fact);
-  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){1,2,3,4}));
+  CU_ASSERT_FATAL (check_ipv4_address (&loc, (uint8_t[]){192,0,2,0}));
   fini (&gv);
 }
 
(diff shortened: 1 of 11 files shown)


src/core/ddsi/tests/locators.c
```diff
@@ -110,12 +110,15 @@ CU_Theory ((enum transport_selector tr), ddsi_locator_from_string, ipv4_invalid)
...
-  CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
+  // The following tests are disabled because they run into a timeout
+  // when calling getaddrinfo, and this timeout depends on the configured
+  // timeout and retries in the OS resolve configuration
+  // snprintf (astr, sizeof (astr), "%s/[1.2.3.4]", fact->m_typename);
+  // res = ddsi_locator_from_string (&gv, &loc, astr, fact);
+  // CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
+  // snprintf (astr, sizeof (astr), "%s/[1.2.3.4]:1234", fact->m_typename);
+  // res = ddsi_locator_from_string (&gv, &loc, astr, fact);
+  // CU_ASSERT_FATAL (res == AFSR_INVALID || res == AFSR_UNKNOWN);
```
k0ekk0ek, line 121: Perhaps it's better to use example IP addresses for tests and instead of commenting these move them into a separate test and use `.disabled = true`? Don't know if that makes sense...

dennis-adlink: What do you mean with 'example IP addresses'? The IP addresses in the lines I commented out are handled as a host name (because of the [ and ]) and looked up via DNS (`ddsrt_gethostbyname`), which obviously fails (and causes the timeout). So I think it doesn't really matter what IP address is used here. 

On disabling the tests: if we plan to fix this in future, I agree that it would make sense to move these tests lines to a separate test and disable that test. But I thinks that not the case here (as there is no reasonable fix because it depends on OS settings), and I don't really like having more tests that report 'disabled' in the test output (which makes it harder to scan through the test output to find failures).

k0ekk0ek: It would be better if the tests use IP addresses ([RFC5737](https://tools.ietf.org/html/rfc5737)) reserved for documentation purposes. Others might be routed. Not that it makes much difference here, it's just good practice. Since you didn't introduce them, I don't require you to change them for this PR. Just thought might as well change it while you're at it(?)

If you move the tests that take a bit longer into a separate test and disable them in Travis, that'd be better. The `disabled` status is not ideal for reading, but I think that's only because the overview is not ideal to start with(?) The fact that these tests will be run locally if not explicitly disabled is added value imho, if not, we should remove them entirely. Either way, commenting them out doesn't seem the best way to go.

dennis-adlink: Ok, I'll replace the 1.2.3.4 addresses (and those which are used as host name with [ ]) with 192.0.2.0. 

About disabling the tests: it's just the other way around: on Travis these tests run fine, because apparently the timeout for DNS resolving is a small value (or DNS resolving is disabled on Travis builds?). On my Ubuntu machine with default settings it gives a timeout because of the name resolving timeout settings.

k0ekk0ek: Since we're trying to mimic `gethostbyname` we could probably get away with checking if there are any non-hostname characters in there and simply refuse to resolve if there are, but I'll have to look into that a bit. I'll put it on my list of things todo. Since I'll have to look into OpenSSL a bit to determine if the Coverity fixes are correct, I'll do it later today or tomorrow.

k0ekk0ek: What about the following change to `ddsi_ipaddr.c`? Would this be acceptable @eboasson? Should take care of one of the time outs?

```diff --git a/src/core/ddsi/src/ddsi_ipaddr.c b/src/core/ddsi/src/ddsi_ipaddr.c
index 4aa408bd..2d731e0b 100644
--- a/src/core/ddsi/src/ddsi_ipaddr.c
+++ b/src/core/ddsi/src/ddsi_ipaddr.c
@@ -138,26 +138,25 @@ enum ddsi_locator_from_string_result ddsi_ipaddr_from_string (const struct ddsi_
     }
   }
 
-#if DDSRT_HAVE_IPV6
-  if (af == AF_INET6)
-  {
-    if (copy[0] == '[') {
-      // strip brackets: last character before the port must be a ']',
-      // in the absence of a port, the last character in the string.
-      ipstr = copy + 1;
-      if (portstr == NULL) {
-        if (copy[len - 1] != ']')
-          return AFSR_INVALID;
-        copy[len - 1] = 0;
-      } else {
-        assert (portstr > copy);
-        if (portstr[-1] != ']')
-          return AFSR_INVALID;
-        portstr[-1] = 0;
-      }
+  if (copy[0] == '[') {
+    // strip brackets: last character before the port must be a ']',
+    // in the absence of a port, the last character in the string.
+    ipstr = copy + 1;
+    if (portstr == NULL) {
+      if (copy[len - 1] != ']')
+        return AFSR_INVALID;
+      copy[len - 1] = 0;
+    } else {
+      assert (portstr > copy);
+      if (portstr[-1] != ']')
+        return AFSR_INVALID;
+      portstr[-1] = 0;
     }
-  }
+#if DDSRT_HAVE_IPV6
+    if (af != AF_INET6)
 #endif
+      return AFSR_INVALID;
+  }
 
   if (ddsrt_sockaddrfromstr(af, ipstr, (struct sockaddr *) &tmpaddr) != 0) {
 #if DDSRT_HAVE_DNS

```

eboasson: I suppose that would work, but I do wonder if it wouldn't make more sense to allow the bracketed notation also for IPv4. It's not like those brackets would cause any problems.

I am also generally in favour of rejecting malformed inputs early. I imagine that it would be easy enough to check whether the the string could possibly be a hostname as you suggested, and one would think that'd be the nicest solution for this particular problem. I suppose you know the rules for hostnames off the top of your head, so I'm not going to bother looking them up right now 😁

src/security/core/src/dds_security_plugins.c
```diff
@@ -211,7 +211,8 @@ dds_return_t dds_security_load_security_library (const dds_security_plugin_confi
   if (lib_ret != DDS_RETCODE_OK)
   {
     char buffer[256];
-    ddsrt_dlerror (buffer, sizeof (buffer));
+    if (ddsrt_dlerror (buffer, sizeof (buffer)) != DDS_RETCODE_OK)
+      (void) snprintf (buffer, sizeof (buffer), "unknown error");
```
k0ekk0ek, line 215: I appreciate the attempt, but I looking at the code for `ddsrt_dlopen` and `ddsrt_dlerror`, I think the real fix is ditching the `dlerror` wrapper entirely and just return the proper return code from `ddsrt_dlopen`. A quick-fix for now could be printing "unknown error" to the buffer in `ddsrt_dlerror` or something, returning `DDS_RETCODE_NOT_FOUND` from `ddsrt_dlerror` makes no sense imho

src/security/builtin_plugins/tests/encode_datawriter_submessage/src/encode_datawriter_submessage_utests.c
```diff
@@ -557,97 +556,73 @@ static bool crypto_decrypt_data(uint32_t session_id, unsigned char *iv, DDS_Secu
...
+  if (key_size == 128 && !EVP_DecryptInit_ex(ctx, EVP_aes_128_gcm(), NULL, NULL, NULL))
   {
-    if (!EVP_DecryptInit_ex(ctx, NULL, NULL, session_key.data, iv))
+    ERR_print_errors_fp(stderr);
+    goto err;
+  }
+  if (key_size == 256 && !EVP_DecryptInit_ex(ctx, EVP_aes_256_gcm(), NULL, NULL, NULL))
+  {
+    ERR_print_errors_fp(stderr);
+    goto err;
```
k0ekk0ek, line 573: Maybe just nitpicking, but what if `key_size` is neither `128` or `256`? We could opt to make it an assert(?) Also, the `ERR_print_errors_fp` is always combined with `goto err;`, you could choose to just call `ERR_print_errors_fp` from the `err:` label... I'm fine with it like this though.

src/core/ddsi/tests/locators.c
k0ekk0ek: So, I thought about this a bit more... Basically it's a bug in some old(er) version of Glibc, so technically it's not our bug to fix and since we haven't had any other complaints, we could choose to leave the tests and implementation as they are. However, since we don't pass AI_IDN, checking the hostname is relatively easy, just check if it's **a)** an IPv4 address or **b)** an IPv6 address or **c)** a classic hostname containing just `[a-zA-Z0-9-]`. That should take care of the problem without breaking things for users. If it does break things, we could choose to treat strings containing UTF-8 characters as IDN names and let `getaddrinfo` worry about it (or use `libidn`, which `getaddrinfo` itself probably also uses) and validate strings containing only ASCII characters the same way. An URI parser I once did can be found [here](https://github.com/ADLINK-IST/opensplice/blob/master/src/abstraction/os/common/code/os_uri.c). It's owned by ADLINK so we can copy the necessary code and re-license or you can just do something like this:
```
for (size_t i=0; name[i]; i++)
  if (name[i] != <valid-hostname-character-or-character-that-occurs-in-ip-address>)
    return <error>
```
Probably close enough for our use-case(?)

src/security/core/src/dds_security_plugins.c
k0ekk0ek: @dennis-adlink, you asked me what I meant with this comment offline, so I'll elaborate. To be honest `ddsrt_dlopen` and `ddsrt_dlclose` are both so-so, I haven't even looked at the other functions yet. For instance, `ddsrt_dlerror` uses an `snprintf`-like signature and actually uses `snprintf` to print to the user provided buffer. There's at least a couple things wrong with that:
1. What if I provide `0` for `buflen`, I'm pretty sure that's undefined behaviour for some platforms
2. If I provide a `buflen` of `1`, that just gets me an empty string, so we must check for the number of bytes required
3. We have a perfectly fine `ddsrt_strlcpy`, no need for "formatting" the string
4. `DDS_RETCODE_NOT_FOUND` just does not make any sense as an error code. The function just shouldn't fail and should print `"unkown error"` itself and return the number of bytes required

So your fix, fixes it for this instance, but next time someone else will make the same mistake. Even if `ddsrt_dlclose` is "successful", the contents (as far as the function signature goes), may contain a completely unusable message.

imho, it would be better to return `DDS_RETCODE_ERROR` from `ddsrt_dlopen` and simply remove `ddsrt_dlclose`, much easier. I should mention that `ddsrt_dlopen` is badly implemented too. Take for instance:

```
  if ((translate) && (strrchr(name, '/') == NULL )) {
    /* Add lib and suffix to the name and try to open. */
#if __APPLE__
    static const char suffix[] = ".dylib";
#else
    static const char suffix[] = ".so";
#endif
    char* libName;
    ddsrt_asprintf( &libName, "lib%s%s", name, suffix);
    *handle = dlopen(libName, RTLD_GLOBAL | RTLD_NOW);
    ddsrt_free(libName);
  }
```
1. Translation is probably done automatically by the linker to begin with
2. If it isn't, wouldn't it make more sense to check if the name ends with `.dylib` or `.so` instead of checking if the path contains a `/`. What if the library is located in the current directory?
3. The return code of `ddsrt_asprintf` isn't checked, so `libName` might just be uninitialized.
3.1. Possibly resulting in passing an uninitialized variable to `dlopen`; and
3.2. Passing an uninitialized pointer to `ddsrt_free`

I don't require you to fix these issues (you didn't introduce them), so I'll take your fix. But if you can spare the time and feel like really fixing it, that'd be even better.

src/security/builtin_plugins/tests/encode_datawriter_submessage/src/encode_datawriter_submessage_utests.c
dennis-adlink: I've added the assert on invalid `key_size` in 7709a3e. I considered moving the `ERR_print_errors_fp` call, but I think that  causes the reported line number in the error message to be the `err:` block and not the line where the error occurred.

src/security/core/src/dds_security_plugins.c
dennis-adlink: @k0ekk0ek thanks for your feedback! 

1-3 regarding `ddsrt_dlerror`: I've refactored the code based on your suggestions (db09700). 

4: I still think an error return code makes sense here: if someone calls the `ddsrt_dlerror` function in case no error related to dynamic library loading occurred, this should return something else than the message 'unknown error'. In `dds_security_plugins.c:215` we know that something went wrong and there should be an error, so that's why I use the 'unknown error' at that point. I've replace the error code with 'PRECONDITION_NOT_MET', which better describes what happened imho. Removing `ddsrt_dlerror` (you wrote `ddsrt_dlclose`, but I assume you meant `ddsrt_dlerror`) completely would make the log messages for library loading less clear: currently it e.g. reports 'library file not found' or 'library file corrupted'. Without the call to `dlerror` we'd not have this information. 

Regarding `ddsrt_dlopen` 1-2: the check is 
"translate == true and not a [absolute/relative] path in the `name` parameter" 
(it checks for _not_ a slash in `name`). E.g. the value of `name` is just something like `dds_security_ac`, which is not correctly handled by dlopen, which requires the full file-name as far as I understand from the manual. So I think the code is correct and useful; i.e. you don't have to add the file extension and `lib` prefix and can have a portable value in configuration when the library is just referenced by its name (which will be used in most cases I think, having the (library) path set correctly). 
3: you're right, I've refactored the code a bit

src/core/ddsi/tests/locators.c
dennis-adlink: Thanks for looking into this. I've implemented (commit 6c835b3) your suggestion of checking for valid characters before passing the value to `getaddrinfo`, I also think that this is sufficient in our situation and we don't need the complete uri parser code for this.

dennis-adlink: @k0ekk0ek, in commit d6206c9 I've replaced `isalnum` in the host name check as you suggested, because of its locale dependent behavior

src/security/core/src/dds_security_plugins.c
dennis-adlink: Some additional refactoring of the dynamic library loading error handling is in commit d6bd655: on Windows it's now using a thread local variable to store the last error, so that the behavior of `ddsrt_dlerror` on Windows is now is equal to the posix and other variants. And this function now returns the number of characters written in the error message (0 in case of no error), or `DDS_RETCODE_NOT_ENOUGH_SPACE` in case the message does not fit in.

k0ekk0ek (approved):

k0ekk0ek:
LGTM @dennis-adlink, if you would be so kind to cleanup the commits a little, I'll be happy to merge. Thanks!

k0ekk0ek:
Much appreciated @dennis-adlink!

PR #574, opened 2020-08-11
Use older gcc version for coverity

eboasson:
The nightly Coverity analysis runs stopped because Coverity's tooling hasn't been updated yet to support gcc 10. This PR changes the build configuration for that run to use the default ubuntu 18.04 compiler (gcc 7) which I guess is supported by coverity.

I don't quite know how to test the cron job without committing this ...


k0ekk0ek:
The issue is probably that it doesn't recognize `gcc-10` as a call to `gcc` and running `cov-configure` would be cleaner, but we can worry about that later...

PR #579, opened 2020-08-20
Add support for Codecov

k0ekk0ek:
By popular demand, this PR adds support for the codecov.io service as previously suggested by @ThijsSassen. I took the the [CMake-codecov](https://github.com/RWTH-HPC/CMake-codecov) modules and created an additional module that can generate the submission file for [codecov.io](https://codecov.io/). `.travis.yml` contains a `submit_to_codecov` block that executes `cmake --build . --target codecov` to generate the file and submits it to the service. For now, I'm publishing results with the same frequency as the Coverity job, which imho is a good place to start. Whether we want to classify PRs based on code coverage results is something we can worry about later(?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,36 +9,62 @@ language: c
 # Coverity Scan quota are not checked as the Coverity enabled build must only
 # run from cron.
 install_coverity: &install_coverity
-  if [ "${COVERITY_SCAN}" = "true" ]; then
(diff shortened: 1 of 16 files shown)


k0ekk0ek:
@eboasson, I'm merging this one too. I tested it locally, the dump file looked right and no actual code is touched. In this case too, I'm 99.99% sure you're fine with the changes.

PR #582, opened 2020-08-24
For the security_timed_dispatcher use the xevent queue to schedule timer events.

MarcelJordense:
Each instance of the security_timed_dispatcher used a separate thread to schedule the timer events for handling expiry of certificates. This thread is now replaced by using the xeventq.

### DIFF
--- a/src/security/builtin_plugins/authentication/src/authentication.c
+++ b/src/security/builtin_plugins/authentication/src/authentication.c
@@ -2068,7 +2068,7 @@ DDS_Security_boolean set_listener(dds_security_authentication *instance, const d
   if (listener)
     dds_security_timed_dispatcher_enable(auth->dispatcher);
   else
-    dds_security_timed_dispatcher_disable(auth->dispatcher);
+    (void) dds_security_timed_dispatcher_disable(auth->dispatcher);
   return true;
 }
 
--- a/src/security/builtin_plugins/authentication/src/authentication.c
+++ b/src/security/builtin_plugins/authentication/src/authentication.c
@@ -120,6 +120,7 @@ typedef struct LocalIdentityInfo
   AuthenticationAlgoKind_t dsignAlgoKind;
   AuthenticationAlgoKind_t kagreeAlgoKind;
   char *permissionsDocument;
+  dds_security_time_event_handle_t timer;
 } LocalIdentityInfo;
 
 typedef struct RemoteIdentityInfo
@@ -133,6 +134,7 @@ typedef struct RemoteIdentityInfo
   DDS_Security_OctetSeq pdata;
   char *permissionsDocument;
   struct ddsrt_hh *linkHash; /* contains the IdentityRelation objects */
+  dds_security_time_event_handle_t timer;
 } RemoteIdentityInfo;
 
 /* This structure contains the relation between a local and a remote identity
@@ -167,8 +169,8 @@ typedef struct dds_security_authentication_impl
   ddsrt_mutex_t lock;
   struct ddsrt_hh *objectHash;
   struct ddsrt_hh *remoteGuidHash;
-  struct dds_security_timed_cb_data *timed_callbacks;
-  struct dds_security_timed_dispatcher_t *dispatcher;
+  struct dds_security_timed_dispatcher *dispatcher;
+  const dds_security_authentication_listener *listener;
   X509Seq trustedCAList;
   bool include_optional;
 } dds_security_authentication_impl;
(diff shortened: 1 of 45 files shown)


src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -599,28 +601,51 @@ static DDS_Security_BinaryProperty_t *hash_value_to_binary_property(const char *
...
   validity_cb_info *info = arg;
   if (kind == DDS_SECURITY_TIMED_CB_KIND_TIMEOUT)
   {
-    assert(listener);
-    dds_security_authentication_listener *auth_listener = (dds_security_authentication_listener *)listener;
-    if (auth_listener->on_revoke_identity)
-      auth_listener->on_revoke_identity((dds_security_authentication *)info->auth, info->hdl);
+    assert(info->auth->listener);
+    SecurityObject * obj = get_identity_info(info->auth, info->hdl);
+    if (obj)
```
eboasson, line 629: It's not clear to me why `obj` can't be freed while the callback is executing

eboasson (commented):
Thanks @MarcelJordense! I have a few comments, but not all commented here because I fiddled with it while reviewing:

* I'm not quite enamored by the decision to pass the events queue in the "add" function for timed cbs, I'd rather see it passed in when the dispatcher is created. That affects some other place, of course ...
* The behaviour of `delete_xevent_callback` is to guarantee that the callback function is not and will not be executing once it returns (it syncs with the event queue processing). So the "busy" flag and the condition variable are not needed here. The "active" flag is synonymous with the presence of an event.

See https://github.com/MarcelJordense/cyclonedds/pull/1

The one other thing is that there appears to be a race condition in expiry handling in the authentication plugin.

MarcelJordense:
@eboasson Thanks for reviewing.
I agree that it would be better to pass the event queue and create the associated event when the dispatcher is created and not when the dispatcher is activated. Currently this event is deleted when the dispatcher is disabled and not when it is freed. The reason for this is that at termination the event queue is deleted before the security plugins are released. Thus either the event should be deleted in the disabling of the dispatcher or by the triggering of the callback when the event queue is deleted.

The busy is used to ensure that no callback is being performed when the user of the dispatcher is cleaning up. May be the use of the busy flag is not the proper solution for this and the user of the dispatcher should implement this synchronization.

eboasson:
> @eboasson Thanks for reviewing.
> I agree that it would be better to pass the event queue and create the associated event when the dispatcher is created and not when the dispatcher is activated. Currently this event is deleted when the dispatcher is disabled and not when it is freed. The reason for this is that at termination the event queue is deleted before the security plugins are released. Thus either the event should be deleted in the disabling of the dispatcher or by the triggering of the callback when the event queue is deleted.

Maybe I'm overlooking something, but my analysis was that `dds_security_timed_dispatcher_free` calls `dds_security_timed_dispatcher_disable` before anything else, and that's where `delete_xevent_callback` is called (and `evt` is reset to a null pointer).

> The busy is used to ensure that no callback is being performed when the user of the dispatcher is cleaning up. May be the use of the busy flag is not the proper solution for this and the user of the dispatcher should implement this synchronization.

I'm not sure I understand this: `busy` is set/cleared only inside the `timed_event_cb`, and the only place where it waits is immediately following `delete_xevent_callback`. So with the latter function guaranteeing to not return until after any pending or concurrent call to `timed_event_cb` has completed, `dds_security_timed_dispatcher_disable` will only ever see `busy == false`, that means it will never wait, thus the claim that the condition variable and the flag can be removed altogether.

MarcelJordense:
> > @eboasson Thanks for reviewing.
> > I agree that it would be better to pass the event queue and create the associated event when the dispatcher is created and not when the dispatcher is activated. Currently this event is deleted when the dispatcher is disabled and not when it is freed. The reason for this is that at termination the event queue is deleted before the security plugins are released. Thus either the event should be deleted in the disabling of the dispatcher or by the triggering of the callback when the event queue is deleted.
> 
> Maybe I'm overlooking something, but my analysis was that `dds_security_timed_dispatcher_free` calls `dds_security_timed_dispatcher_disable` before anything else, and that's where `delete_xevent_callback` is called (and `evt` is reset to a null pointer).
>
The function `dds_security_timed_dispatcher_disable` is also called when the listener in the security plugin is removed (set to NULL). The listener is set to NULL before the security plugins are released and before the event queue is deleted. Thus in that case it is still safe to delete the event because the event queue still exists. The call to `dds_security_timed_dispatcher_disable` in the `dds_security_timed_dispatcher_free` could be removed.

> > The busy is used to ensure that no callback is being performed when the user of the dispatcher is cleaning up. May be the use of the busy flag is not the proper solution for this and the user of the dispatcher should implement this synchronization.
> 
> I'm not sure I understand this: `busy` is set/cleared only inside the `timed_event_cb`, and the only place where it waits is immediately following `delete_xevent_callback`. So with the latter function guaranteeing to not return until after any pending or concurrent call to `timed_event_cb` has completed, `dds_security_timed_dispatcher_disable` will only ever see `busy == false`, that means it will never wait, thus the claim that the condition variable and the flag can be removed altogether.

The function `dds_security_timed_dispatcher_remove` may also call the callback function. In that case there is no protection that a callback is being performed. I admit that this situation probably does not occur because the security listener will be removed after all timers are removed (all identity handles and related certificates associated with local and remote participants are returned).

eboasson:
> > > @eboasson Thanks for reviewing.
> > > I agree that it would be better to pass the event queue and create the associated event when the dispatcher is created and not when the dispatcher is activated. Currently this event is deleted when the dispatcher is disabled and not when it is freed. The reason for this is that at termination the event queue is deleted before the security plugins are released. Thus either the event should be deleted in the disabling of the dispatcher or by the triggering of the callback when the event queue is deleted.
> > 
> > Maybe I'm overlooking something, but my analysis was that `dds_security_timed_dispatcher_free` calls `dds_security_timed_dispatcher_disable` before anything else, and that's where `delete_xevent_callback` is called (and `evt` is reset to a null pointer).
> 
> The function `dds_security_timed_dispatcher_disable` is also called when the listener in the security plugin is removed (set to NULL). The listener is set to NULL before the security plugins are released and before the event queue is deleted. Thus in that case it is still safe to delete the event because the event queue still exists. The call to `dds_security_timed_dispatcher_disable` in the `dds_security_timed_dispatcher_free` could be removed.

That's useful information. It is not always entirely trivial to track exactly where these indirect calls happen.

The listeners can only fire if entities (local or proxy) exist, and by the time the event queue is destroyed, all entities have been removed and no new ones can be added. So one could choose to tie the existence of the "callback xevent" to the existence of events in the dispatcher, but that would tie the implementations together more tightly then I would think is wise.

What would make sense to me is remove these listeners in `q_omg_security_stop`: that functions exists to stop background processing (which events are, I'd say) and at the point it is called, no new entities will be introduced in the system anymore (neither local nor proxy). And that always happens before the event queue is destroyed.

> > > The busy is used to ensure that no callback is being performed when the user of the dispatcher is cleaning up. May be the use of the busy flag is not the proper solution for this and the user of the dispatcher should implement this synchronization.
> > 
> > I'm not sure I understand this: `busy` is set/cleared only inside the `timed_event_cb`, and the only place where it waits is immediately following `delete_xevent_callback`. So with the latter function guaranteeing to not return until after any pending or concurrent call to `timed_event_cb` has completed, `dds_security_timed_dispatcher_disable` will only ever see `busy == false`, that means it will never wait, thus the claim that the condition variable and the flag can be removed altogether.
> 
> The function `dds_security_timed_dispatcher_remove` may also call the callback function. In that case there is no protection that a callback is being performed. I admit that this situation probably does not occur because the security listener will be removed after all timers are removed (all identity handles and related certificates associated with local and remote participants are returned).

I'm not entirely certain of that: you could have the event firing, then a context switch to remove, then actually processing the event. So you could end up with a use-after-free there. (At least I think so, I didn't check everything.) A very simple way to solve that problem is to mark the event as to-be-removed, then reschedule for immediate execution. That way you end up doing all the work in the xevent thread.

Otherwise it requires some thought on getting the synchronization right.

eboasson:
@MarcelJordense I have now also spent some time on this PR, which also really should not have been sitting for so long. I hope we can leave it at a "better late than never" 🙂.

After re-reading the discussion and looking at the code, I have established that my last comment

> you could have the event firing, then a context switch to remove, then actually processing the event [yadda yadda]

really got it wrong. That race condition does not exist. Naturally, I only realised how obviously wrong it is *after* I'd rewritten things as the comment suggested and so I reverted those changes. Serves me right!

However, I did find a race condition in `dds_security_timed_dispatcher_disable`, in that the callback locks the mutex but `delete_xevent_callback` may wait until the callback runs/completes and so must not be called with the dispatcher lock held. I've solved it by calling it after releasing the lock, which of course has some consequences. I did consider the alternative of adding a second mutex, but I don't see why that would be required.

Perhaps you could have a look at it? https://github.com/eboasson/cyclonedds/tree/marcel-secdispatch-pr582

MarcelJordense:
@eboasson Thanks for reviewing and resolving the possible deadlock. Your changes look fine to me.

PR #583, opened 2020-08-27
Write thoughput data of ddsperf to csv file specified on command line via -f option

BartP6703:
Signed-off-by: Bart <bart.poot@adlinktech.com>

### DIFF
--- a/src/tools/ddsperf/ddsperf.c
+++ b/src/tools/ddsperf/ddsperf.c
@@ -50,6 +50,9 @@
 
 #define PINGPONG_RAWSIZE 20000
 
+/* File descriptor to write ddsperf data to */
+static FILE *file = NULL;
+
 enum topicsel {
   KS,   /* KeyedSeq type: seq#, key, sequence-of-octet */
   K32,  /* Keyed32  type: seq#, key, array-of-24-octet (sizeof = 32) */
@@ -1381,12 +1384,18 @@ struct dds_stats {
   const struct dds_stat_keyvalue *discarded_bytes;
 };
 
-static bool print_stats (dds_time_t tref, dds_time_t tnow, dds_time_t tprev, struct record_cputime_state *cputime_state, struct record_netload_state *netload_state, struct dds_stats *stats)
+static bool print_stats (dds_time_t tref, dds_time_t tnow, dds_time_t tprev, struct record_cputime_state *cputime_state, struct record_netload_state *netload_state, struct dds_stats *stats, char *filename)
 {
   char prefix[128];
   const double ts = (double) (tnow - tref) / 1e9;
   bool output = false;
-  snprintf (prefix, sizeof (prefix), "[%"PRIdPID"] %.3f ", ddsrt_getpid (), ts);
+  ddsrt_pid_t pid = ddsrt_getpid ();
+  char hostname[64];
+  if (ddsrt_gethostname (hostname, sizeof (hostname)) != DDS_RETCODE_OK)
+  {
+    snprintf (hostname, sizeof (hostname), "%s", "<unknown>");
+  }
+  snprintf (prefix, sizeof (prefix), "[%"PRIdPID"] %.3f ", pid, ts);
 
   if (pub_rate > 0)
   {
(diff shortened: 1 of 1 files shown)


src/tools/ddsperf/ddsperf.c
```diff
@@ -1381,12 +1384,18 @@ struct dds_stats {
   const struct dds_stat_keyvalue *discarded_bytes;
 };
 
-static bool print_stats (dds_time_t tref, dds_time_t tnow, dds_time_t tprev, struct record_cputime_state *cputime_state, struct record_netload_state *netload_state, struct dds_stats *stats)
+static bool print_stats (dds_time_t tref, dds_time_t tnow, dds_time_t tprev, struct record_cputime_state *cputime_state, struct record_netload_state *netload_state, struct dds_stats *stats, char *filename)
```
eboasson, line 1387: I don't think it makes much sense to pass in the filename here when the `file` is a global variable. It would be more in line with the rest of the code to open the file in `main`, then use it here. (Or, alternatively, to make `file` local to `main` and pass it in here.)

src/tools/ddsperf/ddsperf.c
```diff
@@ -1427,11 +1436,25 @@ static bool print_stats (dds_time_t tref, dds_time_t tnow, dds_time_t tprev, str
...
       const double dt = (double) (tnow - tprev);
       printf ("%s size %"PRIu32" total %"PRIu64" lost %"PRIu64" delta %"PRIu64" lost %"PRIu64" rate %.2f kS/s %.2f Mb/s (%.2f kS/s %.2f Mb/s)\n",
               prefix, last_size, tot_nrecv, tot_nlost, nrecv, nlost,
               (double) nrecv * 1e6 / dt, (double) nrecv_bytes * 8 * 1e3 / dt,
               (double) nrecv10s * 1e6 / (10 * dt), (double) nrecv10s_bytes * 8 * 1e3 / (10 * dt));
+      if (file != NULL)
+      {
+        int64_t tsec = (int64_t)(tnow / 1000000000);
+        int64_t nsec = (int64_t)(tnow % 1000000000);
+        fprintf (file, "%"PRId64".%"PRId64",%s,%"PRIdPID",%.3f,%"PRIu32",%.2f,%.2f\n",
```
eboasson, line 1455: That should be `"%"PRId64".%06"PRId64"` and I suspect the `%s` should be `\"%s\"` but I don't know CSV rules very well.

I furthermore think some of the other fields that get written to `stdout` here are useful; and certainly the CPU, memory and network load data would be good to also collect in the CSV file. They are gathered and printed because they are valuable for understanding what is going on, it follows that a log file specifically for analysis purposes should include them.

src/tools/ddsperf/ddsperf.c
```diff
@@ -1939,11 +1963,12 @@ int main (int argc, char *argv[])
   char netload_if[256];
   double netload_bw = -1;
   double rss_init = 0.0, rss_final = 0.0;
+  char filename[256] = { 0 };
```
eboasson, line 1966: There's no need to copy the file name to the stack, one can just store `optarg`.

eboasson (commented):
Thanks @BartP6703, not logging the data in an easily processed format and instead parsing the console output definitely was a mistake and so this is a welcome addition. However, it would be much better do a bit of additional work and log all the interesting data, rather than just this tiny subset.

src/tools/ddsperf/ddsperf.c
BartP6703: Use the filename in optarg to open the file and use the (global) file descriptor.

src/tools/ddsperf/ddsperf.c
BartP6703: - Changed the format specifier as proposed in the comment.
- Added code to log the latency measurements to csv file.

src/tools/ddsperf/ddsperf.c
BartP6703: Used optarg to open the file.

PR #586, opened 2020-09-02
added OMG DDS C++ API mention to README

joespeed:
added OMG DDS C++ API mention to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 # Eclipse Cyclone DDS
 
 Eclipse Cyclone DDS is a very performant and robust open-source DDS implementation.  Cyclone DDS is developed completely in the open as an Eclipse IoT project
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Thanks @joespeed. It's certainly a good idea to mention Cyclone has a C++ api available too. Before I can merge though, your commit needs to be signed (`git commit -s`), which is the same as appending `<newline>Signed-off-by: Foo Bar <foo.bar@domain><newline>` to you commit message, to keep the Eclipse Contributor Agreement check happy. If you click the `Details` button next to the check, it'll guide you through the process of signing the agreement. If you've already done so, It'll show you why the check failed. Be sure to use the same email address in both places.

eboasson:
Issues not fixed for so long, closing this. If it still needed, feel free to reopen.

PR #589, opened 2020-09-03
Test writecdr/takecdr with custom topic implementation

eboasson:
writecdr, takecdr should at least be tested a little bit. Using a custom topic implementation makes it possible to do this without a dependency on the way the default topic implementation works and hopefully doubles as an example for those who want to play with that (not yet stable) interface.


k0ekk0ek (approved):
:slightly_smiling_face: :+1:

PR #591, opened 2020-09-05
Bits and pieces for IDL compiler

k0ekk0ek:
This PR fixes some memory leaks (delete tokens on Bison parser deletion), fixes a bug or two, adds `idl_strdup` and `idl_strndup` and adds the last bits for handling `#pragma keylist` directives.


eboasson (commented):
I fear the macOS and Windows builds need some TLC. I haven't studied it in detail, but at first glance they seem to be straightforward. Apart from that, I'm ok with merging this.

k0ekk0ek:
Thanks @eboasson, didn't think you'd get around to reviewing that quick! The `winflexbison3` chocolatey package seems to be fixed, so even Windows builds pass :slightly_smiling_face:

PR #593, opened 2020-09-10
Fixed issue with @key annotations

reicheratwork:
Fixed issue with @key annotations not correctly showing up in the parsed tree.


reicheratwork:
Will do

PR #594, opened 2020-09-10
Allow fancy quotes in attribute values in config

eboasson:
Various text editors and email clients like to substitute fancy unicode quotes (U+2018/U+2019 for single ones and U+201C/U+201D for double ones) for the old-fashioned ASCII ones and this has bitten various people when mailing configuration snippets.

This PR makes the config parser accept the fancy ones in UTF-8 representation. This is strictly an extension of a the allowed inputs, and so does not break any existing files.


k0ekk0ek (approved):
Restarted the macOS builds that failed. Changes look fine to me.

PR #596, opened 2020-09-16
Fix some minor bugs

k0ekk0ek:
Running the C++ test cases showed a number of minor bugs that this PR fixes. A memory leak, a bug with escaped identifiers and bug in finalization of declarators with array sizes.


k0ekk0ek:
I'm merging this. There's one macOS build failing on Maven again, but retrying until it works seems like a pointless exercise. I've contact Travis about this btw, lets hope they can quickly resolve the issue.

PR #597, opened 2020-09-16
Iceoryx

evshary:
The pull request is about the first integration of Cyclone DDS+iceoryx.
The integration enables the shared memory based on iceoryx to Cyclone DDS.
For further detail, please refer to [shared_memory.rst](https://github.com/Adlink-ROS/cyclonedds-private/blob/iceoryx/docs/manual/shared_memory.rst).

@eboasson As discussed before, maybe it's better to create a branch instead of merge back to master now. Thanks for all the help you give during the implementation!


eboasson:
Thanks @evshary! I agree that it would be better to put it on a separate branch for now. There are a number of things still to be sorted out before it should go into master. But I think there is one thing that needs to be done before I can merge into an `iceoryx` branch:

* It currently depends on https://github.com/Adlink-ROS/iceoryx_clib, and depending on how quickly Iceoryx gets a C API integrated we may need to do something about that. Right now, I think it would suffice if you give `iceoryx_clib` a compatible license and add copyright headers to the files.

At least that's how I read the rules for the licenses and dependencies (and if I am wrong in that, at least I am erring on the side of caution).

Integration with the CI system would also be good; it would *much* nicer if the CI system includes the Iceoryx integration and verifies that it works. But that can also be done once it is on a branch. So it really is just some paperwork on `iceoryx_clib` that I think should be done first.

On the functional side, a number of points immediately come to mind:
* Support for multiple publishers. I think Iceoryx will gain that soon, but I think a workaround could possibly be done by using per-writer partitions, a wildcard on the reader side, and including the partition names in the Iceoryx topic names. (There are definitely some complications that I am glossing over, so whether the effort would be worth it depends a bit on when Iceoryx gets multi-publisher support.)
* Currently it serializes a sample in `dds_write` (for example), and then if there happen to be local subscriptions connected via Iceoryx, it copies the data into Iceoryx. That's just a `memcpy` so it is pretty fast, but it would be much nicer if it were possible to serialize the data straight to shared memory in this case, with a copy made only if it also has to go out over the network.
* A true zero-copy model is then only a small step: it that case the application would have to reserve the space in shared memory and fill it. If it has to go out on the network as well, see the preceding bullet. Supporting that is the goal 😄 
* On the receiving side there is also this copy "problem". If we could leave the data in Iceoryx's shared memory for a bit longer (that is, until the application has taken the data), then there wouldn't be a need to copy it. It goes without saying that such niceties would require some rework of the internal sample representation, because then it would need to track who owns which bit of memory. That some work would be needed in that area is no surprise, and it would benefit all other places where samples are manipulated as well.
* I also think that the current all-or-nothing approach might be less than ideal, that it would make sense to tie it to the (e.g.) network partition mechanism. That doesn't really change much, it only affects which readers and writers are considered.

These are just some initial thoughts. The performance graph you included says loud and clear that is a step in a very valuable direction, and I think that even with today's limitations it will be of value to some users. Improving it is then simply a logical thing to do.

As a final thought: what we perhaps should be doing is treating the DDSI network protocol and Iceoryx as two different interpretations of a "pub/sub provider", where such a provider is allowed to support only a subset of the topics and QoS settings. The DDSI one would then be a bit special in that it would handle the discovery & always be available as a fallback. But that's just a random musing 😄 .

evshary:
@eboasson Thank you for the comments.
I've added license and copyright header on iceoryx_clib.

For the functional side, I totally agree with your aspects.
All bullets you mentioned here are good topics to be discussed.
The current design is based on the idea of little changes on Cyclone DDS and able to show the power of iceoryx, so there must be some improvements here.
I love the idea of "pub/sub provider", and it worths to have a deeper study :-)

eboasson:
Thanks for updating `iceoryx_clib` @evshary. I'm merging it now in the new `iceoryx` branch. Let's try to get this ready for master as quickly as possible.

PR #599, opened 2020-09-18
Switch to Xcode 11.5

k0ekk0ek:
This should fix the issue with macOS builds. I decided to do some debugging and noticed the issue didn't occur with the newer `xcode12u` `osx_image`. Going over the list, I then noticed `xcode12u` reverted from `14.0.2` back to `14.0.1` which seemed strange to me. Anyway, since Xcode 12.0 is still in beta (and Conan didn't accept it) I decided to revert to `xcode11.5`. It seems to behave much better now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,9 +114,9 @@ ubuntu1804_clang10: &ubuntu1804_clang10
   install:
     - pip install conan --upgrade --user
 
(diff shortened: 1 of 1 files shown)


eboasson (approved):
Strange that this should make such a difference, but a good thing you discovered it!

PR #603, opened 2020-09-23
Added typedef for "struct ser_data" in order to disambiguate an issue in the idlcxx repo that chokes the Windows compiler.

e-hndrks:
Signed-off-by: Erik Hendriks <erik.hendriks@adlinktech.com>

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_serdata.h
@@ -43,6 +43,8 @@ struct ddsi_serdata {
   ddsrt_mtime_t twrite; /* write time, not source timestamp, set post-throttling */
 };
 
+typedef struct ddsi_serdata ddsi_serdata_t;
+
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
@eboasson, can we go ahead and merge once Travis passes?

eboasson:
The only issue with it that I have is that it adds nothing of value to cyclonedds, its only purpose is to work around a limitation of a specific C++ compiler in some generated C++ code. I would think it would be cleaner to handle this in the relevant (C++) repos. That said, the naming is fine (even if the commit message gets it wrong) and if this typedef were ever to be added to the C code, it would have this name.

I'd like some justification why adding it here is the right approach before giving my approval.

k0ekk0ek:
Makes sense. @e-hndrks, any reason this typedef isn't only there in `cyclonedds-cxx`?

e-hndrks:
Thanks, @eboasson. I have decided to move the typedef to the specific C++ file that needs to be disambiguated. I will close this pull request.

PR #604, opened 2020-09-25
Set config programmatically

eboasson:
Title is taken from branch name, because it is being set programmatically whichever way one does it.

What this PR brings is a moderately clean (at least it avoids including Windows.h and only defines symbols with names starting with "dds") header file (`ddsi_config.h`) for including the (internal) configuration structure (`struct ddsi_config`), along with a function to initialize this to the default configuration (`ddsi_config_init_default`) and a function to construct a domain based on a configuration structure rather than based on a configuration given as an XML string.

This is *not* a stable interface, it needs some further cleaning up for that. For those who want to avoid XML, it can be an option today. The likelihood of major changes appears to be pretty small, as the majority of it has been around for a decade.

The configuration contains some lists, arrays of strings and strings. Currently, only a shallow copy is made and the code expects whatever is referenced by the shallow copy to  remain valid while the domain exists. (This may change.) The default initialization uses compile-time constants for these and so is safe.


k0ekk0ek:
Had a quick look (will give it a more thorough look next week). Any reason not to integrate it into `ddsconf` and then adding an additional command line option like `-g` for generate? Just wondering, maybe there's a good reason for that. Like the option to run without the XML parser :+1: Might be an idea to disable the entire XML parser at compile time like with some of the QoSes?

eboasson:
> Had a quick look (will give it a more thorough look next week). Any reason not to integrate it into `ddsconf` and then adding an additional command line option like `-g` for generate? Just wondering, maybe there's a good reason for that.

Two reasons really, and both are of dubious quality: the first is dependencies, ddsconf depends on libddsc (https://github.com/eclipse-cyclonedds/cyclonedds/blob/6f8850f8d074d98810a1f0093ac8040f5b68a339/src/tools/ddsconf/CMakeLists.txt#L13), but I must admit I haven’t looked into the details of that dependency. I imagine it might well be no more than `ddsrt`. The second is the various ways in which the configuration tables are loaded. ddsconf currently has a “unity of purpose”, and that would be broken by the default generation, because it has a very different view of the tables. It moreover needs to have the normal configuration processing at hand. Separating it into a different program seems less likely to create confusion, but that may be looking at things the wrong way.

> Like the option to run without the XML parser 👍 Might be an idea to disable the entire XML parser at compile time like with some of the QoSes?

Yep, that’s the logical next step. It brings some complications in building and testing, though. That’s why I decided to postpone that.

k0ekk0ek:
`ddsconf` depends on `ddsc` because `ddsrt` shouldn't be considered a separate target. It's only separated so that platform-specific tests can be run without the rest of the `ddsc` library. `ddsrt` is an interface target with sources, meaning every target that uses it directly results in `ddsrt` being compiled again. It could be an object target, but that only works properly since CMake 3.13 (from the top of my head).

k0ekk0ek:
Hmmm, you're using the tool to create a source file that is used in `ddsi`, and thus `ddsc`. Makes sense that you can't rely on `ddsc` then :smile:.

k0ekk0ek:
Really like the functionality and agree with the overal changes. Should be really helpful to users with embedded targets too once the compile-time option is in place. Haven't tested it, but since most of the changes are renaming symbols, I don't think that's necessary.

I'd like to propose we stick this into `ddsconf` though. My intent always was for the tool to contain extra features and though I thought something along the lines of a configuration checker was the first addition to arrive, this is a good fit too. Since it's currently generating "schema's", maybe we should add a `-g` option (for generate) and make it accept either `config` or `schema`? I'd probably move the specifics (macro overrides) for each generation mode into it's own `.c` file and leave the command line handling in `ddsconf.c`. Then, instead of depending on `ddsc`, change it to depend on `ddsrt` instead. Maybe have the default for `-g` be `schema` so we don't break scripts that people may have created (unlikely, but still).

Personally, I think that'd be a little cleaner, but if you're really opposed to the idea, I can live with the current implementation.

eboasson:
> Personally, I think that'd be a little cleaner, but if you're really opposed to the idea, I can live with the current implementation.

I am not at all opposed to it; I’ll probably get it done tomorrow.

k0ekk0ek (approved):

PR #606, opened 2020-09-28
Extend content filters to metadata

eboasson:
This adds `dds_set_topic_filter_extended` and `dds_get_topic_filter_extended` which allow selecting from different function prototypes for the filter functions: with or without sample data, and with or without sample info. Note that the filtering interface is unstable and that remains so.


k0ekk0ek (approved):
I have read through it a couple of times and the changes make sense. Given that content filtering was available already I didn't verify overall correctness of the content filter mechanism itself. Let me know if it'd make sense to do so @eboasson, otherwise, consider this PR approved.

eboasson:
Thanks @k0ekk0ek. Fixed the typo and rebased it, I'll wait for the CI (though I don't see how the configuration changes could affect this) before merging.

PR #610, opened 2020-10-02
Check return of dds_register_instance calls

eboasson:
CID 309661 in ddsperf: check added.

CID 309662 in pubsub: that program needs a complete rewrite, so casting to "void" is a reasonable choice.


k0ekk0ek (approved):
LGTM. Feel free to go ahead and merge.

PR #611, opened 2020-10-02
Added the ability to use idl_identifier() function to obtain the identifier of a const.

e-hndrks:
Because idl_name_t was encapsulated in tree.c, the identifier name cannot be accessed in other source files.
Using the existing idl_identifier() function to access the const identifier seemed the most obvious way to go.

Signed-off-by: Erik Hendriks <erik.hendriks@adlinktech.com>

### DIFF
--- a/src/idl/src/tree.c
+++ b/src/idl/src/tree.c
@@ -79,6 +79,8 @@ const char *idl_identifier(const void *node)
     return ((const idl_declarator_t *)node)->name->identifier;
   if (idl_is_forward(node))
     return ((const idl_forward_t *)node)->name->identifier;
+  if (idl_is_const(node))
+    return ((const idl_const_t *)node)->name->identifier;
   return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


src/idl/src/tree.c
```diff
@@ -1383,6 +1385,22 @@ bool idl_is_base_type(const void *node)
...
+#ifndef NDEBUG
+  }
+  else
+  {
+    const idl_const_type_t * const_type = ((const idl_const_t *) node)->type_spec;
+    assert(idl_is_base_type(const_type) || idl_is_string(const_type));
+#endif
+  }
+  return true;
+}
```
k0ekk0ek, line 1402: This check is incorrect. It can also be an enumerator.

k0ekk0ek:
Also, probably best to correct the commit message `idl_entifier()` > `idl_identifier()` and while you're at it, please use present time.

src/idl/src/tree.c
e-hndrks: @k0ekk0ek: Thx for catching that one, must have slipped my mind. Added enumerator and updated PR accordingly.

e-hndrks:
Amended commit with fix and corrected title.

PR #616, opened 2020-10-08
Remove thread pool implementation

dennis-adlink:
When I tried to use the thread pool implementation in ddsrt for a test case, I found out that is has locking issues, e.g. `ddsrt_thread_pool_new_thread` is called with `pool->m_mutex` locked and tries to lock this mutex. As the thread pool is rarely used in practice (it can be enabled in configuration and used with TCP transport), the best option (as agreed with @eboasson) is to remove this feature and remove the broken thread pool implementation.


k0ekk0ek (approved):

k0ekk0ek:
@dennis-adlink, I restarted the Travis builds just to be sure and verified that the generated `cyclonedds.xsd` works. If the builds succeed, I'll merge. Thanks!

k0ekk0ek:
Builds pass. Merging.

PR #618, opened 2020-10-13
Idl const fixes

k0ekk0ek:
This PR adds support for `const` `string` declarations and fixes `const` `char` declarations (in the frontend). I've added two small tests to verify it works, but constant expression evaluation is still far from done. I guess that mainly applies to `integer` expressions, but tests are lacking for `floating point` expressions and more exotic `char` and `string` expressions too, so ymmv. I'll probably fix them in one go after I get around to some small redesign in the tree. Reopening of modules comes to mind and it's probably a good idea to expose the scopes to the backends as well. Stay tuned!

### DIFF
--- a/src/idl/src/expression.c
+++ b/src/idl/src/expression.c
@@ -709,8 +709,8 @@ idl_evaluate(
   idl_retcode_t ret;
   idl_constval_t *constval;
 
-  assert(type & (IDL_BASE_TYPE|IDL_ENUMERATOR));
-  type &= (IDL_ENUMERATOR|IDL_BASE_TYPE | (IDL_BASE_TYPE - 1));
+  assert(type & (IDL_BASE_TYPE|IDL_ENUMERATOR|IDL_STRING));
+  type &= (IDL_STRING|IDL_ENUMERATOR|IDL_BASE_TYPE | (IDL_BASE_TYPE - 1));
 
   /* enumerators are referenced */
   if (type == IDL_ENUMERATOR) {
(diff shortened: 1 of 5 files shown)


k0ekk0ek:
@reicheratwork, can you spare some time to review these changes? I'd like an extra pair of eyes to go over them before I hit merge :slightly_smiling_face:

reicheratwork:
> @reicheratwork, can you spare some time to review these changes? I'd like an extra pair of eyes to go over them before I hit merge 🙂

will do

src/idl/src/expression.c
```diff
@@ -749,7 +749,15 @@ idl_evaluate(
         "Cannot evaluate '%s' as boolean expression", "<foobar>");
       goto err_eval;
     }
-    constval->value.bln = ((idl_literal_t *) expr)->value.bln;
+    constval->value.bln = ((idl_literal_t *)expr)->value.bln;
+  } else if (type == IDL_STRING) {
+    if (expr->mask != (IDL_STRING_LITERAL|IDL_EXPR)) {
+      idl_error(proc, idl_location(expr),
+        "Cannot evaluate '%s' as string expression", "<foobar>");
```
reicheratwork, line 756: I think the expression that was parsed should go here, in stead of <foobar>?

reicheratwork (changes requested):
Looks good, and the only thing that remains to be done is the remaining foobar test string in expression.c

k0ekk0ek:
Actually, it's on purpose :blush: The intent is that `foobar` is replaced with the (scoped) name, but so far I've not implemented that part properly. The idea is to do it in one go with some other changes that I'm planning and leave it as-is for now. The obvious workaround here is that users name all their `const` declarations `foobar`, perhaps putting them in separate modules postfixed with a number... :wink:

reicheratwork:
> Actually, it's on purpose 😊 The intent is that `foobar` is replaced with the (scoped) name, but so far I've not implemented that part properly. The idea is to do it in one go with some other changes that I'm planning and leave it as-is for now. The obvious workaround here is that users name all their `const` declarations `foobar`, perhaps putting them in separate modules postfixed with a number... 😉

So we're both correct :)

I could not find anything wrong except for the scoped name being a placeholder, which is intentional, so according to me, this can be merged

PR #620, opened 2020-10-15
Fix detection of reopened modules

k0ekk0ek:
Small fix for a rookie mistake :sweat_smile: Last node in the list was looked up by checking if the node had a sibling and the duplicate check was located in that loop, needless to say, the last node was never compared to the node added to the list which resulted in a node pointing towards itself, which caused an infinite loop in the C++ backend. @reicheratwork, care to take a quick look?

### DIFF
--- a/src/idl/src/parser.y
+++ b/src/idl/src/parser.y
@@ -981,12 +981,13 @@ static void *push(void *list, void *node)
     return node;
   }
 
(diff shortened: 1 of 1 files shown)


reicheratwork (approved):

k0ekk0ek:
Thanks!

PR #622, opened 2020-10-26
Fix iceoryx's branch CI

evshary:
Let's give it a try and see whether the CI is fixed or not.


k0ekk0ek:
@evshary, builds turn green, so that's a good thing :slightly_smiling_face: Nice! However, for third-party dependencies (like `iceoryx`), Conan is used. While the shared memory transport is in development I'd be OK with doing it this way if a custom version is required. If it's not, I think we should write a Conan recipe for it (cuts build times and ensures users/developers can easily install the requirements too). Of course, I'm willing to lend a hand in that case. Let me know.

evshary:
@k0ekk0ek It's indeed a good way to put third-party dependencies into conan, but I guess we need to upload iceoryx binary to conan server first? I'm not familiar with conan, so I'll be really appreciated if you can help on this or give me some suggestions.

By the way, we need iceoryx_clib currently until iceoryx provides C API. Do you think iceoryx_clib should also be put into conanfile?

k0ekk0ek:
@evshary, we can put both dependencies there. Basically, [Conan](https://conan.io) packages are recipes on how to build the project from source, but it allows you to upload the built binaries to a Conan server for others to use. For this purpose we use [Bintray](https://bintray.com/), a could hosted JFrog Artifactory service. I'll see if I can whip something up. Since Iceoryx uses CMake too, it shouldn't be too much work to get a basic package going.

k0ekk0ek:
@evshary, had a look `iceoryx_clib`. I was unaware that it was a minimal binding to iceoryx. My feeling is that it's unlikely to evolve into a well-supported project in it's own right(?) So it might not make sense to turn it into a package (the package would be more code than the actual sources). Do we have an eta on the C binding for iceoryx? This [PR](https://github.com/eclipse/iceoryx/pull/256) seems to prepare the project a little, but I haven't tested it and the PR mentions that it is *It is not yet functional!*. I don't know how @eboasson feels about this, but it might be easier if we adopt the `iceoryx_clib` sources (2 files) into this repository until iceoryx itself offers a stable C-api?

Would be easiest if we could relicense the two source files. Alternatively, we could kickoff an IP-review.

evshary:
@k0ekk0ek As mentioned [here](https://github.com/eclipse/iceoryx/pull/277#issuecomment-689495501), we might need to wait for [Support for n:m communciation](https://github.com/eclipse/iceoryx/issues/25) in iceoryx. I don't know the eta about this.

And yes, `iceoryx_clib` is the temporary project before C binding for iceoryx is done. It might not worth to put too much effort on packaging (I have no idea about the effort). Relicense `iceoryx_clib` and put it into CycloneDDS repo is fine to me, but I'm not sure how to mix C and C++ in one project (and whether it's suitable to do this). Or maybe we can do like this: Since the iceoryx we use is the `used_by_cyclone`  branch from [the fork of iceoryx](https://github.com/Adlink-ROS/iceoryx/tree/used_by_cyclone), we can put `iceoryx_clib` into it and wait for the official support. I'm not sure which way is better. What do you think?

k0ekk0ek:
@evshary, so the `used_by_cyclone` fork of `iceoryx` has modifications to make things work? In that case (as per my first comment), I'd be fine with keeping it like this for the time being.

evshary:
> @evshary, so the `used_by_cyclone` fork of `iceoryx` has modifications to make things work? In that case (as per my first comment), I'd be fine with keeping it like this for the time being.

Yes, just a simple workaround for [this issue](https://github.com/eclipse/iceoryx/issues/62) and waiting for [Support for n:m communciation](https://github.com/eclipse/iceoryx/issues/25) is done.

k0ekk0ek:
In that case, I'd propose we relicense the `iceoryx_clib` sources and put them in the Cyclone DDS repository and keep building Iceoryx like it is. Then, when the issue is resolved and support for n:m communication is in, switch to a Conan package and remove the `iceoryx_clib` sources when Iceoryx gains a C api. Checking out Iceoryx and building it makes development of the feature easier, because Iceoryx origins/branches can be switched etc and allows for some time to whip up a Conan package. Maybe we can even collaborate with the Iceoryx guys if they're interested. @eboasson, any thoughts/remarks on this?

joespeed:
[Support for n:m communciation](https://github.com/eclipse/iceoryx/issues/25) is closed. Does that help you progress it? @evshary @k0ekk0ek

k0ekk0ek:
It seems the work is continued [here](https://github.com/eclipse/iceoryx/issues/252)(?) Maybe it'll allow us to use `iceoryx` master, but @evshary is the expert.

evshary:
I haven't taken a deeper look at current master, but I think we still need to wait for [Building blocks in iceoryx](https://github.com/eclipse/iceoryx/issues/252) to finish.

budrus:
n:m communication is working on iceoryx master. But we still change the APIs (e.g. waitset is currently changing). For you guys it would make sense to wait for our 0.9 release. With this we provide the new C and C++ APIs. Then we want to have some weeks to gain trust and feedback before making the longterm support release. We hope to have 0.9 ready as Christmas present.

evshary:
I guess the PR is not necessary now. I'll close it.

eboasson:
Thanks @evshary, I know I am terrible at keeping the list of open PRs clean and your help is appreciated!

PR #625, opened 2020-10-27
Import ddsconf when cross compiling

dennis-adlink:
When cross compiling Cyclone, the `ddsconf` executable from the host build should be used when running the target build. The changes in this commit make this possible, by importing the `ddsconf` executable when running a cross compilation build.


k0ekk0ek:
Looks good to me @dennis-adlink, thanks! I'll merge if all the builds pass.

PR #626, opened 2020-10-28
FreeRTOS fix build warnings

dennis-adlink:
This PR resolves the build warnings in the FreeRTOS build, by using `PRI*32` format specifiers and replacing a few instances of `unsigned` by `uint32_t`. The declaration of `main` is added to avoid the 'no previous prototype' warning in case the main function is wrapped and redefined as `_real_main`.

### DIFF
--- a/src/core/ddsi/src/q_receive.c
+++ b/src/core/ddsi/src/q_receive.c
@@ -641,9 +641,8 @@ struct nn_xmsg * nn_gap_info_create_gap(struct writer *wr, struct proxy_reader *
   }
   else
   {
-    unsigned i;
-    ETRACE (wr, " FXGAP%"PRId64"..%"PRId64"/%d:", gi->gapstart, gi->gapend, gi->gapnumbits);
-    for (i = 0; i < gi->gapnumbits; i++)
+    ETRACE (wr, " FXGAP%"PRId64"..%"PRId64"/%"PRIu32":", gi->gapstart, gi->gapend, gi->gapnumbits);
+    for (uint32_t i = 0; i < gi->gapnumbits; i++)
       ETRACE (wr, "%c", nn_bitset_isset (gi->gapnumbits, gi->gapbits, i) ? '1' : '0');
   }
 
@@ -2033,7 +2032,7 @@ static struct ddsi_serdata *remote_make_sample (struct ddsi_tkmap_instance **tk,
       if (gv->logconfig.c.mask & DDS_LC_CONTENT)
         res = ddsi_serdata_print (sample, tmp, sizeof (tmp));
       if (pwr) guid = pwr->e.guid; else memset (&guid, 0, sizeof (guid));
-      GVTRACE ("data(application, vendor %u.%u): "PGUIDFMT" #%"PRId64": ST%x %s/%s:%s%s",
+      GVTRACE ("data(application, vendor %u.%u): "PGUIDFMT" #%"PRId64": ST%"PRIx32" %s/%s:%s%s",
                sampleinfo->rst->vendor.id[0], sampleinfo->rst->vendor.id[1],
                PGUID (guid), sampleinfo->seq, statusinfo, topic->name, topic->type_name,
                tmp, res < sizeof (tmp) - 1 ? "" : "(trunc)");
@@ -3417,7 +3416,7 @@ static void rebuild_local_participant_set (struct thread_state1 * const ts1, str
     qsort (lps->ps, lps->nps, sizeof (*lps->ps), local_participant_cmp);
     lps->nps = (unsigned) dedup_sorted_array (lps->ps, lps->nps, sizeof (*lps->ps), local_participant_cmp);
   }
-  GVTRACE ("  nparticipants %u\n", lps->nps);
+  GVTRACE ("  nparticipants %"PRIu32"\n", lps->nps);
 }
 
 uint32_t listen_thread (struct ddsi_tran_listener *listener)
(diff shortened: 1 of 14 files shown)


src/ddsrt/include/dds/ddsrt/misc.h
```diff
@@ -89,6 +89,12 @@ extern "C" {
   DDSRT_WARNING_GNUC_ON(deprecated-declarations) \
   DDSRT_WARNING_MSVC_ON(4996)
 
+/**
+ * @brief Declare main function to fix the no previous prototype warning
+ *  when main function is wrapped (e.g. in the FreeRTOS port)
+ */
+int main(int argc, char **argv);
+
```
k0ekk0ek, line 97: This seems very odd, I'm 99,99% sure this isn't what we want to do.

src/core/ddsi/src/q_receive.c
```diff
@@ -3468,12 +3468,12 @@ void trigger_recv_threads (const struct ddsi_domaingv *gv)
         ddsrt_iovec_t iov;
         iov.iov_base = &dummy;
         iov.iov_len = 1;
-        GVTRACE ("trigger_recv_threads: %d single %s\n", i, ddsi_locator_to_string (buf, sizeof (buf), dst));
+        GVTRACE ("trigger_recv_threads: %"PRId32" single %s\n", i, ddsi_locator_to_string (buf, sizeof (buf), dst));
```
k0ekk0ek, line 3471: Better use `PRIu32` here. `i` is an `uint32_t`.

src/core/ddsi/src/q_receive.c
```diff
@@ -3468,12 +3468,12 @@ void trigger_recv_threads (const struct ddsi_domaingv *gv)
...
         ddsi_conn_write (gv->xmit_conn, dst, 1, &iov, 0);
         break;
       }
       case RTM_MANY: {
-        GVTRACE ("trigger_recv_threads: %d many %p\n", i, (void *) gv->recv_threads[i].arg.u.many.ws);
+        GVTRACE ("trigger_recv_threads: %"PRId32" many %p\n", i, (void *) gv->recv_threads[i].arg.u.many.ws);
```
k0ekk0ek, line 3476: Same.

src/core/ddsi/src/q_receive.c
```diff
@@ -642,7 +642,7 @@ struct nn_xmsg * nn_gap_info_create_gap(struct writer *wr, struct proxy_reader *
   else
   {
     unsigned i;
-    ETRACE (wr, " FXGAP%"PRId64"..%"PRId64"/%d:", gi->gapstart, gi->gapend, gi->gapnumbits);
+    ETRACE (wr, " FXGAP%"PRId64"..%"PRId64"/%"PRId32":", gi->gapstart, gi->gapend, gi->gapnumbits);
```
k0ekk0ek, line 645: `gapnumbits` is an `uint32_t`, so better use `PRIu32`.

src/core/ddsc/include/dds/dds.h
```diff
@@ -39,6 +39,7 @@ typedef int32_t dds_entity_t;
 #include "dds/ddsrt/time.h"
 #include "dds/ddsrt/retcode.h"
 #include "dds/ddsrt/log.h"
+#include "dds/ddsrt/misc.h"
```
k0ekk0ek, line 42: Probably for declaration of `main`. I'm almost certain that that isn't what we want to do

k0ekk0ek (changes requested):
Thank @dennis-adlink! This seems to be necessary every once in a while, so good thing you looked into it. I have a few small remarks that need to be addressed, most of them trivial. As for the declaration of `main`, that's probably not what we want to do. If the warning isn't fixable in any other way, then it's probably best to leave it for now.

src/core/ddsi/src/q_receive.c
dennis-adlink: Good point! I didn't look further than the `%d` in the old format string

src/ddsrt/include/dds/ddsrt/misc.h
dennis-adlink: I fully agree that this seems very odd, but I think it is correct (as `main` is replaced by something else with a preprocessor define and becomes an ordinary function). I'll add `#ifdef main` around the declaration, maybe that helps? I don't see any alternatives, except for leaving this warning for now (but that means we cannot enable `-Werror` on FreeRTOS builds)

k0ekk0ek: I think we shouldn't (re)define `main` ever, especially not in headers that are public. The compiler is likely complaining about `realmain`, which is a hack in itself, not being declared before it's defined(?) In that case providing a header that's always prepended (there's a `gcc` option for that) could be a fix. Another option could be to rename the symbol. In any case, it's definitely a problem in the FreeRTOS build, therefore the place to fix it is in that specific build.

dennis-adlink: This code is not redeclaring `main`, but redeclaring the function with the rewritten name (`real_main` in this case), as the declaration is also rewritten by the define that is provided with the gcc `-D main=real_main` option. You're right that it should not be in a public header, it should be in a specific place for the FreeRTOS build. 

What I could do is add a header file like this:

```C
#ifndef FREERTOS_MAINWRAP_H
#define FREERTOS_MAINWRAP_H

/**
 * @brief Declare main function to avoid the 'no previous prototype' warning
 *  when main function is wrapped (e.g. in the FreeRTOS port). In that case the
 *  main function is renamed to real_main by using a definition for main passed
 *  to the compiler on the command line (-D main=real_main).
 */
#ifdef main
int main(int argc, char **argv);
#endif

#endif /* FREERTOS_MAINWRAP_H */
```
This header should be in a freertos specific location in the Cyclone source tree (any suggestions?), and that location needs to be in the include_path when running a FreeRTOS build. The header can then be prepended using the gcc `-include` option you mentioned. 

For now I've removed the declaration, so that we can finish this PR and fix the issue later in a separate PR.

k0ekk0ek: @dennis-adlink, the fix is absolutely correct, but it's the wrong repository :slightly_smiling_face: The BSP for this particular build contains a small wrapper that initializes the embedded device. There should be a CMake toolchain file there as well where you can add the `-include` option to specify that a header file of your choosing must be included first, that's probably the best place to add it.

k0ekk0ek:
LGTM @dennis-adlink. If you combine and force-push, I'll merge. Thanks!

k0ekk0ek (approved):

k0ekk0ek:
Thanks, much appreciated @dennis-adlink.

PR #627, opened 2020-10-28
Added packaging configuration to generate a windows installer

ThijsSassen:
This commit adds configuration to create a proper windows installer when using the package target.


k0ekk0ek:
@ThijsSassen, kudos! This is very welcome! Unfortunately, there'll be some due diligence involved before we can merge.

1. The included icon is not the official icon used by the project (https://iot.eclipse.org/projects/). I'm sure we'd much rather use the included one, but if we do, it has to become the official icon and I'm not sure what the process for that is (maybe you can look into that).
2. The `DialogOrder.wxs` includes a license header. It's a good thing that you left it, but I'll probably have to open a CQ. To do so, I'll need to know where this file originated from. Do we have a link or something to the official repository where this is located or was it generated? If it's the latter, and anything other than WiX 3.7 was used (approved in CQ 7712), I'll have to open a CQ in IPZilla.

As for the `Packaging.cmake` changes, I'll approve them, but I do feel that there's a better way for setting those variables. For instance, the generated `build/CPackConfig.cmake` seems to include `CPackProperties.cmake` if it exists. Maybe we should look into that at some point to clean things up a bit. These comments apply to the [PR](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/39) you opened for `cyclonedds-cxx` too.

ThijsSassen:
@k0ekk0ek point 1 i will look into and point 2 i addressed with 2 new commits

ThijsSassen:
This PR is now replaced with #776

PR #628, opened 2020-11-02
Handle failures converting sample into serdata

eboasson:


dennis-adlink:
LGTM, I don't see any issues, except that there is no checking for using `ddsi_serdata_from_sample` for built-in message like PMD, PGM, SPDP etc. But adding checks for that (and especially correctly handling if conversion failed) is probably not worth the effort, because it's very unlike to fail and if it fails then the cause of that should be fixed.

eboasson:
> LGTM, I don't see any issues, except that there is no checking for using `ddsi_serdata_from_sample` for built-in message like PMD, PGM, SPDP etc. But adding checks for that (and especially correctly handling if conversion failed) is probably not worth the effort, because it's very unlike to fail and if it fails then the cause of that should be fixed.

Indeed, the scope of the changes needed for handling failures in those internal cases would be quite significant, and they can't really fail anyway. Eventually something will need to be done about it, and you're quite correct in pointing them out, but for now I think fixing the ones that deal with the API is sufficient.

PR #629, opened 2020-11-02
Refactor serdata includes for C++ friendliness

eboasson:
I think this will make C++ compilers' complaints about C99 constructs that are not also legal C++ in definitions needed for implementing sertopic/serdata go away. I may have missed something, it still needs to be verified with some C++ code. (I imagine that'll be straighforward using the `idlcxx` branch.)

### DIFF
--- a/src/core/ddsi/src/ddsi_eth.c
+++ b/src/core/ddsi/src/ddsi_eth.c
@@ -10,6 +10,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
 #include "ddsi_eth.h"
+#include "dds/ddsi/q_protocol.h" // for NN_LOCATOR_KIND_...
 
 int ddsi_eth_enumerate_interfaces (ddsi_tran_factory_t fact, ddsrt_ifaddrs_t **ifs)
 {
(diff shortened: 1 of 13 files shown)


k0ekk0ek:
Don't see anything wrong with it, but then again I haven't verified the changes either. Once someone does though, feel free to merge.

src/core/ddsi/src/ddsi_eth.c
```diff
@@ -11,16 +11,14 @@
...
 {
     int afs[] = { AF_INET, DDSRT_AF_TERM };
 
     (void)fact;
-    (void)transport_selector;
 
 #if DDSRT_HAVE_IPV6
-    if (transport_selector == DDSI_TRANS_TCP6 ||
-        transport_selector == DDSI_TRANS_UDP6)
+    if (fact->m_kind == DDSI_TRANS_TCP6 || fact->m_kind == DDSI_TRANS_UDP6)
```
dennis-adlink, line 21: `ddsi_tran_factory.m_kind` is set to one of the `NN_LOCATOR_KIND_...` values, so I think we can test for `m_kind` to be `NN_LOCATOR_KIND_UDPv6` or `NN_LOCATOR_KIND_TCPv6` here?

eboasson:
Thanks for checking @dennis-adlink! Those are two embarrassing mistakes ...

PR #630, opened 2020-11-04
Add dds_forwardcdr for sending unmodified serdata

eboasson:
`dds_takecdr` and `dds_readcdr` return a reference to a `ddsi_serdata`, the internal, abstract representation of a sample. One would think one could safely write these again using `dds_writecdr`, but that function (through some accident of history) mimics the regular `dds_write` function and writes the current time and the operation ("write") into it, and so this is not actually such a good idea. It works fine for the case that it originally was supposed to handle ... Secondly, there are no `dds_disposecdr` or `dds_unregistercdr` functions to dispose/unregister data based on such an abstract sample. That is an omission that could easily be fixed by adding these functions.

This PR adds a new function `dds_forwardcdr` (as in "forwarding") that doesn't modify its input. This means (0) that the caller is in full control of timestamp and operation encoded in it, and (1) that it this function can safely be invoked on a `ddsi_serdata` returned by the read/take operations, and thus allows it to be used for forwarding data from a reader to a writer. Hence the name.

(The workaround for this modifying of the input would be to make a copy of the serdata, which is possible with implementation-agnostic code by constructing a new one from the serialised form, but that adds a fair bit of overhead in a place where it, presumably, is undesirable.)

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -1958,7 +1958,8 @@ dds_write_flush(dds_entity_t writer);
  * @brief Write a serialized value of a data instance
  *
  * This call causes the writer to write the serialized value that is provided
- * in the serdata argument.
+ * in the serdata argument.  Timestamp and statusinfo fields are set to the
+ * current time and 0 (indicating a regular write), respectively.
  *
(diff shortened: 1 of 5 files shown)


dennis-adlink:
LGTM!

PR #631, opened 2020-11-05
FIx openssl dependency

mikaelarguedas:
the openssl package provides only the openssl executable and not the development libraries/headers this results in the `ros-foxy-cyclonedds` package to not contain the security plugins.
From http://build.ros2.org/view/Fbin_uF64/job/Fbin_uF64__cyclonedds__ubuntu_focal_amd64__binary/13:
"Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)"

This PR declares a depenency on`libssl-dev` that is providing those.

Should fix https://github.com/ros2/ros2/issues/1051
Would need to be backported to the 0.7.x branch and released into foxy

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -14,7 +14,7 @@
 
     <buildtool_depend>cmake</buildtool_depend>
 
-    <depend>openssl</depend>
+    <depend>libssl-dev</depend>
     <test_depend>libcunit-dev</test_depend>
 
     <doc_depend>doxygen</doc_depend>
--- a/package.xml
+++ b/package.xml
@@ -14,7 +14,10 @@
 
     <buildtool_depend>cmake</buildtool_depend>
 
-    <depend>libssl-dev</depend>
+    <build_depend>libssl-dev</build_depend>
+
+    <exec_depend>openssl</exec_depend>
+
     <test_depend>libcunit-dev</test_depend>
 
     <doc_depend>doxygen</doc_depend>


k0ekk0ek:
Hi @mikaelarguedas. Thanks for the PR! The change looks reasonable to me. Besides the small technical issue that the commit needs a "Signed-off-by:" footer and signing the Eclipse Contributor Agreement (ECA), I have one small question: `libssl-dev` is (to my knowledge) the name of the package in Debian based Linux distributions. Will this work in the ROS2 macOS and Windows builds too? Will things keep working on non-Debian based distributions?

mikaelarguedas:
>  Besides the small technical issue that the commit needs a "Signed-off-by:" footer and signing the Eclipse Contributor Agreement (ECA), 

My bad, did the edit through the github interface instead of a local clone, should be fixed now.

>  I have one small question: libssl-dev is (to my knowledge) the name of the package in Debian based Linux distributions. Will this work in the ROS2 macOS and Windows builds too? Will things keep working on non-Debian based distributions?

This is the name of the rosdep key that is used for packaging. It is meant to be the same for all platforms: https://github.com/ros/rosdistro/blob/77b7ff5a22409ddaf40062df55e781894da44e9d/rosdep/base.yaml#L4028
As right now there is no packaging for MacOS and Windows it shouldnt have any impact on those platforms

k0ekk0ek:
I didn't know the ROS [package.xml](http://wiki.ros.org/catkin/package.xml) worked like that. The change seems entirely reasonable. One last question though, wouldn't it make more sense to keep `openssl` as a `depend` entry (for runtime) and add the `libssl-dev` as a `build_depend` entry. I don't know if that's the intended use. If it isn't than I'm fine merging this, if it is, then can I ask you to update? Thanks!

mikaelarguedas:
`openssl` is the openssl command line tool. I dont think it's ever needed at runtime by CycloneDDS, is it ?
There doesn't seem to be any `libssl` rosdep key defined at the moment. It would more likely be that one that you would want to use as a runtime dependency in the package.xml (and make it correspond to the [libssl1.1](https://packages.ubuntu.com/focal/libssl1.1) package on ubuntu/debian).

As adding such a key to the rosdep database will take multiple days/weeks to be reviewed and merged, I preferred using the one that was already available and make it both build and exec dependency.
I can modify the PR to use `openssl` instead to reduce the runtime footprint, keeping in mind that it's not the dependency cyclone actually needs at runtime.

k0ekk0ek:
I don't think the command line tool is an actual dependency, it's very likely the dependency was introduced because we depend on the openssl package. I think it would be good to separate runtime and build dependencies, so `openssl` as a `depend` and `libssl-dev` as a `build_depend`. We can simply swap `openssl` for `libssl` once/if that's every changed in ROS.

mikaelarguedas:
> I think it would be good to separate runtime and build dependencies, so openssl as a depend and libssl-dev as a build_depend. We can simply swap openssl for libssl once/if that's every changed in ROS.

done in https://github.com/eclipse-cyclonedds/cyclonedds/pull/631/commits/56f54d55d675d869ae71251d5911353021154349

k0ekk0ek:
Thanks @mikaelarguedas, much appreciated!

mikaelarguedas:
You're welcome :)
What are the next steps to get this fixed in the Foxy debian packages ? 
Should I open the same PR on the releases/0.7.x branch ?

eboasson:
@mikaelarguedas I believe that would be the first step, so if you could do that, that'd be great. If I'm not mistaken, that needs to be followed by `bloom`ing it to actually get it into Foxy. We'll ask someone who knows how to do that (@ThijsSassen, perhaps?) to do the next step.

Issue #317, opened 2019-11-11
build static library fails


dirk-thomas:
When build with the CMake define `-DBUILD_SHARED_LIBS=OFF` it outputs the following to `stderr`:

```
You have called ADD_LIBRARY for library ddsc without any source files. This typically indicates a problem with your CMakeLists.txt file
```

I assume as a consequence it doesn't install the following file which fails downstream packages: `share/CycloneDDS/CycloneDDSTargets.cmake`

eboasson:
Thanks @dirk-thomas — another option that doesn't get tested properly. For shared libraries the warning is but a nuisance warning (and removed in the more recent versions of CMake) but considering your observation perhaps it is different for static libraries. Will look into it …

eboasson:
An update (which means I am stupefied how to fix it): one needs to generate the ``...Targets.cmake`` file explicitly and that step is skipped when a static library is generated, and most likely because CMake refuses to generate it:

```
CMake Error: install(EXPORT "CycloneDDS" ...) includes target "ddsc"
  which requires target "ddsrt" that is not in the export set.
```

This appears to be a long-standing issue with interface libraries and static linking in CMake, e.g.: https://gitlab.kitware.com/cmake/cmake/issues/15415#note_333595. I've tried the hack proposed there, but that gives me

```
CMake Error at src/core/CMakeLists.txt:25 (add_library):
  Target "ddsc" links to target "Threads::Threads" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

and so and so forth. My preference by now would be to ditch CMake after "ritually removing it from my laptop" and use a proper makefile instead, but then I would get into real trouble with the rest of the world as well as on Windows, so that's not a really realistic solution.

Perhaps I could generate the ```CycloneDDSTargets.cmake``` by hand, that seems to be the most realistic proposition right now.

How badly do you need it?

dirk-thomas:
> How badly do you need it?

Not urgent at all.

thijsmie:
Superseded by https://github.com/eclipse-cyclonedds/cyclonedds/issues/1313. It is possible to build statically but it needs proper documentation.

PR #633, opened 2020-11-06
Make security_core and ddsrt interface libraries with imported global…

cwyark:
Connect to https://github.com/eclipse-cyclonedds/cyclonedds/issues/317

Can not static build rmw_cyclonedds_cpp due to lack of CycloneDDSTargets.cmake. This is caused by interface libraries (security_core and ddsrt) are interface targets and are not exported. This pull request is to solve exporting interface targets in CycloneDDS.

The tested command : 
```
colcon build --build-base build-static --install-base install-static --packages-ignore rosidl_generator_py rosidl_typesupport_c rosidl_typesupport_cpp --packages-ignore-regex .*fastrtps.* .*connext.* --cmake-args -DBUILD_SHARED_LIBS=OFF --no-warn-unused-cli -DBUILD_TESTING=OFF --packages-up-to rcl
```

### DIFF
--- a/cmake/Modules/Packaging.cmake
+++ b/cmake/Modules/Packaging.cmake
@@ -37,15 +37,13 @@ install(
         "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Version.cmake"
   DESTINATION "${CMAKE_INSTALL_CMAKEDIR}" COMPONENT dev)
 
-if((NOT DEFINED BUILD_SHARED_LIBS) OR BUILD_SHARED_LIBS)
-  # Generates <Package>Targets.cmake file included by <Package>Config.cmake.
-  # The files are placed in CMakeFiles/Export in the build tree.
-  install(
-    EXPORT "${PROJECT_NAME}"
-    FILE "${PROJECT_NAME}Targets.cmake"
-    NAMESPACE "${PROJECT_NAME}::"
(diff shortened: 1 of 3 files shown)


k0ekk0ek:
Hi @cwyark! I haven't tested it myself, but I think it is unlikely that adding `IMPORTED GLOBAL` is the right fix. It essentially states that the library is not built by the project itself. Since `ddsrt` is a an `INTERFACE` library with sources (because OBJECT libraries only work properly since CMake 3.12) it probably still works, but it feels a little off. A quick build of current master doesn't seem to export `ddsrt`, so I'm a little puzzled. Perhaps you can provide the error message?

Sidenote: `ddsrt` must never be exported, the only reason it's a separate target is to verify basic functionality is works on embedded targets.

eboasson:
Ubuntu 20.04 has a sufficiently recent CMake version for OBJECT libraries, it seems (https://packages.ubuntu.com/search?keywords=cmake) so perhaps it would be possible to switch to OBJECT libraries with a fallback to the INTERFACE ones for old CMake versions?

eboasson:
The proposed changes didn't meet @k0ekk0ek's approval (and he is the resident CMake specialist) and nothing much happened with the comments in quite a while, so I'm closing it. That doesn't mean I don't think the underlying problem has gone away, though. Feel free to re-open it if necessary.

PR #634, opened 2020-11-09
Fix including sockets header for LWIP

dennis-adlink:


k0ekk0ek:
Thanks @dennis-adlink!

PR #636, opened 2020-11-13
Compile-time options via CMake-written features.h

eboasson:
Compile-time options (e.g., security, deadline) are configured via CMake options and reflected in a generated file named "features.h" with names like DDS_HAS_...  This commit changes the code to exclusively rely on this generated features.h file.

That means there is no longer a need to specify DDSI_INCLUDE_... on the the compiler command line (via CMake's add_definitions).  Anyone wishing to, e.g., use dds_create_participant_config thus need not worry which set of macros needs to be defined.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,6 +28,8 @@ endif()
 option(ENABLE_SECURITY "Enable OMG DDS Security support" ON)
 option(ENABLE_LIFESPAN "Enable Lifespan QoS support" ON)
 option(ENABLE_DEADLINE_MISSED "Enable Deadline Missed QoS support" ON)
+option(ENABLE_NETWORK_PARTITIONS "Enable network partition support" ON)
+option(ENABLE_SOURCE_SPECIFIC_MULTICAST "Enable support for source-specific multicast" ON)
 if(ENABLE_SECURITY)
   set(DDS_HAS_SECURITY "1")
 endif()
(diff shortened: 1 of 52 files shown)


k0ekk0ek:
I like the changes @eboasson. Much saner way to handle this sort of thing. In the future we could opt to replace `#  cmakedefine VAR @VAR@` with `#  cmakedefine01 VAR` and drop setting the variable to `"1"` explicitly, but I'm totally fine with it like this. Since it's mostly search-replace I went over it rather quickly, so if there's security test output in the builds, it should work as expected(?) Travis didn't run for this PR (which is a little strange), but judging by [the output from your branch](https://travis-ci.org/github/eboasson/cyclonedds/builds/743357146) security tests run fine. Feel free to merge.

k0ekk0ek (approved):

eboasson:
Thanks @k0ekk0ek !

> In the future we could opt to replace `# cmakedefine VAR @VAR@` with `# cmakedefine01 VAR` and drop setting the variable to `"1"` explicitly, but I'm totally fine with it like this.

Ah, but that would define the variable to 0, which means the "if defined" tests are no longer correct. For Cyclone itself, that's no issue, but the `features.h` file was originally introduced for applications, and they may be relying on `#ifdef` ...

> Since it's mostly search-replace I went over it rather quickly, so if there's security test output in the builds, it should work as expected(?) Travis didn't run for this PR (which is a little strange), but judging by [the output from your branch](https://travis-ci.org/github/eboasson/cyclonedds/builds/743357146) security tests run fine. Feel free to merge.

The tests did run, don't know why they would not have been visible to you. There is only test output on failure ... but there are quite a few security-related tests that take the expected time, and many of them can't pass if security isn't enabled.

Issue #638, opened 2020-11-17
roadmap updates


joespeed:
@k0ekk0ek please make these roadmap updates per @eboasson 
https://github.com/eclipse-cyclonedds/cyclonedds/wiki/Roadmap

Cyclone 0.8, Dec 2020 - add: 
* ROS 2 Quality Level 2
   * Automated performance regression testing
   * Formal feature list showing test coverage
   * Each API marked Stable/Evolving/Experimental. APIs in formal feature list must be Stable

Cyclone 0.9, March 2021 - add
* asynchronous mode 
* ROS 2 Quality Level 1

Check with @eboasson about whether the following belong in Long-term milestones, an earlier release or should not be included at this time:
* TrustZone hardware secured DDS Security
* DDS Security logging plugin

k0ekk0ek:
:heavy_check_mark:

Issue #637, opened 2020-11-17
add ROS2 Quality Declaration to cyclonedds repo


joespeed:
@k0ekk0ek please copy this [ROS2 CycloneDDS Quality Declaration](https://github.com/ros2/rmw_cyclonedds/blob/master/CYCLONEDDS_QUALITY_DECLARATION.md) to this repo.

Please update "Cyclone DDS meets all the requirements ..." to say "Cyclone DDS including rmw_cyclonedds meets all the requirements ..."

k0ekk0ek:
:heavy_check_mark:

PR #639, opened 2020-11-17
Add roadmap and quality declaration

k0ekk0ek:
This PR moves the roadmap (suggestions in #638 included) from the wiki to the repository (basically so that external contributors can make suggestions in the form of PRs) and adds a quality declaration as suggested in #637. The latter is simply an import from [rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds) with the first line modified.


k0ekk0ek:
Should be good to go.

PR #654, opened 2020-12-06
Add support for Cirrus CI

k0ekk0ek:
This PR adds build instructions for [Cirrus CI](https://cirrus-ci.org/), a nice independent continuous integration service with some unique features, among them the ability to build with [computing services](https://cirrus-ci.org/guide/supported-computing-services/) from different vendors. Travis is charging nowadays, which is totally fair, but their infrastructure leaves a lot to be desired. i.e. their Windows image is rather messy and FreeBSD support is still beta too. I've contemplated using Azure Pipelines or GitHub actions, but both their agents are written in `C#` and won't run on non-tier-1 platforms. Apart from that, paid computing services or self-hosted runners cannot be shared between the Eclipse Cyclone DDS project and external contributors and I prefer both my own builds as well as PR builds to be as fast as possible.

I fixed the Windows image for Cirrus CI in this [PR](https://github.com/cirruslabs/docker-images-windows/pull/17). It was more work than anticipated, but if the PR is merged everything should be sorted. The Cirrus people were nice enough to [accept a feature request](https://github.com/cirruslabs/cirrus-ci-docs/issues/739), after that's implemented we should be able to use external computing resources for fast builds and contributors can simply use the community cluster, or use computing services of their own. The idea is to have Docker images of our own at some point in the future if necessary. e.g. we can simply create a private Xilinx image to verify FreeRTOS support isn't broken by a PR etc.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,201 @@
+#
+# Copyright(c) 2020 ADLINK Technology Limited and others
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+# v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+ubuntu_2004: &ubuntu_2004
+  container:
+    dockerfile: .cirrus/ubuntu2004/Dockerfile
+  env:
+    PATH: "${HOME}/.local/bin:${PATH}"
+    COV_COMPTYPE: gcc
+    COV_PLATFORM: linux64
+    GENERATOR: "Unix Makefiles"
+    BUILD_TOOL_OPTIONS: "-j 4"
+  pip_cache:
+    folder: ${HOME}/.cache/pip
+  conan_cache:
+    folder: ${HOME}/.conan/data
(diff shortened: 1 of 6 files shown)


.cirrus.yml
```diff
@@ -0,0 +1,201 @@
...
+    folder: ${LOCALAPPDATA}/pip/Cache
+  conan_cache:
+    folder: ${USERPROFILE}/.conan/data
+  maven_cache:
+    folder: ${USERPROFILE}/.m2
+  bootstrap_script:
+    - pip3 install --user conan
+
+env:
+  CYCLONEDDS_URI: '<CycloneDDS><DDSI2E><Internal><EnableExpensiveChecks>all</EnableExpensiveChecks></Internal></DDSI2E></CycloneDDS>'
```
eboasson, line 101: That looks like an old one, on Travis we're using:

```suggestion
  CYCLONEDDS_URI: '<CycloneDDS><Domain><Internal><EnableExpensiveChecks>all</EnableExpensiveChecks><LivelinessMonitoring>true</LivelinessMonitoring></Internal><Tracing><Verbosity>config</Verbosity><OutputFile>stderr</OutputFile></Tracing></Domain></CycloneDDS>'
```

eboasson (approved):
This looks fine to me (also the way Cirrus presents the results and the time it took to run the builds), and I agree with the reasoning behind it as well.

There is a minor divergence from how we run the tests in Travis that I happened to notice, and I don't see why the change would make sense so I have suggested an edit that would bring it in line.

PR #655, opened 2020-12-07
Replace IDL compiler

k0ekk0ek:
This PR reworks the IDL compiler frontend. It's a polished version of the frontend contained in the `idlcxx` branch, but fully supports annotations, expressions and keeps includes (preprocessor still to be modified) on a stack. Apart from that each symbol/node now includes the normalized path of the source file it was parsed from, allowing generators to use it as a key and to determine the relative path to the translation unit. It's not quite done yet, but I'm looking for some initial reviews and findings from others. I'm hoping the new frontend is "done" before christmas, with a bit of luck, we can kick the Java version out before the years end.


src/tools/idlc/idlc.c
```diff
@@ -266,13 +322,19 @@ int32_t idlc_parse(ddsts_type_t **typeptr)
   }
```
dennis-adlink, line 322: Some nitpicking: `fclose` (line 321) on `NULL` value results in undefined behaviour

src/tools/idlc/idlc.c
```diff
@@ -49,59 +49,55 @@ typedef struct {
...
     if (buf == NULL) {
-      retcode = IDL_MEMORY_EXHAUSTED;
+      retcode = IDL_RETCODE_NO_MEMORY;
       return -1;
     }
     /* update scanner location */
-    proc.scanner.cursor = buf + (proc.scanner.cursor - proc.buffer.data);
-    proc.scanner.limit = proc.scanner.cursor + proc.buffer.used;
+    pstate->scanner.cursor = buf + (pstate->scanner.cursor - pstate->buffer.data);
+    pstate->scanner.limit = pstate->scanner.cursor + pstate->buffer.used;
```
dennis-adlink, line 89: I think this should be `pstate->scanner.limit = buf + pstate->buffer.used;`, but the `scanner.limit` is overwritten on line 100 anyway, so it doesn't really matter

src/tools/idlc/idlc.c
```diff
@@ -199,56 +195,116 @@ static int idlc_printf(OUTDEST od, const char *fmt, ...)
...
+    pstate->files = file;
+    pstate->scanner.position.file = (const idl_file_t *)file;
+    pstate->scanner.position.line = 1;
+    pstate->scanner.position.column = 1;
+    pstate->flags |= IDL_WRITE;
   }
 
   if (opts.flags & IDLC_PREPROCESS) {
-    proc.flags |= IDL_WRITE;
+    pstate->flags |= IDL_WRITE | IDL_FLAG_ANNOTATIONS | IDL_FLAG_EXTENDED_DATA_TYPES;
```
dennis-adlink, line 275: `pstate` is not initialized at this point when only running preprocessor

dennis-adlink (commented):
@k0ekk0ek I've looked through the code for idlc and found a few minor issues. I'll continue reviewing the idl library and post my findings for that later.

src/tools/idlc/idlc.c
k0ekk0ek: fixed

src/tools/idlc/idlc.c
k0ekk0ek: fixed

src/tools/idlc/idlc.c
k0ekk0ek: fixed

k0ekk0ek:
Thanks @dennis-adlink! Fixed your comments so far. Didn't give the tool itself much thought (just made sure the compiler didn't croak). Did a small test and it got to the point where it needed the backend, so it should at least somewhat work :wink:

eboasson:
@k0ekk0ek I have taken a brief look (it is only a "draft" PR after all) and it overall looks fine to me.

What I would nonetheless like to suggest is to rewrite some of the `goto`s. While I am not against the use of goto, I prefer to reserve it for error handling and compare-and-swap loops, and (mostly) avoid it for other uses. Here, there are various places (like "goto grammar" — but there are several others like it) that fall in the "rather dubious" category for me.

I think a bit of refactoring would suffice to eliminate them without massively increasing complexity. Would you mind having a look at that?

k0ekk0ek:
Thanks @eboasson! There's definitely some loose ends to tie up (hence the "draft" :slightly_smiling_face:). @reicheratwork noticed an oddity with annotations on constructed types that I have to look into and the way `idlc` itself resolves the path looks off, but as far as I can tell, with the addition of importing command line options from the generator, all the plumbing should be in place. Still aiming for finishing up this week, but it may slip by a couple of days... :crossed_fingers:

k0ekk0ek:
Last push adds a working(?) compiler backend for C. The whole thing needs some polishing still, commits need to be tidied up, some changes need to be a little more self-contained, etc. It definitely does not build without warnings yet either, but this is the latest and greatest. I don't expect changes to the inner structure (except for pragma-keylist, but that's pretty minor).

* Sequences are generated according to the C language specification. The old compiler used the member name of a struct to make sequence definitions unique, that has been changed.
* A pragmatic listener-visitor (`idl_visitor`) is implemented to make constructing backends easier.
* `mcpp` now outputs GCC line directives that the compiler uses to buildup an include stack, which makes it a lot easier to decided which generated headers should be included too.
* ...

k0ekk0ek:
Support for positional parameters `%<position>$<flag>` (e.g. `%1$s`) in `*printf` wrapper functions is now supported too, which is supported since POSIX.2008 and through [`*printf_p` functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/vprintf-functions?view=msvc-160) in Microsoft Windows.

k0ekk0ek:
Last update removes the Java-based compiler and replaces it with working backend on the new compiler. There's still a bit of cleaning up to do, I have to spread the changes over commits, and add some tests, but I thought I'd push the latest-greatest in case people want to have a look-see.

k0ekk0ek:
Latest makes `idl_is_topic` and `idl_is_topic_key` work. The compiler can switch between the two based on the IDL version used. `-v4` gets you IDL4, `-v3.5` gets you IDL3.5 the latter being the default due to existing tests and users. We can figure out if we want to change that further down the road(?) I like the number of LOC changed now better too. There's actually more code going out than there's code going in. There's one last thing to do and that's reporting the actual scoped name in error reports instead of `foobar` :slightly_smiling_face:, apart from that I consider this "done". I'd like to add more tests, especially for the generator, but given that all existing tests pass (locally anyway) and I really want to get rid of the Java compiler, I plan to verify manually this time around and take time to write decent tests after it's in. @dennis-adlink, I processed most of the review comments you provided, so thanks for those. @eboasson, do you agree with the chosen path wrt merging/tests etc? (of course I will clean up the commit history too :wink:). If people want to play with it, that'd be much appreciated. There's only so many quirks I run into by myself...

photex:
Out of curiosity I ran this on our idl files. Our primary idl errors out and I didn't investigate yet to find out why. On the file that did work I noticed that the way the dds_public_impl.h header is included has changed:

```diff
-#include "dds/ddsc/dds_public_impl.h"
+#include "ddsc/dds_public_impl.h"
```

For the file that is breaking here is the output:
```
c:\projects\kopernikus\depot\shared\kdata\idl>\projects\cyclonedds\cmake-build-release\bin\idlc.exe avp.idl
0:0: syntax error
c:/projects/kopernikus/depot/shared/kdata/idl/avp.idl:232: fatal error: File write error
      struct AckermannCommand {
1 error in preprocessor.
```

I've attached the file here: [avp.idl](https://github.com/eclipse-cyclonedds/cyclonedds/files/5933305/avp.txt)

This file defines some common data used by other files which we're `#include`ing. Works with Cyclone 0.7.0, but we're anxious to use this new idlc so we're happy to help test anything or learn what we've done incorrectly that just happens to be working. :D

Also worth noting is that on Windows I forgot to install winflexbison, but I had gow installed from scoop which includes an old and incompatible bison.exe. The configuration didn't catch this but instead gave an error during build. Running `scoop install winflexbison` fixes everything, but it's something to maybe be aware of in case other scoop and gow users run into the problem.

k0ekk0ek:
@photex, thanks so much for giving it a try! Don't know why the file would break the parser with a syntax error yet, there doesn't seem to be anything special in it on first glance, but I'll be sure to give it a try.

k0ekk0ek:
@photex, gave it a quick try and it runs just fine here. Can it be that there's a UTF-8 BOM at the start of the file? Don't know if that's something Microsoft Windows still does, but the `0:0: syntax error` might indicate that there is. I'll give it a quick Windows run and see if I can replicate. Thanks again!

photex:
Good catch!

I ran it under WSL:

```
photex@freya:/mnt/c/projects/kopernikus/depot/shared/kdata/idl$ /mnt/c/projects/cyclonedds/.build/bin/idlc avp.idl
/mnt/c/projects/kopernikus/depot/shared/kdata/idl/avp.idl:0: warning: Unknown encoding: C.UTF-8
    0:0: syntax error
/mnt/c/projects/kopernikus/depot/shared/kdata/idl/avp.idl:232: fatal error: File write error
      };
1 error in preprocessor.
```

I have no idea how to remove that or deal with it.

k0ekk0ek:
@photex, is the file hosted in a public repository anywhere by any chance?

photex:
No, but is there some other way I could send it to you that wouldn't remove the BOM?

k0ekk0ek:
That'd be awesome! If you could tgz/zip it and send it to the address for Jeroen Koekkoek that's in the git log (probably better not to put the address here(?)), I'll have a look.

k0ekk0ek:
@photex, didn't find the byte order mark in your files, so there's definitely something else at play. Seems Windows has some build issues that I didn't pick up on (my Travis credits are depleted). Managed to fix a couple of them, I'll update this early next week. If time permits I'll do a run in WSL too. I'll push a little update asap. Thanks for giving it a try, it's much appreciated :+1:

photex:
🤦 Ok... I get the warning still but I can generate code now. I didn't have the latest from idl-next apparently. Really sorry to have wasted so much of your time this evening!

photex:
So in my kpns.idl, I get an error using bitshifts now. Line 61 in the file I shared with you.

Previously I was setting up constants like:
```
const unsigned long FOO = 1;
const unsigned long BAR = 1 << 2;
const unsigned long BIZ = 1 << 3;
```

I replaced those with simple values instead.

Now it errors out when it reaches the pragma directives: 
```
/mnt/c/projects/kopernikus/depot/shared/kdata/idl/kpns.idl:390:21: Invalid token in #pragma keylist directive
```

Is `FOO::Bar` no longer correct?

Thanks for looking those over. I'll leave you to your weekend now. :D

k0ekk0ek:
@photex, I didn't add the shift operators to the scanner yes. I'll be sure to add them shortly.

The scoped name notation for `#pragma keylist` actually shouldn't work. It wouldn't be too much work to add it but I'm a little hesitant to do so. Intended use is to define it in the same module just after the struct. That's how it works for OpenSplice. Not that we absolutely *must* adhere to that, but that's what's in the documentation. Actually, the documentation says something like `#pragma keylist data-type key1 key2`, so that leaves some room :slightly_smiling_face: @eboasson, any thoughts?

eboasson:
I think it makes vastly more sense to allow a scoped name in `#pragma keylist` than to disallow it.

There's even a reasonable argument for keeping the `#pragma keylist` things separate from the type definitions (it really concerns topics, after all), which is currently possible but would require reopening the module just to do the directive. So you're not changing expressiveness, just making it a bit more friendly.

Besides that, documentation bugs are not uncommon ...

photex:
There is documentation!? :D I just sorta pieced this together from examples and what I found in the repo and worked through trial and error.

The reason I put pragmas at the bottom of the file is that it makes it a bit easier to find the specific topics at a glance.

k0ekk0ek:
I'll add it. Originally I actually intended to do so, but after a couple discussions with other colleagues we thought it'd be better to be strict wrt compatibility with [idlpp](https://github.com/ADLINK-IST/idlpp-cxx) :slightly_smiling_face:. [documentation](download.prismtech.com/docs/Vortex/pdfs/OpenSplice_IDLPreProcGuide.pdf) can be found here if you're curious. Btw, compatiblity is also the reason I only accepts base types and strings as keys for `#pragma keylist` (the `@key` annotation from IDL4 allows for other types to be keys too).

eboasson:
> Btw, compatiblity is also the reason I only accepts base types and strings as keys for #pragma keylist

Frankly, I think that's compatibility gone awry 🙂

Continued support of `#pragma keylist` definitely is important, even with the IDL4 way of using annotations probably taking over in the future. Artificially limiting the set of allowed key types smells to me like a severe case of "because that's how it was done in the past", and that's not a good reason. But I'm happy to be enlightened.

k0ekk0ek:
@photex, with the current state I can generate code for the files you provided. I only take a quick look at the generated code, but the more complex opcodes seem to match. Perhaps you can give it another try?

photex:
idlc finishes without an error. I do get warnings about the lack of newline at the end of each file which is fine of course.

It looks to me that enum output isn't functioning correctly. Here is a portion of the difference in output:

```diff
-typedef enum KPNS_Planner_State
+typedef enum (null)
 {
   KPNS_Planner_NotAvailable,
   KPNS_Planner_Ready,
   KPNS_Planner_Planning,
   KPNS_Planner_Tracking
-} KPNS_Planner_State;
+} KPNS_Planner_Tracking;
```

k0ekk0ek:
Ouch :slightly_smiling_face: I focussed on opcode generation because that was the tricky part. Type generation is reasonably straightforward (famous last words). Thanks for giving it another spin.

photex:
All good! The build is fast and the tool is fast *and* light. I'm very excited about all this work! I'm just happy to help however possible.

eboasson:
@k0ekk0ek once it builds it works until it crashes 😁

The building of course is because I use macOS 11 and I had to put to patches in:
```
diff --git a/src/idl/src/string.c b/src/idl/src/string.c
index c22b6980..43ef1cb4 100644
--- a/src/idl/src/string.c
+++ b/src/idl/src/string.c
@@ -278,7 +278,7 @@ int idl_vfprintf(FILE *fp, const char *fmt, va_list ap)
 #if _WIN32
   /* _vfprintf_p_l supports positional parameters */
   return _vfprintf_p_l(fp, fmt, posix_locale(), ap);
-#elif __APPLE__ || __FreeBSD__
+#elif __FreeBSD__
   return _vfprintf_l(fp, fmt, ap);
 #else
   int ret;
diff --git a/src/tools/idlc/Generate.cmake b/src/tools/idlc/Generate.cmake
index d8ed31db..ae11ccee 100644
--- a/src/tools/idlc/Generate.cmake
+++ b/src/tools/idlc/Generate.cmake
@@ -38,4 +38,5 @@ function(IDLC_GENERATE)
   add_library(${_target} INTERFACE)
   target_sources(${_target} INTERFACE ${_sources} ${_headers})
   target_include_directories(${_target} INTERFACE "${_dir}")
+  add_dependencies(${_target} "${_target}_generate")
 endfunction()
```
I assume you had the `#if __APPLE__` in there for a reason, but the header file doesn't say at what point `vfprintf_l` was introduced, so I took the easy way out and simply dropped the term.

The `add_dependencies` bit is required for XCode generate to work. Without it, CMake gives a really vague error:
```
CMake Error in src/tools/ddsperf/CMakeLists.txt:
  The custom command generating

    /Users/erik/C/cdds/XX/src/tools/ddsperf/ddsperf_types.c

  is attached to multiple targets:

    ddsperf
    ddsperf_types_lib_generate

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".


CMake Error in src/idl/CMakeLists.txt:
  The custom command generating

    /Users/erik/C/cdds/XX/src/tools/ddsperf/ddsperf_types.c

  is attached to multiple targets:

    ddsperf
    ddsperf_types_lib_generate

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".


CMake Generate step failed.  Build files cannot be regenerated correctly.
```
That `add_dependencies` bit used to be there in the old one, otherwise I probably would never have discovered that fix.

Then, if you feed it the IDL file in [cdrtest.zip](https://github.com/eclipse-cyclonedds/cyclonedds/files/5952413/cdrtest.zip) generated by the `src/core/xtests/cdrtest/cdrtest.pl`, it asserts somewhere. I haven't attempted to isolate the problem.

k0ekk0ek:
> once it builds it works until it crashes

Story of my life :wink:. Thanks @eboasson! Probably something I overlooked in the backend. Now working on removing the `<foobar>`'s with actual names for error reporting. Once I finish that, I'll *shake the tree and see what falls out* :grin:

Pro:
@k0ekk0ek this line should probably be changed:

https://github.com/k0ekk0ek/cyclonedds/blob/idl-next/cmake/Modules/Packaging/PackageConfig.cmake.in#L19

to (IdlcGenerate.cmake vs Generate.cmake)

`include("${CMAKE_CURRENT_LIST_DIR}/idlc/Generate.cmake" OPTIONAL RESULT_VARIABLE _IdlcGenerate)`

k0ekk0ek:
Thanks @Pro, I'll include it in the next batch of changes :+1:

k0ekk0ek:
There's still the `assert` in the backend, I'm looking into that tomorrow, as well as the changes proposed by @Pro.

k0ekk0ek:
Second-to-last update before I fumble all of the commits in to one and remove the draft status. This one makes the generator successfully parse and generate opcodes for the IDL files provided by @photex and @eboasson. @reicheratwork kindly provided a test in a PR that I shamelessly copied to test reopening of modules.

The key order specified in the keylist is not yet maintained, so I'll have to look into that. After that I need to add some backwards compatibility changes to the `idlc_generate` CMake function that we ship and remove the "garbage collection", which is itself *garbage*, from the backend and we should be all good(?) :tada:

k0ekk0ek:
Well, changes have been fumbled up into one and in it's current state it builds on Linux and Windows (macOS not tested). As per usual, there were a couple of minor things that popped up last minute, but we should be in pretty good shape. The files @photex provided brought a rather interesting issue to light wrt cleanup on parse errors from within Bison. Btw, I should note that those files will now only compile if you provide `-f case-sensitive`, because they're not actually valid IDL. e.g. `struct S { char s; };` is not valid as IDL is case insensitive. Apart from that:
* Keys now appear in the order that they appear in the `#pragma keylist` directive and whether the `#pragma keylist` directive or `@key` is used, is determined automatically. If both appear, it'll throw a warning and proceed based on annotations. If you must use the pragma method, provide `-f keylist` and no warning will show.
* `idlc_generate` learned a couple of new tricks and is now keyword based. Old-style invocations are recognized and it'll show a warning if that is used. It will work though. It'll also pass `-f case-sensitive` automatically as that was the default behavior with the Java-based compiler.
* ...

I think the compiler is more than good enough to be merged into the `xtypes` branch and hopefully soon that branch can be merged to master. It should support at least everything that the old compiler did and I've seen it generate more correct code in some of the tests I ran. That being said, I'd be surprised if no more bugs pop-up as I've not tested nearly enough as I would've liked. On the other hand, if we wait until it's absolutely perfect, I'll probably never finish. The Java-based compiler itself is also not without it's flaws, the right IDL constructions will make it crash and burn or just generate complete garbage.

@eboasson, can I go ahead and hit merge on this?

Also, the tool can use extra eyes to determine if the options make sense etc, so if any interested people are reading this: please give it a try and report bugs, improvements and whatnot. Thanks!

And before I forget: it'll not do forward declarations or recursive types yet, I'm saving that for the next iteration.

k0ekk0ek:
Just popped into my mind, I broke something while fixing another. Will fix over the weekend.

k0ekk0ek:
This should take care of the build failures (so getting used to Travis not running that I didn't even checked :blush:). Did some cleaning up while I was at it. There's definitely more that can be cleaned up, but I'll do that when I replace Travis. I think in it's current state it's better than the Java based compiler, so we should be good to go.
1. Java based on gets the opcode count wrong (I know it's not used but still)
2. In the cdrtest there's at least one case where it gets alignment wrong (type `a4859`). Because there's an unsigned long long, it should say `8u`, but instead it says `sizeof(char *)`. Which is fine on 64-bit machines, but not on 32-bit, where `sizeof(char *)` would be 4 bytes instead of 8.
3. It generates the wrong code for sequences of bounded strings.

src/idl/src/parser.y
```diff
@@ -0,0 +1,1126 @@
...
+  | IDL_TOKEN_DEFAULT const_expr
+      { $$ = $2; }
+  ;
+
+any_const_type:
+    IDL_TOKEN_ANY
+      { TRY(idl_create_base_type(pstate, &@1, IDL_ANY, &$$)); }
+  ;
+
+annotation_appls:
```
eboasson, line 953: The shift/reduce conflicts are all on "@" and originate here. Bison's shift/reduce conflict handling resolves them as shifts and that'll probably do the right thing here. That means a `%expect 9` could be an option, but it is better to eliminate the conflicts. I think (so you'd better check ...) that this can be done quite easily:

```
annotation_appls:
    %empty
      { $$ = NULL; }
  | nonempty_annotation_appls
      { $$ = $1; }
  ;

nonempty_annotation_appls:
    annotation_appl
      { $$ = $1; }
  | nonempty_annotation_appls annotation_appl
      { $$ = idl_push_node($1, $2); }
  ;
```

(Just removing the first case of `annotation_appl` only eliminates half of the conflicts.)

eboasson (commented):
I dislike warnings in general and I really don't like reduce/reduce conflicts in LR parsers. So there is a little bit of work to be done still (but I think I've done it for you already 🙂).

Given that I have a comment anyway ...

What about the `COMPONENT idlc` in the CMake files? It got added for some reason that I suspect (I don't care to know CMake well enough to really understand why) to be related to optionally building the IDL compiler, which in turn was hacked in to make life easier in ROS 2 for people that don't have a JDK and maven available and don't need it.

Now that the IDL compiler is a just a C program, there seems to be little reason to keep all that around. (Unless `bison`, too, is considered a problematic dependency. One despairs, &c. Next thing we'll have to check in the bison-generated sources like `gcc` used to do way back when IIRC. Come to think of it, I would probably even prefer the latter.)

There's no need to decide on all that before merging it. Nothing's going to break by keeping the IDL compiler as a component at CMake level. So please check out those grammar issues, and then merge it once those are taken care of.

Note that merging it into master will have to wait for a successful run on ROS 2 CI as well.

k0ekk0ek:
Well, the funny thing is `idlc` is not actually a `COMPONENT`. At least, when you read through the CMake files, the `install` command uses either `lib` or `dev` (took some time before I made that connection last saturday in trying to figure out what was wrong with the `CycloneDDS::ddsc` target).

k0ekk0ek:
Warnings fixed, component thingy removed. Fingers crossed :crossed_fingers:

eboasson:
Me again ... when I try to build the "hello world" example using its CMakeLists.export but with this branch as the installed Cyclone, I get:
```
CMake Error at CMakeLists.txt:5 (find_package):
  Found package configuration file:

    /Users/erik/C/cdds-idlc/CC/install/lib/cmake/CycloneDDS/CycloneDDSConfig.cmake

  but it set CycloneDDS_FOUND to FALSE so package "CycloneDDS" is considered
  to be NOT FOUND.


-- Configuring incomplete, errors occurred!
```

that error disappears when I remove the `COMPONENT idlc` part from the `find_package` line (after which I get a warning  that the usage of `idlc_generate` is deprecated, which is excellent).

Is there a way to allow presence of an `COMPONENT idlc` to not cause an error?

PR #659, opened 2020-12-15
Odds and ends

eboasson:
Some minor (this time it is true!) changes.

* One externally visible change: actually treating a failure to get the requested socket buffer size as an error rather than vying for a place in the Underhanded C code contest;
* One performance improvement: the removal of an unnecessary padding clearing loop when concatenating received fragments;
* Two fixes for suppressing warnings from Clang's static analyser.


k0ekk0ek (approved):
LGTM. Curious to see how much impact the performance fix has. Do you remember how much was gained from your tests perhaps?

eboasson:
That padding clearing loop was turned into a `memset` by the compiler that ended up taking about 10% of the CPU time used by the receiving side when doing ddsperf sub / ddsperf pub size 10M over loopback on an M1-based MBP. So definitely worth removing some code for!

PR #664, opened 2021-01-06
add GettingStarted

YixianLv:


k0ekk0ek:
Thanks @YixianLv! Sorry for the delay in review. Documentation is in dire need of attention, so this is very much appreciated :+1: Giving this a go first thing tomorrow.

k0ekk0ek:
@YixianLv, I've taken a decent look and took some time to familiarize myself with the tool(s). Overall, I really like what you've done with the text. It reads well and the instructions are accurate. Some are arguably a bit iffy, like instructing users to use `rm -rf *` to uninstall on Linux, but those are minor details.

However, I'm a bit concerned about build system integration lacking and introducing a new tool (`mdbook`) to the mix. Currently when you configure the project with `-DBUILD_DOCS=On`, it'll automatically search for `Sphinx` and `breathe`, the first one being a documentation generator (like `mdbook`), the latter providing integration with `Doxygen`, the tool used to generate api documentation. Then, when you type `cmake --build . --target docs`, Doxygen and Sphinx are automatically executed to generate documentation.

At this point, a bit of background information might be useful to the passerby:
Like `mdbook`, using [Jupyter Notebook](https://jupyter.org/)'s for interactive examples eventually were internal suggestions, so I took time to familiarize myself with them. Jupyter Notebook's are a way to have interactive examples. They're basically JSON documents containing notes and code, maybe visualation instructions too, that can be run on a server to have an interactive experience. For the Python binding, colleagues are looking into Jupyter Book, which is more-or-less the equivalent of mdbook and Sphinx, but has tight integration with Jupyter Notebooks. Jupyter Book fully supports reStructuredText, which is used by Sphinx, because it's based on Sphinx.

Taking all of this into account, I think the project should stick to Sphinx for now. Especially because we would lose integration with Doxygen, which allows us to reference api functions from the documentation and have them be clickable links in the output. Sphinx also allows us to integrate the actual code for examples by reference, making sure it's always up-to-date.

Of course, this is a decision I can't make by myself, so let's leave this open for a couple of days to give others a chance to comment.

thijsmie:
Superseeded by #1340

PR #666, opened 2021-01-07
Allow re-reaping thread state in thread_states_fini on Windows

eboasson:
On Windows the thread finalization routine for the main thread appears to be
called prior to executing the global destructors.  This triggers an assert in
thread_states_fini if Cyclone shutdown is delayed until some global destructor.

Windows also has the lovely habit of killing all other threads prior to running
the global destructors, so this only helps for the rare case where Cyclone is
kept alive strictly for guard conditions and waitsets.  If a domain was still
running, the killing of threads will corrupt the state and cause errors in
trying to join the killed threads.  Best is self-evidently to use Windows
and/or global destructors only if you must ...


dennis-adlink (commented):
LGTM!

PR #667, opened 2021-01-07
Correct tracemask in config initializer generator

eboasson:
The default config generator initialized the tracemask to the value of
the "valid" field rather than to the value of the "tracemask" field.
This slipped through the test because the test failed to explicitly
check that it is set to 0 prior to forcing it to "config", which it uses
for automatic checking of all other settings.


dennis-adlink (approved):
LGTM

PR #668, opened 2021-01-07
Handling ignored locators

### LINKED ISSUE #663: ddsi_locator_to_string received signal SIGSEGV, Segmentation fault.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/aeb281d1e227e56799ef4671235679120ba66b3f/src/core/ddsi/src/ddsi_plist.c#L463

(gdb) list
269     }
270
271     char *ddsi_locator_to_string (char *dst, size_t sizeof_dst, const ddsi_locator_t *loc)
272     {
273       /* FIXME: should add a "factory" for INVALID locators */
274       if (loc->kind == NN_LOCATOR_KIND_INVALID) {
275         (void) snprintf (dst, sizeof_dst, "invalid/0:0");
276       } else if (loc->tran != NULL) {
277         int pos = snprintf (dst, sizeof_dst, "%s/", loc->tran->m_typename);
278         if (0 < pos && (size_t)pos < sizeof_dst)
(gdb) p loc
$1 = (const ddsi_locator_t *) 0x20c
(gdb) p *loc
Cannot access memory at address 0x20c

eboasson:
Do not set "present" flag for ignored locators  …

It looks like this change was accidentally dropped from fb7034e in cleaning up that commit. Unfortunately I don't have enough historical branches and commits left to be absolutely certain.

This fixes #663.

### DIFF
--- a/src/core/ddsi/src/ddsi_plist.c
+++ b/src/core/ddsi/src/ddsi_plist.c
@@ -2659,7 +2659,7 @@ static dds_return_t init_one_parameter (ddsi_plist_t *plist, nn_ipaddress_params
     ret = entry->op.f.deser (dst, &dstoff, &flagset, entry->present_flag, dd, &srcoff);
   else
     ret = deser_generic (dst, &dstoff, &flagset, entry->present_flag, dd, &srcoff, entry->op.desc);
-  if (ret == 0 && entry->deser_validate_xform)
+  if (ret == 0 && (*flagset.present & entry->present_flag) && entry->deser_validate_xform)
     ret = entry->deser_validate_xform (dst, dd);
   if (ret < 0)
   {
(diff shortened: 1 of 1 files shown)


src/core/ddsi/src/ddsi_plist.c
```diff
@@ -374,11 +374,18 @@ static dds_return_t deser_locator (void * __restrict dst, size_t * __restrict ds
...
+    case DOLOC_INVALID:
+      return DDS_RETCODE_BAD_PARAMETER;
+    case DOLOC_IGNORED:
+      break;
+    case DOLOC_ACCEPTED:
+      *flagset->present |= flag;
+      break;
+  }
   *srcoff += 24;
   *dstoff += sizeof (*x);
```
dennis-adlink, line 388: Should the value of `dstoff` be increased in case the locator is ignored, because the locator will not be in `dst`? (it's not really a problem, because `dstoff` is currently not used after this function).

dennis-adlink (commented):
Looks good, only one minor thing I mentioned above.

src/core/ddsi/src/ddsi_plist.c
eboasson: Looks to me like this is the case for both `srcoff` and `dstoff`. With the value never used after the call, I'd say those really should be pass-by-value. I'll merge this PR as-is, and do a separate PR to change the types, update all (few) specialised deserialisers and update the call site.

PR #669, opened 2021-01-10
Fix use-after-free problems with triggering read conditions

eboasson:
This PR fixes two bugs with similar symptoms:
* The RHC implementation sometimes deferred triggering the read conditions until after unlocking the RHC, but that creates a window during which the read condition has been freed and yet the RHC still tries to use it.

  For readers (and the data available callback) the analogous problem is avoided by delaying the freeing of the reader entity via the GC thread. A similar trick could be used for read conditions, but that is decidedly more complicated.  Therefore, this commit chooses the simpler solution.

* The RHC can calls into the read condition to signal that it has been triggered, which then causes the read condition to signal any attached waitsets.  So it follows that the RHC must not be attempting to signal it by the time the list of attached waitsets is freed.

It adds a test that exercises these code paths a bit & cleans up the read condition test cases a bit by removing some code duplication.

### DIFF
--- a/src/core/ddsc/src/dds_entity.c
+++ b/src/core/ddsc/src/dds_entity.c
@@ -447,8 +447,18 @@ static dds_return_t really_delete_pinned_closed_locked (struct dds_entity *e, en
 
   /* Pin count dropped to one with CLOSING flag set: no other threads still
      in operations involving this entity */
-  dds_entity_observers_signal_delete (e);
+  if (dds_entity_kind (e) == DDS_KIND_WAITSET)
+  {
+    /* This takes care of the rare and deprecated case of a waitset attached
+       to itself. Those only ever get triggered by pinning the waitset, and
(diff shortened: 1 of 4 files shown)


dennis-adlink (commented):
The changes look good to me, only some nit-picks: 
- the comment on dds_rhc_default.c line 153-154 needs an update, because the triggering is not deferred any more
- you could consider adding a comment on `dds_readcond_close`, that explains the read condition should removed from the waitset at this point, and not just when deleting the condition

eboasson:
> The changes look good to me, only some nit-picks:
> 
> * the comment on dds_rhc_default.c line 153-154 needs an update, because the triggering is not deferred any more
> * you could consider adding a comment on `dds_readcond_close`, that explains the read condition should removed from the waitset at this point, and not just when deleting the condition

Thanks & done!

PR #671, opened 2021-01-15
Add serialized data access variants for the read/take_instance operations of the DataReader.

e-hndrks:
Needed to support the C++ implementation of the loaned (zero-copy) read/take operations of a Selector that was set to look for one particular instance by means of its instance handle.


eboasson (approved):
Thanks @e-hndrks!

PR #674, opened 2021-01-18
Remove unnecessary args from plist deserializers

eboasson:
The deserialization functions for the individual parameter values were always invoked with source and destination offset of zero, and the results were always thrown away.  This changes the (internal) interface to not pass the arguments unnecessarily.

### DIFF
--- a/src/core/ddsi/src/ddsi_plist.c
+++ b/src/core/ddsi/src/ddsi_plist.c
@@ -105,7 +105,7 @@ struct piddesc {
        will fit */
     const enum pserop desc[12];
     struct {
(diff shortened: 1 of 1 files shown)


dennis-adlink (commented):
LGTM, the only thing I would consider is moving the code for unsetting a flag (in case of deserialization failure) to `init_one_parameter`, so that the function `deser_generic` can be removed

eboasson:
> LGTM, the only thing I would consider is moving the code for unsetting a flag (in case of deserialization failure) to `init_one_parameter`, so that the function `deser_generic` can be removed

That's a good point, but after trying it out I am of the opinion that the advantage of losing `deser_generic` doesn't outweigh the disadvantage of losing the symmetry and absence of offsets in `init_one_parameter`. Merging it as it is.

PR #675, opened 2021-01-19
Fix CMake package config version file name

roehling:
According to the [CMake Manual](https://cmake.org/cmake/help/latest/command/find_package.html#version-selection), the version file must be named `<ConfigName>Version.cmake`, i.e., `CycloneDDSConfigVersion.cmake`.


k0ekk0ek:
Thanks @roehling!

PR #679, opened 2021-01-25
Move multi-process tests to regular tests

eboasson:
The multi-process tests did not integrate well with the rest, living in a different part of the source tree, using a somewhat complex structure and requiring different asserts, and they only work on platforms that have multi-process support and they are much harder to debug.

They were introduced in the days that there was no way of doing network communication within a process, but that limitation has long since been lifted and there are now many tests that use multiple domain instances to achieve the same result.

This commit moves the interesting tests into the normal testing framework and removes the multi-process framework altogether.


dennis-adlink (commented):
Looks good, nice to have this cleaned up! The documentation can also be removed: https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/multi_process_testing.md 

I think the new approach using `externaldomain` does not need any documentation other than the code comments that are already in.

k0ekk0ek:
@eboasson, I still want to have a look at this. I recall at the time it was necessary primarily for security tests etc. Not sure if it was used for that eventually, but still.

k0ekk0ek:
Right, the original intent was to use this for testing the security implementation. As CTest only supports specifying one executable, the idea was to have a very minimal framework that manages some child processes (hence multi-process is required). The use case for this is testing if processes with and without security can successfully communicate. The idea was: specify a couple of entry points, the executable forks the number of required processes that execute the specified functions and reads the exit codes to determine if the test was successful. Why it has turned into what it is now is a mystery, why no one actually used it for what it is intended for is an even bigger one.

In short: I'm totally fine with moving the current tests. I'm fine with dropping mpt too. I do think we should add some actual tests for security that verify communication between processes with security enabled and processes without security enabled. I don't care if we use mpt for that or not, but doing cross-platform multi-process testing can be a bit of a challenge (without adding yet more dependencies).

For now, it's probably best to merge this and create an issue to implement some tests that refers to this PR(?)

eboasson:
@k0ekk0ek, regarding testing security: we do that (quite a bit of it, actually) by using the same "trick" of instantiating multiple domains in a single process that (then, necessarily) communicate via the network stack. The `ExternalDomainId` setting makes that happen. That is something that wasn't originally possible, but is now used quite extensively.

So I think there is no need to worry about the existence of security tests. (There is always good reason to worry about whether the tests that exist are sufficient, of course.)

k0ekk0ek:
Well, it's not quite the same as having separate processes, but I'll take it. :slightly_smiling_face:  Just wanted to make sure that the idea wasn't lost. In theory it's quite a nifty way to do multi-process testing and I'm sure that at some point we're likely to want to start doing that sort of thing. Feel free to merge :+1:

k0ekk0ek (approved):

PR #683, opened 2021-01-31
Add ANALYZER option to enable static analyzers in GCC and MSVC

k0ekk0ek:
Add support for GCC's (>=10) `-fanalyzer` and MSVC's `/analyze` options to do static analysis during compilation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,6 @@ endif()
 # Enable coloured ouput if Ninja is used for building
 if("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR
    "${CMAKE_C_COMPILER_ID}" STREQUAL "AppleClang")
-  #message(STATUS clang)
   set(wflags "-Wall"
              "-Wextra"
              "-Wconversion"
@@ -124,7 +123,6 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR
     add_compile_options(-Xclang -fcolor-diagnostics)
   endif()
 elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
-  #message(STATUS gcc)
   add_compile_options(-Wall -Wextra -Wconversion -Wmissing-prototypes)
   if(${WERROR})
     add_compile_options(-Werror)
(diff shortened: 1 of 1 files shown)


eboasson (commented):
A sensible addition, @k0ekk0ek! But I do have a few questions:
1. Would it not make more sense to aim for a consistent naming scheme for build options? And is, in light of that, `ANALYZER` the right name? Perhaps it is, perhaps `USE_SANITIZER` should become a simple `SANITIZER`, but in any case the discrepancy doesn't sit wel with me.
2. Does the Visual Studio `/analyzer` switch actually do anything useful if the code isn't annotated with SAL2 annotations? I do like SAL2 in theory, but in practice, it is badly documented, Microsoft-specific and impossible to get right without access to the Visual Studio compilers, and because of that basically unusable in any project that doesn't focus on the MS toolchain.
3. Clang has a lovely static analyzer, too. I dare say it would take its being ignored as an affront 🙂

k0ekk0ek:
1. Had a feeling you were going to say that :slightly_smiling_face: 
2. I tried it real quick and it does spit out a couple of warnings without SAL2 annotations, so I left it in.
3. Clang provides an `--analyze` option but that makes it not-compile. I didn't want to spend too much time on it, but I'll see if I can make it both compile and analyze at the same time (with Clang normally you'd run `make` with `scan-build`, much like the Coverity integration).

k0ekk0ek:
Normal use-case with Clang is definitely to provide `scan-build` before the `make` or `cmake --build .` command. In the past I did try to integrate scanning with Coverity in CMake with [`<LANG>_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html), but it's more trouble than it's worth. It's not platform agnostic and while I got it working in slightly modified form on Windows, writing custom Visual Studio instruction files is definitely no fun :slightly_smiling_face:. The variable is intended for things like ccache. I think we should simply leave out Clang in this particular case, or add a run with `scan-build` enabled too (perhaps modify one of the existing ones).

As for the name, I'm inclined to change `USE_SANITIZER` to `SANITIZER`.

Btw, Visual Studio 2019 has support for the address sanitizer too, so I might have a quick look and see if I can add that too...

Also, I didn't enable a run with `ANALYZER` yet because it did show a couple warnings (might be false-positives), but I'll have a go at that too.

PR #685, opened 2021-02-05
Asynchronous write implementation

sumanth-nirmal:
This PR adds an asynchronous write feature. \
See [Asynchronous Write](https://github.com/eclipse-cyclonedds/cyclonedds/blob/f96d3b7a5e25cf9671b3495abba2388ce9f9353c/docs/dev/async_write.md) doc for the details.


sumanth-nirmal:
@eboasson I still need to update the document, but it should be good for the first round of review!
Now the async write will only happen if a data writer is created with a non-zero-latency budget.

src/core/ddsc/src/dds_writer.c
```diff
@@ -395,6 +398,12 @@ dds_entity_t dds_create_writer (dds_entity_t participant_or_publisher, dds_entit
   dds_topic_allow_set_qos (tp);
   dds_topic_unpin (tp);
   dds_publisher_unlock (pub);
+
+  // start async thread if not already started and the latency budget is non zero
+  if (async_mode && !gv->sendq_running) {
```
eboasson, line 403: There's a race condition that might cause it to start two send threads if you create two writers in parallel. Adding a mutex should do the trick.

src/core/ddsi/src/q_init.c
```diff
@@ -1906,7 +1903,8 @@ void rtps_fini (struct ddsi_domaingv *gv)
 
   xeventq_free (gv->xevents);
 
-  if (gv->config.xpack_send_async)
+  // if sendq thread is started
+  if (gv->sendq_running)
```
eboasson, line 1907: There won't be any races at this point, but if a mutex is added to make setting `sendq_running` and starting the thread an atomic operation, I would probably lock it here, too. At least it'd pacify Coverity.

eboasson (commented):
Thanks @sumanth-nirmal! Resuscitated async sending of entire RTPS messages is definitely a valuable step forward 😀 It will in any case mean that very large messages (like point clouds) can be sent out in the background while all the rest continues, and that's quite an important case.

My own personal interest happens to be more in sending small messages at a high rate, and for that it is important that individual messages get queued and packed. That could also be done by adding RTPS message (a.k.a. `nn_xpack`) merging in the queue, but I think it makes much more sense to work towards queueing samples (a.k.a. `ddsi_serdata`) instead of RTPS messages to facilitate that. But as sending small messages at a high rate seems to be just my personal hobby whereas sending very large messages is a more common pastime, that improvement shouldn't delay this.

This:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/4dd9ba2ab495b2c6ed4406f9709a9a21d3562921/src/core/ddsi/src/q_xmsg.c#L1301

was fine for the original experiment, but I don't want a thread to be spinning at a high rate when there is no work to be done. That means the trigger condition at:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/4dd9ba2ab495b2c6ed4406f9709a9a21d3562921/src/core/ddsi/src/q_xmsg.c#L1364

would need some refinement. That's something that definitely requires some attention. The easiest is trigger as soon as the queue is no longer empty and wait indefinitely. That works fine for those (very) large samples.

src/core/ddsc/src/dds_writer.c
sumanth-nirmal: Fixed this

sumanth-nirmal:
> This:
> 
> https://github.com/eclipse-cyclonedds/cyclonedds/blob/4dd9ba2ab495b2c6ed4406f9709a9a21d3562921/src/core/ddsi/src/q_xmsg.c#L1301
> 
> was fine for the original experiment, but I don't want a thread to be spinning at a high rate when there is no work to be done. That means the trigger condition at:
> 
> https://github.com/eclipse-cyclonedds/cyclonedds/blob/4dd9ba2ab495b2c6ed4406f9709a9a21d3562921/src/core/ddsi/src/q_xmsg.c#L1364
> 
> would need some refinement. That's something that definitely requires some attention. The easiest is trigger as soon as the queue is no longer empty and wait indefinitely. That works fine for those (very) large samples.

@eboasson I have changed the trigger condition so it now waits indefinitely until the queue is not empty.  \
Agree, that we can add the feature of queuing the `ddsi_serdata` samples as an incremental change.

eboasson:
Thanks @sumanth-nirmal !

PR #686, opened 2021-02-05
feat: Add conanfile.py for conan packaging

Pro:
Adds a conanfile.py configuration to build a conan package for cyclone dds.

You can test it locally using:
`conan create -tf scripts/conan/test_package .` in the repo root.

Conan requires Python3, therefore all pip calls are also changed to use pip3.

Since maven is required, this depends on the following two PR:

https://github.com/conan-io/conan-center-index/pull/4491
https://github.com/conan-io/conan-center-index/pull/4488

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,10 @@ ubuntu1804_gcc7: &ubuntu1804_gcc7
   os: linux
   dist: bionic
   compiler: gcc
+  addons:
+    apt:
+      update: true
+      packages: [ python3-pip python3-setuptools ]
   before_install:
     - eval "export CC=gcc CXX=g++"
     - eval "export COV_COMPTYPE=gcc COV_PLATFORM=linux64"
@@ -75,7 +79,7 @@ ubuntu1804_gcc7: &ubuntu1804_gcc7
     - eval "export GENERATOR='Unix Makefiles'"
   install:
     - *install_coverity
-    - pip install conan --upgrade --user
+    - python3 -m pip install conan --upgrade --user
 
 ubuntu1804_gcc10: &ubuntu1804_gcc10
   os: linux
@@ -86,14 +90,14 @@ ubuntu1804_gcc10: &ubuntu1804_gcc10
       update: true
       sources:
         - sourceline: 'ppa:ubuntu-toolchain-r/test'
-      packages: [ gcc-10 g++-10 ]
+      packages: [ gcc-10 g++-10 python3-pip python3-setuptools ]
   before_install:
     - eval "export CC=gcc-10 CXX=g++-10"
     - eval "export COV_COMPTYPE=gcc COV_PLATFORM=linux64"
     - eval "export BUILD_TOOL_OPTIONS='-j 4'"
     - eval "export GENERATOR='Unix Makefiles'"
   install:
-    - pip install conan --upgrade --user
+    - python3 -m pip install conan --upgrade --user
 
 ubuntu1804_clang10: &ubuntu1804_clang10
   os: linux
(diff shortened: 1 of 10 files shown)


eboasson (commented):
Thanks @Pro I really like this (not that I know much about conan ... fortunately @k0ekk0ek does, so I am sure he will do a proper review soon). There are a few remarks that I have, though:

* I would prefer not to have a copy of "hello world" in scripts/conan/test_package. I am sure there is some way to avoid the code duplication. If there isn't, then I would prefer replacing it with an even more stripped down application.
* There should be copyright headers in the new files (e.g., `conanfile.py`)
* Since we're working hard to eliminate maven and java from the sources altogether and the transition should happen "real soon now" (#655) I think it would make sense to this PR on top of #655.

@k0ekk0ek what do you think?

spiderkeys:
Peanut gallery comment: I'm a big Conan user/advocate and would love to have this added :slightly_smiling_face: 

The recipe itself looks pretty good. Regarding the test_package, general convention that I've found in many Conan packages is to simply do the bare minimum in the program to test that the dependency was correctly packaged, rather than demonstrate functional correctness (i.e. include header files and create objects/call functions that demonstrate successful compilation/linking).

k0ekk0ek:
Thanks @Pro. I like Conan a lot myself and we use it too as you've noticed. It's been a while since I last created recipes myself, and things have changed a bit wrt Bincrafters, the ConanCenter etc. The test_package probably shouldn't copy the example, it's much better to only verify packages that depend on it build, like @eboasson and @spiderkeys suggest. I agree with @eboasson too in that it's better to wait until the Java and Maven dependencies have gone.

That being said, the main reason I haven't commented until now is that I couldn't figure out if this package should be maintained within the Eclipse Cyclone DDS repository. Having thought about it a little, I think it's better to add the recipe for building the Cyclone DDS sources to the [ConanCenter](https://github.com/conan-io/conan-center-index) repository. That'd make it much easier for users to consume the package(?) and allows for switching dependencies (and build instructions) between versions(?)

What is the normal procedure for packages these days and how would users consume the Cyclone DDS package?

Pro:
> * I would prefer not to have a copy of "hello world" in scripts/conan/test_package. I am sure there is some way to avoid the code duplication. If there isn't, then I would prefer replacing it with an even more stripped down application.

> Regarding the test_package, general convention that I've found in many Conan packages is to simply do the bare minimum in the program to test that the dependency was correctly packaged, rather than demonstrate functional correctness (i.e. include header files and create objects/call functions that demonstrate successful compilation/linking).

My idea behind having the whole hello world in there was to verify if the IDL compiler is working inside the conan package, and the corresponding linking and loading of the libraries is correct.

> * There should be copyright headers in the new files (e.g., `conanfile.py`)

I can add them. If we would move it to conan-center, then there probably have to be different ones.

> * Since we're working hard to eliminate maven and java from the sources altogether and the transition should happen "real soon now" (#655) I think it would make sense to this PR on top of #655.

Definitely! Getting rid of the Java Dependency is quite nice! Then I'd suggest to wait until #655 is in, and then rework this PR.

> That being said, the main reason I haven't commented until now is that I couldn't figure out if this package should be maintained within the Eclipse Cyclone DDS repository. Having thought about it a little, I think it's better to add the recipe for building the Cyclone DDS sources to the ConanCenter repository. That'd make it much easier for users to consume the package(?) and allows for switching dependencies (and build instructions) between versions(?)
> 
> What is the normal procedure for packages these days and how would users consume the Cyclone DDS package?

Good question...
Having the conanfile.py in the Cyclone DDS Repo helps to avoid breaking changes. This is probably also the only advantage I see...

On the other hand, conan-center handles all the creation of conan packages and has various hooks to check for incompatibilities and best-practices. For conan-center, the community needs to create new PRs when Cyclone DDS is updated/new version released, to get a conan package for this new version. And conan users can much easier get the pre-compiled conan package.

So I guess the conan-center approach would be better.
If you agree, I can create a new PR to the conan-center-index which is adding cyclone-dds there. (They typically require such a confirmation if a non-maintainer is adding a package).

k0ekk0ek:
Feel free to issue a PR against conan-center-index (I'm sure @eboasson agrees). I'll help out where I can. Let me know once the PR is issued, I'll have a look at it and give my :+1: (comments if applicable). One quick remark in advance :slightly_smiling_face: , the package is probably  better named `cyclonedds` rather than `cyclone-dds` for consistency.

k0ekk0ek:
@Pro, just wanted to inform you the new Bison based compiler has been merged to the master branch. If you still plan on creating a Conan package, the Java dependency has been removed.

k0ekk0ek:
@Pro, I'm closing this PR. Feel free to refer to it or open an issue if you need help with packaging.

PR #689, opened 2021-02-15
Bug fix in freeing loaned samples

dennis-adlink:
Loaned samples are not correctly freed in `dds_delete_reader`. This could lead to leaking memory when a reader is deleted right after reading data, before the application has returned the loaned data (the return loan will result in a precondition_not_met error in that case).


eboasson (approved):
Thanks @dennis-adlink!

PR #690, opened 2021-02-16
Support abbreviating attribute names in XML

eboasson:
This allows abbreviating attributes names in XML read directly from `CYCLONEDDS_URI`, just like abbreviating element names is already allowed. Some elements have only a single attribute with a long name, and I found it really annoying to have to type those out in full 🙂


dennis-adlink (approved):
LGTM, did some basic testing and it works fine

PR #691, opened 2021-02-16
Stressing DATA_AVAILABLE listeners a bit under reader creation/deletion

eboasson:
This adds a test for DATA_AVAILABLE listeners while creating/deleting readers with a writer publishing data as fast as it can. The original reasons for looking into this turned out to be completely different and have already been addressed (007308f230f75192be380f73dd3c62c8b2d5bf5a and 421f41983d5d032bef371663cf4e418599f3eb38), but this test in turn laid bare some other interesting little things:

* A proxy writer's reorder buffer may still contain data when the last reader is deleted. That data wasn't removed (holding on to memory for longer than necessary), and then made available for delivery on receiving the first heartbeat after matching a new reader (which correctly wasn't delivered, but incorrectly leaked away). This PR contains a commit that drops that data on deleting that last reader.
* An assertion in the out-of-sync to in-sync transition of a reader was obviously incorrect, because an out-of-order reader can legitimately be ahead of its proxy writer if the writer is writing quickly and packet loss results in not yet having delivered the last sample written before the writer matched a new reader. It turns out the code handles the case correctly, so it is an only an `assert` that needs to be deleted.
* The HEARTBEAT processing in the presence of out-of-sync readers is relatively expensive, and so a writer that generates HEARTBEATs at a very high rate induces rather a lot of CPU usage in the readers. This stress test happens to hit that path, as the loopback interface is fast enough to really get tons of samples in separate packets per second, each with a heartbeat piggybacked simply because it is in another packet and readers are only allowed to request a retransmit in response to a heartbeat. There's a commit here that limits the heartbeat rate for this reason to 10kHz. That's still rather much; the reasoning is that this means the change won't affect "normal" usage.

The  "marginally improve reporting by discstress writer" is just a tiny fix for the error reporting it does on detecting an issue which it still does (#275).

### DIFF
--- a/src/core/ddsc/tests/CMakeLists.txt
+++ b/src/core/ddsc/tests/CMakeLists.txt
@@ -24,6 +24,7 @@ set(ddsc_test_sources
     "builtin_topics.c"
     "cdr.c"
     "config.c"
+    "data_avail_stress.c"
     "discstress.c"
     "dispose.c"
     "domain.c"
--- a/src/core/ddsc/tests/data_avail_stress.c
+++ b/src/core/ddsc/tests/data_avail_stress.c
@@ -1,0 +1,236 @@
+/*
+ * Copyright(c) 2020 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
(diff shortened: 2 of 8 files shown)


dennis-adlink (commented):
Looks good, I did some additional manual testing on windows and linux and didn't find any issues. A minor issue in discstress is duplicate lines in the output in case of an error, which can be fixed by replacing discstress.c:181 with: 
`for (int i = *logidx; i < LOGDEPTH; i++)`

eboasson:
Thanks @dennis-adlink 

> Looks good, I did some additional manual testing on windows and linux and didn't find any issues. A minor issue in discstress is duplicate lines in the output in case of an error, which can be fixed by replacing discstress.c:181 with:
> `for (int i = *logidx; i < LOGDEPTH; i++)`

Good point, I'll update the PR with that change

PR #693, opened 2021-02-18
Cyclone config tool

thijsmie:
This commit adds a very minimal tool that can print the path to binary, include and library directories for CycloneDDS. When you make cyclone-config available on your path it allows tools that depend on CycloneDDS, such as the Python API, to easily find the libraries and headers. It was inspired by tools like python-config and root-config which perform a similar function.

### DIFF
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 set(CMAKE_INSTALL_TOOLSDIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/tools")
+add_subdirectory(cyclone-config)
 add_subdirectory(pubsub)
 add_subdirectory(ddsls)
 add_subdirectory(ddsconf)
--- a/src/tools/cyclone-config/CMakeLists.txt
+++ b/src/tools/cyclone-config/CMakeLists.txt
@@ -1,0 +1,28 @@
+#
+# Copyright(c) 2021 ADLINK Technology Limited and others
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+# v. 1.0 which is available at
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+
+IF (WIN32)
+  string(REPLACE "/" "\\\\" CYCLONE_CONFIG_PREFIX ${CMAKE_INSTALL_PREFIX})
+ELSE()
+  set(CYCLONE_CONFIG_PREFIX ${CMAKE_INSTALL_PREFIX})
+ENDIF()
+
+configure_file(src/cyclone-config.h.in ${CMAKE_CURRENT_BINARY_DIR}/src/cyclone-config.h)
+add_executable(cyclone-config src/cyclone-config.c)
+set_target_properties(cyclone-config PROPERTIES LINKER_LANGUAGE C)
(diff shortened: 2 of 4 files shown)


k0ekk0ek:
Hi @thijsmie! I'm assuming this is for [cyclonedds-python](https://github.com/eclipse-cyclonedds/cyclonedds-python). A decent Python binding is very welcome! However, I'm a bit reluctant to pull these changes, as I don't understand the problem it solves. I mean, I have an idea, but I'm not sure this actually solves anything. You probably want the Python side to find the library right? If we install, it's all based on a prefix. Let's assume we install to `/foo/bar`. The `bin` directory in that case is `/foo/bar/bin`, the library directory would be `/foo/bar/lib`. On \*NIX platforms you then have to setup the `PATH` environment variable to include `/foo/bar/bin` so that it'll print `/foo/bar/lib` when you ask it where the library is. You could've just set `LD_LIBRARY_PATH` instead(?) And that's only necessary if you install to a custom location because if it's located in the system wide directory, it'll find it without needing anything. The real problem is then probably Windows, where there's no such thing as a default path. However, on Windows, the `dll` is likely located in the `bin` directory as well, so it's already in the `PATH` if the tool is. So far then the tool does not add functionality, right? Apart from that, paths are determined at compile time, but the installation might go to an entirely different place (very common practice when creation packages). At that point the tool will provide the wrong information.

Maybe you can explain the problem you're trying to solve a bit?

thijsmie:
Hi @k0ekk0ek! I realize this indeed might need a bit more of an introduction, so I'll try to reason why I like the idea of a tool.

Firstly, when you type "pip install cyclonedds" there is C code that needs to be compiled. For this access to include directories is needed. We could (reasonably) assume the headers are located at ../include from the library or tool location. This requires me to include this path logic into the installation process since the python part is not in CMAKE so there is no CMAKE_PREFIX_PATH. If I don't want to rely on path logic I can ask the user to set LD_LIBRARY_PATH and LIBRARY_PATH correctly. This can also cause mismatches if the user has multiple Cyclone installations and they are not ordered the same in LD_LIBRARY_PATH and LIBRARY_PATH.

The cyclone-tool gives the user a single source of what Cyclone is going to be used, a single command they can run (and thus mentioned in the installation process) that verifies they are running the correct instance and version of Cyclone. For me personally this has already been helpful: I have both system and user installs of Cyclone and to switch I simply symlink the cyclone-config tool I want in my ~/.bin directory.

This highlights one of the points I am less sure about. Compiling the CMAKE install path into the executable itself makes the tool executable relocatable and the install fixed. You can obviously turn this around and make the tool return paths based on it's own location. That might be preferable with the package creation practice you mentioned.

In short: No, the tool doesn't provide any new information. It just moves the logic of locating Cyclone libraries, binaries and headers out of dependants of Cyclone into Cyclone itself.

k0ekk0ek:
Thanks for the explanation, that's very helpful. `cyclone-config` is not the way to go here, but I understand the problem you're trying to solve. Let me study how pip does things and let's see if we can come up with something we both like. Since you just studied all of this, do you by any chance have a link with pip examples or something?

thijsmie:
There are a lot of things configurable about the python installation process, but here most important is the [C extension](https://docs.python.org/3/distutils/apiref.html#distutils.core.Extension). Secondly, locating the dds library at run-time is also a pain since the python find_library function [does not consult LD_LIBRARY_PATH](https://bugs.python.org/issue2936#msg67505) and [does not return full paths on linux](https://bugs.python.org/issue21042). I can roll my own implementation there but that just comes down to iterating directories and checking if the library exists. I then also have to duplicate that code because I need it both in the library and in the installation process.

eboasson:
> cyclone-config is not the way to go here, but I understand the problem you're trying to solve.

I am not so sure that it is not the way to go, it'd have to be `cyclonedds-config`, but that's not the bit being discussed: and that is that the issue basically exists for any project not using CMake. The way I see it, for those projects, either one expects, e.g., `CYCLONEDDS_HOME` to be set in the environment and bases everything on that, or one uses a tool like the proposed one.

Python extensions are simply one of many cases, and so whether or not a tool like this is a good thing or not cannot be answered by studying those extensions.

PR #698, opened 2021-03-05
Rewrite add_specific_mac fixing Coverity warning

eboasson:
The rewrite is vastly more complicated than necessary to solve CID
325141 (which was a false positive anyway) but addresses a number of
other surprising details in the code.  In particular, enlarging the
buffer before searching for the header/footer submessages means that the
bounds checking during the search was incorrect.

The error checking in the code suggests that the input may be invalid,
but at the same time it the contents are locally generated and can
therefore be trusted.  All of the checks and bailing out on error can
most likely be replaced by assertions, but that requires verifying that
these failures really cannot occur under any circumstances.

### DIFF
--- a/src/security/builtin_plugins/cryptographic/src/crypto_transform.c
+++ b/src/security/builtin_plugins/cryptographic/src/crypto_transform.c
@@ -253,33 +253,6 @@ static Header_t *add_rtps_header (trusted_crypto_buffer_t *buffer, const Header_
   return dst;
 }
 
-static void *trusted_crypto_buffer_find_submessage (const trusted_crypto_buffer_t *buffer, enum SubmessageKind submessageId, size_t offset)
-{
-  assert (offset <= buffer->length);
-  const unsigned char *ptr = buffer->contents + offset;
-  while (sizeof (SubmessageHeader_t) <= (size_t) (buffer->contents + buffer->length - ptr))
-  {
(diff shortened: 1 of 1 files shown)


src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -745,38 +748,45 @@ add_specific_mac(
...
-  struct receiver_specific_mac *rcvmac = &postfix->receiver_specific_macs._buffer[index];
+  // appending may force reallocation
+  trusted_crypto_buffer_append (buffer, sizeof (struct receiver_specific_mac));
+  struct trusted_crypto_footer * const footer = (struct trusted_crypto_footer *) (buffer->contents + footer_offset);
+  const uint32_t index = ddsrt_fromBE4u (footer->postfix.receiver_specific_macs._length);
+
+  // Coverity gets upset by using a byteswapped value without checking it on the assumption that
+  // byteswapping implies external input ...
+  const size_t receiver_specific_macs_offset = CRYPTO_HMAC_SIZE + 4;
+  if (index > (footer->header.octetsToNextHeader - receiver_specific_macs_offset) / sizeof (struct receiver_specific_mac))
```
dennis-adlink: Maybe use `length` instead of `index` here, because this is validating the current list of macs, not if there is room for one more

PR #700, opened 2021-03-06
Add support for Azure Pipelines

k0ekk0ek:
This PR adds build scripting for building on Azure Pipelines. Of course, this is intended to replace Travis CI, which leaves much to be desired. The free plan with Azure offers up-to-date environments for Linux, macOS and Windows and gives users 10 concurrent jobs. Besides providing better build images out-of-the-box, builds are also faster. I've added caching of both Python and Conan packages and switched to using [JFrog ConanCenter](https://conan.io/center/) instead of Bincrafters because Bincrafters is hosted on [JFrog Bintray](https://bintray.com/), which is currently deprecated and will be closed. Also, both packages were obsoleted and did not receive updates. Necessary changes to CMake modules have been done to cope with a "bug" in the CUnit package, which has been reported [here](https://github.com/conan-io/conan-center-index/issues/4789). Another downside is that ConanCenter does not offer 32-bit packages, which is why I've disabled the 32-bit Windows build for now. Simply because building the packages both takes too long (dependencies must be built as well) and failed.

This PR also adds the `-DANALYZER` option which has previously been discussed in #683. In line with the comments on that PR, a Clang build with `scan-build` has been added too and the `USE_SANITIZER` option has been renamed to just `SANITIZER`. I'm sure build times will decrease even further once the Java dependencies have been removed.

Both Codecov and Coverity Scan integrations are working, but I'm hoping that eventually we can include support for reporting both coverage and test results through Azure. Especially keeping a records of test results, like Azure allows you to do, will be a nice addition.

Once this has been merged, we'll have to ask the @eclipsewebmaster to add the necessary pipelines.


eboasson (commented):
Thanks @k0ekk0ek for figuring out all that gobbledygook! I simply assume that it all works 🙂

It LGTM except for the following:
- release builds tend to expose other warnings (especially gcc, but also variables only used if not `NDEBUG`), so I think would be better to replace the gcc-10 debug build by a release build in the PR thing
- building without security, &c., is relatively likely to expose some build problems as well if some major refactoring happens, perhaps it would be an option to add a Linux debug+release builds for that in the PR case? It's not likely to affect one platform differently from the others.

I'm simply worried that PRs could pass the CI but that merging it would then make master fail its builds ... that'd be very unpleasant. Making the above two changes would, IMHO, reduce that risk by quite a bit.

k0ekk0ek:
@eboasson, merged the `push.yml` and `pull-request.yml` into `default.yml`. Cancelled the Travis build, because I very much doubt it'll validate the correctness of Azure Pipelines configurations :sweat_smile: I triggered a run of this on my own repo, you can view the results [here](https://dev.azure.com/k0ekk0ek/cyclonedds/_build/results?buildId=330&view=results) if you want. The *Coverity Scan* and *Codecov* pipelines are a bit harder to trigger, but I've tested them both.

eboasson:
Very good! I presume you have also verified that a test failure results in the run being marked as failed? 🙂

If so, please go ahead and merge it.

k0ekk0ek:
Looking at the results [here](https://dev.azure.com/k0ekk0ek/cyclonedds/_build/results?buildId=331&view=results), I'd say that works as intended :slightly_smiling_face:

eboasson:
> Looking at the results [here](https://dev.azure.com/k0ekk0ek/cyclonedds/_build/results?buildId=331&view=results), I'd say that works as intended 🙂

Indeed! Thanks 🙂

PR #702, opened 2021-03-08
Regenerate sources if single idl in target changes

### LINKED ISSUE #697: Accounting for dependencies in cmake between IDL files included by the #include directive
In the file https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/idlc/cmake/IdlcGenerate.cmake the following is present:
```
    add_custom_command(
      OUTPUT   "${_source}" "${_header}"
      COMMAND  "${IDLC_DIR}/${IDLC}"
      ARGS     ${IDLC_ARGS} ${ABS_FIL}
      DEPENDS  "${ABS_FIL}" idlc
      COMMENT  "Running idlc on ${FIL}"
      VERBATIM)
```
As a result, recompilation does not take into account the modification dates of the included IDL files. The reassembly may not be complete. How to fix it?

k0ekk0ek:
Fixes #697.

### DIFF
--- a/src/idlc/cmake/IdlcGenerate.cmake
+++ b/src/idlc/cmake/IdlcGenerate.cmake
@@ -47,6 +47,12 @@ function(IDLC_GENERATE _target)
      set(IDLC_ARGS)
   endif()
 
+  set(_files)
+  foreach(FIL ${ARGN})
+    get_filename_component(ABS_FIL ${FIL} ABSOLUTE)
+    list(APPEND _files ${ABS_FIL})
+  endforeach()
+
   set(_dir "${CMAKE_CURRENT_BINARY_DIR}")
   set(_sources)
   set(_headers)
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Doesn't `IMPLICIT_DEPENDS` work like ordinary `DEPENDS` for non-Makefile generators? In that case, I'd say it sounds like a further improvement to use that instead.

Longer-term, with the new IDL compiler, we'll be able to do it properly.

k0ekk0ek:
Documentation says `IMPLICIT_DEPENDS` is ignored by generators other than Makefile generators. Reason I did it like this is to have the exact same behavior everywhere and avoid unknowingly breaking things on macOS and Windows. I'd be fine with adding it if you prefer. The only way it'd currently break would be if someone includes an idl file that's not in the target.

eboasson:
> Documentation says `IMPLICIT_DEPENDS` is ignored by generators other than Makefile generators.

I should have known, it's CMake after all ... that means `IMPLICIT_DEPENDS` becomes a mess. Let's just use `DEPENDS` for now, and put some effort into getting it right with the new compiler as soon as possible.

k0ekk0ek:
Thanks @eboasson!

PR #703, opened 2021-03-08
Allow users to build without IDL compiler

k0ekk0ek:
Make IDL compiler optional again to save time compiling on ROS buildfarm.


eboasson (commented):
I am fine with making it optional again, not because I agree that it might take a meaningful amount of time during the build or that `bison` would be a problematic dependency, but because it doesn't cost us anything and I don't want to delay merging the compiler.

What I would like you to double check is whether these changes are really all that are needed. It is quite possible, you did some spring cleaning and so comparing with the old set of conditions involving `BUILD_IDLC` may not be relevant. But if you could just double-check it ...

Other than that it is approved.

k0ekk0ek:
Made inclusion of `Generate.cmake` depend on whether or not there's a `CycloneDDS::idlc` target. PR didn't build because of Travis macOS credits being depleted again, but I'm confident it works the same on all platforms.

PR #705, opened 2021-03-09
Various fixes triggered by static analysis

eboasson:
All pretty trivial, except the refactoring in the access_control plugin.

### DIFF
--- a/src/core/ddsc/tests/typelookup.c
+++ b/src/core/ddsc/tests/typelookup.c
@@ -164,6 +164,8 @@ static endpoint_info_t * find_typeid_match (dds_entity_t participant, dds_entity
   return result;
 }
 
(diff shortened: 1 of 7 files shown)


k0ekk0ek:
@eboasson, this looks good to me. I have not done a test run, but I have read it thoroughly. From what I can tell, the outcome of the checks should be the same. I really like the refactor of the access plugin. One can actually tell what it's supposed to do in your version :slightly_smiling_face: If Travis thinks it's ok too, feel free to merge.

eboasson:
Curious why it failed, I guess it has something to do with the reason #707 is needed. Anyway: https://travis-ci.org/github/eboasson/cyclonedds/builds/761615538 is a CI run for this particular commit.

I'd prefer to see #707 sorted before merging more stuff, so I'll wait a little while longer.

PR #706, opened 2021-03-09
Access control deny rules

eboasson:
It turns out that there are two separate issues with the handling of "deny" rules in the access control plugin, which I found while doing #705 (it should come as no surprise that this PR sits on top of that one, but it is too important a detail to be hidden in some other PR).

The first is that the access control plugin implemented the wrong test: if a deny rule says partition "A" is denied than any reader/writer that includes "A" in its set of partition is to be denied (as per common sense and per DDS Security 1.1, section 9.4.1.3.2.3.2.4). Instead of implementing this, it used the same test as for allow rules, but for those, the test is whether *all* partitions are allowed.

The second is that the spec is totally broken on the subject of allow & deny rules, except in the case where wildcards are avoided like the plague, where allow = "*" and nothing is denied, or some other cases that I can't be bothered to come up with. The reason is that the spec says the matching must be done using `fnmatch` as defined in POSIX 1003.2-1992, Section B.6 (naturally it omits listing the options to be used ...). But that means if, e.g., allow = "?" (i.e. any single character name), a partition of "*" is allowed, but "*" is empathically not a single-character partition name.

That however will have to be the subject of some other PR. This one simply notes the problem in a comment and implements the specified behaviour.

### DIFF
--- a/src/security/builtin_plugins/access_control/src/access_control.c
+++ b/src/security/builtin_plugins/access_control/src/access_control.c
@@ -1194,7 +1194,7 @@ is_partition_in_criteria(
     current_partition = current_partitions->partition;
     while (current_partition != NULL)
     {
-      if (ac_fnmatch(current_partition->value, partition_name))
+      if (ac_fnmatch(current_partition->value ? current_partition->value : "", partition_name))
         return true;
       current_partition = (struct string_value *)current_partition->node.next;
     }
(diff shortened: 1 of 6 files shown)


dennis-adlink (approved):
Good catch, the wrong rules for testing deny rules! I've looked through the changes and don't see any issues. When running the partition test you added, in ran into an issue with incorrect thread states when deleting domains. I will make a separate PR for the fix.

PR #708, opened 2021-03-11
Fix for thread_state issue

dennis-adlink:
This PR adds a fix for an issue with thread_state, which was found when testing the changes for PR #706 (the issue is not related to the changes in that PR). See the commit message for details on the issue and the fix. 

Thanks @eboasson for helping me analyze the issue!


eboasson (approved):
Thanks for fixing this @dennis-adlink !

PR #715, opened 2021-03-16
Implement support for @value and fix evaluation of duplicate annotations

k0ekk0ek:
Implement support for `@value` annotation (handling of `any` type was somewhat problematic) and fix evaluation of duplicate annotations that don't have parameters, which would lead to a segmentation fault.


k0ekk0ek:
Build results [here](https://dev.azure.com/k0ekk0ek/cyclonedds/_build/results?buildId=352&view=results).

k0ekk0ek:
Added figuring out what the default discriminator is when finalizing a union, which is useful when writing backends for modern bindings like C++. I still want to add a couple of tests for unions with a non-enum discriminator (which I'll do tomorrow), but the overall logic is ready for review. @e-hndrks can you spare some time to look over my changes?

src/idl/src/annotation.c
```diff
@@ -132,27 +133,51 @@ annotate_autoid(
...
+    idl_error(pstate, idl_location(annotation_appl),
+      "@value cannot be applied to '%s' elements", idl_construct(node));
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
+  type = idl_type(const_expr);
+  if (type == IDL_OCTET || (type & IDL_INTEGER_TYPE)) {
+    idl_intval_t intval = idl_intval(const_expr);
+
+    if ((intval.type & IDL_UNSIGNED) && intval.value.ullng > INT32_MAX) {
```
e-hndrks, line 160: Line 160: Shouldn't the check be that intval.value.llng > UINT32_MAX?

e-hndrks (changes requested):

e-hndrks (approved):

src/idl/src/annotation.c
k0ekk0ek: Good catch.

k0ekk0ek: It's been resolved in the latest version. Thanks @e-hndrks!

k0ekk0ek:
Switch to Azure made getting the checks green a bit more complicated, hence the force pushes. @eboasson, @e-hndrks was kind enough to review the changes on this PR. If you're ok with it too, I'd like to proceed and merge these changes.

eboasson:
@k0ekk0ek, sure, please go ahead with that.

PR #718, opened 2021-03-18
Add xcheck config option for xevent checks

dennis-adlink:
Analysing test runs for topic discovery with a profiler revealed that asserts on the check nontimed_xevent_in_queue take a significant amount of time in these runs. This commit makes this specific check an 'expensive check', and adds a new category 'xevent' in the xchecks configuration option so that it can be explicitly enabled. In CI these checks are disabled.


k0ekk0ek:
The fact that *Codecov* and *Coverity Scan* do not pass has to with the fact that `pr: none` is not set in those pipelines. The other builds (the ones that are used to verify PRs actually run) all pass. While I haven't run it with a profiler myself, it does seems to shave some time off even from Azure build times. Thanks @dennis-adlink!

k0ekk0ek (approved):

PR #724, opened 2021-03-23
Fix handling of (bounded) strings

k0ekk0ek:
That didn't take long... A colleague trying out the new compiler found this. This fixes support for typedefs of strings and handling of bounded strings.


eboasson (approved):
LGTM, but I don't know the innards well enough to make that a very meaningful comment.

Issue #619, opened 2020-10-15
add public license badge(s)


joespeed:
please add public license badge(s) a'la Zenoh https://github.com/eclipse-zenoh/zenoh 
<img width="542" alt="image" src="https://user-images.githubusercontent.com/8238588/96058102-4ea38580-0e3f-11eb-8d95-2be5d2fa2f50.png">

PR #727, opened 2021-03-23
Fix Coverity Scan job and use Azure Pipelines badge

k0ekk0ek:
Title says it all. I removed the platform variable because it interfered with the builds in a previous commit, but forgot about the use of it to download and install Coverity. Also, this replaces the Travis CI badge with an Azure Pipelines one.


k0ekk0ek:
I threw in a couple of license badges too (as per request #619)

eboasson (approved):

PR #731, opened 2021-03-25
Add message authentication/encryption to perftest

eboasson:
What it says on the tin. Running with "perftest -S sign" enables RTPS message authentication, running with "-S encrypt" additionally adds payload encryption.

(This builds on #730)


dennis-adlink (approved):
LGTM

PR #733, opened 2021-03-26
Fix use of idlc in cross compilation scenarios

k0ekk0ek:
Like the title.


eboasson (approved):
It looks fine to me, but I have no experience doing cross builds with CMake myself yet, so take that for what it's worth.

The failing test would appear to be unrelated.

PR #740, opened 2021-04-01
Iceoryx logging and resource cleanup

MatthiasKillat:
Fixes the problem that a reader could be deleted while still being used in a shared memory monitoring callback.

Adapts iceoryx listener usage to a small iceoryx API change.

Reintroduces setting the shared memory logging capability.

Prepare restructuring of shared memory init code but this will be postponed postponed until the conditional use of shared memory (depending on QoS and types) has been added. Currently it just isolates the logging translation logic for iceoryx.

### DIFF
--- a/docs/manual/shared_memory.rst
+++ b/docs/manual/shared_memory.rst
@@ -63,8 +63,7 @@ Save the following xml as cyclonedds.xml.
 SubQueueCapacity, SubHistoryRequest and PubHistoryCapacity can be optionally set if Shared Memory is enabled.
 SubQueueCapacity controls how many samples a reader using shared memory can hold before the least recent is discarded.
 PubHistoryCapacity defines how many samples a shared memory writer will keep to send to late-joining subscribers.
-SubHistoryRequest is the number of samples a late-joining reader will request from a writer (this can be at most 
-as many as were send and at most PubHistoryCapacity).
+SubHistoryRequest is the number of samples a late-joining reader will request from a writer (this can only be as many as were sent and at most PubHistoryCapacity).
 
 The Loglevel controls the output of the iceoryx runtime and can be set to:
 verbose, debug, info, warn, error, fatal and off
--- a/docs/manual/shared_memory.rst
+++ b/docs/manual/shared_memory.rst
@@ -55,6 +55,7 @@ Save the following xml as cyclonedds.xml.
               <SubQueueCapacity>256</SubQueueCapacity>
               <SubHistoryRequest>16</SubHistoryRequest>
               <PubHistoryCapacity>16</PubHistoryCapacity>
+              <LogLevel>info</LogLevel>
           </SharedMemory>
       </Domain>
   </CycloneDDS>
@@ -63,7 +64,11 @@ SubQueueCapacity, SubHistoryRequest and PubHistoryCapacity can be optionally set
 SubQueueCapacity controls how many samples a reader using shared memory can hold before the least recent is discarded.
 PubHistoryCapacity defines how many samples a shared memory writer will keep to send to late-joining subscribers.
 SubHistoryRequest is the number of samples a late-joining reader will request from a writer (this can be at most 
-as many as were send and at most PubHistoryCapacity).   
+as many as were send and at most PubHistoryCapacity).
+
+The Loglevel controls the output of the iceoryx runtime and can be set to:
+verbose, debug, info, warn, error, fatal and off
+in order of decreasing output level. It is set to info by default. 
 
 Now we start to run Cyclone DDS with shared memory.
 In the 1st terminal we will start RouDi.
--- a/src/core/ddsc/src/shm_monitor.c
+++ b/src/core/ddsc/src/shm_monitor.c
@@ -44,12 +44,12 @@ void shm_monitor_init(shm_monitor_t* monitor)
 
 void shm_monitor_destroy(shm_monitor_t* monitor) 
 {
-    //note: we must ensure no readers are actively using the monitor anymore,    
     shm_monitor_wake_and_disable(monitor);
-    while(monitor->m_state == SHM_MONITOR_RUNNING); //spin until callbacks are no longer processed
+    // waiting for the readers to be detaced is not necessary, they will be detached when the listener is destroyed
+    // but if we for some reason need this we could do it like this 
+    // while(monitor->m_state == SHM_MONITOR_RUNNING || monitor->m_number_of_attached_readers > 0);
+    while(monitor->m_state == SHM_MONITOR_RUNNING);
 
-    //ICEORYX_TODO: is it ok to deinit while readers are still attached?
-    //              they should be automatically detached
     iox_listener_deinit(monitor->m_listener);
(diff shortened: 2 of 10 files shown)


docs/manual/shared_memory.rst
```diff
@@ -55,6 +55,7 @@ Save the following xml as cyclonedds.xml.
               <SubQueueCapacity>256</SubQueueCapacity>
```
reicheratwork, line 55: I think we should add in this documentation which settings have default values and what these are (if any).

For instance, the setting SubHistoryRequest will fall back to the value 16 if nothing is set in the configuration file.

src/core/ddsc/src/shm_monitor.c
```diff
@@ -80,7 +80,7 @@ dds_return_t shm_monitor_wake_and_enable(shm_monitor_t* monitor)
 dds_return_t shm_monitor_attach_reader(shm_monitor_t* monitor, struct dds_reader* reader) 
 {
 
-    if(iox_listener_attach_subscriber_event(monitor->m_listener, reader->m_iox_sub, SubscriberEvent_HAS_DATA, shm_subscriber_callback) != ListenerResult_SUCCESS) {
```
reicheratwork, line 83: Seems like the name of this enum value has changed.

We will need to keep track of which commit of iceoryx this iteration is compiled against.
Until a tag on both iceoryx and cyclonedds is created linking the two together, this will remain uncertain.

docs/manual/shared_memory.rst
```diff
@@ -63,7 +64,11 @@ SubQueueCapacity, SubHistoryRequest and PubHistoryCapacity can be optionally set
 SubQueueCapacity controls how many samples a reader using shared memory can hold before the least recent is discarded.
 PubHistoryCapacity defines how many samples a shared memory writer will keep to send to late-joining subscribers.
 SubHistoryRequest is the number of samples a late-joining reader will request from a writer (this can be at most 
-as many as were send and at most PubHistoryCapacity).   
+as many as were send and at most PubHistoryCapacity).
```
reicheratwork, line 67: grammar correction: perfect tense of "to send" is "sent", change

src/core/ddsc/src/shm_monitor.c
```diff
@@ -44,12 +44,12 @@ void shm_monitor_init(shm_monitor_t* monitor)
 
 void shm_monitor_destroy(shm_monitor_t* monitor) 
 {
-    //note: we must ensure no readers are actively using the monitor anymore,    
     shm_monitor_wake_and_disable(monitor);
-    while(monitor->m_state == SHM_MONITOR_RUNNING); //spin until callbacks are no longer processed
+    // waiting for the readers to be detaced is not necessary, they will be detached when the listener is destroyed
+    // but if we for some reason need this we could do it like this 
+    // while(monitor->m_state == SHM_MONITOR_RUNNING || monitor->m_number_of_attached_readers > 0);
+    while(monitor->m_state == SHM_MONITOR_RUNNING);
```
reicheratwork, line 51: This will cause this thread to spin at 100% CPU usage, is this required due to timing?
Maybe adding even a small (1 millisecond) sleep will reduce the cpu overhead by a large amount, while still keeping a quick cleanup.

reicheratwork (changes requested):
Looks okay, but the reasoning behind some implementations is not 100% clear to me, mainly the spin wait in shm_monitor_destroy.
Also some things will need further explanation or documentation, like the configuration settings and keeping track which commits of iceoryx and cyclone belong together.

docs/manual/shared_memory.rst
MatthiasKillat: True, I set the current maximum value for all of them explicitly. Especially for queue size this makes sense I think to reduce the chance of data loss. Note that the majority of shared memory is only used if the queue actually contains data, since the queue itself just stores pointer equivalents.

We are currently limited by some hardcoded values of iceoryx. This will likely be improved. Documentation will be a separate PR.

src/core/ddsc/src/shm_monitor.c
MatthiasKillat: It has changed and I needed it to build with the latest master. We had no release tag at this point but in the release the changed enum will be required.

docs/manual/shared_memory.rst
MatthiasKillat: You are right, will change.

src/core/ddsc/src/shm_monitor.c
MatthiasKillat: I will look into it, it should suffice to wait for the thread to join.

src/core/ddsi/include/dds/ddsi/shm_init.h
```diff
@@ -0,0 +1,31 @@
+/*
```
MatthiasKillat, line 1: Note that the file might get a more general name later. The idea is to collect everything needed for setting up and shutting down shared memory support here (i.e. init and destruction) as well as some helper methods. Monitoring shared memory activity (i.e. running while using it) is done in sh,_monitor.h, however.

docs/manual/shared_memory.rst
MatthiasKillat: Will be done in a documentation PR. Probably best once the features are complete.

src/core/ddsc/src/shm_monitor.c
```diff
@@ -80,7 +80,7 @@ dds_return_t shm_monitor_wake_and_enable(shm_monitor_t* monitor)
 dds_return_t shm_monitor_attach_reader(shm_monitor_t* monitor, struct dds_reader* reader) 
 {
 
-    if(iox_listener_attach_subscriber_event(monitor->m_listener, reader->m_iox_sub, SubscriberEvent_HAS_DATA, shm_subscriber_callback) != ListenerResult_SUCCESS) {
+    if(iox_listener_attach_subscriber_event(monitor->m_listener, reader->m_iox_sub, SubscriberState_HAS_DATA, shm_subscriber_callback) != ListenerResult_SUCCESS) {
```
eboasson, line 83: This looks weird to me: `iox_listener_attach_subscriber_event` and `SubscriberState_HAS_DATA`. Looking at the code and the enums, I would have expected `SubscriberEvent_DATA_RECEIVED` instead.

MatthiasKillat: Right, corrected via changing last commit and force-push.

src/core/xtests/rhc_torture/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 idlc_generate(RhcTypes RhcTypes.idl)
+include(${CMAKE_CURRENT_LIST_DIR}/../../../../cmake/Modules/CUnit.cmake)
```
k0ekk0ek, line 13: This probably shouldn't be here. I'm guessing this is required for setting the path property for the test, but pulling in the entire `CUnit.cmake` module seems like overkill.

PR #744, opened 2021-04-06
Replace type identifier field by pointer to type lookup meta data

dennis-adlink:
Replace the type identifier field in (proxy) endpoints and topic definitions by a pointer to the type-lookup meta data record. This change is to prepare for the implementation of the XTypes type system, in which a type identifier can't be used as an identifier in certain cases, because a minimal and complete version of the type identifier exists.


eboasson:
@dennis-adlink this seems to be at least partly irrelevant in the light of #904. As it appears to be an optimisation, it makes little sense to merge it now with if it becomes irrelevant on merging #904. (I certainly could've been more responsive on this PR, I'm sorry about that ...)

dennis-adlink:
You're right, and new insights made that 904 replaces all changes in this PR and make it no longer relevant, so I'll close this.

PR #747, opened 2021-04-06
Added new API to facilitate a publisher loan.

e-hndrks:


dejanpan:
@eboasson was this PR also included in https://github.com/eclipse-cyclonedds/cyclonedds/pull/751?

eboasson:
@dejanpan not yet, @e-hndrks was at that point still going to make a few changes and the impact of this on the rest should be minimal.

eboasson (approved):
Looks fine to me, a simple sanity comes out well and I'm not worried about the macOS builds because that, firstly, looks more like a problem at Azure than a problem with our stuff; and, secondly, it builds fine for me locally on macOS.

PR #748, opened 2021-04-06
Update writecdr for iceoryx

sumanth-nirmal:
This MR adds the following:

1. Updates `dds_writecdr_impl` to work with iceoryx if SHM is enabled
2. Fix compiler warning when setting ddsi_sertype init flags

### DIFF
--- a/src/core/ddsc/src/dds_write.c
+++ b/src/core/ddsc/src/dds_write.c
@@ -63,7 +63,7 @@ dds_return_t dds_writecdr (dds_entity_t writer, struct ddsi_serdata *serdata)
   }
   serdata->statusinfo = 0;
   serdata->timestamp.v = dds_time ();
-  ret = dds_writecdr_impl (wr->m_wr, wr->m_xp, serdata, !wr->whc_batch);
+  ret = dds_writecdr_impl (wr->m_wr, wr->m_xp, serdata, !wr->whc_batch, wr);
   dds_writer_unlock (wr);
   return ret;
 }
@@ -83,7 +83,7 @@ dds_return_t dds_forwardcdr (dds_entity_t writer, struct ddsi_serdata *serdata)
     dds_writer_unlock (wr);
     return DDS_RETCODE_ERROR;
   }
-  ret = dds_writecdr_impl (wr->m_wr, wr->m_xp, serdata, !wr->whc_batch);
+  ret = dds_writecdr_impl (wr->m_wr, wr->m_xp, serdata, !wr->whc_batch, wr);
   dds_writer_unlock (wr);
   return ret;
 }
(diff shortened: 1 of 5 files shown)


src/core/ddsc/src/dds_write.c
```diff
@@ -335,6 +335,21 @@ dds_return_t dds_writecdr_impl (struct writer *ddsi_wr, struct nn_xpack *xp, str
...
+  if (wr) {
+    if ((wr->m_entity.m_domain->gv.config.enable_shm && iox_pub_has_subscribers(wr->m_iox_pub)) &&
+        (dinp->iox_chunk != NULL))
+    {
+      iceoryx_header_t * ice_hdr = dinp->iox_chunk;
+      ice_hdr->guid = ddsi_wr->e.guid;
+      ice_hdr->tstamp = dinp->timestamp.v;
+      ice_hdr->data_kind = (unsigned char)dinp->kind;
+      ddsi_serdata_get_keyhash (dinp, &ice_hdr->keyhash, false);
+      iox_pub_publish_chunk (wr->m_iox_pub, ice_hdr);
```
eboasson, line 348: This, I believe, causes the Iceoryx to take over ownership of the chunk, and so it may be freed from this point onward. That makes it problematic to push the serdata into the WHC and to perform local delivery.

IINM the local readers will get it from Iceoryx instead of via Cyclone's own short-circuiting, so the problem is solvable by looking ahead whether there is a need to insert the data in the WHC and to serialize the data if there is/if there are remote readers.

Iceoryx is only used for volatile data (at least for the time being), and so doing `write_sample_gc` prior to call `iox_pub_publish_chunk` should force serialization if there are readers. So if the corresponding `to_ser`/`to_ser_ref` call caches the serialized representation it should be safe.

eboasson (commented):
Thanks @sumanth-nirmal! It looks fine for the most part, but I think there's a race condition with Iceoryx releasing the memory while it may still be in use by Cyclone.

As I've already started doing stupid things in merging, I'm not going to do anything else right now ... https://github.com/eboasson/cyclonedds/tree/iceoryx-with-master should have everything included (also this PR) except for the multi-homing (there's an "interesting" detail there in how to treat the Iceoryx locator, because it doesn't have an associated network transport).

Moreover, that branch survives CI https://dev.azure.com/erikboasson/Cyclone%20DDS/_build/results?buildId=28&view=results (that is without Iceoryx) and it builds locally with Iceoryx. There's a chance it works ... a ringing endorsement I dare say. You might want to check it out.

src/core/ddsc/src/dds_write.c
sumanth-nirmal: @eboasson Ah, correct the ownership of the chunk is transferred to Iceoryx after `iox_pub_publish_chunk`, also I don't think the local readers will be updated by Iceoryx (at least I have not been able to see this behavior in the code). I am not sure now how to overcome this problem of updating local readers without any copy!

eboasson:
Included in #751.

PR #749, opened 2021-04-06
Fix for discstress test failure

dpotman:
These changes fix the issue that causes the discstress test to fail (issue #275), particularly when running this test with some packet loss and using an increased writer linger duration. As I'm not sure of the side effects this fix may have, I've put this PR in draft state. It needs some more research (and tests) to find out the consequences of the changed behavior with respect to sending and handling heartbeats and gaps, e.g. because a reader can get ahead of its proxy writer, as mentioned in issue #275. 

The changes in the way heartbeats and gaps are send and handled by the proxy writer and reader:
- A reader is only set in-sync after it has received a directed heartbeat.
- Readers keep sending pre-emptive nacks until a directed heartbeat for the match with the writer is received, so that the reader will properly set its initial sequence number, which can be higher than the proxy writer's current sequence number in case of a volatile reader that joined later.
- The writer sends directed heartbeats with a seqno of at least 1 + the seqno at the moment of matching the (volatile) reader.
- Handle a gap message only for the targeted reader if the reader is not in sync. If the reader is in sync, also apply the gap to the
  proxy writer's reorder buffer

### DIFF
--- a/src/core/ddsc/tests/CMakeLists.txt
+++ b/src/core/ddsc/tests/CMakeLists.txt
@@ -33,6 +33,7 @@ set(ddsc_test_sources
     "entity_hierarchy.c"
     "entity_status.c"
     "err.c"
+    "fastpath.c"
     "filter.c"
     "instance_get_key.c"
     "instance_handle.c"
--- a/src/core/ddsc/tests/fastpath.c
+++ b/src/core/ddsc/tests/fastpath.c
@@ -1,0 +1,245 @@
+/*
+ * Copyright(c) 2006 to 2018 ADLINK Technology Limited and others
+ *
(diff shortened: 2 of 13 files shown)


thijsmie:
Hey @dpotman,

This PR is now quite out of date and in draft for more than a year. Is it still relevant?

dpotman:
Yes, it is still relevant in the sense that the issue still exists and causes CI builds to fail every now and then, and the changes from this PR make the situation better (at least it gives less test failures on my dev machine). But more work needs to be done to improve and validate the fix. I don't expect this to be done in short term, so I'll close this PR for now and we can re-open it once the issue gets priority.

PR #751, opened 2021-04-07
Combine most things in iceoryx branch

eboasson:
This merges `master` into `iceoryx` and also integrates with #748 and #739. I intend to push it onto the branch shortly rather than creating yet another merge commit, but I figure having a PR makes it easier to track things.

Various changes have been made in addition:

- local subscriptions now get the data via Iceoryx
- `writecdr_impl` delays the publishing via Iceoryx until after publishing on the network, this gives the code a chance to generate a serialized representation (if needed); taken with the previous point, this means that Iceoryx can safely take ownership of the chunk
- `writecdr_impl` now always consumes the Iceoryx chunk when it returns OK, rather than doing so only if there are subscribers in Iceoryx
- the Iceoryx dependency is now private, after changing a few pointers in `ddsi_serdata.h` to `void *`
- QoS checks for Iceoryx compatibility now really on the fully initialized QoS object after inheriting from topic/default (so it actually works)
- it adds a CI build on Linux that includes Iceoryx support (it doesn't enable it, so it merely validates that it builds)
- it adds actually building the Hello World example to the CI, rather than limiting it to running CMake; this also fixes some typos in setting sanitizer options in that case
- it fixes build issues on Windows
- it removes the sleeps added to a number of test cases to work around a race condition with Iceoryx on deleting a reader
- it asserts that the async send queue is empty only after joining the async send thread (I think having those two lines in the wrong order was a really old bug)
- it adds a typedef keyword for the enum definition for the Iceoryx memory allocation wrapper, which avoids defining a global variable that was never meant to exist

I guess I'm probably forgetting something, but that's the bulk of it. Some rudimentary tests (on macOS, no less!) suggest that it works, but it really needs some TLC still.

Multi-homing support isn't integrated yet.


eboasson:
Note: the ECA check fails on known commits that were merged a long time ago after having established that the author really does have a valid ECA on file. They have been re-checked a number of times, so that's not going to be blocking issue.

eboasson:
Note 2: it seems rather unlikely the one failing test has anything to do with this.

PR #752, opened 2021-04-07
Add is_loan_available method to dds_data_allocator

sumanth-nirmal:


src/core/ddsc/src/dds_data_allocator.c
```diff
@@ -126,3 +126,30 @@ void dds_data_allocator_free (dds_data_allocator_t *data_allocator, void *ptr)
...
+#ifdef DDS_HAS_SHM
+  dds_entity * e;
+
+  if (DDS_RETCODE_OK != dds_entity_pin(entity, &e)) {
+    return ret;
+  }
+
+  switch (dds_entity_kind(e)) {
+    case DDS_KIND_READER:
+      ret = ((struct dds_reader *) e)->m_entity.m_domain->gv.config.enable_shm;
```
eboasson, line 142: Actually, it would be better to do a different check: whether `((struct dds_reader *) e)->->m_iox_sub` is not a null pointer (and the same for the writer). It used to be this test, but @reicheratwork changes to make it dependent on the QoS did change that. Looking at this, I realize there are now a number of places where the change still needs to be made:

* `dds_write_impl` (though it is conceivable that @e-hndrks's PR affects that)
* `dds__reader_data_allocator_init` and `dds__writer_data_allocator_fini`

eboasson (commented):
Apart from one issue — and one that is not exclusive to this PR, but that does matter — it looks good to me. CI runs are looking good, too.

Would you have time/be willing to fix this one issue in the various places? I think it makes sense to do it as this PR, seeing that it is all very closely related.

src/core/ddsc/src/dds_data_allocator.c
sumanth-nirmal: @eboasson Will fix this and other instances

eboasson:
Thanks @sumanth-nirmal!

It looks like the macOS builds on Azure have developed a problem, which I assume to be a problem on their side because they just stay in a "queued" state forever. (@e-hndrks ran into the same issue). Perfect timing, as it were. Fortunately, I have a mac and so I can confirm it builds fine there as well, as was to be expected.

PR #753, opened 2021-04-07
Fix for unittests and removed unnecessary code

reicheratwork:
Fixed unitests for dispose/(un)register by disabling shared memory for these tests.
Removed unnecessary shared memory global mutex related code

### DIFF
--- a/src/core/ddsc/tests/data_avail_stress.c
+++ b/src/core/ddsc/tests/data_avail_stress.c
@@ -88,7 +88,23 @@ static void setup (bool remote, struct writethread_arg *wrarg)
 {
   // Creating/deleting readers while writing becomes interesting
   // only once the network is lossy.
+#ifdef DDS_HAS_SHM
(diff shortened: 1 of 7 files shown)


k0ekk0ek:
Thanks @reicheratwork! Not sure if all of this still applies to master, the logic for shm seems ok (even though I haven't given it a proper look), but I'm not sure we should disable testing the regular path if someone compiles with support for Iceoryx?

eboasson:
@k0ekk0ek I don't think it is wise to disable shm usage in some tests like this. If the test doesn't make sense with shm enable it should be skipped, and else I think it is better to have it fail than to test something else than one would expect. It looks to me like these tests are all of the second category.

The second commit I like very much and that one I'll merge. I'll make a new PR with that one commit rebased on master, merge that new one, and close this PR.

eboasson:
Via #856.

Issue #741, opened 2021-04-01
CMake Windows warning CMP0115


emersonknapp:
This is making the ROS2 buildfarm go yellow on the nightlies - see for example https://ci.ros2.org/view/nightly/job/nightly_win_deb/1943/consoleText

The warning output is as follows

```
CMake Warning (dev) at src/idl/CMakeLists.txt:41 (add_library):
  Policy CMP0115 is not set: Source file extensions must be explicit.  Run
  "cmake --help-policy CMP0115" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  File:

    C:/ci/ws/src/eclipse-cyclonedds/cyclonedds/src/idl/src/hashid.c.in
This warning is for project developers.  Use -Wno-dev to suppress it.
```

eboasson:
@k0ekk0ek just from the name of the file involved I think this relates to #735. Are you ok with me leaving it for you to look into in more detail?

k0ekk0ek:
@eboasson, absolutely, but I missed the notification. I'll have a go at it.

k0ekk0ek:
I think this particular issue is fixed by #737. Looking at the description provided in #756, I think CMake previously searched for `hashid.c` and found `hashid.c.in`. The error of course being it should use the generated version. I've tried to reproduce locally with CMake 3.20, but the warning didn't occur. The warning also isn't in the latest ROS builds, so I think we can safely close this. @emersonknapp, do you agree?

emersonknapp:
Yep, this was opened to make ROS builds green - I don't have any particular input beyond that :)

k0ekk0ek:
Closing this issue. Thanks for the feedback.

PR #737, opened 2021-03-31
Ignore unsupported pragmas

k0ekk0ek:
Like the title. Includes fix in #735 by @clalancette to speedup the process. Happy to rebase after his change has been merged, or alternatively, if he's OK with it we can just merge all of this PR.

PR #756, opened 2021-04-08
Declare hashid.c file as GENERATED to solve cmake warning

j-rivero:
CMake does not consider the configured file as a file with extension in newer versions. While [upstream fix it](https://gitlab.kitware.com/cmake/cmake/-/issues/21993), I think there is no hurt on declaring it as GENERATED.

Tested: [Build Status](https://ci.ros2.org/job/ci_windows/14407/)

close: #741

### DIFF
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -37,6 +37,8 @@ foreach(func ddsrt_md5_init ddsrt_md5_append ddsrt_md5_finish)
 endforeach()
 
 configure_file(src/hashid.c.in ${CMAKE_CURRENT_BINARY_DIR}/hashid.c @ONLY)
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Thanks @j-rivero. However, I believe this issue is not currently not present anymore. Like I mentioned on #741 and #737, I believe the error came from the fact that I forgot to remove the `src/hashid.c` entry from the list of sources (I just copied it before). Are you still experiencing the issue with the current HEAD of master?

j-rivero:
Thanks Jeroen:

> Are you still experiencing the issue with the current HEAD of master?

I just checked the nightly series and seems like the error is gone [since April 8th](https://ci.ros2.org/view/nightly/job/nightly_win_deb/1951/). So you should be right, seems not a problem right now. Closing this.

PR #757, opened 2021-04-09
Add statusinfo to iceoryx header

eboasson:
Without this, disposes/unregisters will be misinterpreted as writes.

@e-hndrks or @reicheratwork, care to review?


e-hndrks (approved):
It looks all good to me. No further comments.

PR #758, opened 2021-04-09
Build src/core/xtests only if BUILD_TESTING

eboasson:
This removes the hard dependency on CUnit that (I think) was introduced to work around a "rpath" issue in the Iceoryx libraries.


clalancette:
CI seems happy with this: https://github.com/ros2/rmw_cyclonedds/pull/300#issuecomment-816759045 .  I also verified locally that if I remove `libcunit1-dev`, it compiles now.

PR #759, opened 2021-04-09
Fix for mixing shared memory and network readers

reicheratwork:
Mixing iceoryx enabled readers with non-enabled readers caused
the writer to fall back to using the UDP broadcast, even though
it should use iceoryx for the readers that support it.
This is fixed by removing the process-id from the iceoryx locator,
which made each process's iceoryx instance unique, which it isn't.
Also the iceoryx locator is added to the list of multicast locators,
which also necessitated restructuring this function a little.

### DIFF
--- a/src/core/ddsi/src/q_ddsi_discovery.c
+++ b/src/core/ddsi/src/q_ddsi_discovery.c
@@ -897,20 +897,31 @@ static void add_locator_to_ps (const ddsi_locator_t *loc, void *varg)
 }
 
 #ifdef DDS_HAS_SHM
-static void add_iox_locator_to_ps(const ddsi_locator_t* loc, struct add_locator_to_ps_arg *arg)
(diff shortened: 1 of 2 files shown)


eboasson:
Commit da75c2fd188c4fa19b07e91185d9fb23bfd84c5e (among other things) prioritizes Iceoryx locators and essentially eliminates the distinction between uni- and multicast locators in all the processing (other than that it disallows multicast addresses when AllowMulticast disallows them). That commit has been rebased many times but it has retained its original commit date, which is a bit misleading by now. Suffice to say that it was updated more recently than August 2020, and that the rebasing to the version that integrates with Iceoryx introduced this prioritization.

Furthermore, Iceoryx isn't really unicast or multicast, it is proper pub/sub messaging. In this context, it makes more sense to treat it as a unicast locator than as a multicast locator. It would be safe to have it twice, it just wouldn't have any effect and waste some bandwidth during discovery.

So I'm closing this PR as no longer relevant.

PR #763, opened 2021-04-16
Fix defects reported by Coverity Scan

k0ekk0ek:
Title says it all really. Fix a bunch of defects report by the Coverity Scan service.


k0ekk0ek:
I'm not sure about the `/* coverity[report_constant_logical_operands: FALSE] */` annotations, meaning, I know they're supported but I don't know if it's the right attribute as information is scarce for the Coverity Scan service.

eboasson (approved):
LGTM, thanks @k0ekk0ek!

PR #764, opened 2021-04-16
Install the pdb files for msvc

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>


eboasson (commented):
Installing pdb files makes sense to me, but I can't oversee the consequences of these changes to the CMake files well enough to approve. @k0ekk0ek @ThijsSassen any ideas on this subject?

k0ekk0ek:
I'm definitely in favor of installing the pdb files, but I'll have to read up on the how a bit before I can say anything useful about this particular PR. I'll look into it this afternoon/tomorrow-ish.

k0ekk0ek:
I've looked at this in somewhat more detail and I'm not sure it's the right way. Not because the install command is wrong or anything like that, I'm just not sure what projects normally do in this situation. There might be some problems if people use different builds configurations, but that's a problem we have with the way we install binaries too, I'm inclined to accept the changes and fix other problems should they come-up. One minor remark though, can you add `COMPONENT dev` to install rules?

lygstate:
> I've looked at this in somewhat more detail and I'm not sure it's the right way. Not because the install command is wrong or anything like that, I'm just not sure what projects normally do in this situation. There might be some problems if people use different builds configurations, but that's a problem we have with the way we install binaries too, I'm inclined to accept the changes and fix other problems should they come-up. One minor remark though, can you add `COMPONENT dev` to install rules?

gocha

k0ekk0ek:
Merged. Thanks @lygstate!

PR #765, opened 2021-04-16
Getting bison to be optional.

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ In order to build Cyclone DDS you need a Linux, Mac or Windows 10 machine (or, w
   * [OpenSSL](https://www.openssl.org/), preferably version 1.1 or later if you want to use TLS over
     TCP.  You can explicitly disable it by setting ``ENABLE_SSL=NO``, which is very useful for
     reducing the footprint or when the FindOpenSSL CMake script gives you trouble;
-  * [Bison](https://www.gnu.org/software/bison/) parser generator.
+  * [Bison](https://www.gnu.org/software/bison/) parser generator(optional).
 
 On Ubuntu ``apt install bison`` should do the trick for getting Bison installed, and the rest should
 already be there.  On Windows, installing chocolatey and ``choco install winflexbison3`` should get
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -14,13 +14,16 @@ cmake_minimum_required(VERSION 3.7)
 
 include(GenerateExportHeader)
 
-find_package(BISON 3.0.4 REQUIRED)
+find_package(BISON 3.0.4)
 find_package(Threads QUIET REQUIRED)
 
 set(binary_dir "${CMAKE_CURRENT_BINARY_DIR}")
 set(source_dir "${CMAKE_CURRENT_SOURCE_DIR}")
 
-bison_target(parser src/parser.y "${binary_dir}/parser.c")
+if (BISON_FOUND)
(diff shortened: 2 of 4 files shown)


eboasson (commented):
This is an old trick to avoid having lex/flex/yacc/bison available on the build machine (if I am not mistaken even `gcc` used it at some point in the past) but that doesn't make it attractive to commit generated files. At the very least, committing generated files requires the CI to fail if the committed file is out of date.

Before contemplating making another exception, I'd like to at least understand what the problem is with requiring `bison` for building Cyclone.

@k0ekk0ek, care to share your opinion?

k0ekk0ek:
@eboasson, you beat me to it :slightly_smiling_face: I was contemplating what the best response would be.

Thanks for your interest @lygstate. PRs are always appreciated even if only to see people are using the project. Bison is a reasonably lightweight dependency and only required if you want to compile Cyclone DDS from source. Bison was specifically chosen over e.g. ANTLR because it does not introduce runtime dependencies. Generally speaking, I'm against storing generated code in the repository for the same reason I wouldn't store the resulting binaries in the repository. All of that is of course without knowing what the exact problem is that this PR is trying to solve. Perhaps there's a very good reason? If it's merely for usability, I'm very much inclined to say that we should offer binary releases instead (we're already planning to so eventually, hopefully 0.8, but that's not a guarantee). Perhaps you can you elaborate a bit?

lygstate:
> @eboasson, you beat me to it 🙂 I was contemplating what the best response would be.
> 
> Thanks for your interest @lygstate. PRs are always appreciated even if only to see people are using the project. Bison is a reasonably lightweight dependency and only required if you want to compile Cyclone DDS from source. Bison was specifically chosen over e.g. ANTLR because it does not introduce runtime dependencies. Generally speaking, I'm against storing generated code in the repository for the same reason I wouldn't store the resulting binaries in the repository. All of that is of course without knowing what the exact problem is that this PR is trying to solve. Perhaps there's a very good reason? If it's merely for usability, I'm very much inclined to say that we should offer binary releases instead (we're already planning to so eventually, hopefully 0.8, but that's not a guarantee). Perhaps you can you elaborate a bit?

Hi, on Windows, bison need to be download separately, and windows have no default package manager, that's why I am trying to getting bison to be optional.

k0ekk0ek:
@lygstate, I suspected that was the reason. For Windows the situation might not be ideal (though both manual and installation via Chocolatey work pretty well). To be honest, the situation was worse with dependencies on Java and ANTLR. I think the best way forward is for the Cyclone DDS project to start offering installers, which is "easy" because there are no third-party runtime dependencies, so that consumers of the project aren't required to build it from source and thus have no need to install Bison. People that want to contribute to the development of Cyclone DDS of course do need to have Bison installed, but I think it's a very light dependency to have. @ThijsSassen, can you issue a PR like this one for Eclipse Cyclone DDS? Perhaps if that's in place we can upload binaries for the upcoming 0.8 release?

On a sidenote: a Conan package is in the works on #686, once that's done, consuming Cyclone DDS is as easy as `conan install cyclonedds` and you'd be all set.

With that, I'd like to close this PR, if @eboasson agrees.

k0ekk0ek:
I'm going to assume @eboasson agrees based on his former answer and close this issue. @lygstate, we're working on distributing installers, hopefully that'll work for you too. Thanks for the interest, sorry I can't accept this particular PR.

PR #767, opened 2021-04-21
Various fixes + merge master

eboasson:
Merging master brings a lot of fixes for various warnings from various static analyzers and adds those to the CI as well. So that's good.

The fixes:

* No longer exclusively look at the Iceoryx locator when it is present in the SEDP data, but also take the UDP ones into account. This allows remote machines to also have Iceoryx enabled and still communicate. The writer address set calculation exhibits a strong preference for Iceoryx locators and penalizes the use of locators that would additionally deliver data to those readers via the network.
* Ignore Iceoryx locators advertised by readers in calculating the addresses to use if the writer isn't actually using Iceoryx. This allows a writer that has a QoS unsuitable not yet mapped to Iceoryx to still communicate with readers on the same host that do use a QoS compatible with Iceoryx.
* Do not include Iceoryx locators in the actual writer address set. When some locator is present, the sample enters the network transmit path, and while all that ultimately leads to doing nothing at all, that does potentially force serialization and is in any case a lot of unnecessary work that was meant to be avoided by integrating Iceoryx at a high level in the stack.
* It now uses the same conditions for adding multicast locators to the SEDP messages when Iceoryx is enabled as have always been used for the SPDP messages.
* It only checks the validity of the port number in received locators *after* determining that it really understands the locator (the difference is in vendor-specific locators).
* It adds options to control the address of the Iceoryx locator and Iceoryx "service name" from the configuration. This is meant for testing (the test isn't included yet in the PR) but may also come in useful in normal usage.

Addressing some of the issues required doing things similarly to what my already rather long-lived "multi home" branch did and that has been used by various people in a variety of settings without issue already. Rather than reinventing the wheel, this simply builds on that. So with this PR, Cyclone also becomes capable of using multiple network interfaces simultaneously. The default behaviour should be unchanged except in some subtle cases (the only known one so far is that when only the loopback interface is enabled, discovery messages advertising only non-loopback addresses are now ignored — but that really sounds a like a bug fix!).

A likely improvement to the behaviour would be to enable loopback + ethernet by default and to disable the multicast loopback on the ethernet interface. That should play really nicely with the weighting of the various locators and seamlessly result in Iceoryx within the machine, ethernet multicast outside. But that is not in this PR.


reicheratwork (approved):

reicheratwork (approved):

eboasson:
Thanks for reviewing @reicheratwork !

PR #771, opened 2021-04-22
Add idl_is_keyless for C++ generator

k0ekk0ek:
Title says it all. Simple function required originally implemented specifically for the C++ generator, but might be useful for other generators too. Also, adding these few lines here saves having to re-implement a couple of functions in cyclonedds-cxx.

### DIFF
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -53,6 +53,7 @@ add_library(
     src/tree.c
     src/visit.c
     src/thread.c
+    src/print.c
     ${CMAKE_CURRENT_BINARY_DIR}/hashid.c
     ${BISON_parser_OUTPUT_SOURCE})
 
--- a/src/idl/include/idl/print.h
+++ b/src/idl/include/idl/print.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright(c) 2021 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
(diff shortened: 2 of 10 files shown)


k0ekk0ek:
Latest commit adds `IDL_PRINT` and `IDL_PRINTA` to lower the amount of bookkeeping needed in generators. Previously this was implemented for the C generator only in the form of the `AUTO` macro which accepted a dynamically allocated string, copied that to stack and freed the passed argument. That API didn't work for C++ and was strange to offer as a more general purpose interface.

eboasson (approved):
I can't say I am really enamored by the growing amount of platform-dependent "stuff", the level at which this depends on global variables (making it thread-local doesn't make it non-global in my mind) and the convoluted nature of the macros, but you do get some bonus points for squeezing it all in a macro so you can use `alloca`, though the bonus is halved for using `alloca` ...

The bit that bothers me most is not so much that the code is convoluted, as that I am not convinced the machinery is necessary in the first place. That is a discussion we have had before and I'm not going to repeat it now. Given the choices made in the past, I think this is a pragmatic step that I can live with.

But Visual Studio doesn't like it yet, so you will have to sort that out before you can merge it.

k0ekk0ek:
I know, it isn't pretty. I was just looking for a way to quickly generate throwaway type names and such without having to worry about memory allocation failure as that would make everything more complex. Even more so for the C++ generator. I could've just done more bookkeeping on all memory allocated (list or (region) allocator), but that isn't exactly pretty either and requires cleanup. I would've preferred using GCC's cleanup attribute, but nothing like it exists for compilers other than Clang and GCC. Honestly, in this case the overhead of memory management in C is just too much. Maybe supporting backends written in LUA would've made sense, but that introduces yet another dependency.

Maybe I just overlooked the obvious simple and elegant solution(?) If so, please let me know, if only for the learning experience.

k0ekk0ek:
For completeness, I compared the output for `cdrtest/xxx.idl` and some other IDL files. No differences, so I'm hitting the merge button.

PR #768, opened 2021-04-21
Update data allocator API to also support allocation on the heap

sumanth-nirmal:
This MR updates the `dds_data_allocator` API to take a flag which forces the allocation on the heap regardless of the SHM

PR #772, opened 2021-04-22
Revert data allocator api change

eboasson:
It is unfortunate that I merged #768 without considering the complications *at this point in time* for the ROS 2 release process. There's nothing wrong with the changes, this is merely coordination with ROS 2.

@sumanth-nirmal, there are two ways to move this forward again: one is to first get approval for the corresponding RMW change, the other is to do by adding a function to construct a "heap-only" allocator instead of adding a parameter. Both can work as far as I am concerned. So it is merely a delay. My apologies for my mistake.


k0ekk0ek (approved):

sumanth-nirmal:
@eboasson Even I didn't realize that ROS2 is using the `iceoryx` branch instead of a tag. I will create a new MR with the changes for the data allocator, and then we can decide if we will merge immediately or wait until we get all the corresponding changes together.

PR #773, opened 2021-04-25
Add API to initialise data allocator to allocate on heap

sumanth-nirmal:
New API which will intialise the data_allocator to allocate on the heap,
with out the context of any entity (reader or writer). This is
considered as a special case with entity 0.

Signed-off-by: Sumanth Nirmal <sumanth.724@gmail.com>

### DIFF
--- a/src/core/ddsi/src/q_init.c
+++ b/src/core/ddsi/src/q_init.c
@@ -1102,8 +1102,7 @@ static int iceoryx_init (struct ddsi_domaingv *gv)
 {
   shm_set_loglevel(gv->config.shm_log_lvl);
   
-  char str[128];
-  char *sptr = str;
+  char *sptr;
   uint32_t pid = (uint32_t) ddsrt_getpid ();
   
   ddsrt_asprintf (&sptr, "iceoryx_rt_%d_%ld", pid, gv->tstart.v);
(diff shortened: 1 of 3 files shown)


sumanth-nirmal:
@eboasson Alternative for this https://github.com/eclipse-cyclonedds/cyclonedds/pull/772. This shouldn't cause any compatibility issues with ROS2.

src/core/ddsi/src/q_init.c
```diff
@@ -1164,7 +1163,7 @@ static int iceoryx_init (struct ddsi_domaingv *gv)
     gv->loc_iceoryx_addr.kind = NN_LOCATOR_KIND_SHEM;
     gv->loc_iceoryx_addr.port = 0;
     memset ((char *) gv->loc_iceoryx_addr.address, 0, sizeof (gv->loc_iceoryx_addr.address));
-    ddsrt_strlcpy ((char *) gv->loc_iceoryx_addr.address, sptr, strlen (sptr));
+    ddsrt_strlcpy ((char *) gv->loc_iceoryx_addr.address, sptr, strlen (sptr) + 1);
```
eboasson, line 1166: Interesting find! The original code would've replaced the final 4 bits of the MAC address by 0.

That said, I have two issues with the code here:
* the original as well as the new code abuses `ddsrt_strlcpy`, because the third parameter really should be the size of the destination;
* locators are not strings anyway, so I think the conversion to string is quite useless.

I think it would make more sense altogether to do:
```C
    memset (gv->loc_iceoryx_addr.address, 0, sizeof (gv->loc_iceoryx_addr.address));
    if (ddsrt_eth_get_mac_addr (gv->interfaces[if_index].name, gv->loc_iceoryx_addr.address))
    {
      GVERROR ("Unable to get MAC address for iceoryx\n");
      return -1;
    }
    gv->loc_iceoryx_addr.kind = NN_LOCATOR_KIND_SHEM;
    gv->loc_iceoryx_addr.port = 0;
```

eboasson (commented):
I just ran the ROS 2 CI to double-check what we knew already, so the API change is approved.

I'd have merged it now if it weren't for the change in the code dealing with the MAC address, because it seems wrong to perpetuate an API misuse and it moreover seems to me that the conversion to string is an unnecessary complication that we shouldn't try to keep.

eboasson:

PR #775, opened 2021-04-26
Some fixes IDL compiler for parsing annotations

dennis-adlink:
- fix a bug in annotation_appl parameter handling
- typo fix in extensibility annotation enum value for appendable
- enable tests for parsing extensibility annotations

### DIFF
--- a/src/idl/src/scope.c
+++ b/src/idl/src/scope.c
@@ -452,10 +452,12 @@ idl_resolve(
   if (pstate->flags & IDL_FLAG_CASE_SENSITIVE)
     ignore_case = 0u;
 
-  if (pstate->parser.state == IDL_PARSE_ANNOTATION_APPL_PARAMS)
+  if (scoped_name->absolute)
+    scope = pstate->global_scope;
+  else if (pstate->parser.state == IDL_PARSE_ANNOTATION_APPL_PARAMS)
     scope = pstate->annotation_scope;
   else
-    scope = (scoped_name->absolute) ? pstate->global_scope : pstate->scope;
+    scope = pstate->scope;
   assert(scope);
 
   for (size_t i=0; i < scoped_name->length && scope;) {
--- a/src/idl/src/scope.c
+++ b/src/idl/src/scope.c
@@ -452,7 +452,10 @@ idl_resolve(
   if (pstate->flags & IDL_FLAG_CASE_SENSITIVE)
     ignore_case = 0u;
 
-  scope = (scoped_name->absolute) ? pstate->global_scope : pstate->scope;
+  if (pstate->parser.state == IDL_PARSE_ANNOTATION_APPL_PARAMS)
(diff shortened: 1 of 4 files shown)


src/idl/src/scope.c
```diff
@@ -452,7 +452,10 @@ idl_resolve(
   if (pstate->flags & IDL_FLAG_CASE_SENSITIVE)
     ignore_case = 0u;
 
-  scope = (scoped_name->absolute) ? pstate->global_scope : pstate->scope;
+  if (pstate->parser.state == IDL_PARSE_ANNOTATION_APPL_PARAMS)
+    scope = pstate->annotation_scope;
+  else
+    scope = (scoped_name->absolute) ? pstate->global_scope : pstate->scope;
```
k0ekk0ek, line 458: I'm not sure this is correct. It reads like the current scope isn't searched if something can't be found in the annotation declaration scope. I'm also not sure what the spec says about this... Should the scenario below work?

```
module m1 {
  @annotation a1 {
    unsigned int32 i1;
  };
  module m2 {
    const unsigned int32 i2 = 1;
    @a1(i1 = i2) struct s1 {
      char c1;
    };
  };
};
```

k0ekk0ek (changes requested):
Thanks @dennis-adlink! I mostly agree, but there's a few minor things. Next to whether the proper scope is searched (we can ask @e-hndrks for this too), I also think that the name being looked up should be leading(?). If an absolute name is specified, we should start searching from the global scope and choose the annotation scope or current scope on fallback. Perhaps, we should also add some simple tests for those scenarios. Of course, since you're correcting my work I'm definitely willing to help out there :slightly_smiling_face:

src/idl/src/scope.c
dennis-adlink: Good question, I'm not sure if it should and I can't really find the answer in the spec. The specs states that enums that are defined in the annotation body may be used unscoped, which could imply that a scoped name should be used for const expressions from outside the annotation body scope (i.e. `i1 = m2::i2') and that works with the changes I've added in commit 7b1f554. This commit also adds an additional test for annotation parameter scope.

src/idl/src/tree.c
```diff
@@ -2493,7 +2493,7 @@ static void *iterate_annotation(const void *ptr, const void *cur)
   const idl_node_t *node = cur;
   assert(root);
   if (node) {
-    assert(idl_parent(root) == node);
+    assert(idl_parent(node) == root);
```
k0ekk0ek, line 2496: Ah, this probably saves me some work. Noticed a bug yesterday that may be caused by this :+1:

k0ekk0ek (approved):

k0ekk0ek:
I approve, but it seems Clang-Tidy doesn't yet agree. If you fix that one, I'm happy to merge. Thanks @dennis-adlink!

PR #777, opened 2021-04-28
Delete ddsls

YixianLv:


README.md
```diff
@@ -166,6 +166,8 @@ The numbers above were measured on Mac running a 4.2 GHz Intel Core i7 on Decemb
 these numbers you can see how the roundtrip is very stable and the minimal latency is now down to 17
 micro-seconds (used to be 25 micro-seconds) on this HW.
 
+**Note:** Some tools that used to be here have been moved to the [python binding](https://github.com/eclipse-cyclonedds/cyclonedds-python/tree/master/src/cyclonedds/tools).
```
eboasson, line 169: It is difficult to find a good spot for a note like this. I think it would be better to have a list of language bindings near the beginning and point out there that the python we've started migrating the tools there. That is a bit out of the scope of this PR, so perhaps the more practical approach is for you to simply remove this sentence. (But if you do want to take a shot at it, do feel free!)

YixianLv: Yeah, you're right. I can't seem to find a nice place to add it, so I'm just deleting the note for now.

eboasson:
So now  we have a conflict on `src/tools/ddsls/CMakeLists.txt` ... can you update your branch to current master (which involves fixing the conflict, which is trivial because the file still needs to be deleted!) and do a force-push to update the PR?

Thanks!

eboasson:
Hi @YixianLv I'm afraid something went wrong with that, because now the diff is far bigger than it should be and there are some new conflicts. Given that the change really is trivial, it would perhaps make the most sense to just reset the branch (as in `git reset --hard`) to current master and delete the files again.

eboasson:
I think I know what the problem is: your making this change on something that is not the current master of cyclone — indeed it has this "Merge pull request #5 from eclipse-cyclonedds/master" in there, which refers to a positively ancient pull request. And while I was typing this, you did something and now there is an "Initial commit" again

What I would do is make sure you are on this `dev` branch and that there is nothing uncommitted that can't be thrown away, and if that is the case, do (while checking everything works out fine because I may have made a typo!):
```
git fetch https://github.com/eclipse-cyclonedds/cyclonedds
git reset --hard FETCH_HEAD
git rm src/tools/ddsls/*
```
and delete that one line in `src/tools/CMakeLists.txt`. That should clean things up quite well.

eboasson:
Thanks @YixianLv !

PR #778, opened 2021-04-28
Improved shared memory documentation

reicheratwork:
Expanded documentation on configuration and shared memory
limitations
Added documentation of loan mechanism
Rewrote documentation to no longer reference colcon


docs/manual/shared_memory.rst
```diff
@@ -1,54 +1,130 @@
...
-The shared memory implementation is based on `Eclipse iceoryx <https://projects.eclipse.org/proposals/eclipse-iceoryx>`_.
+This documentation is intended to describe the details of supporting shared memory exchange in Cyclone DDS, which is based on `Eclipse iceoryx <https://projects.eclipse.org/proposals/eclipse-iceoryx>`_.
 
-*****
 Build
-*****
+--------------
 
-The following step is under Ubuntu 18.04, but other Linux distribution should also work, too.
+The following steps were done on Ubuntu 20.04.
```
MatthiasKillat, line 11: I have not checked whether they work on e.g. Ubuntu 18.04 or other linux distributions, but they should. Maybe we can state this somewhere but it is not critical I guess, people will try it since it is a natural assumption.

docs/manual/shared_memory.rst
```diff
@@ -1,54 +1,130 @@
...
+  cmake --build build --config Debug --target install
+
+When the compiler has finished, you should have built both iceoryx and Cyclone DDS. Their respective files can be found in the install directories which were created in the directories git has made for you.
+
+Configuration
+--------------
+
+There are two levels of configuration for Cyclone DDS with shared memory support, the shared memory service (iceoryx) level, and the Cyclone DDS level.
+
+The performance of the shared memory service is strongly dependant on how well its configuration matches up with the use cases it will be asked to support, and large gains in performance can be made by configuring it correctly.
```
MatthiasKillat, line 44: dependent

docs/manual/shared_memory.rst
```diff
@@ -1,54 +1,130 @@
...
+.. code-block:: toml
+
+  [general]
+  version = 1
+
+  [[segment]]
+
+  [[segment.mempool]]
+  size = 16448
+  count = 32768
```
MatthiasKillat, line 61: The user must be aware that by using this kind of configuration iceorx will only be able to allocate memory of this size, i.e. larger allocations will fail (I think it throws an error if the required size does not even exist). In turn, smaller allocations may waste a lot of memory.  So currently it is advised to have a variety of sizes, e.g. powers of two or, if the type sizes are well known exactly those sizes ( + header size).

docs/manual/shared_memory.rst
```diff
@@ -1,54 +1,130 @@
...
+
+    * *warn*
+
+    * *error*
+
+    * *fatal*
+
+    * *off*
+
+Below follows an example of a Cyclone DDS configuration file supporting shared memory exchange:
```
MatthiasKillat, line 121: I wonder if we can get rid of these config options by incorporating them into the QoS.

The question then is what to do if the QoS request more than possible with iceoryx (fallback to max possible?, error? warning and do not use iceroyx?). Might be worth considering but for now this describes the settings as they are.

docs/manual/shared_memory.rst
```diff
@@ -60,78 +136,181 @@ Save the following xml as cyclonedds.xml.
...
+
+.. code-block:: C
+
+  message_type *loaned_sample;
+  dds_return_t status = dds_loan_sample(writer, (void**)&loaned_sample);
+
+If *status* returns **DDS_RETCODE_OK**, then *loaned_sample* will contain a pointer to the memory pool object, in all other cases, *loaned_sample* should not be dereferenced.
+For requesting loaned samples, the writer used to request the loaned sample should be of the same data type as the sample that you are writing in it, since necessary information about the data type is supplied by the writer.
+
+The user is limited in this case by the maximum number of outstanding loans, defined by **MAX_PUB_LOANS** (default set to 8). This is the maximum number of loaned samples that can be outstanding process wide from the shared memory, before some must be returned (handed back to the publisher through *dds_write*) before requesting new loaned samples.
```
MatthiasKillat, line 281: I think this limit is not process wide but per publisher, as the constant name suggests.

docs/manual/shared_memory.rst
```diff
@@ -60,78 +136,181 @@ Save the following xml as cyclonedds.xml.
...
+
+After a loaned sample has been returned to the shared memory pool (at the moment, this can only be done by invoking *dds_write*), dereferencing the pointer is undefined behaviour.
+
+If the user is not able to use the loan mechanism, as *dds_write* will also write to the shared memory service if it is configured to do so. Though in that case, the overhead of the additional copy step in publication is still incurred.
```
MatthiasKillat, line 285: typo: a *dds_write*
Maybe explain the copy step, i.e. that in this case write loans internally and performs a copy into loaned memory. May be obvious though.

docs/manual/shared_memory.rst
```diff
@@ -144,33 +323,8 @@ Here is some tips for you to get started.
...
-*********
-
-- Support DDS QoS:
-  Please refer to the `Limitations`_.
-- Support true zero copy:
-  Please refer to the `Limitations`_.
-- Extend configuration options for Shared Memory
-- Add data and measurements of performance improvements
+--------------
+
```
MatthiasKillat, line 328: Due to some implementation details it is not true zero copy yet but this is work in progress and will be resolved, so we should  probably not list it here as this should be resolved in the next days.

I am not aware of other planned work here (except fixing bugs, improving performance), but it might be that we will have something to add here later.

MatthiasKillat (commented):
Looks correct and from my experience cyclonedds with shared memory can be used like this. Minor typos and a little detail in some places could be added as remarked but is not strictly required.

docs/manual/shared_memory.rst
reicheratwork: You are correct on this point, but my fear is that we may drag in too much detail whose place should be the iceoryx documentation.

Maybe putting in as little detail as possible and linking to the relevant section in the iceoryx documentation for the rest is a good compromise?

docs/manual/shared_memory.rst
reicheratwork: After looking in the code: you are correct.

docs/manual/shared_memory.rst
reicheratwork: The asterisks on both sides of dds_write are .rst style markers, causing the text to show up as italics in the generated documentation.

About the extra copy step, where do we draw the line between documenting necessary details and trivialities? In this case, I think it will cause no harm going into a little more detail.

docs/manual/shared_memory.rst
```diff
@@ -1,54 +1,130 @@
...
 .. code-block:: bash
 
-  mkdir -p ~/cyclone_iceoryx_ws/src
-  cd ~/cyclone_iceoryx_ws/src
-  git clone https://github.com/eclipse-cyclonedds/cyclonedds.git -b iceoryx
   git clone https://github.com/eclipse-iceoryx/iceoryx.git -b master
-  cd ~/cyclone_iceoryx_ws
-  colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
+  cd iceoryx
+  cmake -Bbuild -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=install -DBUILD_ALL=True -DBUILD_SHARED_LIBS=ON -Hiceoryx_meta
```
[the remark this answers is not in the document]
reicheratwork: Yep.

docs/manual/shared_memory.rst
reicheratwork: I do not think the QoS settings are the correct place for configuration options in my opinion.

docs/manual/shared_memory.rst
reicheratwork: You are correct that the last steps needed to make it fully zero copy are not in there.

But this touches on another point, I personally don't think that the documentation is the correct place to have a list of "expected features" or "bugs we plan to solve", that should be more something for the github issues page.

docs/manual/shared_memory.rst
MatthiasKillat: Maybe in the future, I do not think this is necessary right now. It should hopefully run on a wide variety of Linux distributions with any somewhat recent gcc or clang compiler.

docs/manual/shared_memory.rst
MatthiasKillat: Sure. The user just needs be aware it is important for optimal use and where to find the information. That is the case here I think.

docs/manual/shared_memory.rst
MatthiasKillat: Sure, it is fine as it reflects the current state, no changes needed (or requested). I was just wondering if the current state could be improved and what the consequences would be on this document.

docs/manual/shared_memory.rst
MatthiasKillat: Surely it is not the ideal place here. But it  might be nice for the user to know that improvements are planned here which may lead to performance improvements.

MatthiasKillat (approved):
The documentation is correct as far as I can see and should enable a user to get started using cyclonedds with shared memory support via iceoryx.

docs/manual/shared_memory.rst
MatthiasKillat: I think the explanation level of detail is fine. This is always subjective but I tend more to the "explaining more avoids future questions/bug reports" side. :-)

docs/manual/shared_memory.rst
MatthiasKillat: Maybe I missed something, but it still says process wide. Shouldn't  it be per writer?

eboasson:
Thanks @MatthiasKillat and @reicheratwork! Now that a bunch of Iceoryx-integration related fixes got merged, I think it is about time I merge this PR as well 🙂

PR #779, opened 2021-04-29
Add missing build_depend on JDK for IDLC in package.xml

k0ekk0ek:
The main branch deleted the Java based IDL compiler and there where quite a number of commits that shouldn't land in 0.7.x, so this is quick copy of #643, but with only that change. Thanks @JWhitleyWork!

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -15,6 +15,7 @@
     <buildtool_depend>cmake</buildtool_depend>
 
     <build_depend>libssl-dev</build_depend>
+    <build_depend>java</build_depend>
 
     <exec_depend>openssl</exec_depend>
 


k0ekk0ek:
@JWhitleyWork, we've started a build on the ROS farm (https://ci.ros2.org/view/All/job/ci_launcher/8339/console), but I'm not sure that's the correct way(?) aarch64 and macOS failed, but since there are no code changes, I think this change might not be the cause. Is there some way we can verify this doesn't break the ROS builds?

k0ekk0ek:
@JWhitleyWork, it seems we just started the build incorrectly. A new one is started and can be tracked at https://ci.ros2.org/view/All/job/ci_launcher/8342/console. Let's see what happens. Sorry for the inconvenience.

PR #781, opened 2021-04-29
Suppress heartbeat/ack/historical data if Iceoryx, add test for CI

eboasson:
* Heartbeats and ACKs are not needed if iceoryx is used, but these were not fully suppressed.
* Volatile readers don't expect to get historical data, hence we shouldn't request it from Iceoryx.
* Adds a test of many combinations, with "fake" machines and processes by using 4 domain instances and overriding iceoryx locators and service names.
* It also fixes a few formatting issues with iceoryx locators in the trace.

(I hope it is not premature, CI is still underway and there may even be build failures on some platforms still, but I wanted to push it out for review.)

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -116,7 +116,9 @@ steps:
   - bash: |
       set -e -x
       cd build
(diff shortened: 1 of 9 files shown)


eboasson:

PR #782, opened 2021-04-30
Fix IDL_PRINTA redirect for Visual Studio

k0ekk0ek:
Title says it all? This snuck in because the C code generator does not use the user_data pointer, but MSVC builds for C++ fail.


k0ekk0ek:
@eboasson, I'm pretty sure you don't agree but approve :wink: Merging this so I can continue working on *cyclonedds-cxx*.

PR #786, opened 2021-05-06
ddsi_ipaddr_is_nearby_address: ignore non-IP ones

eboasson:
This fixes an assert(0) in converting a locator to an IP address when
the interface is not actually using IP addressing, e.g., when it is the
virtual one representing Iceoryx.

In a release build it would behave correctly because the interface
address would be all-zero and hence the address families in the struct
sockaddr would be different.


eboasson:
CI test failure is unrelated.

eboasson:

PR #787, opened 2021-05-06
Allow Iceoryx use for best-effort readers

eboasson:
This doesn't break the QoS matching rules for DDS while allowing the
case of a reliable writer delivering data to a best-effort reader via
Iceoryx.


eboasson:
CI test failure is unrelated.

eboasson:

PR #788, opened 2021-05-06
Avoid serialization in write for volatile writers

### LINKED ISSUE #783: dds_write when used with iceoryx performs serialization even if there are only shared memory readers
If shared memory (via iceoryx) is used a `dds_write` call will still serialize the data even if there is no need to (e.g. volatile and no network readers).

This precludes achieving zero copy transport when e.g. `dds_loan_sample` is used before calling `dds_write`.

Note rmw_cyclonedds is using a different API (already expecting a serdata) which avoids this problem. 

Potential solution: 
Perform lazy serialization, i.e. only serialize once it is clear the data has to be send over the network. This requires changes of serdata itself and in the logic of the `dds_write` implementation if shared memory is used.

MatthiasKillat:
- Extensive refactoring in dds_write.c to better handle shared memory paths
- dds_write_impl will not serialize for a writer with volatile QoS if iceoryx is available and there are no network readers (at this moment) - this is another step towards true zero copy transmission
-   added serdata construction to support this case

Closes #783

### DIFF
--- a/src/core/ddsc/src/dds_write.c
+++ b/src/core/ddsc/src/dds_write.c
@@ -410,120 +410,6 @@ static dds_return_t deliver_data (struct writer *ddsi_wr, dds_writer *wr, struct
   return ret;
 }
 
-#if 0
-// old implementation, remove later
-static dds_return_t dds_writecdr_impl_common (struct writer *ddsi_wr, struct nn_xpack *xp, struct ddsi_serdata *dinp, bool flush, dds_writer *wr)
-{
-  // consumes 1 refc from dinp in all paths (weird, but ... history ...)
-  struct thread_state1 * const ts1 = lookup_thread_state ();
-  struct ddsi_tkmap_instance *tk;
-  struct ddsi_serdata *dact;
-  int ret = DDS_RETCODE_OK;
-  int w_rc;
-
-  if (ddsi_wr->type == dinp->type)
-  {
-    dact = dinp;
-    // dact refc: must consume 1
-    // dinp refc: must consume 0 (it is an alias of dact)
-  }
-  else if (dinp->type->ops->version == ddsi_sertype_v0)
-  {
-    // deliberately allowing mismatches between d->type and ddsi_wr->type:
-    // that way we can allow transferring data from one domain to another
-    dact = ddsi_serdata_ref_as_type (ddsi_wr->type, dinp);
-    // dact refc: must consume 1
-    // dinp refc: must consume 1 (independent of dact: types are distinct)
-  }
-  else
-  {
-    // hope for the best (the type checks/conversions were missing in the
-    // sertopic days anyway, so this is simply bug-for-bug compatibility
-    dact = ddsi_sertopic_wrap_serdata (ddsi_wr->type, dinp->kind, dinp);
-    // dact refc: must consume 1
-    // dinp refc: must consume 1
-  }
-
-  if (dact == NULL)
-  {
-    // dinp may not be NULL, so this means something bad happened
-    // still must drop a dinp reference
-    ddsi_serdata_unref (dinp);
-    return DDS_RETCODE_ERROR;
-  }
-
-  thread_state_awake (ts1, ddsi_wr->e.gv);
-
-  // retain dact until after write_sample_gc so we can still pass it
-  // to deliver_locally
-  ddsi_serdata_ref (dact);
-
-  tk = ddsi_tkmap_lookup_instance_ref (ddsi_wr->e.gv->m_tkmap, dact);
-  // write_sample_gc always consumes 1 refc from dact
-  w_rc = write_sample_gc (ts1, xp, ddsi_wr, dact, tk);
-  if (w_rc >= 0)
-  {
-    /* Flush out write unless configured to batch */
-    if (flush && xp != NULL)
-      nn_xpack_send (xp, false);
-    ret = DDS_RETCODE_OK;
-  }
-  else
-  {
-    if (w_rc == DDS_RETCODE_TIMEOUT)
-      ret = DDS_RETCODE_TIMEOUT;
-    else if (w_rc == DDS_RETCODE_BAD_PARAMETER)
-      ret = DDS_RETCODE_ERROR;
-    else
-      ret = DDS_RETCODE_ERROR;
-  }
-
-  bool suppress_local_delivery = false;
-#ifdef DDS_HAS_SHM
-  if (wr && ret == DDS_RETCODE_OK) {
-    // Currently, Iceoryx is enabled only for volatile data, so data gets stored in the WHC only
-    // if remote subscribers exist, and in that case, at the moment that forces serialization of
-    // the data.  So dropping iox_chunk is survivable.
-    if (wr->m_iox_pub != NULL && dinp->iox_chunk != NULL)
-    {
-      iceoryx_header_t * ice_hdr = dinp->iox_chunk;
-      // Local readers go through Iceoryx as well (because the Iceoryx support code doesn't exclude
-      // that), which means we should suppress the internal path
-      suppress_local_delivery = true;
-      ice_hdr->guid = ddsi_wr->e.guid;
-      ice_hdr->tstamp = dinp->timestamp.v;
-      ice_hdr->statusinfo = dinp->statusinfo;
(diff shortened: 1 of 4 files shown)


src/core/ddsc/src/dds_write.c
```diff
@@ -270,19 +271,147 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
   return rc;
 }
```
MatthiasKillat, line 273: General note: we can improve some of the function names, bool variable condition names and potentially optimize the code path in dds_write_impl even more (wrt. branching and condition evaluation), but this was secondary to eliminate the serialization and general consolidation of the different ways of writing.

There is still a lot of room for improvement and refactoring.

src/core/ddsc/src/dds_write.c
```diff
@@ -270,19 +271,147 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
+  }
+  return dact;
+}
+
+static dds_return_t deliver_data (struct writer *ddsi_wr, dds_writer *wr, struct ddsi_serdata *d, struct nn_xpack *xp, bool flush) {
+  struct thread_state1 * const ts1 = lookup_thread_state ();
+  thread_state_awake (ts1, ddsi_wr->e.gv);
+
+  struct ddsi_tkmap_instance *tk = ddsi_tkmap_lookup_instance_ref (ddsi_wr->e.gv->m_tkmap, d);
+  // write_sample_gc always consumes 1 refc from dact
```
MatthiasKillat, line 329: @eboasson We need to check the overall refcounting balance with this refactoring.

src/core/ddsc/src/dds_write.c
```diff
@@ -270,19 +271,147 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
   struct writer *ddsi_wr = wr->m_wr;
-  struct ddsi_serdata *d;
-  dds_return_t ret = DDS_RETCODE_OK;
-  int w_rc;
+  int ret = DDS_RETCODE_OK;
 
   if (data == NULL)
     return DDS_RETCODE_BAD_PARAMETER;
 
-  /* Check for topic filter */
```
MatthiasKillat, line 285: The topic filter code is also a candidate to be factored out.

src/core/ddsi/src/ddsi_serdata_default.c
```diff
@@ -452,6 +452,23 @@ static struct ddsi_serdata* serdata_default_from_iox(const struct ddsi_sertype*
...
+  const struct ddsi_sertype_default *t = (const struct ddsi_sertype_default *)type;
+  struct ddsi_serdata_default *d = serdata_default_new(t, kind);
+
+  if(d == NULL) 
+    return NULL;
+
+  gen_keyhash_from_sample (t, &d->keyhash, sample);
+
+  struct ddsi_serdata *serdata = (struct ddsi_serdata*) d;
+  serdata->iox_chunk = SHIFT_BACK_TO_ICEORYX_HEADER(sample);
```
MatthiasKillat, line 469: This assumes that sample is preceeded by an iceoryx header (which is the case currently as far as I can see). If this assumption is broken in the future it will cause memory access problems.

src/core/ddsc/src/dds_write.c
```diff
@@ -270,19 +271,147 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
+      // the chances of detecting them
+      d->iox_chunk = NULL;
+      iox_pub_publish_chunk (wr->m_iox_pub, ice_hdr);
+      return true; //we published the chunk
+    }
+    return false; // we did not publish the chunk
+}
+#endif
+
+static struct ddsi_serdata *convert_serdata(struct writer *ddsi_wr, struct ddsi_serdata *dinp) {
```
MatthiasKillat, line 297: Rename dinp and dact.

src/core/ddsc/src/dds_write.c
```diff
@@ -270,19 +271,147 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
+  {  
+    return DDS_RETCODE_ERROR;
+  }
+
+  if(dact != dinp) {
+    ddsi_serdata_unref(dinp);
+    ddsi_serdata_ref(dact);
+  }
+
+  #ifdef DDS_HAS_SHM
```
MatthiasKillat, line 382: Could also be moved to conversion.

dkroenke:
@MatthiasKillat 
there are tests failing with the following error message:
```
core/ddsi/include/dds/ddsi/q_thread.h:221: thread_state_awake_to_awake_no_nest: Assertion `(vt & VTIME_NEST_MASK) == 1' failed.
```

src/core/ddsc/src/dds_write.c
MatthiasKillat: Opted to leave is as.

src/core/ddsc/src/dds_write.c
MatthiasKillat: Balanced refcounts (hopefully, requires further checking by review and CI)

src/core/ddsc/src/dds_write.c
MatthiasKillat: Refactored conditions somewhat, volatile writer check is now an assert as per discussion with @eboasson

MatthiasKillat:
Rebased onto master. @sumanth-nirmal
New target branch is master.
It is now not really possible to integrate it cleanly in the iceoryx branch anymore unless the iceoryx branch gets synchronized with master as well.

MatthiasKillat:
@eboasson it appears on Windows some seemingly unrelated test CUnit_ddssec_secure_communication_multiple_readers fails.

On macOS it is some build failure

> src/parser.y:307:9: error: use of undeclared label 'yyreturn'

but I think this should also be a problem on master (I have no mac to check it).

eboasson:
@MatthiasKillat thanks! Regarding the two failures

> Windows some seemingly unrelated test CUnit_ddssec_secure_communication_multiple_readers fails

I've seen that happen before in this way and I am positive it is an independent issue. It only ever seems to happen on CI, and the difficulty is that I haven't (yet) found a good way to determine whether it is merely bad luck and raising the timeout for the test is the correct solution, or whether there is an interesting problem hidden in there somewhere. So ... that test will probably be flakey for a while longer.

> On macOS it is some build failure
> src/parser.y:307:9: error: use of undeclared label 'yyreturn'

That's a new one, but definitely unrelated. I looks like something went awry with the bison that is (or gets) installed there. No point in holding up this PR for that.

What I did notice (and find odd, I don't see how it could have happened) is that `src/core/ddsi/include/dds/ddsi/ddsi_serdata.h` is effectively completely rewritten in the diff, but it is only formatting changes. `git log --stat` for the commit in my branch (that you cherry picked it from) says:
```
 src/core/ddsi/include/dds/ddsi/ddsi_serdata.h | 36 ++++++++++++++++--------------------
```
but for the one on this branch it says:
```
 src/core/ddsi/include/dds/ddsi/ddsi_serdata.h | 387 +++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------
```
Again, as far as I can tell it really it simply a combination of the one substantial change and the formatting. Those formatting changes shouldn't be in this PR and I'd really like you to see if you can re-cherry-pick that commit.

eboasson:
> The reason were changes introduced in master (conflicting with the commit). Rebasing caused some modifications and I use clang format on modified code. Maybe it would be useful to have a clang format run by git hooks. Then the formatting would be the same for everyone.

Ah, that explains. @k0ekk0ek and I discuss this every now and then, with me clinging to non-machine aesthetics. Sometimes I think I'm ever so slowly losing my grip and I therefore suppose that the day is coming that we'll simply enforce formatting using clang format. But not today — not yet!

> Let us see whether the pipeline is agreeable now. :-)
> Now we appear to have some resource problem on macOS
> RSS grew too much (3543040.000000 -> 7737344.000000)

It is agreeable to me. That particular error occurs in like 1% (if that many) of the runs on macOS. There are cases where the RSS legitimately grows a bit when you're writing 100kS/s & ping-ponging in parallel, packet loss could do that for example. I am loath to raise the threshold, because I really want the root cause ...

Anyway, I like the PR as it is 🙂 Thank you!

PR #791, opened 2021-05-07
Support stdin as input for compiler + some fixes

k0ekk0ek:
This PR fixes building the compiler on Visual Studio versions before 2019, takes `-fcase-sensitive` into account for identifiers too. Someone nagged me for `True` not working for booleans :wink: and that same someone found stdin didn't work as input. I actually intended for stdin to work from the start, but forgot about it while writing the generator. If stdin is used as input, it'll now generate `a.h` and `a.c` as output. Allowing the user to specify the output file with `-o` should probably be on the list too.

Next up is processing @reicheratwork's comments on [#71](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/71) in *cyclonedds-cxx*. I'll add generating a `a.hpp` as part of that PR as well.

### DIFF
--- a/src/idl/include/idl/stream.h
+++ b/src/idl/include/idl/stream.h
@@ -17,6 +17,8 @@
 
 #include "idl/export.h"
 
+IDL_EXPORT FILE *idl_fopen(const char *pathname, const char *mode);
+
 IDL_EXPORT int idl_fprintf(FILE *fp, const char *fmt, ...);
 
 IDL_EXPORT int idl_vfprintf(FILE *fp, const char *fmt, va_list ap);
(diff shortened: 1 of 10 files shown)


src/idl/include/idl/processor.h
```diff
@@ -125,10 +126,19 @@ struct idl_builtin_annotation {
   const idl_annotation_callback_t callback;
 };
 
+typedef struct idl_config idl_config_t;
+struct idl_config {
+  const char *working_directory;
+  const char *path; /**< <stdin>, <string> or pathname */
```
eboasson, line 132: So what's going to happen if I create a file `<stdin>`?

Answer: I'll get `a.c` and `a.h` instead of `<stdin>.c` and `<stdin>.h` 😢 A contrived case, admittedly, but I think it would not be harder to define an enum for file/stdin/string and then you'll be protected from such annoying people.

k0ekk0ek:
That's how I intended on doing it before, but in that scenario I had to update the `idl_source_t` too so that it can be represented properly. It's the better way, so I'll change it. Thanks @eboasson!

k0ekk0ek:
I've opted to split stdin support and the rest so that the changes can be merged into cyclonedds-cxx.

reicheratwork (approved):

PR #792, opened 2021-05-07
Update allocator API for special case of heap allocation

sumanth-nirmal:
This PR aims to use a special entity handle which will be treated as an allocation on the heap by the dda_data_allocator API


sumanth-nirmal:
@eboasson Do you mind reviewing this?

eboasson (approved):
Thanks @sumanth-nirmal, this is a sensible improvement. The API changes are such that the only really visible change is that passing 0 for the entity in `dds_data_allocator_init` no longer works, and that is the point of the exercise.

Before merging it, I'll await the results of the pro forma (short) ROS 2 CI run I triggered, but that's more as a habit-formed exercise than because I am concerned there may be issues.

PR #797, opened 2021-05-12
Fixes for idl parser related to enums

reicheratwork:
IDL enums are constructed types:
- the function idl_is_annotation_appl now also returns true
  when the annotation is made on an enum
- the function no_specific_key now returns true when types
  other than structs and unions are entered


k0ekk0ek:
LGTM. Thanks @reicheratwork!

PR #802, opened 2021-05-17
Automated documentation generation in CI

thijsmie:
This pull requests automates the generation and publication of documentation for CycloneDDS. It is accompanied by a new repository that should be created somewhere to house the generated html, see for example [my personal minimal example](https://github.com/thijsmie/cddstemp.github.io). The documentation is versioned by tag (where any 'subtags' are stripped off, for example 0.4.3rc1 will be published under 0.4.3) and master will be named 'latest'.

Requirements to merge:
 * Sister repo for html
 * SSH key with write access to sister repo added to Azure Pipelines Secure files
 * Environment variables KNOWN_HOSTS_ENTRY, SSH_PUBLIC_KEY and DOCUMENTATION_REPOSITORY


k0ekk0ek:
Mostly looks good to me. I'll have to ask the @eclipsewebmaster to create a eclipse-cyclonedds.github.io repository. Also, I've been reading up on permissions and the "sister" repository mechanism.

Don't know if you have the answer ready, but is it possible to restrict write actions to a certain directory based on the deploy key? And, what if instead of a new repository we use an orphaned branch, e.g. `gh-pages` as source for the page? Of course, that also works best if we can assign a specific deploy key that's only allowed to push to the `gh-pages` branch.

k0ekk0ek:
Having just one repo arguably makes it easier for people to determine where to open pull request for site updates. At least, I hope so.

thijsmie:
Yes, we can also have a orphaned `gh-pages` branch. However, if we have multiple backends also publishing their docs it might be nice to have those in one repo together that is not the main one and the eventual 'cyclonedds' website also needs a home. Right now my 'static site generator' construction also relies on having two branches but that is not necessarily the final mechanism.

As far as I am aware no mechanisms exists to limit which directory can be written to. What I can imagine is that we restrict pushes to a single branch and let the "website CI" deal with this access management from said secondary branch to the live website branch.

k0ekk0ek:
@thijsmie, I like most of it, but there's a publishing mechanism that I want to give a try first. Would it be possible to strip the pipeline from this PR first? That way I can accept most of this PR. Should my idea not work, I'll happily accept the second part too. As a matter of fact, I'll use the instructions either way, just not quite in the way it's done in this PR.

k0ekk0ek:
For informational purposes: I've looked into building documentation and then updating the website/documentation page. With Azure Pipelines the cleanest method (I think, coordinated with @thijsmie) is to have each project publish a documentation artifact from a *build-documentation* pipeline and then have website generation pipeline that is triggered by the aforementioned *build-documentation* pipeline. The website generated is then published to an orphaned branch named *gh-pages*, which is then used to serve the website from (branch can be chosen from GitHub settings).

PR #803, opened 2021-05-17
Fix broken assert in network partition config

eboasson:
The blame for the screw-up this fixes is mine and mine alone, but I do have to vent a bit of frustration at the tools brightly declaring that they have found a null pointer dereference when they haven't, declaring the code unacceptable because of that, and that leading to actually broken code (fortunately not in a release build) and yet more time wasting on tool pacification ...


k0ekk0ek:
Without worrying about the logic too much, I can see the removed assert is indeed incorrect. So, LGTM.

PR #806, opened 2021-05-19
Improve handling of external vs internal addresses

eboasson:
This PR improves the behaviour of the ExternalNetworkAddress setting, which is intended to allow advertising an address different from the actual address in discovery messages. As a background, the current behaviour is rather simple: specifying an ExternalNetworkAddress means that gets advertised in SPDP/SEDP and used without further consideration by other nodes.

That leads to a problem in cases where the other processes should sometimes use this "external address" and sometimes the normal address, e.g., the external one *if and only if* on a different host. It turns out that containers have this issue; I'm sure that a bit of thinking will yield plenty of interesting scenarios with NAT'ing firewalls where the boundary should not be the host but the local network, &c. I'm not sure all can be solved with a simple setting ...

One of the commits rewrites discovered locators that contain the configured external address to use the actual interface address. It would be a weird machine indeed where sending to the interface address does not work while sending to some random other address does work to get a packet to another process on the same machine.

Another broadens the notion of what a local address is to include those addresses matching the external address. This affects the heuristics for figuring out to which interface another is likely to be connected (which matters for multicast), as well as whether the loopback interface can possibly be used .

Finally, there is one that simply moves it out of the globals and into the network interface description and reworks the setting to effectively ignore loopback interfaces. It used to be disallowed to configure an external address if there were multiple network interfaces, now it is only disallowed if it there are multiple non-loopback ones. It could equally be done by allowing one to be specified for each interface independently, but that would require extending the configuration elements and can be done at a later time.

### DIFF
--- a/src/core/ddsi/src/ddsi_ownip.c
+++ b/src/core/ddsi/src/ddsi_ownip.c
@@ -308,6 +308,7 @@ int find_own_ip (struct ddsi_domaingv *gv, const char *requested_address)
       interfaces[n_interfaces].netmask.port = NN_LOCATOR_PORT_INVALID;
       memset(&interfaces[n_interfaces].netmask.address, 0, sizeof(interfaces[n_interfaces].netmask.address));
     }
+    interfaces[n_interfaces].extloc = interfaces[n_interfaces].loc;
     interfaces[n_interfaces].mc_capable = ((ifa->flags & IFF_MULTICAST) != 0);
     interfaces[n_interfaces].mc_flaky = ((ifa->type == DDSRT_IFTYPE_WIFI) != 0);
     interfaces[n_interfaces].point_to_point = ((ifa->flags & IFF_POINTOPOINT) != 0);
--- a/src/core/ddsi/src/ddsi_ownip.c
+++ b/src/core/ddsi/src/ddsi_ownip.c
@@ -308,6 +308,8 @@ int find_own_ip (struct ddsi_domaingv *gv, const char *requested_address)
       interfaces[n_interfaces].netmask.port = NN_LOCATOR_PORT_INVALID;
       memset(&interfaces[n_interfaces].netmask.address, 0, sizeof(interfaces[n_interfaces].netmask.address));
     }
+    // Default external (i.e., advertised in discovery) address to the actual interface
+    // address.  This can subsequently be overridden by the configuration.
     interfaces[n_interfaces].extloc = interfaces[n_interfaces].loc;
     interfaces[n_interfaces].mc_capable = ((ifa->flags & IFF_MULTICAST) != 0);
     interfaces[n_interfaces].mc_flaky = ((ifa->type == DDSRT_IFTYPE_WIFI) != 0);
--- a/src/core/ddsi/src/ddsi_raweth.c
+++ b/src/core/ddsi/src/ddsi_raweth.c
@@ -168,7 +168,7 @@ static int ddsi_raweth_conn_locator (ddsi_tran_factory_t fact, ddsi_tran_base_t 
   {
     loc->kind = NN_LOCATOR_KIND_RAWETH;
     loc->port = uc->m_base.m_base.m_port;
-    memcpy(loc->address, uc->m_base.m_base.gv->extloc.address, sizeof (loc->address));
+    memcpy(loc->address, uc->m_base.m_base.gv->interfaces[0].loc.address, sizeof (loc->address));
     ret = 0;
   }
   return ret;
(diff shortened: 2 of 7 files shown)


gbiggs:
This PR fixed a problem I had with a container that, when ExternalNetworkAddress was used, could talk to DDS participants on other hosts, but could not see any other participants in the same container. Without this fix, I cannot use DDS in a network involving a container and NAT/port-forwarding.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -168,7 +168,7 @@ static int ddsi_raweth_conn_locator (ddsi_tran_factory_t fact, ddsi_tran_base_t
   {
     loc->kind = NN_LOCATOR_KIND_RAWETH;
     loc->port = uc->m_base.m_base.m_port;
-    memcpy(loc->address, uc->m_base.m_base.gv->extloc.address, sizeof (loc->address));
+    memcpy(loc->address, uc->m_base.m_base.gv->interfaces[0].loc.address, sizeof (loc->address));
```
dennis-adlink, line 171: This function is used for setting `pp->m_locator` (q_entity.c:1087), which should get the `extloc` because it is (with specific config settings) used in discovery and for getting the locators that are passed to the receive threads, which should get the internal address (`loc`). Both TCP and UDP implementations are using `loc`, so I agree with changing this also for raw eth. But `pp->m_locator` needs the external address, so maybe add a 'ext' flag to this function?

dennis-adlink (commented):
LGTM, a few minor remarks.

src/core/ddsi/src/ddsi_ownip.c
```diff
@@ -308,6 +308,7 @@ int find_own_ip (struct ddsi_domaingv *gv, const char *requested_address)
       interfaces[n_interfaces].netmask.port = NN_LOCATOR_PORT_INVALID;
       memset(&interfaces[n_interfaces].netmask.address, 0, sizeof(interfaces[n_interfaces].netmask.address));
     }
+    interfaces[n_interfaces].extloc = interfaces[n_interfaces].loc;
```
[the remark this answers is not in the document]
eboasson: A good suggestion, that's one of those things that are totally obvious while you're making the change but that may not be so obvious to someone reading the code.

src/core/ddsi/src/ddsi_raweth.c
eboasson: I'm wondering if the "real" problem isn't that the per-participant socket is listening on all interfaces but isn't treated similarly to the other cases in `get_participant_builtin_topic_data`:
```C
    if (pp->e.gv->config.many_sockets_mode == DDSI_MSM_MANY_UNICAST)
    {
      locators_add_one (&def_uni, &pp->m_locator, NN_LOCATOR_PORT_INVALID);
      locators_add_one (&meta_uni, &pp->m_locator, NN_LOCATOR_PORT_INVALID);
    }
    else
    {
      for (int i = 0; i < pp->e.gv->n_interfaces; i++)
      {
        if (!pp->e.gv->xmit_conns[i]->m_factory->m_enable_spdp)
        {
          // skip any interfaces where the address kind doesn't match the selected transport
          // as a reasonablish way of not advertising iceoryx locators here
          continue;
        }
        // FIXME: should have multiple loc_default_uc/loc_meta_uc or compute ports here
        locators_add_one (&def_uni, &pp->e.gv->interfaces[i].extloc, pp->e.gv->loc_default_uc.port);
        locators_add_one (&meta_uni, &pp->e.gv->interfaces[i].extloc, pp->e.gv->loc_meta_uc.port);
      }
    }
```
Should that not be something like:
```C
      for (int i = 0; i < pp->e.gv->n_interfaces; i++)
      {
        if (!pp->e.gv->xmit_conns[i]->m_factory->m_enable_spdp)
          continue;
        if (pp->e.gv->config.many_sockets_mode == DDSI_MSM_MANY_UNICAST)
          defport = metaport = pp->m_locator.port;
        else
        {
          defport = pp->e.gv->loc_default_uc.port;
          metaport = pp->e.gv->loc_meta_uc.port;
        }
        locators_add_one (&def_uni, &pp->e.gv->interfaces[i].extloc, defport);
        locators_add_one (&meta_uni, &pp->e.gv->interfaces[i].extloc, metaport);
      }
```
Then it always advertises the same set of addresses regardless of the `many_sockets_mode`, and it only affects the port numbers. That seems far more sensible than changing the set of addresses as well.

dennis-adlink: Although I don't know all details of the different many-sockets modes, it sounds good to make the set of advertised addresses consistent with non-msm-many mode.

dennis-adlink (approved):

PR #810, opened 2021-05-25
Bitmask support in idl compiler front-end

dennis-adlink:
Adds the implementation for the bitmask data type in the idl compiler front-end, including some basic tests. This also adds the @bit_bound annotation, which can be used with bitmask and enum types. Note that the current idlc back-end does not yet support the bitmask type, this will be added later.

### DIFF
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -38,37 +38,38 @@
   (IDL_BASE_TYPE | IDL_TEMPL_TYPE | IDL_CONSTR_TYPE | IDL_TYPEDEF)
 
 #define IDL_CONSTR_TYPE \
-  (IDL_STRUCT | IDL_UNION | IDL_ENUM)
+  (IDL_STRUCT | IDL_UNION | IDL_ENUM | IDL_BITMASK)
 
 #define IDL_TEMPL_TYPE \
   (IDL_SEQUENCE | IDL_STRING | IDL_WSTRING | IDL_FIXED_PT)
 
 /* miscellaneous */
-#define IDL_KEYLIST (1llu<<37)
-#define IDL_KEY (1llu<<36)
-#define IDL_INHERIT_SPEC (1llu<<35)
-#define IDL_SWITCH_TYPE_SPEC (1llu<<34)
-#define IDL_LITERAL (1ull<<33)
+#define IDL_KEYLIST (1llu<<38)
+#define IDL_KEY (1llu<<37)
+#define IDL_INHERIT_SPEC (1llu<<36)
+#define IDL_SWITCH_TYPE_SPEC (1llu<<35)
+#define IDL_LITERAL (1ull<<34)
 /* declarations */
-#define IDL_MODULE (1llu<<32)
-#define IDL_CONST (1llu<<31)
-#define IDL_MEMBER (1llu<<30)
-#define IDL_FORWARD (1llu<<29)
-#define IDL_CASE (1llu<<28)
-#define IDL_CASE_LABEL (1llu<<27)
+#define IDL_MODULE (1llu<<33)
+#define IDL_CONST (1llu<<32)
+#define IDL_MEMBER (1llu<<31)
+#define IDL_FORWARD (1llu<<30)
+#define IDL_CASE (1llu<<29)
+#define IDL_CASE_LABEL (1llu<<28)
 /* if explicit default is specified */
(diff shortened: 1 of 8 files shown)


src/idl/include/idl/tree.h
```diff
@@ -397,6 +399,23 @@ struct idl_enum {
...
   idl_enumerator_t *enumerators;
+  uint16_t bit_bound;
+  idl_extensibility_t extensibility;
+};
+
+typedef struct idl_bit_value idl_bit_value_t;
+struct idl_bit_value {
+  idl_node_t node;
+  struct idl_name *name;
+  uint16_t position;
```
k0ekk0ek, line 410: Probably the same issue as with the bit_bound member? Normally this would be initialized to `0`, but if `@position` is used and we provide two of the same values, that'd go unnoticed.

src/idl/src/parser.y
```diff
@@ -837,6 +842,27 @@ enumerator:
       }
   ;
 
+bitmask_dcl: bitmask_def { $$ = $1; } ;
+
+bitmask_def:
+    "bitmask" identifier '{' bit_values '}'
```
k0ekk0ek, line 848: I don't see a token added for `"bitmask"`? I'm a bit surprised it works. Does it show up in the token table?

src/idl/src/tree.c
```diff
@@ -772,6 +777,16 @@ uint32_t idl_bound(const void *node)
...
+uint16_t idl_bit_bound(const void *node)
+{
+  idl_mask_t mask = idl_mask(node);
+  if ((mask & IDL_BITMASK) == IDL_BITMASK)
+    return ((const idl_bitmask_t *)node)->bit_bound;
+  if ((mask & IDL_ENUM) == IDL_ENUM)
+    return ((const idl_enum_t *)node)->bit_bound;
+  return 0u;
+}
+
```
k0ekk0ek, line 789: We could opt to support this through `idl_is_bounded` and `idl_bound`. Not sure if that's desirable, but have you considered that possibility? There's of course the maximum of `64` that can be returned for a `@bit_bound`, but maybe it's pragmatic to repurpose those functions...

src/idl/tests/annotation.c
```diff
@@ -446,3 +446,50 @@ CU_Test(idl_annotation, parameter_scope)
 #undef M
 #undef A
 #undef TA
+
+
+#define BM(i) "@bit_bound(" i ") bitmask MyBitMask { @position(0) flag0 };"
```
k0ekk0ek, line 451: This test is faulty I'd say? `@position` is simply ignored, so the value of `flag0` is always `0`.

k0ekk0ek (changes requested):
Thanks @dennis-adlink! Ok with it in general, but there's some minor things you should probably take a look at.

src/idl/src/parser.y
dennis-adlink: The token was already in [parser.y](https://github.com/eclipse-cyclonedds/cyclonedds/blob/5fa204dd61620b24c394ef88d9040944ecbbb470/src/idl/src/parser.y#L270)

src/idl/src/tree.c
dennis-adlink: Good point, I've merged the functions `idl_bit_bound` and `idl_bound`. I think `idl_is_bounded` for enum and bitmask types does not really make sense, as these types always have a bounded bit size.

src/idl/tests/annotation.c
dennis-adlink: You're right, I've refactored this test a bit and added more test cases, and I fixed the `@position` annotation implementation.

src/idl/include/idl/tree.h
```diff
@@ -397,6 +399,23 @@ struct idl_enum {
...
+};
+
+typedef struct idl_bitmask idl_bitmask_t;
+struct idl_bitmask {
+  idl_node_t node;
+  struct idl_name *name;
+  idl_bit_value_t *bit_values;
+  uint16_t bit_bound;
```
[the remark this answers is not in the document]
dennis-adlink: I've added setting the default values for `bit_bound` in `idl_create_bitmask` and `idl_create_enum`, and I also added checking that the number of enumerators and bit position are within the specified bit_bound value.

src/idl/include/idl/tree.h
dennis-adlink: See above, I've added some additional checks

PR #816, opened 2021-05-31
Add pipeline for publishing documentation

k0ekk0ek:
This PR builds on #802 and adds a pipeline for generating documentation for *latest* and specific tags/releases and publishes as a build artifact. The idea is that the artifact is picked up by a pipeline for eclipse-cyclonedds.github.io, which is triggered by this pipeline, which then regenerates the website ensuring documentation is always up-to-date.


thijsmie (commented):
Looks good to me, I like the way it is set up here.

PR #820, opened 2021-06-02
libidl version number independent of project

eboasson:
We can't guarantee that LD_LIBRARY_PATH doesn't place an older libidl.so
ahead of the newly built one, and this can cause the IDL compiler to
fail when invoked as part of the build process.  Bumping the version
number often & independently of the cyclone library is one way to solve
it.

### DIFF
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -57,10 +57,15 @@ add_library(
     ${CMAKE_CURRENT_BINARY_DIR}/hashid.c
     ${BISON_parser_OUTPUT_SOURCE})
 
+# We can't guarantee that LD_LIBRARY_PATH doesn't place an older libidl.so ahead of the
+# newly built one, and this can cause the IDL compiler to fail when invoked as part of the
+# build process.  Bumping the version number often & independently of the cyclone library
+# is one way to solve it (another is to prepend the directory containing the newly built
+# one to LD_LIBRARY_PATH during the build process, and there may be others.)
 set_target_properties(
   idl PROPERTIES
-  VERSION ${PROJECT_VERSION}
-  SOVERSION ${PROJECT_VERSION_MAJOR})
+  VERSION 1
+  SOVERSION 1)
 
 generate_export_header(idl EXPORT_FILE_NAME include/idl/export.h)
 


k0ekk0ek:
I don't get the problem... We also cannot guarantee that nobody puts a libddsc.so in LD_LIBRARY_PATH that'll mess with the build process or that someone changes the path between configure time and build time to point to an incompatible Bison version etc. I'm very hesitant to include this as an actual fix.

eboasson:
> I don't get the problem... We also cannot guarantee that nobody puts a libddsc.so in LD_LIBRARY_PATH that'll mess with the build process or that someone changes the path between configure time and build time to point to an incompatible Bison version etc. I'm very hesitant to include this as an actual fix.

One way to run into this problem is when you install Galactic from debian packages and then try to build a new cyclone as part of an "overlay" ROS 2 build. (Apparently we are not the only one with this type of problem). Another is if you have installed some version of cyclone in the past and included it in `LD_LIBRARY_PATH`. So I think it is not entirely theoretical.

Whether that means it must be solved, and if yes, via library versioning is a different story. Using a static library would work also, as would prepending the build directory to `LD_LIBRARY_PATH` when running `idlc` during the cyclone build. (I don't think these are attractive options, but they do work.)

I made this PR primarily to get comments, and for that it worked! 😁

k0ekk0ek:
One way is to apply the same *hack*. i.e. set `LD_LIBRARY_PATH` path as part of the `idlc_generate` step. I'll have a look to see if I can come up with something

k0ekk0ek:
The plot thinkens... Just trying to reproduce this problem by having a different version in `LD_LIBRARY_PATH` doesn't fail the build because the executables have `RPATH` set by default (both in build and install mode). So, does the ROS build system change the `RPATH` settings?

eboasson:
> The plot thinkens... Just trying to reproduce this problem by having a different version in `LD_LIBRARY_PATH` doesn't fail the build because the executables have `RPATH` set by default (both in build and install mode). So, does the ROS build system change the `RPATH` settings?

That sounds highly unlikely, it just runs CMake ... I also find this:
```
Set runtime path of "/home/erik/ros2g_cyc_overlay/install/cyclonedds/bin/idlc" to "$ORIGIN/../lib"
```
which to me suggests it isn't messing with the rpath.

eboasson:
It is ultimately a known issue with building overlays of non-leaf packages in combination with `colcon --merge-install`. While this would fix the problem, I agree with @k0ekk0ek that bumping the version number on an internal shared library for each small change is not a good plan.

PR #826, opened 2021-06-07
Switch Coverity Scan build to Clang

k0ekk0ek:
While not strictly necessary, switch the Coverity Scan build for Cyclone DDS to Clang too. That way both configurations are the same.


k0ekk0ek:
For good measure, I've verified for *cyclonedds-cxx*, I'm assuming Clang gets picked up without problems for C code too.

eboasson (approved):

PR #828, opened 2021-06-07
Fix C.UTF-8 warnings generated by idlc

k0ekk0ek:
This little PR adds very rudimentary support for the so-called [XPG syntax](https://www.gnu.org/software/libc/manual/html_node/Locale-Names.html) to mcpp. This fixes the `warning: Unknown encoding: C.UTF-8` warnings on some systems.

### DIFF
--- a/src/tools/idlpp/src/mbchar.c
+++ b/src/tools/idlpp/src/mbchar.c
@@ -367,6 +367,7 @@ const char *    set_encoding(
     const char *    loc = "";
     int     alias;
     size_t  len;
+    size_t  pos;
     char    norm[ NAMLEN];
             /*
              * Normalized name (removed 'xxxxx.', stripped '_', '-', '.'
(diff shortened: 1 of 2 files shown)


src/tools/idlpp/src/mbchar.c
```diff
@@ -381,11 +382,19 @@ const char *    set_encoding(
             mcpp_fputc( '\n', ERR);
         }
     }
+    /* XPG syntax: language[_territory[.codeset]][@modifier]    */
     len = strlcpy( norm, name, sizeof(norm));
-    if (len >= 5 && norm[ 5] == '.')
-        memmove( norm, norm + 5, (len - 5) + 1);
+    for (pos = 0; pos < len && norm[pos] != '@'; pos++)
```
eboasson, line 387: According the man pages, `strlcpy` always returns the `strlen(source)`, which means this loop will read beyond `norm` if `strcspn(name,"@")>=sizeof(norm)`.

Otherwise, fine with me, if a bit convoluted. (I'd have factored it out into a separate function and used `strchr`, and made `norm` a pointer into a array of `char`, because that'd save me from having to do the math on a `memmove` 😁)

k0ekk0ek: Good point! Thanks @eboasson

PR #836, opened 2021-06-15
Fixes for errors and warnings on FreeRTOS

eboasson:
* Build errors when no IPv6 support available
* Build errors when assertions enabled and no SSM support available
* `uint32_t` vs `int` in a printf-style format string using `%*.*s` to print only a prefix of a string
* Warnings from LwIP redefining macros from `errno.h`

Plus a second attempt at eliminating two coverity false positives.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_domaingv.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_domaingv.h
@@ -183,11 +183,9 @@ struct ddsi_domaingv {
   /* number of selected interfaces. */
   int n_interfaces;
   struct nn_interface interfaces[MAX_XMIT_CONNS];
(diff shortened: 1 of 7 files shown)


k0ekk0ek:
This [one](https://dev.azure.com/eclipse-cyclonedds/cyclonedds/_build/results?buildId=448&view=logs&j=adc611be-d2d6-58a9-dbb6-278cbdb63baf&t=29bca4b7-68d9-5b33-bb02-0e147609d8ac&l=2556) might be interesting?

eboasson:
Yes, it is definitely interesting, but unfortunately there is no clue as to where it crashed. I don't think it is related to these changes.

k0ekk0ek (approved):

PR #838, opened 2021-06-16
Fix dds_find_topic_scoped handling of topics in different participants

### LINKED ISSUE #837: Issue with participant cleanup after finding topic created in one participant through another
This was tested on master:93c8f8420f11d898854902bd6386c4ae5c09e9a9

To recreate:
Modify unittest ddsc_topic_find_local.domain (topic_find_local.c) and replace:
`ret = dds_delete (topic1);` with `ret = dds_delete (g_participant1);` (line 70)
and:
`ret = dds_delete (topic2);` with `ret = dds_delete (g_participant2);` (line 72)

This will cause a assert failure in the function `dds_participant_delete` on line 54 of `dds_participant.c` indicating that the topics belonging to the participant have not yet been disposed of correctly.

By switching `g_participant1` and `g_participant2` in the changed unittest, the assert failure disappears.

This is caused by the reference count of the topic in topic1 being raised by the function `dds_find_topic_scoped` which creates a topic which is referenced in topic2 but contained in topic1.

To fix:
Probably requires rewriting parts of `find_local_topic_pp`

eboasson:
This adds a test that reliably reproduces #837 and a fix to `dds_find_topic_scoped` for it. The problem was that it cross-referenced the `ktopic` pointer from one participant to another, where those were supposed to be local to the participant. So doing:
```C
p1 = dds_create_participant(...);
p2 = dds_create_participant(...);
t1 = dds_create_topic(p1,...);
t2 = dds_find_topic_scoped(p2, DDS_FIND_SCOPE_LOCAL_DOMAIN, ...);
```
meant `t2` pointed to (the refcounted) `t1->m_ktopic` living in `p1->m_ktopics`. A call to `dds_delete(p1)` at this point causes it to delete all contained entities and it consequently expects to end up with an empty set of these "ktopic"s. The assertion failure is `t2` referencing `p1`s "ktopic" prevented the set from becoming empty.

If you first delete `p2` (or `t2`) the reference no longer exists and there is no problem.

This fixes #837

### DIFF
--- a/src/core/ddsc/src/dds_topic.c
+++ b/src/core/ddsc/src/dds_topic.c
@@ -94,10 +94,12 @@ static struct ktopic_type_guid * topic_guid_map_refc_impl (const struct dds_ktop
   return m;
 }
 
+#if 0 // not currently needed, symmetry arguments suggest it should exist but the compiler disagrees
 static void topic_guid_map_ref (const struct dds_ktopic * ktp, const struct ddsi_sertype *sertype)
 {
   (void) topic_guid_map_refc_impl (ktp, sertype, false);
 }
+#endif
 
 static void topic_guid_map_unref (struct ddsi_domaingv * const gv, const struct dds_ktopic * ktp, const struct ddsi_sertype *sertype)
 {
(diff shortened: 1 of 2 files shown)


dennis-adlink (approved):
LGTM

eboasson:
@k0ekk0ek shouldn't the "publish documentation" check simply pass while reporting ok? It is definitely correct that nothing got published for a PR, but it is a bit of shame that it means the overall CI verdict is "failed". Just wondering; going ahead and merging this now with @dennis-adlink's approval.

PR #839, opened 2021-06-16
Do not publish documentation on pull requests

k0ekk0ek:
Signed-off-by: Jeroen Koekkoek <jeroen@koekkoek.nl>


eboasson (approved):
"All checks passed" — that means it does what it says on the tin 🙂

I'm glad I didn't look into it yesterday evening, I wasn't aware that it would be this simple.

PR #840, opened 2021-06-17
Fix mistyped ddsrt_mtime_t.

thijsmie:
Ran into this on CI, I can't really test it locally due to lack of a mac but it is a fairly obvious typo.
Error log: https://dev.azure.com/tmiedema/57989905-544b-4946-93ca-4f71890be2ab/_apis/build/builds/195/logs/16


k0ekk0ek:
I guess nobody uses macOS versions that old anymore? Good catch!

k0ekk0ek (approved):

eboasson:
> I guess nobody uses macOS versions that old anymore? Good catch!

I guess not 🙂 but I prefer this simple fix over deleting the compatibility code. Thanks @thijsmie !

thijsmie:
Yeah I am not quite certain how I actually managed to trigger said code since it is running on Mac 10.15, but I think that might be a configuration problem on my end.

PR #842, opened 2021-06-18
Add support for building with mingw-w64

k0ekk0ek:
Still a draft until I actually manage to get it working, but I thought it might be useful to open this so that people can see we're working on it.


k0ekk0ek:
Building works. There's a bunch of warnings, but it generates a `.dll` and successfully builds executables. No further testing has been done, but maybe this helps someone.

k0ekk0ek:
Should now build with MinGW as well as MSVC. I introduced two extra macros in `dds/export.h`, namely `DDS_INLINE_EXPORT` and `DDS_EXTERN_INLINE_EXPORT` that get expanded differently based on whether MinGW or MSVC is used. There's still a bunch of warnings to fix though. `ddsconf` and `idlc` seem to work, but that's all I've tried to run.

k0ekk0ek:
This PR makes Cyclone DDS compile cleanly on both MinGW and MSVC. I haven't compiled the examples on MinGW yet because binaries for MinGW builds aren't available, but the examples, tools, etc all work fine. Tested with `ddsperf` locally. One MSVC build, one MinGW build and that works perfectly, so this is ready for review. I'm hoping to get Clang for Windows running as well so we can have a Clang-Tidy build available for Windows. The build error on other platforms I'll take a look at shortly.

k0ekk0ek:
Almost there. Locally I can build everything just fine (including tests and SSL), but enabling the tests showed a couple warnings, so those I'll have to fix. The 32-bit Windows build is waiting for a couple of fixes in the CUnit conan package so it can be cross compiled to 32-bit without needing to build msys2, m4 and libtool ([PR](https://github.com/conan-io/conan-center-index/pull/6595)).

k0ekk0ek:
Cleaned up some additional things here and there that got in the way. Builds fail due to the introduction of a trivial warning. Only thing left to do is figure out why the thread cleanup tests fail, all other tests pass. At least, they do so when testing in a Windows VM.

k0ekk0ek:
Small update to TLS for libidl and update of cache key. If the tests pass, I'll hit merge.

k0ekk0ek:
Tests passed. Hitting merge! :slightly_smiling_face:

PR #847, opened 2021-06-22
Added QNX OS support

LanderU:
Co-authored-by: Ahmed Sobhy <asobhy@blackberry.com>
Signed-off-by: LanderU <lander.usategui@gmail.com>


LanderU:
Hi Cyclonedds developers! We (Apex.AI and BB) are actievely using CycloneDDS in QNX and we would like to contribute back with the necesary changes to support QNX OS.

k0ekk0ek:
Hi @LanderU! About time I responded to your PR :slightly_smiling_face: I only had a quick look, but the changes look fine at first glance and we're definitely willing to merge QNX support. Much appreciated! One question I do have, and I've mentioned this off line to some people as well, is: Would it be possible to integrate a QNX build in our CI setup? I know QNX is a commercial operating system, so it's not straightforward, but past experiences with other (embedded) builds taught me builds for operating systems not tested on a regular base tend to break. i.e. we have some FreeRTOS builds (that we haven't gotten around to putting in Azure Pipelines yet) that fail every so often because the changes aren't tested when the PR is issued. Even having just the build without running any tests would help a lot. Perhaps we can do something like a proprietary Docker container to trigger a cross build?

eboasson (approved):
This is much appreciated and the changes look fine.

I do want to second @k0ekk0ek's request regarding integrating a build in the CI. With so few changes needed for QNX, I think the risks are much smaller, but things that can go wrong just tend to go wrong at some point ... It would be great if we can avoid that.

LanderU:
@k0ekk0ek @eboasson, thanks a lot for reviewing. I do agree with you that we should setup a CI for QNX. I cannot speak by Black Berry, so not sure if this is feasible. @asobhy-qnx, would you comment on that?
Btw, there is a CI that build Cyclone DDS https://gitlab.com/qnx/ros2/ros2_qnx at is building everynight the Cyclone DDS versions for galactic/rolling https://gitlab.com/qnx/ros2/ros2_qnx/-/pipeline_schedules.

eboasson:
> there is a CI that builds Cyclone DDS https://gitlab.com/qnx/ros2/ros2_qnx at is building everynight the Cyclone DDS versions for galactic/rolling https://gitlab.com/qnx/ros2/ros2_qnx/-/pipeline_schedules

That's good to know. Detecting problems after introducing them is at least something. Still, we have the same for Free-RTOS and so I have learnt that it is quite annoying to accidentally break things and have to fix them again, as compared to the CI catching it prior to merging a PR ... so hopefully @asobhy-qnx can help figure out how to fully integrate it in the CI pipeline.

LanderU:
@eboasson @k0ekk0ek, I think we should merge this as it is. We are testing this every day internally, we'll report/fix any failure in QNX builds. At least till we've a response about the CI from @asobhy-qnx 

What do you think?

asobhy-qnx:
Yes, sorry about the delay. I am hoping that we can work out the build server license approval soon.

eboasson:
@LanderU Given that the changes required are really very small, the likelihood of accidentally breaking things seems small. I think that makes it reasonable to merge it while relying on your test set-up, but, as @k0ekk0ek mentioned earlier, we have been bitten by a similar arrangement with FreeRTOS and so are a bit hesitant now.

If @k0ekk0ek agrees with my tentative yes, we'll merge it. I expect it'll be couple more days before he responds.

k0ekk0ek:
True, the changes are small, I've stated so myself. I don't expect something breaks right away. However, once they're in I'd feel obligated to ensure things keep somewhat working and currently there's no way for us to verify. Also, once they're in, supporting the other platforms will be slightly more complicated (it's an extra `ifdef` in various places after all).

Since it's a strictly non-technical reason why the build isn't available yet, I'm in favor of merging only after we have a build available so there's some incentive to get that last part done too.

eboasson:
I am done with waiting for a QNX build to get integrated into the CI. The changes are small and in perfectly reasonable places and I see no downside to including them. It simply means QNX becomes a "known to work" platform, not a supported one.

LanderU:
Thanks a lot @eboasson!

PR #852, opened 2021-06-23
Fixes for the remaining Coverity issues

eboasson:
I think this will take care of all outstanding issues.


k0ekk0ek (approved):
I tried to keep the number of changes to a minimum until fixes from other active mcpp forks are merged, but useful changes no doubt. Thanks for taking care of them @eboasson!

PR #860, opened 2021-06-28
Fix warn '[-Wmisleading-indentation]' in if statement

homalozoa:
Signed-off-by: Homalozoa <nx.tardis@gmail.com>


eboasson (commented):
Thanks @homalozoa, but this one has already been fixed in `master`, by commit a0036df

PR #861, opened 2021-06-28
Fix warn '[-Wmisleading-indentation]' in if statement

homalozoa:
Signed-off-by: Homalozoa <nx.tardis@gmail.com>


eboasson (commented):
Actually, the problem is that the `assert` was added without adding a pair of braces. Commit a0036df fixes it on the master branch. I'll do a PR for cherry-picking that onto the `iceoryx` branch.

PR #863, opened 2021-06-29
Support const expressions in idl_type_spec

k0ekk0ek:
Addition to help fix [issue 108 in cyclonedds-cxx](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/108).


eboasson (approved):
Looks reasonable to me. A quick check of the callers in the generic code and the C backend confirms that a null pointer return is not (currently) a useful case.

The CI issues are unrelated.

PR #864, opened 2021-06-30
Add missing dds_ prefix to is_loan_available

eboasson:
It remains (for now) available under the unprefixed name as deprecated
export.


dennis-adlink (approved):
LGTM

PR #865, opened 2021-06-30
Fix assert(thread_is_asleep()) receiving short msg

eboasson:


dennis-adlink (approved):
LGTM

PR #871, opened 2021-07-06
Add the option to prematurely return a publisher loan using the already existing dds_return_loan function.

e-hndrks:
This functionality is only supported when CycloneDDS is built using the ENABLE_SHM macro, otherwise it will return DDS_RETCODE_UNSUPPORTED.

### DIFF
--- a/src/core/ddsc/src/dds_read.c
+++ b/src/core/ddsc/src/dds_read.c
@@ -13,6 +13,7 @@
 #include <string.h>
 #include "dds__entity.h"
 #include "dds__reader.h"
+#include "dds__writer.h"
 #include "dds/ddsi/ddsi_tkmap.h"
 #include "dds/ddsc/dds_rhc.h"
 #include "dds/ddsi/q_thread.h"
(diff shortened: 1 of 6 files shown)


src/core/ddsc/src/dds_read.c
```diff
@@ -509,32 +510,36 @@ dds_return_t dds_take_next_wl (dds_entity_t reader, void **buf, dds_sample_info_
...
+  if ((ret = dds_entity_pin (entity, &p_entity)) < 0) {
     return ret;
-  } else if (dds_entity_kind (entity) == DDS_KIND_READER) {
-    rd = (dds_reader *) entity;
-  } else if (dds_entity_kind (entity) != DDS_KIND_COND_READ && dds_entity_kind (entity) != DDS_KIND_COND_QUERY) {
-    dds_entity_unpin (entity);
+  } else if (dds_entity_kind (p_entity) == DDS_KIND_READER) {
+    rd = (dds_reader *) p_entity;
+  } else if (dds_entity_kind (p_entity) == DDS_KIND_WRITER) {
+    ret = dds_return_writer_loan((dds_writer *) p_entity, *buf);
```
eboasson, line 527: `*buf` might be invalid here — something surprising enough that I felt it necessary to add a comment in the `bufsz <= 0`. (I do remember that my rationale was that this allows you to call `dds_return_loan` after a failed read/take operation simply passing in the latters result.)

src/core/ddsc/src/dds_read.c
```diff
@@ -13,6 +13,7 @@
 #include <string.h>
 #include "dds__entity.h"
 #include "dds__reader.h"
+#include "dds__writer.h"
```
eboasson, line 16: I have a bit of a problem with retaining `dds_return_loan` as a function in a file so strongly associated with reading data when after extending it to also deal with loans on writers. I think it would make more sense to move the dispatching code to `dds_entity.c` and keep only the reader-specific parts here, for example in a function `dds_return_reader_loan`.

src/core/ddsc/src/dds_write.c
```diff
@@ -115,6 +121,36 @@ dds_return_t dds_loan_sample(dds_entity_t writer, void** sample)
...
+dds_return_t dds_return_writer_loan(dds_writer *writer, void *sample)
+{
+#ifndef DDS_HAS_SHM
+  (void) writer;
+  (void) sample;
+  return DDS_RETCODE_UNSUPPORTED;
+#else
+  dds_return_t ret;
+
+  if (!sample)
```
eboasson, line 133: This being not an external API, I think it makes more sense to require `sample` to be non-null.

eboasson (commented):
Thanks @e-hndrks, this almost got approved, but there is one issue with input validation and there are moreover two stylistic matters. Perhaps you could have a look?

src/core/ddsc/src/dds_read.c
MatthiasKillat: Alternatively the loaning code could get its on compilation unit and would contain functions for returning loans for different entities (general and specific). We lose the de facto object oriented notion then if we move reader and writer specific implementations also there. Furthermore the functions to obtain the loan would be elsewhere, which is also confusing.

src/core/ddsc/src/dds_write.c
e-hndrks: Hi @eboasson, thanks for your review. I processed your review comments in a subsequent commit, so you can clearly see the diffs I made. If you agree, could you squash commit this feature into the master?

e-hndrks: BTW: There is still one failing testcase, but that seems to be unconnected to this Pull Request. It is a testcase that test security plugins, and it complains about an expired CRL. Maybe you or @k0ekk0ek can take care of that?

eboasson:
@e-hndrks there were some details I noticed, rather than bothering you with it I decided to go ahead and open #886 (naturally, it builds on your work, it doesn't replace it).

eboasson:
Merged via #886.

PR #876, opened 2021-07-08
Add support for reloading configuration

k0ekk0ek:
This PR contains the boilerplate for reloading the configuration for selected options, namely: peers, tracemask and verbosity. It's not quite done yet. There's one function I still have to implement and I'm thinking about using atomic operations for setting the tracemask, but I thought I'd make a draft PR to get initial comments.


thijsmie:
This draft PR is now quite out of date, and as far as I know we have received no further requests for this functionality. Thanks for the initial work, we might look into it again in the future if it comes up again.

PR #877, opened 2021-07-09
Adds custom property to ddsc target to indiciate if SHM is supported

sumanth-nirmal:
Adds a custom property `CYCLONE_DDS_SHM_SUPPORT_IS_AVAILABLE` to target `CycloneDDS::ddsc` and exported with the target. The language bindings can get this property for info on if Cyclone DDS has shared memory support.

Some more info can be found in this discussion https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/86#issuecomment-875116934

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -51,6 +51,17 @@ target_include_directories(
 # SOVERSION should increase on incompatible ABI change
 set_target_properties(ddsc PROPERTIES VERSION ${PROJECT_VERSION} SOVERSION ${PROJECT_VERSION_MAJOR})
 
+# define target property to indicate if Cyclone DDS is compiled with SHM support
+define_property(TARGET
+    PROPERTY SHM_SUPPORT_IS_AVAILABLE
+    BRIEF_DOCS "Indicator whether SHM support is available with current Cyclone DDS build configuration."
+    FULL_DOCS "Indicator whether SHM support is available with current Cyclone DDS build configuration."
+    )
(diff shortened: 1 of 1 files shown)


eboasson (commented):
It does what I think it should do (the installed `CycloneDDSTargets.cmake` gains a `CYCLONE_DDS_SHM_SUPPORT_IS_AVAILABLE` property), but despite having had to deal with CMake for quite some time now, it still trips me up often enough that I'd like a second opinion. @k0ekk0ek, perhaps you can have a look, too?

Secondly, I checked and we seem to more commonly use `CYCLONEDDS` rather than `CYCLONE_DDS`, despite the latter technically being a better way of transcribing the official project name (which is "Cyclone DDS").

Perhaps that issue can be side stepped by removing the `CYCLONE_DDS_` prefix is altogether? I am not sure about scoping rules for properties in CMake, but there certainly are some things that can be scoped with the project name. `CycloneDDS::SHM_SUPPORT_IS_AVAILABLE` would definitely be nicer. Or even the more old-fashioned `HAS_SHM` 🙂

sumanth-nirmal:
@eboasson Agree, since this is a property on target it can be accessed as `CycloneDDS::ddsc CYCLONE_DDS_SHM_SUPPORT_IS_AVAILABLE`, but from your comment, I think it will be more meaningful if I rename this to `SHM_SUPPORT_IS_AVAILABLE`, so it accessed as `CycloneDDS::ddsc  SHM_SUPPORT_IS_AVAILABLE`. 
Do you also agree?

PR #880, opened 2021-07-14
Increase the CRL expiration date.

clalancette:
When I wrote the tests for CRL, I forgot that I used the
"recommended" practice of 30 days for the CRL.  Since that
has now expired, some of the CRL tests are failing.

Instead, increase the CRL time up to 73000 days, which makes
it valid for 200 years.  This isn't recommended best practice
for a real deployment, but should keep the tests from ever
failing in this way again.

I also had to regenerate the CRL when doing this and update
the tests to use that CRL file.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


eboasson:
Thanks @clalancette! I'll cherry-pick this onto master as the "permanent" fix there as it way simpler than generating a CRL on the fly and we'll both long be retired in 200-odd years 😀.

Issue #867, opened 2021-06-30
Default message "using network interface..." is printed as error


EttoreVidotto:
Hello everyone,
The default mentioned message is printed using `GVWARNING`. It is not clear to me why, but the message is then printed as an error, instead of as an information, as it happens, for example, for all the configuration parameters when the example `cyclonedds.xml` file is used.
![CycloneDDS-Warning](https://user-images.githubusercontent.com/79527119/123985089-78004380-d9c5-11eb-803d-c0ad6f19f758.PNG)
Is it really necessary to treat this message as a warning? Could the function `GVLOG(DDS_LC_CONFIG, ...)` be used instead as for the other settings?
Thank you in advance, any answer would be appreciated.
Best regards,
Ettore Vidotto

k0ekk0ek:
Hi @EttoreVidotto! Sorry for taking so long to respond. I think warnings are always printed to stderr by default, but you might be right in that it's better printed as info. I'll have a look in the coming days.

PR #895, opened 2021-07-28
Default message "using network interface..." is printed as info

EttoreVidotto:
Fix for issue #867


k0ekk0ek:
Hi @EttoreVidotto! Sorry for the long overdue reply. Changes LGTM. @eboasson, any reason we shouldn't merge this?

PR #896, opened 2021-07-28
Sanitize automatic enablement of Iceorxy

k0ekk0ek:
This PR makes the `ENABLE_SHM` option a tri-state that prefers compiling with support for Eclipse Iceoryx if nothing is specified. Meaning if `AUTO` and Iceoryx is not found, it'll compile without SHM. If `ON`, `REQUIRED` is passed to `find_package` resulting in an error if Iceoryx is not found. If `OFF`, no attempt to find Iceoryx is made. The philosophy here being: if the user specifies he wants support for shared memory and it's not available, the configure steps fails with a hard error. This also removes the Linux/macOS check, after all, if running on Windows, Iceoryx cannot be available (for now), so it's more feature based. A bit like `autoconf` is you will.

@MatthiasKillat, @sumanth-nirmal: Can you guys do a quick review too?


sumanth-nirmal (approved):
The change looks good and seems to do what it is supposed to do.

thijsmie:
We will approach this differently (and this now has conflicts).

Issue #891, opened 2021-07-21
Status mask is not effective if it is previously set to none


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
- GCC 7.5.0
- Cyclone DDS master (c1c32493ebe5ef5b0cb37d9845501a1282c4174f)
- Cyclone DDS CXX master [(544703f)](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/544703f72ad3d24d548be0906a082cf95ed3a94d)

**Observed result or behaviour:**

When we create a status condition for a reader and set the status mask to `none` first, and then set the status mask to some status (for eg. `requested_incompatible_qos`) and attach it to a waitset, then the wait set doesn't get triggered for this status condition. The below example should show the behavior

```cpp
TEST(tooch, event)
{
  using DDSType = test_msgs::msg::dds_::Strings_;
  dds::domain::DomainParticipant dp(0);
  dds::pub::Publisher pub(dp);
  dds::sub::Subscriber sub(dp);
  dds::topic::Topic<DDSType> topic(dp, "test");
  dds::pub::qos::DataWriterQos dw_qos{};
  dw_qos.policy<dds::core::policy::Durability>().kind(
    dds::core::policy::DurabilityKind::VOLATILE);
  dds::pub::DataWriter<DDSType> dw(pub, topic, dw_qos);
  dds::sub::qos::DataReaderQos dr_qos{};
  dr_qos.policy<dds::core::policy::Durability>().kind(
    dds::core::policy::DurabilityKind::TRANSIENT_LOCAL);
  dds::sub::DataReader<DDSType> dr(sub, topic, dr_qos);

  dds::sub::cond::ReadCondition rc(dr,
    dds::sub::status::SampleState::not_read());
  dds::core::cond::StatusCondition sc_w(dw);
  sc_w.enabled_statuses(dds::core::status::StatusMask::offered_incompatible_qos());
  dds::core::cond::StatusCondition sc_r(dr);
  sc_r.enabled_statuses(dds::core::status::StatusMask::none());
  sc_r.enabled_statuses(dds::core::status::StatusMask::requested_incompatible_qos());

  DDSType msg{};
  msg.string_value_("tooch");
  //dw.write(msg);

  dds::core::cond::WaitSet ws{};
  //ws.attach_condition(rc);
  ws.attach_condition(sc_w);
  ws.attach_condition(sc_r);

[...]

PR #900, opened 2021-08-04
Waitsets, listeners and masks

eboasson:
This PR changes — "fixes" would be a nicer way of putting it, but there is always the possibility of mistakes ... — the way waitsets, listeners and status masks interact. Some of this met the spec but caused trouble nonetheless, some did not meet the spec but caused suprisingly little trouble.

* The spec says that prior to listener invocation, the corresponding status is reset (preventing waitsets from triggering). Cyclone implemented this, but this caused complications for https://github.com/ros2/rmw_cyclonedds/pull/256. Sometimes it makes sense to reset the status, but there is definitely also value in supporting listeners without side-effects. This makes it configurable with the new `dds_lset_..._arg` functions.
* The status mask did too much: it did indeed allow one to configure which events should trigger waitsets, but it did this by suppressing too much: neither status updates nor listener invocations took place for statuses disabled by the mask. This was clearly a bug, e.g., https://github.com/eclipse-cyclonedds/cyclonedds/issues/891. (If you don't want a listener, don't install one; if you don't care about a status, don't look at it.)

Both referenced issues should be addressed by this PR.

In the process it also changes the implementation of the `DATA_ON_READERS` event because maintaining it in the subscriber in combination with the first point turned out to dramatically lower performance on an M1-based MBP (`ddsperf -TOU` in a throughput test dropping below 3MS/s is clearly unacceptable!). Now, for subscribers not attached to any waitset, it computes the `DATA_ON_READERS` status on demand from the subscriber's readers.

`DATA_ON_READERS` is supposed to be raised on along with `DATA_AVAILABLE` on some reader (of the subscriber), and supposed to be reset by the first operation on some reader (of the subscriber) that resets `DATA_AVAILABLE`. When it is attached to a waitset, this is what it does; when not attached to a waitset, it actually it returns `DATA_ON_READERS` if there is some reader that has `DATA_AVAILABLE` set.

Either behaviour seems usable for applications, and the change in behaviour seems acceptable considering that `D_O_R` only becomes really valuable for "group ordered" and "group coherent" data, which aren't supported yet anyway.

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -477,16 +477,18 @@ dds_get_guid (dds_entity_t entity, dds_guid_t *guid);
  *
  * @param[in]  entity  Entity on which the status has to be read.
  * @param[out] status  Returns the status set on the entity, based on the enabled status.
- * @param[in]  mask    Filter the status condition to be read (can be NULL).
+ * @param[in]  mask    Filter the status condition to be read, 0 means all statuses
  *
  * @returns A dds_return_t indicating success or failure.
  *
  * @retval DDS_RETCODE_OK
  *             Success.
  * @retval DDS_RETCODE_BAD_PARAMETER
- *             The entity parameter is not a valid parameter.
+ *             The entity parameter is not a valid parameter, status is a null pointer or
+ *             mask has bits set outside the status range.
  * @retval DDS_RETCODE_ILLEGAL_OPERATION
- *             The operation is invoked on an inappropriate object.
+ *             The operation is invoked on an inappropriate object or mask has status
+ *             bits set that are undefined for the type of entity.
  * @retval DDS_RETCODE_ALREADY_DELETED
  *             The entity has already been deleted.
  */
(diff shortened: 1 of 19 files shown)


dennis-adlink:
I agree with the changes in how the status mask works, the existing approach has always confused me somewhat. I've checked the code changes and couldn't find any issues (although I didn't look into the data-on-readers part in detail, as I don't have much experience in using this). 

The name of the `dds_lset_..._arg` functions confused me a bit, because of the extra option 'reset', which I didn't expect based on the name (but of course it is not really a problem, because the function signature makes clear what it does). 

PS: while testing the changes of this PR, I found another issue (#905) which I think is completely unrelated

eboasson:
@mauropasse I just realised that you may have missed this PR, which I think should solve the problems with the EventsExecutor on Cyclone. I would appreciate it if you could give it a try.

The default behaviour is unchanged, so to get the desired behaviour you need to change the way you configure the listeners. It is but a small change, e.g., instead of `dds_lset_data_available(listener, dds_listener_callback)`, you'll need `dds_lset_data_available_arg(listener, dds_listener_callback, &sub->user_callback_data, false)`: the `false` tells it not to reset the event on invoking the callback and as the argument pointer is now configurable per listener, you need to pass the correct listener argument in as well.

mauropasse:
Hi @eboasson, thanks for your work!
I've tested your branch, now I have succesful test results both with `EventsExecutor` as well as the waitset based `SingleThreadedExecutor`.
I created an internal PR to use the new APIs: https://github.com/irobot-ros/rmw_cyclonedds/pull/26
Still didn't have time to get through the inners of this PR, but great that it works!

PR #906, opened 2021-08-08
Make sure to use buffer with truncated name for pthread_setname_np

christophebedard:
Otherwise names with a length of > 16 characters (including null character) will silently not be used on Linux.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>


eboasson (approved):
Thanks!

PR #909, opened 2021-08-17
Duplicate increment of input buffer position upon array key readout

thijsmie:
Ran into this while fuzzy testing the Python serializer for key correctness: this statement is duplicated between the removed line and the function ``dds_is_get_bytes``, resulting in incorrect keys or even segfaults when a type has multiple keys and an array key occurs before another key.


eboasson (approved):
Thanks for catching that one!

PR #910, opened 2021-08-18
Subtle CDR key calculation fixes

thijsmie:
Fix some subtle ways the key calculation can go wrong with unlikely datatypes:
 * Incorrect behaviour when an array of bounded strings precedes a key
 * Incorrect behaviour when a sequence of Union or Struct precedes a key and is empty.
 * Incorrect behaviour when a sequence of bounded strings precedes a key and is not empty.
 * Fixed IDL not taking alignment into account when determining if a type has a key that is always <= 16 bytes (thanks @eboasson)
 * Allow small bounded strings to stay below 'MD5-hashing-threshold' in IDL.


eboasson (approved):
Ouch, but thank you very much!

PR #913, opened 2021-08-24
Fix the wrong description about the default value of ParticipantIndex

t0ny-peng:
This PR fixes a wrong description in the Config doc, on which I spent several hours to debug..


eboasson (approved):
I can see why this documentation bug was so costly. Thanks for persevering and submitting a fix!

It'd probably be a good thing to go over this document again and update it. Where there's one bug, there are usually more ... and it has not been updated in a very long time. I'll make an issue for that.

PR #919, opened 2021-08-30
Additional annotation parsing

reicheratwork:

### DIFF
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -61,6 +61,8 @@
 /* if no explicit default is specified and range is not covered */
 #define IDL_IMPLICIT_DEFAULT_CASE_LABEL (IDL_DEFAULT_CASE_LABEL | 2u)
 #define IDL_ENUMERATOR (1llu<<26)
+#define IDL_DEFAULT_ENUMERATOR (IDL_ENUMERATOR | 1u)
+#define IDL_IMPLICIT_DEFAULT_ENUMERATOR (IDL_DEFAULT_ENUMERATOR | 2u)
 #define IDL_DECLARATOR (1llu<<25)
 /* annotations */
 #define IDL_ANNOTATION (1llu<<24)
@@ -387,6 +389,7 @@ struct idl_enum {
   idl_node_t node;
   struct idl_name *name;
   idl_enumerator_t *enumerators;
+  idl_enumerator_t *default_enumerator;
   IDL_ANNOTATABLE(idl_extensibility_t) extensibility;
 };
 
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -283,6 +283,7 @@ struct idl_member {
   /* metadata */
   IDL_ANNOTATABLE(bool) key;
   IDL_ANNOTATABLE(bool) optional;
+  IDL_ANNOTATABLE(const idl_literal_t*) value;
   IDL_ANNOTATABLE(uint32_t) id;
 };
 
@@ -482,6 +483,7 @@ IDL_EXPORT bool idl_identifier_is(const void *node, const char *identifier);
 IDL_EXPORT const idl_name_t *idl_name(const void *node);
 IDL_EXPORT uint32_t idl_array_size(const void *node);
 IDL_EXPORT uint32_t idl_bound(const void *node);
+IDL_EXPORT const idl_literal_t *idl_default_value(const void *node);
 
 /* navigation */
 IDL_EXPORT void *idl_ancestor(const void *node, size_t levels);
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -282,6 +282,7 @@ struct idl_member {
   idl_declarator_t *declarators;
   /* metadata */
   IDL_ANNOTATABLE(bool) key;
+  IDL_ANNOTATABLE(bool) optional;
   IDL_ANNOTATABLE(uint32_t) id;
 };
 
--- a/src/idl/src/annotation.c
+++ b/src/idl/src/annotation.c
@@ -130,6 +130,40 @@ annotate_autoid(
 }
 
 static idl_retcode_t
+annotate_optional(
+  idl_pstate_t *pstate,
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  const idl_const_expr_t *const_expr;
+  idl_member_t *mem = (idl_member_t*)node;
+  bool value = true;
+
+  assert(annotation_appl);
+
+  if (!idl_is_member(node)) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@optional can only be assigned to members");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  } else if (mem->key.value) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@optional cannot be assigned to key members");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
+  if (annotation_appl->parameters) {
+    const_expr = annotation_appl->parameters->const_expr;
+    assert(const_expr);
+    value = ((const idl_literal_t*)const_expr)->value.bln;
+  }
+
+  mem->optional.annotation = annotation_appl;
+  mem->optional.value = value;
+
+  return IDL_RETCODE_OK;
+}
+
+static idl_retcode_t
 annotate_value(
   idl_pstate_t *pstate,
   idl_annotation_appl_t *annotation_appl,
@@ -250,8 +284,13 @@ annotate_key(
   }
 
   if (idl_mask(node) & IDL_MEMBER) {
-    ((idl_member_t *)node)->key.annotation = annotation_appl;
-    ((idl_member_t *)node)->key.value = key;
+    if (((idl_member_t*)node)->optional.value) {
+      idl_error(pstate, idl_location(annotation_appl),
+        "@key cannot be applied to optional members");
+      return IDL_RETCODE_SEMANTIC_ERROR;
+    }
+    ((idl_member_t*)node)->key.annotation = annotation_appl;
+    ((idl_member_t*)node)->key.value = key;
   } else if (idl_mask(node) & IDL_SWITCH_TYPE_SPEC) {
     ((idl_switch_type_spec_t *)node)->key.annotation = annotation_appl;
     ((idl_switch_type_spec_t *)node)->key.value = key;
@@ -430,12 +469,11 @@ static const idl_builtin_annotation_t annotations[] = {
       "instructs to automatically allocate identifiers to elements that have "
       "not been assigned a 32-bit unsigned identifiers explicitly.</p>",
     .callback = &annotate_autoid },
-#if 0
   { .syntax = "@annotation optional { boolean value default TRUE; };",
     .summary =
-      "<p>Set optionality on any element that makes sense to be optional.</p>",
-    .callback = annotate_optional },
-#endif
+      "<p>Indicates that the annotated member may be in a NULL state, not"
+      "containing any value.</p>",
+    .callback = &annotate_optional },
   { .syntax = "@annotation value { any value; };",
     .summary =
       "<p>Set a constant value to any element that may be given a constant "
--- a/src/idl/src/annotation.c
+++ b/src/idl/src/annotation.c
@@ -149,6 +149,10 @@ annotate_optional(
     idl_error(pstate, idl_location(annotation_appl),
       "@optional cannot be assigned to key members");
     return IDL_RETCODE_SEMANTIC_ERROR;
+  } else if (mem->value.annotation) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@optional cannot be assigned to members with explicit default values");
+    return IDL_RETCODE_SEMANTIC_ERROR;
   }
 
   if (annotation_appl->parameters) {
@@ -346,6 +350,51 @@ set_nested(
 }
 
 static idl_retcode_t
+annotate_default(
+  idl_pstate_t *pstate,
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  idl_const_expr_t *value;
+  idl_member_t *mem = (idl_member_t*)node;
+  idl_type_spec_t *mem_spec = mem->type_spec;
+
+  assert(annotation_appl);
+  assert(annotation_appl->parameters);
+
+  if (!idl_is_member(node)) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@default can only be assigned to members");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  } else if (mem->optional.value) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@default cannot be set on optional members");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
+  value = annotation_appl->parameters->const_expr;
+  assert(idl_is_literal(value));
+
+  /*check whether type of literal matches and falls inside spec of member*/
+  idl_type_t mem_type = idl_type(mem_spec);
+  if (mem_type != idl_type(value)) {
+    idl_retcode_t ret = IDL_RETCODE_OK;
+    idl_literal_t *literal = NULL;
+    if ((ret = idl_evaluate(pstate, value, mem_type, &literal)))
+      return ret;
+
+    assert(literal);
+    annotation_appl->parameters->const_expr = literal;
+    literal->node.parent = (idl_node_t*)annotation_appl->parameters;
+  }
+
+  ((idl_member_t *)node)->value.annotation = annotation_appl;
+  ((idl_member_t *)node)->value.value = annotation_appl->parameters->const_expr;
+
+  return IDL_RETCODE_OK;
+}
+
+static idl_retcode_t
 annotate_nested(
   idl_pstate_t *pstate,
   idl_annotation_appl_t *annotation_appl,
@@ -504,6 +553,11 @@ static const idl_builtin_annotation_t annotations[] = {
       "<p>Specify a data member is part of the key for the object whose type "
       "is the constructed data type owning this element.</p>",
     .callback = annotate_key },
+  { .syntax = "@annotation default { any value; };",
+    .summary =
+      "<p>Specify the value with which the annotated member should be default"
+      "initialized.</p>",
+    .callback = &annotate_default },
 #if 0
   { .syntax = "@annotation must_understand { boolean value default TRUE; };",
     .summary =
@@ -511,10 +565,6 @@ static const idl_builtin_annotation_t annotations[] = {
       "making use of that piece of data.</p>",
     .callback = annotate_must_understand },
   /* units and ranges */
-  { .syntax = "@annotation default { any value; };",
-    .summary =
-       "<p>Specify a default value for the annotated element.</p>",
-    .callback = annotate_default },
   { .syntax = "@annotation range { any min; any max; };",
     .summary =
       "<p>Specify a range of allowed value for the annotated element.</p>",
--- a/src/idl/src/annotation.c
+++ b/src/idl/src/annotation.c
@@ -395,6 +395,25 @@ annotate_default(
 }
 
 static idl_retcode_t
+annotate_default_literal(
+  idl_pstate_t *pstate,
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  assert(pstate);
+
+  if (!idl_is_enumerator(node)) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@default_literal can only be assigned to enumerators");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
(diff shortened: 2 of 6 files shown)


k0ekk0ek:
Thanks @reicheratwork! Better to have a member in the corresponding nodes to indicate if something's optional so that each backend can use it. Also, would be nice to verify `@optional` can only be assigned to the correct types (probably structs and unions only?).

reicheratwork:
> Thanks @reicheratwork! Better to have a member in the corresponding nodes to indicate if something's optional so that each backend can use it. Also, would be nice to verify `@optional` can only be assigned to the correct types (probably structs and unions only?).

Can't @optional not be assigned to basic types as well?
If i recall correctly, it is an annotation to members, right?

reicheratwork:
@k0ekk0ek in my most recent push, I added an idl_boolean_t is_optional to the idl_member_t class and created a utility function checking for this, so that the different backend implementations will not have to create their own

src/idl/src/annotation.c
```diff
@@ -133,6 +133,31 @@ annotate_autoid(
...
+
+  assert(annotation_appl);
+  if (annotation_appl->parameters
+   && annotation_appl->parameters->const_expr
+   && idl_mask(annotation_appl->parameters->const_expr) != (IDL_LITERAL | IDL_BOOL)) {
+        idl_error(pstate, idl_location(annotation_appl),
+          "@optional cannot have non-boolean values assigned");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
```
dennis-adlink, line 157: Shouldn't `((idl_member_t *)node)->is_optional` be set at this point?

src/idl/src/annotation.c
```diff
@@ -261,6 +286,29 @@ annotate_key(
...
+  if (annotation_appl->parameters
+   && annotation_appl->parameters->const_expr) {
+        idl_error(pstate, idl_location(annotation_appl),
+          "@default can only be assigned to values of the following types: ");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }*/
+
+  return IDL_RETCODE_OK;
+}
+
```
dennis-adlink, line 311: Maybe better to move the @default implementation to a different PR (including the changes below for @default in the annotations array)? Or fix the implementation in this PR (and change the PR title)

src/idl/src/parser.y
```diff
@@ -812,8 +812,11 @@ case_label:
   ;
 
 element_spec:
-    type_spec declarator
-      { TRY(idl_create_case(pstate, LOC(@1.first, @2.last), $1, $2, &$$)); }
+    annotations type_spec declarator
+      { TRY(idl_create_case(pstate, LOC(@1.first, @3.last), $2, $3, &$$));
+        TRY_EXCEPT(idl_annotate(pstate, $$, $1), free($$)); }
+  | type_spec declarator
+      { TRY(idl_create_case(pstate, LOC(@1.first, @2.last), $1, $2, &$$));}
```
dennis-adlink, line 819: @optional cannot be applied to union members

src/idl/src/tree.c
```diff
@@ -772,6 +772,59 @@ uint32_t idl_bound(const void *node)
...
+bool idl_is_optional(const void *node)
+{
+  if (idl_is_declarator(node)) {
+    const idl_node_t *parent = idl_parent(node);
+    if (idl_is_member(parent))
+      return idl_is_optional(parent);
+  } else {
+    idl_annotation_appl_t *annotation_appl = NULL;
+    IDL_FOREACH(annotation_appl, ((const idl_node_t*)node)->annotations) {
+      if (0 != strcmp(idl_identifier(annotation_appl->annotation), "optional"))
```
dennis-adlink, line 784: Although we don't have a coding style guide and slightly different styles are used through the code base, I'd prefer a consistent coding style in a single code file, so `strmp(..) != 0` or just `strcmp(...)`

src/idl/src/tree.c
```diff
@@ -772,6 +772,59 @@ uint32_t idl_bound(const void *node)
...
+      } else {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+const idl_literal_t *idl_has_default(const void *node)
+{
```
dennis-adlink, line 801: Move to a separate PR?

dennis-adlink: With the function name `idl_has_default` I would expect a bool that indicates if the node has a default (which is what the function `idl_is_default` below does). Maybe a better name would be `idl_get_default_value` or something similar?

src/idl/src/tree.c
```diff
@@ -772,6 +772,59 @@ uint32_t idl_bound(const void *node)
...
+  return NULL;
+}
+
+bool idl_is_default(const void *node)
+{
+  const idl_literal_t *lit = idl_has_default(node);
+
+  if (lit
+   && idl_mask(lit) == (IDL_LITERAL | IDL_ANY))
+    return lit->value.bln;
```
dennis-adlink, line 823: Does this also work if an explicit default value `0` is applied to a member (`@default(0)`), this function should return true in that case?

src/idl/tests/annotation.c
```diff
@@ -67,6 +67,124 @@ CU_Test(idl_annotation, id_member)
...
+  assert(u);
+  IDL_FOREACH(uc, u->cases) {
+    CU_ASSERT_PTR_NOT_NULL_FATAL(uc);
+    CU_ASSERT_FATAL(idl_is_case(uc));
+    CU_ASSERT_EQUAL(idl_is_optional(uc), n++ > 0);
+  }
+
+  idl_delete_pstate(pstate);
+
+  const char *errstr[] = {"@optional stuct s {\n"
```
dennis-adlink, line 136: Typo `stuct` ?

src/idl/tests/annotation.c
```diff
@@ -67,6 +67,124 @@ CU_Test(idl_annotation, id_member)
...
+                          "  char c;\n"
+                          "};\n",
+                          "enum e {\n"
+                          "  e_0,\n"
+                          "  @optional e_1,\n"
+                          "  e_2,\n"
+                          "  e_3\n"
+                          "};\n"};
+
```
dennis-adlink, line 145: Maybe add a test for invalid arguments, like `@optional(1)`, `@optional('true')` ?

src/idl/tests/annotation.c
```diff
@@ -67,6 +67,124 @@ CU_Test(idl_annotation, id_member)
...
+  idl_declarator_t *d;
+  idl_union_t *u;
+  idl_case_t *uc;
+  const char structstr[] = "struct s {\n"
+                           "  char c_none;\n"
+                           "  @optional(false) char c_false;\n"
+                           "  @optional char c_default;\n"
+                           "  @optional(true) char c_true;\n"
+                           "  @optional char c_1_default, c_2_default;\n"
+                           "};\n";
```
dennis-adlink, line 89: Add test some cases for other member types, e.g. aggregated types, sequences, arrays?

dennis-adlink (commented):
I've reviewed these changes and found a few issues that need a fix, mainly the incomplete @default implementation and using the is_optional field instead of iterating through the annotations in `idl_is_optional`. And some nit-picking on the consistency of the coding style.

reicheratwork:
> I've reviewed these changes and found a few issues that need a fix, mainly the incomplete @default implementation and using the is_optional field instead of iterating through the annotations in `idl_is_optional`. And some nit-picking on the consistency of the coding style.

Thanks for taking the time to look over this PR! I will take your comments into account, though I will probably rebase the current changes onto #923 when it has been merged into master. After which some of the new features' implementation may be quite different.

k0ekk0ek:
The `IDL_ANNOTATABLE` changes are in the master branch.

reicheratwork:
> The `IDL_ANNOTATABLE` changes are in the master branch.

Super!
I will rebase this onto the latest changes and then rework it into the IDL_ANNOTATABLE syntax

src/idl/include/idl/tree.h
```diff
@@ -299,6 +299,7 @@ struct idl_member {
   idl_declarator_t *declarators;
   /* metadata */
   idl_boolean_t key;
+  idl_boolean_t is_optional;
```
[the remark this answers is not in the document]
reicheratwork: You are correct, in the newest iteration this is fixed.
However, some member names cannot be chosen as they are reserved keywords (such as default), so there I may need to use little workarounds.

src/idl/src/parser.y
reicheratwork: You are correct, this has been removed.

src/idl/tests/annotation.c
reicheratwork: You are correct.

src/idl/tests/annotation.c
reicheratwork: The number of test cases will be expanded a lot in the next iteration.

src/idl/tests/annotation.c
reicheratwork: Will do.

src/idl/src/annotation.c
```diff
@@ -133,6 +133,31 @@ annotate_autoid(
...
 }
 
+static idl_retcode_t
+annotate_optional(
+  idl_pstate_t *pstate,
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  if (!idl_is_member(node)
+   && !idl_is_case(node)) {
```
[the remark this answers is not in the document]
reicheratwork: This  is corrected in the new iteration.

src/idl/src/annotation.c
reicheratwork: This is no longer relevant, after rebasing I have changed to using the IDL_ANNOTATABLE macro

src/idl/src/annotation.c
```diff
@@ -306,6 +341,85 @@ set_nested(
...
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+  mem_spec = ((idl_member_t*)node)->type_spec;
+
+  value = annotation_appl->parameters->const_expr;
+  if (!value || !idl_is_literal(value)) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@default requires a value");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
```
k0ekk0ek, line 368: This can be an `assert`. The parser checks the syntax and since there's no default, it should throw an error if there's no value specified. Probably good to add a small test for that (if there's none yet). If the test fails, we (or I) should fix the bug there.

src/idl/src/annotation.c
```diff
@@ -306,6 +341,85 @@ set_nested(
...
+    default:
+    {
+      idl_mask_t value_type = (idl_mask(value) ^ IDL_LITERAL);
+      if (idl_mask(mem_spec) != value_type) {
+        idl_error(pstate, idl_location(annotation_appl),
+          "value type of @default does not match member");
+        return IDL_RETCODE_SEMANTIC_ERROR;
+      }
+    }
+  }
```
k0ekk0ek, line 395: This too, should have been done by the parser already.

src/idl/src/annotation.c
```diff
@@ -129,6 +129,41 @@ annotate_autoid(
...
+  if (annotation_appl->parameters) {
+    const_expr = annotation_appl->parameters->const_expr;
+    assert(const_expr);
+    if (idl_mask(const_expr) != (IDL_BOOL | IDL_LITERAL)) {
+      idl_error(pstate, idl_location(annotation_appl),
+        "@optional cannot have non-boolean values assigned");
+      return IDL_RETCODE_SEMANTIC_ERROR;
+    } else {
+      value = ((const idl_literal_t*)const_expr)->value.bln;
+    }
```
k0ekk0ek, line 158: This is probably checked by the parser already. Like in one of my earlier comments, let's add a couple of small tests and fix the parser if it doesn't behave correctly already.

src/idl/src/annotation.c
```diff
@@ -306,6 +341,85 @@ set_nested(
...
+  idl_pstate_t *pstate,
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  assert(pstate);
+
+  if (!idl_is_enumerator(node)) {
+        idl_error(pstate, idl_location(annotation_appl),
+          "@default_literal can only be assigned to member nodes");
+        return IDL_RETCODE_SEMANTIC_ERROR;
```
k0ekk0ek, line 414: Indentation.

src/idl/include/idl/tree.h
```diff
@@ -378,6 +380,7 @@ struct idl_enumerator {
      mapped to a native data type capable of representing a maximally-sized
      enumeration */
   IDL_ANNOTATABLE(uint32_t) value;
+  IDL_ANNOTATABLE(bool) is_default;
```
k0ekk0ek, line 383: In closing the `enum` declaration we should check if there's only one default. Plus, I'm thinking it may actually make more sense to have a `default_enumerator` member in the `enum` too/instead?

k0ekk0ek: The `default_enumerator` could then point to the first enumerator is not explicit default is provided? (Probably works like that, but I'm not a 100% sure)

k0ekk0ek: Maybe introduce `IDL_DEFAULT_ENUMERATOR` that equals `IDL_ENUMERATOR | 1u`. The (implicit) default case for unions works like that as well. Don't know if it makes sense, but it might be worth looking into that.

src/idl/src/annotation.c
reicheratwork: You're correct that this should not really be checked here.
I will check whether the parser catches this, or if this requires a fix someplace else

src/idl/src/annotation.c
reicheratwork: I think that there might be a misunderstanding here, this is checking whether the type in the literal (the parameter field of the `@default` annotation) is compatible with that of the type it is being annotated to (so it should reject something like this: `@default('a') long l;`

src/idl/include/idl/tree.h
reicheratwork: My attempt here was to modify the existing interface as little as possible, but your approach should result in less duplicated functionality, as the different backends will not have to implement their own checks

reicheratwork: I will look into the default member approach

src/idl/src/annotation.c
reicheratwork: You're correct that this should not really be checked here.
I will check whether the parser catches this, or if this requires a fix someplace else

src/idl/src/annotation.c
reicheratwork: This is correctly being checked by the parser at a previous step, and therefore I will replace this check with a simpler assert

src/idl/src/annotation.c
reicheratwork: This is correctly being checked by the parser at a previous step, and therefore I will replace this check with a simpler assert

src/idl/src/annotation.c
reicheratwork: Will be fixed

src/idl/tests/annotation.c
```diff
@@ -45,6 +45,196 @@ parse_string(uint32_t flags, const char *str, idl_pstate_t **pstatep)
...
+    {IDL_RETCODE_OK,                  true,   "struct s { @optional(true) char c; };"                 },
+    {IDL_RETCODE_OK,                  true,   "struct s { @optional(true) char c_1, c_2, c_3; };"     },
+    {IDL_RETCODE_OK,                  true,   "struct s { @optional(true) char c_1, c_2, c_3; };"     },
+    {IDL_RETCODE_OK,                  true,   "struct s { @optional sequence<double> s_d; };"         },
+    {IDL_RETCODE_OK,                  true,   "typedef sequence<long> seq_long;\n"
+                                              "struct s { @optional seq_long s_l_a[15]; };"           },
+    {IDL_RETCODE_SEMANTIC_ERROR,      false,  "@optional struct s { char c; };"                       },
+    {IDL_RETCODE_ILLEGAL_EXPRESSION,  false,  "struct s { @optional(\"true\") char c; };"             },
+    {IDL_RETCODE_ILLEGAL_EXPRESSION,  false,  "struct s { @optional(123) char c; };"                  },
+    {IDL_RETCODE_ILLEGAL_EXPRESSION,  false,  "struct s { @optional(0.1) char c; };"                  },
```
k0ekk0ek, line 98: I think it'd be nicer to cover these in more generic annotation tests(?)

src/idl/tests/annotation.c
```diff
@@ -45,6 +45,196 @@ parse_string(uint32_t flags, const char *str, idl_pstate_t **pstatep)
...
+    {IDL_RETCODE_SEMANTIC_ERROR,      false,  "enum e {  e_0, @optional e_1};"                        },
+    {IDL_RETCODE_SYNTAX_ERROR,        false,  "union u switch (long) { case 0: @optional double d; };"}
```
k0ekk0ek, line 100: Same for this one because we're actually testing the syntax rather than something `@optional` specific.

src/idl/tests/annotation.c
reicheratwork: My reasoning behind putting the tests for the correct parameter type in the annotation is that this is something that is unique to the @optional annotation, and in my opinion, having a whole forest of different checks of matching the parameter type with the annotation type would be less descriptive.

src/idl/tests/annotation.c
reicheratwork: Same reasoning as previous comment.

src/idl/include/idl/tree.h
reicheratwork: Changed to a default_enumerator pointer in the idl_enum class

src/idl/tests/annotation.c
```diff
@@ -45,6 +45,210 @@ parse_string(uint32_t flags, const char *str, idl_pstate_t **pstatep)
...
+    {"struct s { @default(123) string str; };",              IDL_RETCODE_SEMANTIC_ERROR,  false, IDL_NULL,    NULL}, //parameter type mismatch (int vs string)
+    {"struct s { @default(\"false\") boolean b; };",         IDL_RETCODE_SEMANTIC_ERROR,  false, IDL_NULL,    NULL}, //parameter type mismatch (string vs bool)
+    {"struct s { @default(123) boolean b; };",               IDL_RETCODE_SEMANTIC_ERROR,  false, IDL_NULL,    NULL}, //parameter type mismatch (int vs bool)
+    {"union u (switch long) {\n"
+     "case 1: @default(123) short s;\n"
+     "case 2: string s; };",                                 IDL_RETCODE_SYNTAX_ERROR,    false, IDL_NULL,    NULL}, //unions do not allow defaults on values
+    /* skipping this test as idl_create_annotation_appl leaks memory if idl_resolve cannot resolve the scoped name (https://github.com/eclipse-cyclonedds/cyclonedds/issues/950)
+      {"@default(e_0) enum e { e_0, e_1, e_2, e_3 };",         IDL_RETCODE_SEMANTIC_ERROR,  false, IDL_NULL,    NULL}  //setting default on enums is done through @default_literal
+    */
```
k0ekk0ek, line 187: Better enable this and indicate the test is known to fail using the [WILL_FAIL property](https://cmake.org/cmake/help/latest/prop_test/WILL_FAIL.html). Then, when the fix for #950 is in, the test will pass (causing the build to fail). The advantage being that we're actively notified when a change is required.

src/idl/tests/annotation.c
```diff
@@ -45,6 +45,210 @@ parse_string(uint32_t flags, const char *str, idl_pstate_t **pstatep)
...
+          CU_ASSERT_EQUAL(idl_mask(en), IDL_ENUMERATOR);
+        }
+      }
+    }
+
+    idl_delete_pstate(pstate);
+  }
+}
+
+CU_Test(idl_annotation, enum_default)
```
k0ekk0ek, line 236: If you can find the time, please rename the testcase to `default_literal`. That way it's directly obvious in the logs/terminal what this test is meant to test (not that it's super hard to guess, but still).

src/idl/tests/annotation.c
reicheratwork: The issue is not here that the test fails, it is that there is a memory leak that is triggered by this test, and that the builds in Azure that check for memory leaks will fail if this test is done.

src/idl/tests/annotation.c
reicheratwork: Will do.

src/idl/tests/annotation.c
k0ekk0ek: I know, but CTest will mark it as a failure. Since `WILL_FAIL` is a CTest property, it'll get us the desired behavior right? If you prefer you may set the test to disabled. It's just that commenting it out only ensures it's never looked after again. At least, there's tons of commented out tests (or test ideas) for libidl that would've caught a bunch of things but they were never looked at or thought of again... I prefer to have some kind of notification that something is still off, at least that way no one forgets.

k0ekk0ek (approved):

k0ekk0ek:
Just a couple of warnings in the tests, if those are solved, I'll happily merge

reicheratwork:
> Just a couple of warnings in the tests, if those are solved, I'll happily merge

That one escaped my notice, I will fix it immediately.

PR #922, opened 2021-09-01
Add IDL compiler option to specify the output directory for the generated code (fixes #801)

### LINKED ISSUE #801: IDL compiler option to specify the output directory for the generated code
## Brief feature description

Currently, the IDL compiler doesn't have an option to specify the output directory for the generated code, the compiler generates the code in the directory from where the compiler is invoked. It will benefit by having a compiler option to specify the output directory. Is it possible to add this option to `idlc`?

The old java based compiler has an option `-d` to specify the output directory  (IIRC)

## Detailed information

In some use-cases its common to have IDL files in one place and the generated code in another place (usually somewhere in the build directory), instead of moving the generated files manually to the desired place its beneficial to use the compiler option to directly generate the code in the desired location.

obiltschnig:
This PR adds a `char *outdir;` member to `idl_pstate`, along with option handling and setting the output directory in `idlc_generate`. A separate PR for `cyclonedds-cxx` will be submitted as well.

### DIFF
--- a/src/idl/include/idl/processor.h
+++ b/src/idl/include/idl/processor.h
@@ -56,6 +56,7 @@ struct idl_pstate {
   idl_file_t *paths; /**< normalized paths used in include statements */
   idl_file_t *files; /**< filenames used in #line directives */
   idl_source_t *sources;
+  char *outdir; /**< output directory */
   idl_scope_t *global_scope, *annotation_scope, *scope;
   void *directive;
   idl_node_t *builtin_root, *root;
--- a/src/idl/src/processor.c
+++ b/src/idl/src/processor.c
@@ -226,6 +226,7 @@ void idl_delete_pstate(idl_pstate_t *pstate)
     /* buffer */
     if (pstate->buffer.data)
       free(pstate->buffer.data);
+    free(pstate->outdir);
     free(pstate);
   }
 }
--- a/src/tools/idlc/src/generator.c
+++ b/src/tools/idlc/src/generator.c
@@ -356,7 +356,12 @@ idlc_generate(const idl_pstate_t *pstate)
   }
 
   file = sep ? sep + 1 : path;
-  if (idl_isabsolute(path) || !sep)
+  if (pstate->outdir) {
+    if (!(dir = idl_strdup(pstate->outdir))) {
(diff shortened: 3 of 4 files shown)


k0ekk0ek:
Hi @obiltschnig. Thanks for the PR. We've actually had some requests to add this from others, so the feature is very welcome :+1: There's two things I'd like to change before merging though.
1. The `outdir` is better not placed in the `pstate` structure. The compiler basically consists of two parts, the parser and the tool itself. The latter uses the former and ties it to a specific backend. e.g. libidlcxx in the case of C++. The `outdir` is very specific to the tool. The parser (libidl) is meant to be usable as a standalone IDL parser too and I'd like to keep the separation between the two very strict. Of course, this raises a slight problem for options, like the output directory, that are useful to multiple backends. We could make a separate options struct and pass that to the `generate` function that's called in the backends(?)
2. Currently, if you specify an absolute path it outputs to `<basename>.h`. If a relative path, e.g. `foo/bar/baz.idl` is provided the compiler outputs to `foo/bar/<basename>.h`. I think we should keep that intact even when specifying an output directory(?) It's debatable. But some of our users have files with the same basename and I can see it being very useful to them if the output directory they specify is the base directory and the same directory structure is created there. Perhaps you have thoughts on this?

Thanks

k0ekk0ek:
@obiltschnig, I've been working on this myself for a bit and I think we should allow the user to specify a prefix directory. So something like `idlc -p foo bar.idl`. In the example a file named `bar.c` would be generated in `foo`. Which basically does the same thing as your patch, but would generate `foo/baz/bar.c` for `idlc -p foo baz/bar.idl`. In my copy I'm exposing the config struct for idlc so that generators have access to them. There's a couple of other issues I have to tackle too before I can submit a PR though.

obiltschnig:
@k0ekk0ek sorry for not getting back earlier. Being able to specify a prefix directory in addition to setting the output directory directly would be nice. In my case, I'm putting the `.idl` files in an `idl` directory, but need the generated files to be in `src` or `include`. However, for now my project has postponed the transition from OpenSplice to Cyclone due to other missing features, so no urgency here.

Issue #926, opened 2021-09-03
ddsrt_malloc size is incorrect in find_own_ip


12li-wei:
in find_own_ip:
    maxq_list = ddsrt_malloc (max_interfaces * sizeof (*interfaces));
is sizeof(*interfaces) correct?

I think it should be sizeof (*maxq_list)

eboasson:
Absolutely!

It has not been causing problems because `sizeof(void*)` ≥ `sizeof(size_t)`, but that doesn't make it right.

Do you want to make a PR for it? In that case, you also need to fill out the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) (you only need to do that once) before we are allowed to accept the contribution. There is a lot more information on the [Eclipse Foundation's process for contributing to a project](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git).

eboasson:
Fixed via #928.

PR #928, opened 2021-09-03
ddsrt_malloc size for maxq_list is incorrect in find_own_ip

12li-wei:
ddsrt_malloc size for maxq_list is incorrect in find_own_ip fix bug #926


eboasson (approved):
Thanks!

PR #929, opened 2021-09-03
Improve diagnostic when idlc loads plugin

jobol:


k0ekk0ek:
Hi @jobol. I guess we can improve on the error message. Probably best to reflect what went wrong in the return code and print a message from `idlc.c` in that case and only differentiate between failing to load the `generate` symbol and the library. Should be enough(?) Maybe printing a more detailed error message can be done, but we'd need to improve the message under Windows too. There's some code in `ddsrt` that we can use as an example. I'm also interested in the why. Are you developing a custom backend or are you using C++ or Python and experience problems loading the plugin and this would help users(?)

jobol:
Hi @k0ekk0ek I'm studying how to bridge a DDS system with our solution. I found your idlc compiler when looking around at DDS solutions. That idlc is really good IMHO and its support for plugin is probably well suited for my needs. So when working to implment a plugin, based on the existing generator, it was difficult to know what functions was missing in my plugin that wasn't able to load. dlerror is good because it prints the missing symbol.

For windows, perhaps the below snippet could work:
```
   FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ALLOCATE_BUFFER, NULL, GetLastError(), 0, (LPTSTR) & msg_buf, 0, NULL);
```

But if you try to load more than one plugin, it is not possible to get the right diagnostic. The standard function access can not be used because dlopen implement a path search that would be a pain to emulate.

k0ekk0ek:
Thanks @jobol! That's actually a nice compliment, because that's exactly why it's setup the way it is. I'm glad it's useful to someone by itself :+1:. I see why the addition would be useful. I think we can take the dlerror implementation that's there in `ddsrt`, replace the part where it calls `ddsrt_strerror_r` with the actual Windows code and that'll work just fine(?) I'd set a thread-specific variable from the openlib function if it fails, provide a function to retrieve the error in a platform agnostic way and print it from idlc.

Of course, the amount of code duplication between `ddsrt` and `idl`/`idlc` is yet again increasing, so I'll (or we as a project) will have to come up with a way to reduce it, but let's worry about that later.

jobol:
You must first decide how to open the plugin. In both implementation, there is a fallback to a second load. It is not compatible with error report. Either the name looks like a path (directory separator or known extension) or the name is build from indication but using the 2 is an issue here for reporting the error.

k0ekk0ek:
I really like the possibility for users to simply specify `c`, `cxx` or `py`. It's probably best if we just `fprintf` for now from `plugin.c`. Eventually using `idl_error` would be best, but before that's usable from the actual generator (currently it's only used within `libidl`) that function has to be modified to take location flags instead of always requiring a location.

Also, it maybe useful to give some background on how `idlc`, `libidl`, etc are intended. `libidl` is intended as a standalone thread-safe IDL parsing library that's useful to any kind of project requiring an IDL parser. `idlc` is specifically intended for use within Cyclone DDS. If `idlc` works for you right now with some minor tweaks that don't impact the way it functions for Cyclone DDS, I'm very happy to apply those. However, should you want to have more exotic features in the actual tool, you may be better served by reusing `libidl` and `idlpp` (fork of mcpp with lots of fixes that is purposely kept as close to the original as possible) and copying the sources to `idlc`. Especially because things like #936 are about to land.

Anyway, thought I'd give some background info. If you implement using `dlerror` on *nix platforms and use the `FormatMessage` stuff on Windows, I'm happy to merge the changes.

Small sidenote: Before I can merge any contribution, we do require the Eclipse Contributor Agreement to be signed. If you click the Details link next to the check it'll guide you through the process.

Thanks :+1:

jobol:
> I really like the possibility for users to simply specify `c`, `cxx` or `py`.

I really like it too.

Thank you for the detailed explanation. It helps much.

After that interesting discussion I will submit a patch in the coming weeks.

I'm not closing, I could but prefer let it opened until my patch.

k0ekk0ek:
We can leave it open :+1:

src/tools/idlc/src/plugin.c
```diff
@@ -105,17 +127,34 @@ idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang)
...
+    handle = openlib(path);
+  } else {
+    handle = openlib(lang);
+    if (handle) {
+      path = lang;
+    } else {
+      handle = openlib(path);
+    }
+  }
+  if (handle) {
```
k0ekk0ek, line 141: I think trying to `openlib(path)` is more than enough since `path` is assigned the "generated" library name.

src/tools/idlc/src/plugin.c
```diff
@@ -105,17 +127,34 @@ idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang)
...
     generate = loadsym(handle, "generate");
     if (generate) {
       plugin->handle = handle;
       plugin->generate = generate;
       plugin->generator_options = loadsym(handle, "generator_options");
       plugin->generator_annotations = loadsym(handle, "generator_annotations");
     } else {
+      fprintf(stderr, "no symbol 'generate' found in %s\n", path);
```
k0ekk0ek, line 149: (**nitpick**) please change to "Symbol 'generate' not found in %s\n".

src/tools/idlc/src/plugin.c
```diff
@@ -105,17 +127,34 @@ idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang)
...
       closelib(handle);
     }
   }
+  else {
+    char errmsg[300];
+    liberror(errmsg, sizeof(errmsg));
+    fprintf(stderr, "can't load %s: %s\n", path, errmsg);
```
k0ekk0ek, line 156: (**nitpick**) please change to "Cannot load generator %s\n".

k0ekk0ek (changes requested):

k0ekk0ek:
Thanks @jobol. Looks mostly fine, if the last changes are in I'm more than happy to merge.

jobol:
I'm going to update the pull request tomorow

src/tools/idlc/src/plugin.c
jobol: Yes I agree. I'll change that.

src/tools/idlc/src/plugin.c
jobol: I'm going to keep both %s: one for the effective name used in loadlib and one for the diagnostic message

jobol:
IMHO failures are not because of the patch. are they?

k0ekk0ek (approved):
Sorry @jobol, this escaped my attention. No, they're not. I'll merge the changes. Thanks!

PR #931, opened 2021-09-06
Cache Eclipse Iceoryx

k0ekk0ek:
This PR caches Eclipse Iceoryx builds if enabled for that pipeline and pins the version use to v1.0.1. By setting the `iceoryx_branch` and `iceoryx_repository` variables the respective version and origin can be overruled.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -73,15 +73,20 @@ steps:
   - template: /.azure/templates/install-conan.yml
   - bash: |
       set -e -x
-      if [[ "${ICEORYX}" = 'on' ]] ; then
-        sudo apt install libacl1-dev libncurses5-dev pkg-config && \
-          git clone https://github.com/eclipse-iceoryx/iceoryx.git && \
-          cd iceoryx && \
-          cmake -Bbuild -Hiceoryx_meta -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=${BUILD_TYPE} && \
-          cmake --build build && \
-          sudo cmake --build build --target install
-      fi
-    name: iceoryx
+      sudo apt install libacl1-dev libncurses5-dev pkg-config
+      git clone --depth 1 \
+                --branch "${ICEORYX_BRANCH:-v1.0.1}" \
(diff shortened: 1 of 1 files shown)


.azure/templates/build-test.yml
```diff
@@ -71,24 +71,41 @@ steps:
       path: $(pip_cache)
     name: cache_pip
   - template: /.azure/templates/install-conan.yml
+  - task: Cache@2
+    inputs:
+      key: iceoryx | $(iceoryx_repository) | $(iceoryx_branch) | $(Agent.OS) | $(arch) | $(cc) | $(build_type)
```
eboasson, line 76: The CI run says
```
Entry found at fingerprint: `iceoryx|$(iceoryx_repository)|$(iceoryx_branch)|Linux|x86_64|gcc-10|Debug`
```
which makes me think that something is amiss with the expansion of variables in the key. Surely that should have been like "iceoryx|https://github.com/eclipse-iceoryx/iceoryx.git|v1.0.1|..."?

k0ekk0ek: Probably depends on whether the variables are available(?) Saw the same message but figured it doesn't matter much as the key is the same if the variables aren't there and it'd change if they are, which is basically what I was trying to accomplish: make sure there isn't a cache hit if someone specifies one of the variables. Of course, we could just set them globally, that'd work too.

eboasson: > it doesn't matter much as the key is the same if the variables aren't there and it'd change if they are

Fair enough, but then one does need to be certain that this really is the case. Sometimes the variables are used in lower case and sometimes in upper case, so it isn't immediately obvious (to me) that they will be.

But if that all works fine, then I'm happy with it.

k0ekk0ek: > Sometimes the variables are used in lower case and sometimes in upper case
That bit depends on whether they're referenced as environment variables or pipeline variables. Every pipeline variable is automatically made available in the environment in uppercase. Then there's also yaml compile-time and runtime variables... It probably makes total sense, but it does get a little complex

k0ekk0ek: I'll remove the caching part. It saves some time, but we can worry about that later too.

PR #932, opened 2021-09-06
Fix dlopen_path test on normal VS/Xcode

eboasson:
This restores the directories used before the MinGW support was added. For some reason, in the CI the VS and Xcode builds do store the output in directories not suffixed with Release or Debug, but this is not usually the case.  As a consequence, on a typical build with those build systems, this test failed.

### DIFF
--- a/src/ddsrt/tests/CMakeLists.txt
+++ b/src/ddsrt/tests/CMakeLists.txt
@@ -36,9 +36,32 @@ endif()
 
 if(HAVE_DYNLIB)
   list(APPEND sources dynlib.c)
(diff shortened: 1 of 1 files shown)


eboasson:
Note: also needs backporting to `releases/0.8.x` as it is a show stopper for ROS 2 (https://github.com/ros2/ros2/pull/1174)

k0ekk0ek:
Well, I tried to clean this up, guess I'll have to go the whole nine yards. There's a lot wrong with what was originally there and what's restored with this PR, but it did "work". For the short-term, let's restore it.

k0ekk0ek (approved):

PR #809, opened 2021-05-25
Dereference as void* to avoid crash with -O2

k0ekk0ek:
This fixes #808.

Signed-off-by: Jeroen Koekkoek <jeroen@koekkoek.nl>

Issue #808, opened 2021-05-22
Segmentation fault during idlc generating


paeunt:
Hi. I got this error when building CyclongDDS with 'Release' build type. Both master branch and iceoryx branch.But it is OK with 'Debug' build type. BTW, The GCC version is 11.1. 

I don't have a clue. 

`[ 80%] Building C object src/tools/idlpp/CMakeFiles/idlpp.dir/src/system.c.o
[ 80%] Linking C static library ../../../lib/libidlpp.a
[ 80%] Built target idlpp
[ 80%] [BISON][parser] Building parser with bison 3.7.6
[ 80%] Building C object src/idl/CMakeFiles/idl.dir/src/symbol.c.o
[ 81%] Building C object src/idl/CMakeFiles/idl.dir/src/directive.c.o
[ 81%] Building C object src/idl/CMakeFiles/idl.dir/src/expression.c.o
[ 82%] Building C object src/idl/CMakeFiles/idl.dir/src/file.c.o
[ 82%] Building C object src/idl/CMakeFiles/idl.dir/src/processor.c.o
[ 82%] Building C object src/idl/CMakeFiles/idl.dir/src/scanner.c.o
[ 83%] Building C object src/idl/CMakeFiles/idl.dir/src/string.c.o
[ 83%] Building C object src/idl/CMakeFiles/idl.dir/src/annotation.c.o
[ 83%] Building C object src/idl/CMakeFiles/idl.dir/src/scope.c.o
[ 84%] Building C object src/idl/CMakeFiles/idl.dir/src/tree.c.o
[ 84%] Building C object src/idl/CMakeFiles/idl.dir/src/visit.c.o
[ 85%] Building C object src/idl/CMakeFiles/idl.dir/src/thread.c.o
[ 85%] Building C object src/idl/CMakeFiles/idl.dir/src/print.c.o
[ 85%] Building C object src/idl/CMakeFiles/idl.dir/hashid.c.o
[ 86%] Building C object src/idl/CMakeFiles/idl.dir/parser.c.o
[ 86%] Linking C shared library ../../lib/libidl.so
[ 86%] Built target idl
[ 87%] Building C object src/tools/idlc/CMakeFiles/idlc.dir/src/idlc.c.o
[ 87%] Building C object src/tools/idlc/CMakeFiles/idlc.dir/src/plugin.c.o
[ 87%] Building C object src/tools/idlc/CMakeFiles/idlc.dir/src/options.c.o
[...]

PR #933, opened 2021-09-07
Dereference as void* to avoid crash with -O2

homalozoa:
Ref: [pr 809](https://github.com/eclipse-cyclonedds/cyclonedds/pull/809)

This fix cyclonedds building in ArchLinux with GCC 11.1. Related issue https://github.com/eclipse-cyclonedds/cyclonedds/issues/808

Signed-off-by: Homalozoa <xuhaiwang@xiaomi.com>

### DIFF
--- a/src/idl/src/tree.c
+++ b/src/idl/src/tree.c
@@ -277,7 +277,7 @@ create_node(
   node->describe = methods->describe;
   node->references = 1;
 
-  *((idl_node_t **)nodep) = node;
+  *((void **)nodep) = (void *)node;
   return IDL_RETCODE_OK;
 }
 


k0ekk0ek:
Hi @homalozoa! Thanks! Seems like the same sort of change I did on the master branch :+1: Before I can hit merge though, may I request you sign the Eclipse Contributor Agreement? It's a bit of overhead for such a small change, but unfortunately it's a required step.

homalozoa:
hi @k0ekk0ek . Thanks for your reply. Yes, it's the same pr as you did before. I've already signed ECA. Please check again.

k0ekk0ek:
Thanks! I think this can definitely go in functionality wise. Not sure if a cherry-pick would be better wrt tracking etc. If @eboasson approves, I'll hit merge.

eboasson (approved):

PR #934, opened 2021-09-07
Remove VS analyzer and add example for using built-in topics

eboasson:
This adds an example of using the built-in topics by monitoring
DCPSTopic for 10s.  Whether or not DCPSTopic is available is determined
by the ENABLE_TOPIC_DISCOVERY build-time flag, and the code checks for
the case where that is disabled.

Whether DCPSTopic behaves like one expects is further governed by the
Discovery/EnableTopicDiscoveryEndpoints setting.  If it is disabled,
there is no topic discovery over the network and this program would not
really show any topics.

Because of that, it also monitors DCPSPublication and DCPSSubscription
and prints a warning if some endpoints are discovered when no topics are
discovered.  That check is simple enough and has the added benefit of
also introducing these two topics in the sample example.


k0ekk0ek (approved):
:+1:

dennis-adlink (approved):
LGTM

eboasson:
After trying whether Visual Studio's analyzer could perhaps be satisfied with a pound of a flesh, then considering that it might have gone metric and that inflation may require a bit more than a pound anyway, I tried whether a kilo would suffice, only to find that that too would not do. I wondered if perhaps it had gone vegan, but no such luck.

In short I've given up on what, in the absence of SAL2 annotations, really is no more than a worthless piece of junk. This PR now proposes removing the Visual Studio analyzer from the runs, because otherwise the CI will keep failing on a false positive.

k0ekk0ek:
Fine with me :+1:

PR #936, opened 2021-09-07
Add python idl locator to idlc.

thijsmie:
Since python packages can be installed to multiple locations (system, user or virtual environment) there can be multiple idlpy libraries on your system. Which one should be used depends on which python is active in  your shell environment. Sadly, activating a python installation does not change library loading paths so regular library locating mechanisms will not work.

In cyclonedds-python there is a module called `cyclonedds.__idlc__` which prints the full path to the idlpy library it is installed with. This PR adds a locator mechanism to idlc which tries to run this module when executing `idlc -l py`. This kind of mechanism will probably also be needed for other language backends down the line.

### DIFF
--- a/src/tools/idlc/src/plugin.c
+++ b/src/tools/idlc/src/plugin.c
@@ -21,6 +21,8 @@
 static const char sep[] = "/\\";
 static const char lib[] = "";
 static const char ext[] = "dll";
+#define SUBPROCESS_POPEN _popen
+#define SUBPROCESS_PCLOSE _pclose
 #else
 #include <dlfcn.h>
 static const char sep[] = "/";
@@ -30,6 +32,8 @@ static const char ext[] = "dylib";
 #else
 static const char ext[] = "so";
 #endif
+#define SUBPROCESS_POPEN popen
+#define SUBPROCESS_PCLOSE pclose
 #endif
 
 #include "plugin.h"
@@ -86,6 +90,63 @@ static void liberror(char *buffer, size_t bufferlen)
   buffer[bufferlen - 1] = 0; /* ensure final zero in all cases */
 }
 
+#define SUBPROCESS_PIPE_MEMORY_LIMIT 1024 * 1024
+static int run_library_locator(const char *command, char **out_output) {
+  size_t output_size = 0;
+  size_t output_pt = 0;
+  char *output = NULL;
+  FILE *pipe;
+  int ret = 0;
+  bool success = true;
+  int c;
+
+  if ((pipe = SUBPROCESS_POPEN (command, "r")) == NULL) {
+    // broken-pipe
+    return -1;
+  }
+
+  while ((c = fgetc(pipe)) != EOF) {
+    if (c == '\n' || c == '\r') {
+      break;
+    }
+
+    if (output_pt == output_size) {
+      output_size += 128;
+      if (output_size > SUBPROCESS_PIPE_MEMORY_LIMIT) {
+        success = false;
+        break;
+      }
+
+      char* new = (char*) realloc(output, output_size);
+      if (!new) {
+        success = false;
+        break;
+      }
+      output = new;
+    }
+
+    output[output_pt++] = (char) c;
+  }
+
+  ret = SUBPROCESS_PCLOSE (pipe);
+
+  if (success && output != NULL && ret == 0) {
+    // ensure proper string termination (might be newline)
+    output[output_pt] = '\0';
+    *out_output = output;
+
+    return 0;
+  }
+
+  // error
+  if (output) {
+    free(output);
+  }
+
+  return -1;
+}
+
+
 extern int idlc_generate(const idl_pstate_t *pstate);
 
 int32_t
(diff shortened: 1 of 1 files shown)


src/tools/idlc/src/plugin.c
```diff
@@ -21,6 +21,10 @@
 static const char sep[] = "/\\";
 static const char lib[] = "";
 static const char ext[] = "dll";
+#define SUBPROCESS_POPEN _popen
+#define SUBPROCESS_PCLOSE _pclose
+#define SUBPROCESS_MODE "rt"
```
k0ekk0ek, line 26: I'm wondering if text-mode is right. Normally we open files binary so we have actual bytes, because Windows might play tricks. The how-and-what escapes me, but might be good to give that a check if you haven't already.

src/tools/idlc/src/plugin.c
```diff
@@ -64,6 +72,59 @@ static void *loadsym(void *handle, const char *symbol)
...
+  char *output = (char*) malloc(output_size);
+  size_t output_pt = 0;
+  FILE *pipe;
+  int ret = 0;
+  int c;
+
+  if (!output) {
+    // out-of-memory
+    return -1;
+  }
```
k0ekk0ek, line 87: I'd set `output` to NULL and the `output_size` to `0` so there's only one place where allocation is done (in the loop). I'm not against this, but it does make things slightly simpler.

src/tools/idlc/src/plugin.c
```diff
@@ -64,6 +72,59 @@ static void *loadsym(void *handle, const char *symbol)
...
+      if (output_size > SUBPROCESS_PIPE_MEMORY_LIMIT) {
+        free(output);
+        SUBPROCESS_PCLOSE (pipe);
+        return -1;
+      }
+
+      char* new = (char*) realloc(output, output_size);
+      if (!new) {
+        free(output);
+        SUBPROCESS_PCLOSE (pipe);
```
k0ekk0ek, line 113: I wonder if this actually kills the process too. If don't read the output the process might block and I suspect `pclose` (judging by the documentation) simply waits. Maybe good to give that a quick try. The chances of this situation arising is very small of course, but still.

k0ekk0ek (changes requested):

k0ekk0ek:
Thanks @thijsmie! Nice clean solution. I have some small remarks/questions, but overall I agree with the changes.

src/tools/idlc/src/plugin.c
thijsmie: You were right about this one,  in 'r' mode it mimicks UNIX and the windows 'rt' mode does some weird things wrt line endings. The code below was insensitive to the differences, but no need to make things more complex than needed.

src/tools/idlc/src/plugin.c
thijsmie: I looked into this and you are right that this can cause a hanging process in the way you describe. There is no nice  standard library function that allows us to work around this, by setting a timeout and then killing the subprocess, which is how I would approach it to make it fully 'safe'. I have seen a few solutions, I could implement one but not without significantly increasing the code needed for this PR, effectively doing a full 'subprocess' library again.

k0ekk0ek: I figured that would be the case. In PR #842 I dropped a bunch of code that implements exactly this from `ddsrt`. I propose ignoring the issue for now. The chance of this happening is very small and it's not in the core of Cyclone DDS. Eventually, I think we should look into it, but we're starting to maintain a separate abstraction layer for `idlc` (the work I'm doing for #922 also introduces some extra platform specific stuff). I'm going to take a small step back and figure out how/if we can share a bit more code between the two and then revisit this. Would you mind creating an issue so we don't forget?

@eboasson, do you agree with accepting the process isn't killed on timeout?

eboasson: I can learn to stop worrying and love a hanging process 🙂

src/tools/idlc/src/plugin.c
```diff
@@ -64,6 +70,59 @@ static void *loadsym(void *handle, const char *symbol)
...
+      output_size += 128;
+      if (output_size > SUBPROCESS_PIPE_MEMORY_LIMIT) {
+        free(output);
+        SUBPROCESS_PCLOSE (pipe);
+        return -1;
+      }
+
+      char* new = (char*) realloc(output, output_size);
+      if (!new) {
+        free(output);
```
k0ekk0ek, line 102: `output` can be `NULL` here if the first allocation fails.

src/tools/idlc/src/plugin.c
```diff
@@ -64,6 +70,59 @@ static void *loadsym(void *handle, const char *symbol)
...
+
+  if (output == NULL) {
+    // no data received
+  } else {
+    // ensure proper string termination (might be newline)
+    output[output_pt] = '\0';
+  }
+
+  *out_output = output;
+  *out_return_code = SUBPROCESS_PCLOSE (pipe);
```
k0ekk0ek, line 120: I'd just return `-1` if the exit code is not `0`. Also, just return `-1` if the there's no output. If no output is received, we cannot continue anyway. So, if `0` is returned from this function there must be `output`, otherwise, if the process return code is non-zero, free `output`(?) Much easier.

src/tools/idlc/src/plugin.c
```diff
@@ -64,6 +70,59 @@ static void *loadsym(void *handle, const char *symbol)
...
+      }
+      output = new;
+    }
+
+    output[output_pt++] = (char) c;
+  }
+
+  if (output == NULL) {
+    // no data received
+  } else {
```
k0ekk0ek, line 114: Or, `if (output != NULL)`?

src/tools/idlc/src/plugin.c
```diff
@@ -84,8 +143,38 @@ idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang)
...
+        as part of the cyclonedds python package and it can very well be that multiple installations
+        are present on the system, by use of virtual environments, user installs and system installs.
+        Sadly activating a python distribution does not set a any library loading paths. However,
+        the python cyclonedds package has a __idlc__ module that prints the path when executed. The
+        'python3' executable is always guaranteed to point to the active python so we always get the
+        correct idlpy library.
+   */
+  if (idl_strcasecmp(lang, "py") == 0) {
+    int retcode;
+    if (run_library_locator("python3" SUBPROCESS_EXEC " -m cyclonedds.__idlc__", &file, &retcode) != 0) {
```
k0ekk0ek, line 157: According to the MSDN page `_popen` uses the command processor and executes the command as the first command in that processor. Probably easier to remove `SUBPROCESS_EXEC`. My reasoning here being that it may actually cause problems when executed in Cygwin shells etc(?)

src/tools/idlc/src/plugin.c
thijsmie: I reworked the structure of the function to avoid having to check the validity of these pointers everywhere.

k0ekk0ek (approved):

Issue #937, opened 2021-09-07
ExplicitlyPublishQosSetToDefault breaks reader and/or writer


gsnuffles:
I have a reader and a writer that communicate perfecly fine, but when I set `ExplicitlyPublishQosSetToDefault` to true on the reader or both of them, they do not communicate anymore and I get the error messages:

```
1631028312.388135 [1] dq.builtin: invalid parameter list (vendor 1.16, version 2.1): pid 74 (TYPE_CONSISTENCY_ENFORCEMENT) invalid, input = 1,0,255,255,0,0,0,0
1631028312.388187 [1] dq.builtin: SPDP (vendor 1.16): invalid qos/parameters
```

Any idea why this is happening?

Edit: Adding to my confusion is the inconsistency,  that this does not always happen.

eboasson:
Looks like a bug to me, and this appears to solve it:
```
diff --git a/src/core/ddsi/src/ddsi_plist.c b/src/core/ddsi/src/ddsi_plist.c
index 8a7aa2f0..082c05e1 100644
--- a/src/core/ddsi/src/ddsi_plist.c
+++ b/src/core/ddsi/src/ddsi_plist.c
@@ -542,7 +542,7 @@ static dds_return_t ser_type_consistency (struct nn_xmsg *xmsg, nn_parameterid_t
   char * const p = nn_xmsg_addpar_bo (xmsg, pid, 8, bo);
   const uint16_t kind = ddsrt_toBO2u (bo, (uint16_t) x->kind);
   memcpy (p, &kind, 2);
-  size_t offs = sizeof (x->kind);
+  size_t offs = sizeof (kind);
   p[offs + 0] = x->force_type_validation;
   p[offs + 1] = x->ignore_sequence_bounds;
   p[offs + 2] = x->ignore_string_bounds;
```
`x->kind`4 bytes in memory (an enum) where the wire format requires two bytes. That means it incorrectly introduces two bytes of padding, left uninitialized (in your case the `255,255`). Booleans are expected in those bytes, and so it correctly rejects these values as invalid input. If you have a chance to confirm that this indeed fixes it, I'd be grateful. The deserializer doesn't have this issue.

It hasn't been seen because the CI runs with the default setting for `ExplicitlyPublishQosSetToDefault` and the value happens to always be the default value. (That should change with #904, because that introduces the code where non-default values for type consistency enforcement actually become meaningful.)

gsnuffles:
[...]

PR #938, opened 2021-09-07
Use correct offset when serializing type consistency qos

dennis-adlink:
The serialize function for the Type Consistency QoS was incorrectly using the size of the internal `kind` field for calculating the offset of the option flags, which is 32 bits. This should be the size of the 16 bits `kind` field from the QoS definition. 

This PR fixes issue #937.


k0ekk0ek:
Thanks @dennis-adlink!

Issue #941, opened 2021-09-09
Update serdata functions to handle sequences/strings with shared memory


MatthiasKillat:
### Feature

To support bounded and unbounded sequences as well as strings with shared memory we need to modify the serdata functions.

The goal is to be able serialize the sample into shared memory even if the data type has no fixed size (instead of using the network) for local readers.

### Details

The following is required:
 
1. Update the flag in iceoryx user header to indicate whether the iceorxy memory chunk is holding the data in serialized/deserialized format.
2. Modify the serdata function `ddsi_serdata_from_loaned_sample()` to serialize the data from the sample into the iceoryx memory chunk, when the type is not fixed
3. Modify the serdata function `ddsi_serdata_from_iox()` to deserialize the data from the iceoryx memory chunk to the sample, when the type is not fixed

### Technical Assumption

The work should be based on https://github.com/eclipse-cyclonedds/cyclonedds/pull/788, which needs to be merged before in this case.

sumanth-nirmal:
@MatthiasKillat and @eboasson The changes for this are merged in https://github.com/eclipse-cyclonedds/cyclonedds/pull/1000, we could probably close this issue, no?

thijsmie:
Was merged and closing was already suggested :smile:.

PR #943, opened 2021-09-09
use iceoryx user header

elfenpiff:
After merge please merge the cyclone cxx adjustments: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/159

### DIFF
--- a/src/core/ddsc/src/dds_write.c
+++ b/src/core/ddsc/src/dds_write.c
@@ -30,6 +30,7 @@
 #ifdef DDS_HAS_SHM
 #include "dds/ddsi/shm_sync.h"
 #include "dds/ddsi/q_addrset.h"
+#include "iceoryx_binding_c/chunk.h"
 #endif
 
 #ifdef DDS_HAS_SHM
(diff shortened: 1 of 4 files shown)


src/core/ddsc/src/dds_write.c
```diff
@@ -71,25 +72,26 @@ static void *create_iox_chunk(dds_writer *wr)
...
     void *sample;
     uint32_t sample_size = wr->m_topic->m_stype->iox_size;
-    uint32_t chunk_size = DETERMINE_ICEORYX_CHUNK_SIZE(sample_size);
     while (1)
     {
-      enum iox_AllocationResult alloc_result = iox_pub_loan_chunk(wr->m_iox_pub, (void **) &ice_hdr, chunk_size);
+      enum iox_AllocationResult alloc_result = 
+        iox_pub_loan_aligned_chunk_with_user_header(wr->m_iox_pub, &sample, sample_size, IOX_C_CHUNK_DEFAULT_USER_PAYLOAD_ALIGNMENT, sizeof(iceoryx_header_t), 8);
+      iox_chunk_header_t * iox_chunk_header = iox_chunk_header_from_user_payload(sample);
+      ice_hdr = (iceoryx_header_t*)iox_chunk_header_to_user_header(iox_chunk_header);
```
MatthiasKillat, line 80: I think there is no function to get the user header directly from the sample, is there?

Could be changed in iceoryx API in the next release. But if we do so we should also think about whether we really want to expose the internal chunk header in the user API (I would restrict it to a private header).

src/core/ddsi/include/dds/ddsi/q_xmsg.h
```diff
@@ -61,15 +61,11 @@ struct iceoryx_header {
    uint32_t data_size;
    unsigned char data_kind;
    ddsi_keyhash_t keyhash;
+   bool does_contain_serialized_data;
```
MatthiasKillat, line 64: `contains_serialized_data`? or something shorter, but more importantly:

It appears not to be set anywhere as of now, is this correct?

I will then set it when I need to determine whether the data is serialized and has to be deserialized (for https://github.com/eclipse-cyclonedds/cyclonedds/issues/941).

Could currently be set to false whenever we fill the iceoryx chunk (but we will evaluate it nowhere until https://github.com/eclipse-cyclonedds/cyclonedds/issues/941).

MatthiasKillat (approved):
Header and payload data handling looks ok.
CI failure requires investigation but should not be caused by the changes.

Unused flag is no problem (will use it shortly), otherwise we can remove it or set it to false when we fill the chunk (with non-serialized data) and true otherwise.

@eboasson

src/core/ddsi/include/dds/ddsi/q_xmsg.h
k0ekk0ek: Or just `serialized` would work too. Perhaps make it an enum eventually because of the various serialization options?

elfenpiff: @k0ekk0ek @MatthiasKillat

I like expressive variable names and prefer a `is`, `has`, `does` ... prefix for boolean variable names. If one reads such variables later in the code during a review the reviewer knows instantly that a boolean is handled here. Therefore, I suggest to leave the name as it is.

At the moment we are not planning to have multiple serialization options and a bool seems sufficient at the moment. If this changes in the future we can still adapt it.

MatthiasKillat:
@elfenpiff I rebased master (required moderate adaptations) and fixed an obvious problem where the header still was accessed by casting the iox_chunk in serdata_default_from_iox but the assert still fires sporadically. 

This was definitely a problem but there are other problems it seems. Not entirely sure about the cause, will have a look later.

MatthiasKillat:
@eboasson Should be alright now, from a functionality point of view. I have not run it together with say ROS2 rolling yet.

eboasson (approved):

PR #944, opened 2021-09-10
Documentation has two different "default" values for ParticipantIndex

nbbrooks:
https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/manual/options.md#cycloneddsdomaindiscoveryparticipantindex

```
The default is auto. The participant index is part of the port number calculation and if predictable port numbers are needed and fixing the participant index has no adverse effects, it is recommended that the second be option be used.

The default value is: "none".
```

I believe the actual default is none, not sure if other files need changing though.

### DIFF
--- a/docs/manual/options.md
+++ b/docs/manual/options.md
@@ -116,8 +116,6 @@ This element specifies the DDSI participant index used by this instance of the C
 
  * none:, which causes it to use arbitrary port numbers for unicast sockets which entirely removes the constraints on the participant index but makes unicast discovery impossible.
 
-The default is auto. The participant index is part of the port number calculation and if predictable port numbers are needed and fixing the participant index has no adverse effects, it is recommended that the second be option be used.
-
 The default value is: "none".
 
 


k0ekk0ek:
Hi @nbbrooks. You're right, there's a discrepancy between the documentation and what's actually being used. Thanks for taking the time to open a PR. Much appreciated! However, the `options.md` file is automatically generated so we need to make some additional changes. If you'll remove the lines from `ddsi_config.h` https://github.com/eclipse-cyclonedds/cyclonedds/blob/de984036af79ce652ec3abe582e09b76920727db/src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h#L1772, the files are properly generated if you build the `schema` target. The changed files `options.md`, `cyclonedds.rnc` and `cyclonedds.xsd` can then simply be copied over and submitted in the same PR. Also, before we can accept PRs the Eclipse Contributor Agreement must be signed. If you click the details link from checks, it'll guide you through the process.  Let me know if you need additional help.

nbbrooks:
I thought that might be the case - I will update the PR accordingly!

I do have a question related to this I can put as a feature request if it makes sense. If AllowMulticast is set to false, is there a way to warn the user if they are using the default ParticipantIndex? Maybe this already does happen and the instance I troubleshooted just had the warning/error scrolled off way up at the top. If not, I could see this being a common mistake people make during configuration.

k0ekk0ek:
Probably best to do that in a separate issue so it's tracked separately(?) That way, it doesn't get lost if we merge the PR. I don't know the answer from the top of my head, but I can have a quick look.

k0ekk0ek:
`Don't think it reports a warning, at least, judging by `rtps_config_prep` in `q_init.c`. Adding a `DDS_ILOG` seems trivial enough. @eboasson, do you think it makes sense adding such a  warning message?

eboasson:
> If AllowMulticast is set to false, is there a way to warn the user if they are using the default ParticipantIndex? Maybe this already does happen and the instance I troubleshooted just had the warning/error scrolled off way up at the top. If not, I could see this being a common mistake people make during configuration.

It is indeed an issue. I had counted on `AllowMulticast` = `spdp` to be almost universally preferred over `false`, but that may have been a miscalculation.

> Don't think it reports a warning, at least, judging by `rtps_config_prep` in q_init.c. Adding a `DDS_ILOG` seems trivial enough. @eboasson, do you think it makes sense adding such a warning message?

Yes, but strictly speaking there are cases where there is no advantage to changing `ParticipantIndex` even when `AllowMulticast` is `false`: it only matters for being discovered via unicast, so if you go for "clients" discovering a "server" (and not each other) the clients wouldn't need it.

So I think the proper way to do it is to do the warning only if `ParticipantIndex` hasn't been explicitly set to `none`. That's perfectly feasible, but a little more involved. Come to think of it, if one does that, why not go one step further and simply set it to `auto` *and* add `localhost` to the Peers section if no peers are specified either?

nbbrooks:
How would you configure a client-server discovery model with unicast? Maybe I missed something in the documentation, but I did not realize cyclone supported that!

Both of those proposals sound good to me, I think the 2nd one is great!

eboasson:
> How would you configure a client-server discovery model with unicast?

There were scare quotes in there for a reason. I know of various DDS-based systems using 100s of nodes in a strict hub-and-spoke system. Often the hub nodes are called "servers" and the spokes clients, or sometimes workers, or whatever really. The key point in these systems is that the hub nodes see all the spokes, but the spokes doen't need to discover each other.

One way to configure this is to disable any discovery message on the hub and to put the hub's address in the Peer list on the spokes: then the hub will learn of the existence of the spokes and respond.

There are nicer ways of course, where there is still multicast-based discovery, with multiple hub nodes and still using multicast from the hub nodes to the spokes as well as from the spokes to the hub nodes, while still preventing the spokes from discovering each other and the IP stack from routing the multicasts to them. But that it really is just a variation on the idea that the spokes ping the hubs.

> Maybe I missed something in the documentation, but I did not realize cyclone supported that!

Nope, you didn't miss anything — but there is also no need for cyclone (or any DDS implementation for that matter) to do anything special to support that because the DDSI protocol has enough flexibility to allow completely transparent discovery and routing. We can already do this with https://github.com/eclipse-zenoh/zenoh-plugin-dds but there are still some gaps to be filled in its functionality.

thijsmie:
Superseded by #1428.

Issue #915, opened 2021-08-25
Error while cross compiling with yocto


mtudan:
Hi,

I am trying to cross compile `cyclonedds` `0.8.0rc1` release using yocto project but I got the following configuration error:
```
| -- Configuring incomplete, errors occurred!
| See also "/home/ubuntu/yocto/build/tmp/work/cortexa7t2hf-neon-linux-gnueabi/cyclonedds/0.8.0rc1-r0/build/CMakeFiles/CMakeOutput.log".
| WARNING: exit code 1 from a shell command.
| ERROR: Execution of '/home/ubuntu/yocto/build/tmp/work/cortexa7t2hf-neon-linux-gnueabi/cyclonedds/0.8.0rc1-r0/temp/run.do_configure.29481' failed with exit code 1:
| -- The C compiler identification is GNU 10.2.0
| -- Detecting C compiler ABI info
| -- Detecting C compiler ABI info - done
| -- Check for working C compiler: /home/ubuntu/yocto/build/tmp/work/cortexa7t2hf-neon-linux-gnueabi/cyclonedds/0.8.0rc1-r0/recipe-sysroot-native/usr/bin/arm-linux-gnueabi/arm-linux-gnueabi-gcc - skipped
| -- Detecting C compile features
| -- Detecting C compile features - done
| -- Setting build type to 'RelWithDebInfo' as none was specified.
| -- Linux/MacOS system found, trying to find iceoryx...
| -- iceoryx not found, falling back to loopback communication.
| -- Found OpenSSL: /home/ubuntu/yocto/build/tmp/work/cortexa7t2hf-neon-linux-gnueabi/cyclonedds/0.8.0rc1-r0/recipe-sysroot/usr/lib/libcrypto.so (found version "1.1.1j")
| -- Building with OpenSSL support
| CMake Error at src/tools/ddsconf/CMakeLists.txt:13 (find_program):
|   Could not find DDSCONF_EXECUTABLE using the following names: ddsconf
```
It seems that the error comes form the following `CMakeLists.txt` file:
https://github.com/eclipse-cyclonedds/cyclonedds/blob/0.8.0rc1/src/tools/ddsconf/CMakeLists.txt#L12

sjames:
Cyclone needs the ddsconf executable to build, and this needs to be built for the host.  Here is how I solved it.

[...]

PR #945, opened 2021-09-13
examples: add support for yocto cross compiling

mtudan:
This commit extends the cmake files of examples in order to support
yocto builds. In order to cross compile CycloneDDS, the native
build is required. Because of that, the find_package() function
was only able to find the target version of the CycloneDDS package,
and the target_link_libraries() function was using ddsc library
from the native build.

Fix this default behavior with proper paths when searching for
the CycloneDDS package and the ddsc shared library.

Signed-off-by: Matija Tudan <tudan.matija@gmail.com>


mtudan:
This pull request fixes the https://github.com/eclipse-cyclonedds/cyclonedds/issues/915 issue.

k0ekk0ek:
Hi @mtudan. Thanks for the PR! I glanced over the changes and I'm wondering if we're not trying to solve the issue in the wrong place. Have you tried compiling with a CMake Toolchain file?  e.g. `find_package(CycloneDDS REQUIRED PATHS ${STAGING_DIR_NATIVE} NO_CMAKE_FIND_ROOT_PATH)` seems like the type of setting you'd do in a toolchain file so the rest can remain the same.

mtudan:
Hi @k0ekk0ek, well yocto will prepare this file for you using the `cmake.bbclass`. But the problem is it will first append the `${STAGING_DIR_HOST}` and then the `${STAGING_DIR_NATIVE}` variables to the `CMAKE_FIND_ROOT_PATH` variable:
https://github.com/openembedded/openembedded/blob/master/classes/cmake.bbclass#L48

This will cause the `find_package()` function to find the CycloneDDS in the `${STAGING_DIR_HOST}` location (which is the target build) instead in the `${STAGING_DIR_NATIVE}` location (which is the required, native version).

So the following line fixes that:
```
find_package(CycloneDDS REQUIRED PATHS ${STAGING_DIR_NATIVE} NO_CMAKE_FIND_ROOT_PATH)
```
but it will simultaneously cause the `target_link_libraries()` function to try to link with the `libddsc.so` from the native build, because it uses the library from the CycloneDDS package. That's why we need that additional line to tell cmake to link with the target version of the library:
```
find_library(DDSC_LIB NAMES ddsc REQUIRED PATHS ${STAGING_DIR_HOST}/usr/lib NO_DEFAULT_PATH)
```

k0ekk0ek:
Hi @mtudan. I think the issue that we're actually trying to solve is that `idlc` is required to build the examples but it finds the Cyclone DDS targets exported for `${STAGING_DIR_HOST}` rather than `${STAGING_DIR_NATIVE}`. Which is actually what you'd want, but the problem is that the IDL compiler is then not available. It just so happens that a colleague of mine is running into the same problem (except we're trying to do macOS ARM builds on x86_64). Your patch fixes the problem for the examples, but I doubt it's the right fix because users that have their own projects cannot use it. Or rather, they'd run into the same problem and the fix is not automatic. I'll have read up a bit more on the subject, please allow me some time to do so.

mtudan:
That's exactly what happens. Feel free to propose the right place or fix for the `idlc` bug.

k0ekk0ek:
@mtudan, @thijsmie pointed me towards the [CMake Wiki on cross compiling](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/CrossCompiling), which contains a nice sample of how to fix the situation where a project wants to build and import `idlc` (our situation for cyclonedds-python). Looking through this and taking one of his suggestions into consideration, I think it's wise to just patch `idlc_generate` so that it uses `find_program(IDLC_EXECUTABLE idlc REQUIRED NO_CMAKE_FIND_ROOT_PATH)` if `CMAKE_CROSSCOMPILING` is true (just use `CycloneDDS::idlc` if not cross compiling). Then, judging by the [documentation on find_program](https://cmake.org/cmake/help/latest/command/find_program.html) the command should search the proper directories in order, we could add `STAGING_DIR_HOST` to `PATHS` for the `find_program` call if `YOCTO_CROSSCOMPILING` is true.

Note that I've not tested any of this yet.

@thijsmie, the same scenario probably applies to `ddsconf`.

k0ekk0ek:
@mtudan, a draft implementation by @thijsmie is available in #948. Once that's in, Yocto cross-compilation problems are probably gone too.

mtudan:
@k0ekk0ek thank you for the update, will watch for the update on that one. When do you expect it to be merged?

k0ekk0ek:
Shouldn't be too long, it's only a small change. My guess is early next week-ish or so.

k0ekk0ek:
Hi @mtudan, I think your problem has been resolved by #948. Can you verify if that's indeed the case?

mtudan:
Hi @k0ekk0ek, issue resolved, we can close this one. Thank you!

k0ekk0ek:
@thijsmie is the one who deserves the credit here :slightly_smiling_face: Thanks for checking :+1:

PR #947, opened 2021-09-14
Add compiler option to specify prefix directory

k0ekk0ek:
This PR add the `-p` command line option (short for prefix) to the IDL compiler to achieve the desired functionality mentioned in #801 and #922. Instead of letting a user simply specify the output directory I've chosen to name it the prefix directory because the structure is kept if the specified path is relative. For instance `baz/bar/foo.*` is generated if the file passed along is `baz/bar/foo.idl`. This is meant to help out with e.g. ROS2 use cases where some files have the same name. `foo.*` is generated if a full path to the file is specified or the relative path to `foo.idl` does not contain any directories. This PR let's the user specify a prefix different from the current working directory. So, running `idlc -p foobar baz/bar/foo.idl` generates `foo.*` in `foobar/baz/bar`.

This draft PR is mainly to get some input. I still have to write tests etc, I've only run it once myself. You've been warned :slightly_smiling_face:

### DIFF
--- a/src/tools/idlc/src/generator.c
+++ b/src/tools/idlc/src/generator.c
@@ -337,16 +337,16 @@ idl_retcode_t
 idlc_generate(const idl_pstate_t *pstate)
 {
   idl_retcode_t ret = IDL_RETCODE_NO_MEMORY;
+  int cnt;
   const char *sep, *ext, *file, *path;
-  char empty[1] = { '\0' };
   char *dir = NULL, *basename = NULL;
   struct generator generator;
 
   assert(pstate->paths);
   assert(pstate->paths->name);
-  path = pstate->sources->path->name;
+  path = pstate->sources->file->name;
   /* use relative directory if user provided a relative path, use current
-     word directory otherwise */
+     work directory otherwise */
   sep = ext = NULL;
   for (const char *ptr = path; ptr[0]; ptr++) {
     if (idl_isseparator((unsigned char)ptr[0]) && ptr[1] != '\0')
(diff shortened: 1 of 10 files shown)


k0ekk0ek:
@sumanth-nirmal, can you have a quick look to see if the `-p` option works for you? If so, I'll finish this up and do the same for the C++ generator.

k0ekk0ek:
Only thing I'm not 100% sure about is whether the "relative" behavior should be triggered by a flag instead. @eboasson, can you give this one a quick look?

src/tools/idlc/src/generator.c
```diff
@@ -357,8 +357,13 @@ idlc_generate(const idl_pstate_t *pstate)
 
   file = sep ? sep + 1 : path;
   if (idl_isabsolute(path) || !sep)
-    dir = empty;
-  else if (!(dir = idl_strndup(path, (size_t)(sep-path))))
+    cnt = idl_asprintf(&dir, "%.*s", idlc_config->prefix);
```
sumanth-nirmal, line 360: Doesn't `%.*s` expect an integer? I don't think this is working when the `idlc` is invoked with absolute path

sumanth-nirmal (changes requested):
Thanks for adding this feature with prefix option in IDL compiler, this is definitely useful (specifically with ROS2 msg structure). I quickly gave this a try and it seems to work when the IDL is passed with relative path, but I was unable to get this working when IDL is passed with the absolute path

src/tools/idlc/src/generator.c
k0ekk0ek: Yep, that's exactly why an absolute path gives problems currently. I have fixed it locally, but I was more interested in whether this works for the ROS usecase. Expect a mergeable version soon. Thanks!

ichernev:
Well, to be fair I think both behaviors should be present (final output directory, aka `-d` (rtiddsgen like)), and prefix (`-p`, this PR). Also `-d` is the more simple option, so it wouldn't make sense to have `-p` without `-d`. Ideally both, i.e you want a prefix, but not the full one on the command line, so something similar to -C (change dir before begining) + `-d` (for direct output) + `-p` (for relative output, to -C directory, if necessary).

But then, the idea is that you'd have multiple idl files and you want to convert them all at once (also preserving the dir structure), so a recursive call makes sense here, then the two sensible options are source dir and target dir (which is `-d` and the source file being a dir). In a sense just `-r` recursive option + `-d` (raw target dir) would be enough.

I think that makes more sense than supporting `-C`, `-p` and `-d`, and is also more useful (otherwise you have to write a bash loop to iterate deep into a file hierarchy).

PR #949, opened 2021-09-17
Avoid some undocumented interfaces in bison

eboasson:
This fixes incompatibilities with bison 3.8:

* returning user-defined values from yyparse is not supported, this
  works aruond that by storing a more meaningful return value in the
  parser state object;

* the token name table generated by %token-table is deprecated and has
  been for quite some time, and it seems that now the accompanying
  mapping to internal token numbers is gone as well, this uses a hack to
  work around that.

A follow-up fix is needed to remove the use of the token name table
entirely.  The hack here is attractive only in that it provides a quick
solution for the failing CI runs on macOS.

### DIFF
--- a/src/idl/include/idl/processor.h
+++ b/src/idl/include/idl/processor.h
@@ -116,6 +116,7 @@ struct idl_pstate {
       IDL_PARSE_UNKNOWN_ANNOTATION_APPL_PARAMS
     } state;
     void *yypstate; /**< state of Bison generated parser */
+    idl_retcode_t result; /**< return value from idl_parse */
   } parser;
 };
 
--- a/src/idl/include/idl/processor.h
+++ b/src/idl/include/idl/processor.h
@@ -116,7 +116,6 @@ struct idl_pstate {
       IDL_PARSE_UNKNOWN_ANNOTATION_APPL_PARAMS
     } state;
     void *yypstate; /**< state of Bison generated parser */
-    idl_retcode_t result; /**< return value from idl_parse */
   } parser;
 };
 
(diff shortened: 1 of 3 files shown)


src/idl/include/idl/processor.h
```diff
@@ -116,6 +116,7 @@ struct idl_pstate {
       IDL_PARSE_UNKNOWN_ANNOTATION_APPL_PARAMS
     } state;
     void *yypstate; /**< state of Bison generated parser */
+    idl_retcode_t result; /**< return value from idl_parse */
```
k0ekk0ek, line 119: I purposely left out the result from the parser state. Initial versions by Frans did that too and that'd mean we have to make sure it's always correct because it's not immediately obvious when this one can be used. Scanner, parser, etc. I think I prefer adding a thread-local variable in `parser.y` and calling an additional function to invoke the parser that returns the value of said variable. Or perhaps extending the Bison parser signature to accept a pointer to a result value. It's somewhat slower and not as simple/elegant as simply returning a negative value, but overal limits the scope in which errors can be made in deciding when to use the return value and/or `result` member in the parser state (if that makes sense).

k0ekk0ek:
Thanks @eboasson. Shame the `yytokname` is now actually being removed. I liked that Bison let me define the names and keywords in a single place, but the documentation did warn that this was to happen at some point. Other than the `result` member in `struct idl_pstate`, I'm definitely ok with the changes. Perhaps we should create a GitHub issue to change the token name stuff. There's a couple of other smallish problems that need fixing (including ZeroC mcpp patches etc). I should probably do them all in one go next week or so.

Issue #952, opened 2021-09-18
Library names are too generic


roehling:
The libidl library from the 0.8.0 release conflicts with the CORBA libidl in Debian. Generally, the library names are far too generic IMO. Especially since package name are derived from library names, people looking for CycloneDDS will probably miss it in Debian, as they will likely use the search term "cyclone" and thus miss the libddsc packages (the source package does have the name cyclonedds, but it has much less visibility than the binary package names).

I would suggest to add the cyclone prefix to the library names, yielding libcycloneddsc and libcycloneidl. This is much more distinctive. Given that the CMake config is named CycloneDDS already, this change is pretty transparent for library users.

eboasson:
I think this would be a sensible change. It would then probably also be wise to look at changing the include directory name.

The only concerns I have are with backwards compatibility, but is true that for those using the CMake config, it at least would not affect the sources.

roehling:
You could install the headers to `include/CycloneDDS` instead of `include` and adjust the `target_include_directories()` to use `$<INSTALL_INTERFACE:include/CycloneDDS>`. This isolates the header files from conflicting implementations but stays compatible with existing source code.

roehling:
I fixed the issues with the PR (side note: is there a way to add alternate email addresses to the ECA?).

Currently, I cannot update the CycloneDDS packages in Debian because of this issue, and I'm contemplating options:

1. I can backport my proposed fix for the Debian packaging of the 0.8.0 release, preserving source code compatibility for downstream projects using CycloneDDS.
2. I can link `libidl` statically into `idlc` and not package it (nor its headers), preserving binary compatibility for programs linked against `libddlsc` only.
3. I can skip the 0.8.0 release completely, maintaining the outdated 0.7.0 release until this issue is addressed in a later release.

I would prefer solution 1, because I value source code compatibility higher than the ability to ship pre-compiled binaries to Debian (besides, anyone shipping pre-compiled binaries will probably bundle them with their own CycloneDDS libraries anyway).

Solution 3 would mean no built-in support for CycloneDDS in the Debian-released ROS 2 packages. While unfortunate, the impact of this decision would be low in the grand scheme of things, as the vast majority of ROS 2 users currently work with the OSRF-shipped CycloneDDS packages anyway.

k0ekk0ek:
[...]

PR #953, opened 2021-09-18
Emit shared object files with libcyclone* names

roehling:
This is my proposed fix for issue #952. As you can see, the required changes are minimally invasive: except for the actual file names of the shared libraries on disk, nothing changes.


roehling:
I'm closing this, because it is obsolete now.

PR #963, opened 2021-09-27
Add member ID field population and validation

reicheratwork:
- Move member_id field to the declarator, since these have unique ids, not
  the enveloping member containing type and declarators
- Add autoid annotation to modules
  - autoid and inherits from parent nodes if none is found at the current
    node
  - defaults to sequential autoids if none whatsoever are present
  - member ids are populated according to id already set, or the autoid
    for the struct, then ids are checked for collisions
- Add and restructure unittests for member id fields

### DIFF
--- a/src/idl/src/tree.c
+++ b/src/idl/src/tree.c
@@ -996,194 +996,175 @@ static int compare_declarator(const void *lhs, const void *rhs)
     return 0;
 }
 
-idl_retcode_t
-idl_propagate_autoid(
-  idl_pstate_t *pstate,
-  idl_node_t *list,
-  idl_autoid_t toset)
+static void
+assign_id(
+  idl_declarator_t *declarator,
+  const idl_declarator_t *last,
+  idl_autoid_t autoid)
 {
-  assert(list);
+  assert(declarator);
 
-  idl_node_t *node = NULL;
+  /* id assigned through @id or @hashid annotation */
+  if (declarator->id.annotation)
+    return;
+
+  assert(declarator->name);
+  assert(declarator->name->identifier);
+
+  if (autoid == IDL_HASH)
+    declarator->id.value = idl_hashid(declarator->name->identifier);
+  else if (last) /* identifiers silently overflow */
+    declarator->id.value = (last->id.value + 1) & 0x0FFFFFFFu;
+  else
+    declarator->id.value = 0u;
+}
+
+static idl_declarator_t *
+next_member(const void *node, const idl_declarator_t *declarator)
+{
+  const idl_struct_t *derived = node;
+
+  assert(idl_type(node) == IDL_STRUCT);
+  if (!declarator) {
+    /* iteration is depth first, see if struct is derived */
+    while (derived && derived->inherit_spec)
+      derived = derived->inherit_spec->base;
+  } else if (declarator->node.next) {
+    return (idl_declarator_t *)declarator->node.next;
+  } else {
+    const idl_struct_t *base;
+    const idl_member_t *member = (const idl_member_t *)declarator->node.parent;
+    assert(member);
+    if (member->node.next)
+      return (idl_declarator_t *)((const idl_member_t *)member->node.next)->declarators;
+    base = (const idl_struct_t *)member->node.parent;
+    if (base == node)
+      return NULL;
+    while (derived->inherit_spec && derived->inherit_spec->base != base)
+      derived = derived->inherit_spec->base;
+  }
+
+  assert(derived);
+  /* (intermediate) structs without members are supported in IDL4 */
+  while (!derived->members && derived != node) {
+    const idl_struct_t *base = derived;
+    derived = node;
+    while (derived->inherit_spec && derived->inherit_spec->base != base)
+      derived = derived->inherit_spec->base;
+  }
+  assert(derived);
+  if (!derived->members)
+    return NULL;
+
+  return (idl_declarator_t *)derived->members->declarators;
+}
+
+/* FIXME: does not take into account single inheritance for union types as
+          introduced by the DDS-XTypes specification section  7.2.2.4.5 */
+static idl_declarator_t *
+next_case(const void *node, const idl_declarator_t *declarator)
+{
+  const idl_case_t *branch;
+
+  assert(idl_type(node) == IDL_UNION);
+  if (!declarator)
+    return ((const idl_union_t *)node)->cases->declarator;
+  assert(!declarator->node.next);
+  branch = (const idl_case_t *)declarator->node.parent;
+  assert(branch);
+  branch = (const idl_case_t *)branch->node.next;
+  if (!branch)
+    return NULL;
+
+  return (idl_declarator_t *)branch->declarator;
+}
+
+static idl_retcode_t
+assign_field_ids(idl_pstate_t *pstate, void *node)
+{
   idl_retcode_t ret = IDL_RETCODE_OK;
-  IDL_FOREACH(node, list) {
-    if (idl_is_module(node)) {
-      idl_module_t *mod = (idl_module_t*)node;
-      if (!mod->autoid.annotation)
-        mod->autoid.value = toset;
-      ret = idl_propagate_autoid(pstate, mod->definitions, mod->autoid.value);
-    } else if (idl_is_struct(node)) {
-      idl_struct_t *str = (idl_struct_t*)node;
-      if (!str->autoid.annotation)
-        str->autoid.value = toset;
-      ret = idl_propagate_autoid_constr_type(pstate, (idl_node_t*)str);
-    } else if (idl_is_union(node)) {
-      idl_union_t *u = (idl_union_t*)node;
-      if (!u->autoid.annotation)
-        u->autoid.value = toset;
-      ret = idl_propagate_autoid_constr_type(pstate, (idl_node_t*)u);
-    }
-    if (ret)
+  idl_autoid_t autoid = IDL_SEQUENTIAL;
+  idl_declarator_t *declarator = NULL, *last = NULL;
+  idl_declarator_t **sorted = NULL; /* not "const" to silence MSVC */
+  idl_declarator_t *(*iterate)(const void *, const idl_declarator_t *);
+  size_t length = 0;
+
+  switch (idl_type(node)) {
+    case IDL_UNION:
+      autoid = ((const idl_union_t *)node)->autoid.value;
+      iterate = &next_case;
+      break;
+    default:
+      assert(idl_type(node) == IDL_STRUCT);
+      /* structs without members are supported in IDL4 */
+      if (!((const idl_struct_t *)node)->members)
+        return IDL_RETCODE_OK;
+      autoid = ((const idl_struct_t *)node)->autoid.value;
+      iterate = &next_member;
       break;
   }
 
-  return ret;
-}
-
-static void
-process_declarator_id(
-  idl_declarator_t *decl,
-  idl_declarator_t *last,
-  idl_autoid_t autoid)
-{
-  assert(decl);
-
-  if (decl->id.annotation) {
-    /*this declarator already has an ID annotation*/
-    return;
-  } else if (autoid == IDL_HASH) {
-    assert(decl->name->identifier);
-    decl->id.value = idl_hashid(decl->name->identifier);
-  } else if (last) {
-    /*after conferring with @e-hndrks it was decided that overflow would not be flagged as an error*/
-    decl->id.value = (last->id.value + 1) & 0x0FFFFFFFu;
-  } else {
-    decl->id.value = 0;
-  }
-}
-
-static idl_declarator_t*
-next_decl_struct(idl_declarator_t *decl, idl_declarator_t **last, idl_node_t *current) {
-  assert(decl);
-  assert(last);
-  assert(current);
-
-  idl_declarator_t *next_decl = idl_next(decl);
-  if (next_decl)
-    return next_decl;
-
-  idl_member_t *mem = idl_parent(decl);
-  assert(idl_is_member(mem));
-  idl_member_t *next_mem = idl_next(mem);
-
-  if (next_mem)
-    return next_mem->declarators;
-
-  idl_struct_t *str = idl_parent(mem);
-  assert(str);
-
-  /*store the last declarator of the parent of current*/
-  if ((idl_struct_t*)current != str && *last == NULL)
-    *last = decl;
-
-  if (str->inherit_spec) {
-    assert(idl_is_struct(str->inherit_spec->base));
-    idl_struct_t *base = str->inherit_spec->base;
-    assert(base->members);
-
-    return base->members->declarators;
+  /* assign field identifiers */
+  while ((declarator = iterate(node, declarator))) {
+    assert(declarator->node.parent);
+    if (((const idl_node_t *)declarator->node.parent)->parent == node)
+      assign_id(declarator, last, autoid);
+    last = declarator;
+    length++;
   }
 
-  return NULL;
-}
+  /* theoretically there are no fields to assign identifiers to */
+  if (!length)
+    return IDL_RETCODE_OK;
 
-static idl_declarator_t*
-next_decl_union(idl_declarator_t *decl, idl_declarator_t **last, idl_node_t *current) {
-  (void) last;
-  (void) current;
-  assert(decl);
-  idl_case_t *_case = idl_parent(decl);
-  assert(idl_is_case(_case));
-  _case = idl_next(_case);
-  if (_case)
-    return _case->declarator;
-  else
-    return NULL;
-}
+  /* check for duplicate field identifiers */
+  if (!(sorted = malloc(length * sizeof(*sorted))))
+    return IDL_RETCODE_NO_MEMORY;
 
-typedef idl_declarator_t* (*itfunc_ptr)(idl_declarator_t*,idl_declarator_t**,idl_node_t*);
+  assert(!declarator);
+  for (size_t count=0; (declarator = iterate(node, declarator)); count++)
+    sorted[count] = declarator;
+  qsort(sorted, length, sizeof(*sorted), &compare_declarator);
 
-idl_retcode_t
-idl_propagate_autoid_constr_type(
-  idl_pstate_t *pstate,
-  idl_node_t *constr_type)
-{
-  idl_retcode_t ret = IDL_RETCODE_OK;
-  size_t n_ids;
-  idl_declarator_t *decl = NULL, *first = NULL, *last = NULL;
-  idl_autoid_t aid = IDL_SEQUENTIAL;
-  /*function which iterates over declarators, returns the next declarator describing an entity
-    which needs a unique member id assigned, or NULL if it is done iterating
-    - the first parameter is the current declarator
-    - the second parameter will point to the last declarator of any parent after a full run over the
-      entire inheritance tree
-    - the third parameter is the pointer of the current struct, it is used to determine whether to
-      store anything in the second parameter*/
-  itfunc_ptr iterate_func = NULL;
-
-  /*initialize decl, aid and iterate_func*/
-  if (idl_is_struct(constr_type)) {
-    idl_struct_t *str = (idl_struct_t*)constr_type;
-    iterate_func = &next_decl_struct;
-    aid = str->autoid.value;
-    if (str->members)
-      first = str->members->declarators;
-  } else if (idl_is_union(constr_type)) {
-    idl_union_t *u = (idl_union_t*)constr_type;
-    iterate_func = &next_decl_union;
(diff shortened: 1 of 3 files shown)


src/idl/tests/annotation.c
```diff
@@ -455,93 +455,152 @@ CU_Test(idl_annotation, redefinition)
...
-  CU_ASSERT_EQUAL(s1->autoid, IDL_AUTOID_SEQUENTIAL);
-  idl_delete_tree(tree);
+  for (size_t i = 0; i < sizeof(tests)/sizeof(id_test_t); i++)
+    test_id(tests[i]);
 }
 
-// x. autoid twice
-// x. autoid (HASH)
-// x. autoid (SEQUENTIAL)
+CU_Test(idl_annotation, autoid_module)
```
k0ekk0ek, line 585: These tests aren't a good fit for the `test_id` function IMHO. I'd like to see some that try to break the assignment of ids. So different branches etc.

src/idl/tests/annotation.c
```diff
@@ -455,93 +455,152 @@ CU_Test(idl_annotation, redefinition)
...
+    {"@autoid(HASH) struct s { char c; @hashid(\"c\") char d; };",                  IDL_RETCODE_SEMANTIC_ERROR, {0},              {0},            {0}}                                      //clashing hashid fields
+  };
 
-#if 0
-CU _ Test(idl_annotation, autoid_struct)
+  for (size_t i = 0; i < sizeof(tests)/sizeof(id_test_t); i++)
+    test_id(tests[i]);
+}
+
+CU_Test(idl_annotation, autoid_inheritance)
```
k0ekk0ek, line 563: I think it would be good to cover some weird scenarios here. So a derived of a derived that has a higher id than the intermediate. That sort of thing.

src/idl/src/annotation.c
```diff
@@ -38,17 +38,18 @@ annotate_id(
...
-    } else if (member->id.annotation) {
+    } else if (member->declarators->id.annotation) {
       idl_error(pstate, idl_location(annotation_appl),
         "@id conflicts with earlier annotation");
       return IDL_RETCODE_SEMANTIC_ERROR;
     }
-    member->id.annotation = annotation_appl;
-    member->id.value = literal->value.uint32;
+    member->declarators->id.annotation = annotation_appl;
+    member->declarators->id.value = (literal->value.uint32 & 0x0FFFFFFFu);
```
k0ekk0ek, line 52: Why the `0x0FFFFFFFu`? Ensuring it's within range?

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+idl_retcode_t
+propagate_autoid_struct(
+  idl_pstate_t *pstate,
+  idl_struct_t *str)
+{
+  assert(str);
+
+  size_t n_ids = 0;
+  uint32_t last = 0xFFFFFFFF;
+  bool last_is_inherited = false;
```
k0ekk0ek, line 1036: I'd switch this to a `const idl_declarator_t *` and see if the ancestor of that matches the current.

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+    else
+      current = NULL;
+  }
+
+  qsort(id_array, n_ids, sizeof(*id_array), &compare_id);
+
+  for (size_t i = 1; i < n_ids; i++) {
+    if (id_array[i-1] == id_array[i]) {
+      idl_error(pstate, loc_array[i],
+        "Member id '0x%.07x' is already assigned", id_array[i]);
```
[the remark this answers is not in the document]
k0ekk0ek: I'm thinking most people would annotate in decimal form? Not sure on that though...

k0ekk0ek (changes requested):

src/idl/src/annotation.c
```diff
@@ -107,26 +109,34 @@ annotate_autoid(
...
   if (annotation_appl->parameters) {
     enumerator = idl_identifier(annotation_appl->parameters->const_expr);
     if (strcmp(enumerator, "HASH") == 0) {
       autoid = IDL_HASH;
     } else if (strcmp(enumerator, "SEQUENTIAL") == 0) {
       autoid = IDL_SEQUENTIAL;
+    } else {
+      idl_error(pstate, idl_location(annotation_appl),
+        "@autoid does not support '%s' as a parameter", enumerator);
+      return IDL_RETCODE_SEMANTIC_ERROR;
```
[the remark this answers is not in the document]
reicheratwork: This will be corrected.

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+    }
+    if (ret)
+      break;
+  }
+
+  return ret;
+}
+
+idl_retcode_t
+propagate_autoid_struct(
```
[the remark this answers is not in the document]
reicheratwork: The way it works at the moment is that the inheriting structs will each proceed with their own id list, and in their own manner of automatically populating the id fields.
Which at the most basic level is sequential ids (Xtypes v1.3 7.3.1.2.1.1).
So if a base struct has as last member with id 123, then all structs inheriting from the base will have as first member id 124 (if nothing is explicitly set).

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+  idl_declarator_t *decl = NULL;
+  idl_retcode_t ret = IDL_RETCODE_OK;
+
+  idl_struct_t *current = str;
+  while (current) {
+    IDL_FOREACH(mem, current->members) {
+      IDL_FOREACH(decl, mem->declarators) {
+        n_ids++;
+        if (current != str
+         && !last_is_inherited)
```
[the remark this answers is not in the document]
reicheratwork: Okay.

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+  idl_node_t *node = NULL;
+  idl_retcode_t ret = IDL_RETCODE_OK;
+  IDL_FOREACH(node, list) {
+    if (idl_is_module(node)
+     && !((idl_module_t*)node)->autoid.annotation) {
+      idl_module_t *mod = (idl_module_t*)node;
+      mod->autoid.value = toset;
+      ret = propagate_autoid(pstate, mod->definitions, toset);
+    }
+    if (idl_is_struct(node)
```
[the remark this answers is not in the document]
reicheratwork: I will restructure this branching.

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+  if (*(const uint32_t*)lhs < *(const uint32_t*)rhs)
+    return -1;
+  else if (*(const uint32_t*)lhs > *(const uint32_t*)rhs)
+    return 1;
+  else
+    return 0;
+}
+
+idl_retcode_t
+propagate_autoid(
```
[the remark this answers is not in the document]
reicheratwork: I will rename this function 'idl_propagate_autoid', it cannot be static as it is used in annotation.c and processor.c.

src/idl/src/tree.c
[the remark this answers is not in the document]
reicheratwork: I will rename this function 'idl_propagate_autoid', it cannot be static as it is used in annotation.c and processor.c.

src/idl/src/tree.c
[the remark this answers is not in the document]
reicheratwork: I chose to use hex notation here as ids can originate from hashes and explicit assignment of numbers, and in the case of hashes, the decimal representation made less sense in my eyes.

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+
+  if (!n_ids)
+    return ret;
+
+  uint32_t *id_array = malloc(sizeof(uint32_t)*n_ids);
+  const idl_location_t **loc_array = malloc(sizeof(const idl_location_t*)*n_ids);
+  if (!id_array || !loc_array) {
+    ret = IDL_RETCODE_NO_MEMORY;
+    goto finish;
+  }
```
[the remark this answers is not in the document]
reicheratwork: By just storing the pointer to the declarators and modifying the comparison function, I only need a single array.
This is fixed in a different manner.

src/idl/src/tree.h
```diff
@@ -59,6 +59,17 @@ idl_create_unary_expr(
...
+idl_retcode_t
+propagate_autoid(
+  idl_pstate_t *pstate,
+  idl_node_t *list,
+  idl_autoid_t toset);
+
+idl_retcode_t
+propagate_autoid_struct(
+  idl_pstate_t *pstate,
+  idl_struct_t *str);
```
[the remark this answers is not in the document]
reicheratwork: It is also used in calculating the ids in a struct if it has the @autoid annotation on it (annotation.c:annotate_autoid)

src/idl/src/tree.c
```diff
@@ -983,6 +984,136 @@ static const char *describe_struct(const void *ptr)
...
+            assert(decl->name->identifier);
+            decl->id.value = idl_hashid(decl->name->identifier);
+          } else {
+            decl->id.value = (last + 1) & 0x0FFFFFFFu;
+          }
+        }
+
+        last = decl->id.value;
+        loc_array[n_ids] = idl_location(mem);
+        id_array[n_ids++] = decl->id.value;
```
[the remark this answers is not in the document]
reicheratwork: You are correct.
This makes another request simpler as well.

src/idl/tests/annotation.c
```diff
@@ -455,93 +455,152 @@ CU_Test(idl_annotation, redefinition)
...
-    { "struct s { @id char c; };", IDL_RETCODE_SEMANTIC_ERROR }, // @id without const-expr
-    { "struct s { @id(1) @id(1) char c; };", IDL_RETCODE_OK }, // duplicate @id
-    { "struct s { @id(1) @id(2) char c; };", IDL_RETCODE_SEMANTIC_ERROR }, // conflicting @id
-    { "struct s { @id(1) @hashid char c; };", IDL_RETCODE_SEMANTIC_ERROR } // @id and @hashid
-  };
+  idl_pstate_t *pstate = NULL;
+  idl_retcode_t ret = parse_string(IDL_FLAG_ANNOTATIONS, test.s, &pstate);
+  CU_ASSERT_EQUAL(ret, test.ret);
+  if (ret == IDL_RETCODE_OK
+   && ret == test.ret) {
```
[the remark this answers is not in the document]
reicheratwork: Okay

src/idl/src/annotation.c
reicheratwork: This is to generate the rollover of the member-id values, as the consumers of the member ids (the xtypes implementations) only use the 28 least significant bits.

src/idl/tests/annotation.c
```diff
@@ -455,93 +455,152 @@ CU_Test(idl_annotation, redefinition)
...
+      IDL_FOREACH(mem, s->members) {
+        IDL_FOREACH(decl, mem->declarators) {
+          CU_ASSERT_TRUE_FATAL(n < sizeof(test.aid)/sizeof(test.aid[0]));
+          if (test.annotation_present[n]) {
+            CU_ASSERT_PTR_NOT_NULL(decl->id.annotation);
+          } else {
+            CU_ASSERT_PTR_NULL(decl->id.annotation);
+          }
+          if (decl->id.value != test.id[n]) {
+          CU_ASSERT_EQUAL(decl->id.value, test.id[n]);
```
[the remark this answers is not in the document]
reicheratwork: Check.

src/idl/tests/annotation.c
reicheratwork: I will make additional tests

src/idl/src/tree.c
reicheratwork: This approach has been changed.

src/idl/src/parser.y
```diff
@@ -814,8 +814,10 @@ case_label:
   ;
 
 element_spec:
-    type_spec declarator
-      { TRY(idl_create_case(pstate, LOC(@1.first, @2.last), $1, $2, &$$)); }
+    annotations type_spec declarator
+      { TRY(idl_create_case(pstate, LOC(@1.first, @3.last), $2, $3, &$$));
+        TRY_EXCEPT(idl_annotate(pstate, $$, $1), free($$));
+      }
```
k0ekk0ek, line 820: Can union element-spec non-terminals be annotated? I was under the impression that only the switch-type-spec and the union itself can be annotated? Perhaps due to the `@id` and `@hashid` annotations?

src/idl/src/tree.c
```diff
@@ -983,6 +984,204 @@ static const char *describe_struct(const void *ptr)
...
+  idl_pstate_t *pstate,
+  idl_struct_t *str)
+{
+  assert(str);
+
+  size_t n_ids = 0;
+  idl_member_t *mem = NULL;
+  idl_declarator_t *decl = NULL,
+                   *strlast = NULL,
+                   *last = NULL;
```
k0ekk0ek, line 1067: Style.

src/idl/src/tree.c
```diff
@@ -983,6 +984,204 @@ static const char *describe_struct(const void *ptr)
...
+    if (current->inherit_spec)
+      current = current->inherit_spec->base;
+    else
+      current = NULL;
+  }
+
+  if (!n_ids)
+    return ret;
+
+  idl_declarator_t **id_array = malloc(sizeof(idl_declarator_t *)*n_ids);
```
k0ekk0ek, line 1091: `const`

src/idl/src/tree.c
```diff
@@ -983,6 +984,204 @@ static const char *describe_struct(const void *ptr)
...
+    free((void*)id_array);
+
+  return ret;
+}
+
+idl_retcode_t
+idl_propagate_autoid_union(
+  idl_pstate_t *pstate,
+  idl_union_t *u)
+{
```
k0ekk0ek, line 1137: There's quite a bit of duplication between all these functions. How about just writing some iteration functions that can be passed as a function argument? A bit inline with what `idl_iterate` does.

src/idl/src/parser.y
k0ekk0ek: Maybe you have a link to an example or the spec or something?

src/idl/tests/annotation.c
k0ekk0ek: Still like to see tests with some additional modules.

src/idl/src/parser.y
reicheratwork: This is described in the XTypes v1.3 spec table 21.

I will also add this to the files.

src/idl/src/tree.c
reicheratwork: You are right, i will fix this.

src/idl/src/tree.c
reicheratwork: There is some code being duplicated here, true.
But the actions between the two functions (for instance, which items they iterate over, unions directly over cases, structs over members and declarators and everything that is inherited) are different enough IMO that creating a more generic implementation would only add complexity, which is why I chose this more explicit approach.

src/idl/src/tree.c
reicheratwork: Check.

src/idl/tests/annotation.c
reicheratwork: I will create more tests for autoid on modules

src/idl/src/tree.c
reicheratwork: I switched to a different approach: not keeping track of the actual last id value, but just keeping the last declarator (which has the id value).

src/idl/src/tree.c
```diff
@@ -983,6 +984,202 @@ static const char *describe_struct(const void *ptr)
...
+static void
+process_declarator_id(
+  idl_declarator_t *decl,
+  idl_declarator_t *last,
+  idl_autoid_t autoid)
+{
+  assert(decl);
+
+  if (decl->id.annotation) {
+    /*this declarator already has an ID annotation*/
```
k0ekk0ek, line 1045: Why not just return immediately?

src/idl/src/tree.c
```diff
@@ -983,6 +984,202 @@ static const char *describe_struct(const void *ptr)
...
+  } else if (autoid == IDL_HASH) {
+    assert(decl->name->identifier);
+    decl->id.value = idl_hashid(decl->name->identifier);
+  } else if (last) {
+    decl->id.value = (last->id.value + 1) & 0x0FFFFFFFu;
```
k0ekk0ek, line 1050: Might be possible to overflow? Come to think of it, did we add a test for that?

src/idl/src/tree.c
k0ekk0ek: The logic is exactly the same. Determine the number of declarators, allocate, assign then sort. So implementing a static function to return the next declarator to iterate is more elegant IMHO. It'll even save complexity if you'd only use it for `struct` types...

src/idl/src/tree.c
```diff
@@ -983,6 +984,202 @@ static const char *describe_struct(const void *ptr)
...
+  while (current) {
+    IDL_FOREACH(mem, current->members) {
+      IDL_FOREACH(decl, mem->declarators) {
+        n_ids++;
+        if (current != str && !last)
+          strlast = decl;
+      }
+    }
+    if (!last)
+      last = strlast;
```
k0ekk0ek, line 1078: To work around inheritance for iteration wrt struct vs union, we can just always set last like with union traversal and then check if we shouldn't set it to `NULL` after if the ancestor of the declarator is the same as the node we started at.

src/idl/src/parser.y
k0ekk0ek: Much appreciated. The spec is rather unclear where annotations can/cannot be applied so :+1:

src/idl/src/tree.c
k0ekk0ek: My bad. :+1:

k0ekk0ek:
Just a small number of small remarks, other than that changes look good.

src/idl/src/tree.c
reicheratwork: I will change this.

src/idl/src/tree.c
reicheratwork: It is possible to overflow.
After discussions with @e-hndrks it was decided that overflow was something that would not be flagged as an error.

k0ekk0ek: Fine with that. Please place some small comment above this line so that's explicit. :+1:

src/idl/src/tree.c
reicheratwork: Made a more generic implementation, so this is no longer applicable.

k0ekk0ek:
@dennis-adlink, do you mind reviewing this? I've just pushed a patch that takes into account empty structs in the inheritance line. Perhaps, while you're at it, look over the other changes too?

k0ekk0ek:
Some static analysis builds are rightfully complaining about a possible `malloc` of 0 bytes. I'll fix those tomorrow, shouldn't matter too much for the logic though.

dennis-adlink (commented):
LGTM, good to have this extensive set of tests for id/autoid annotations! Only issue I found is the auto-id calculation (which is not part of this PR), see comment below.

src/idl/src/tree.c
```diff
@@ -983,6 +984,184 @@ static const char *describe_struct(const void *ptr)
...
+}
+
+static idl_declarator_t *
+next_case(const void *node, const idl_declarator_t *declarator)
+{
+  const idl_case_t *branch;
+
+  assert(idl_type(node) == IDL_UNION);
+  if (!declarator)
+    return ((const idl_union_t *)node)->cases->declarator;
```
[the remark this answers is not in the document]
k0ekk0ek: Good point. Didn't know about XTypes adding support for `union` inheritance. Added a comment with the section number.

src/idl/src/tree.c
```diff
@@ -983,6 +984,184 @@ static const char *describe_struct(const void *ptr)
...
+
+  /* id assigned through @id or @hashid annotation */
+  if (declarator->id.annotation)
+    return;
+
+  assert(declarator->name);
+  assert(declarator->name->identifier);
+
+  if (autoid == IDL_HASH)
+    declarator->id.value = idl_hashid(declarator->name->identifier);
```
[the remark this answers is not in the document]
k0ekk0ek: Probably easiest if I create an issue for this.

PR #966, opened 2021-09-28
Enable shared memory with transient local qos

### LINKED ISSUE #965: Enable shared memory support for transient local readers and writers
Enable iceoryx shared memory support for transient local readers and writers.
We also allow best effort writers, i.e. have no restriction on reliability.

MatthiasKillat:
Closes #965

Note: based on https://github.com/eclipse-cyclonedds/cyclonedds/pull/943
Merge after this PR

### DIFF
--- a/src/core/ddsc/src/dds_reader.c
+++ b/src/core/ddsc/src/dds_reader.c
@@ -482,15 +482,13 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
     return false;
   }
 
-  uint32_t sub_history_req = cfg->sub_history_request;
-
   return (NULL != qos &&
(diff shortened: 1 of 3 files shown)


MatthiasKillat:
@elfenpiff @eboasson I am not sure this is sufficient to correctly support transient local writers but I think it is necessary. We will check with further tests in the C++ API.

Before we merge this we need to merge  #943 anyway.

It is therefore a draft and NOT to be merged as of now.

MatthiasKillat:
The PR will be abandoned and recreated in the near future. This is preferable due to some internal API change as well as the upcoming iceoryx 2.0 API (which is needed to allow DDS compliant transient local reader matching).

MatthiasKillat:
Replaced by https://github.com/eclipse-cyclonedds/cyclonedds/pull/1180

PR #969, opened 2021-09-29
Fix for issue #968

### LINKED ISSUE #968: GUID byte ordering swapped
In the initialization of the ddsi_guid_prefix in the rtps_init function, the ordering of the bytes in the GUID does not follow that of the specification (e.g. the vendor ID should be first two bytes).
This is due to the different manners of addressing the bytes, in 4 byte units (as the union branch u) or as 1 byte units (as the union branch s), which causes an implicit byte swap being done.

The GUID should have the vendor ID (which for cyclonedds should be 0x01, 0x10) as the bytes 1/16 and 2/16, and 13-15/16 should be a counter of the lokal entities created.
E.g.: one of the GUIDs retrieved from a builtin topic `2d 61 10 01 9a 27 b5 b0 26 43 4d 2f 00 00 01 02`

Looking at this, it seems that the bytes for the vendor id are swapped and at the 3rd and 4th bytes.

reicheratwork:
- Add a byte swap to the participant GUID prefix
- This fixes #968


eboasson (approved):
Nice catch! That no-one ever noticed it before pretty much confirms my suspicion that the change in the specification that introduced the rule was a mistake, throwing away 2 bytes out of only 12 for a system-wide unique identifier for not much benefit.

Anyway, I'm glad that with this change we're finally throwing away the correct 2 bytes 🙂 Thanks @reicheratwork!

PR #970, opened 2021-09-30
Support anonymous types in union cases

k0ekk0ek:
What it says in the title. Unions got left behind for some reason...

### DIFF
--- a/src/tools/idlc/src/types.c
+++ b/src/tools/idlc/src/types.c
@@ -46,7 +46,7 @@ emit_implicit_sequence(
     if (idl_is_sequence(type_spec))
       return IDL_VISIT_REVISIT | IDL_VISIT_TYPE_SPEC;
   } else {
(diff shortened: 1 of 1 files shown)


src/tools/idlc/src/types.c
```diff
@@ -108,7 +108,7 @@ generate_implicit_sequences(
   memset(&visitor, 0, sizeof(visitor));
   visitor.visit = IDL_MEMBER | IDL_SEQUENCE;
```
dennis-adlink, line 109: This also needs `IDL_CASE`, and in `emit_implicit_sequence` the assert should also accept a case (line 49)

k0ekk0ek: Added them, probably got lost when I removed the debug code. Thanks!

dennis-adlink (approved):

PR #972, opened 2021-09-30
Fix incorrect bitshift in filetime_to_time.

thijsmie:
I love single-character merge requests 😄


eboasson (approved):
That is quite an underhanded way of undercounting time used!

PR #973, opened 2021-10-01
Add checks on inheritance

reicheratwork:
- Check that there are no clashes between member names
- Check that extensibility matches between inheritance
  ([XTypes 1.3] 7.2.2.4.5)
- Check that inherited keylist is not modified
  ([XTypes 1.3] 7.2.2.4.5)
- Add unittests to check for these limitations


src/idl/tests/inheritance.c
```diff
@@ -17,27 +17,76 @@
...
+    CU_ASSERT_STRING_EQUAL(idl_identifier(derived), "derived");
+    CU_ASSERT_EQUAL(derived->extensibility.value, test.derived_ext);
+
+    CU_ASSERT(derived->inherit_spec && derived->inherit_spec->base == base);
+
+    idl_delete_pstate(pstate);
+  }
+}
+
+CU_Test(idl_inheritance, extensibility_conflict)
```
k0ekk0ek, line 70: This test mixes and matches too much functionality that has very little to do with each other IMHO. Name clashes are part of scoping checks while extensibility checks are part of the code added in this commit. Both have nothing to do with keys. I think it's wise to split that up.

src/idl/src/processor.c
```diff
@@ -380,7 +411,17 @@ idl_retcode_t idl_parse(idl_pstate_t *pstate)
...
-  return ret;
+  /*check that the extensibilities of inherited and inheriting match*/
+  idl_visitor_t visitor;
+  memset(&visitor, 0, sizeof(visitor));
+
+  visitor.visit = IDL_STRUCT;
+  visitor.accept[IDL_ACCEPT_STRUCT] = &check_inheritance;
+
+  if (ret == IDL_RETCODE_OK)
+    return idl_visit(pstate, pstate->root, &visitor, NULL);
```
k0ekk0ek, line 422: While this technically works. The visitor pattern is meant for consumers. This works, but is slower because it traverses much more than just structs. Also, I'd like to keep the split intact.

src/idl/src/tree.c
```diff
@@ -990,6 +990,34 @@ idl_finalize_struct(
...
+              idl_error(state, idl_location(decl),
+                "Member name %s clashes with declaration in inheritance", decl->name->identifier);
+              return IDL_RETCODE_SEMANTIC_ERROR;
+            }
+          }
+        }
+      }
+    }
+  }
+
```
k0ekk0ek, line 1020: This should really be implemented in the scoping implementation. There's imports already that check for names etc. I meant to implement it there for inheritance too, but it must've slipped my mind.

k0ekk0ek (changes requested):
Thanks @reicheratwork! A useful addition for sure, but I have my doubts about the correctness of the check. If a derived struct is mutable, then it seems to me each base struct can either be final, appendable or mutable. If it's appendable, then the base struct should be final as is the case when the derived struct is final itself. At least, that would make the most sense from a technical perspective if all members from the base struct are inserted first like with C structs. @e-hndrks, @dennis-adlink, what do you guys think?

src/idl/src/tree.c
e-hndrks: @k0ekk0ek The spec states that an inheriting struct must either have the same extensibility as its parent or have no explicit extensibility type specified. But I agree that the approach as proposed in section 7.2.2.4.5 in the XTypes spec causes potential backward incompatibility issues when people add attributes to non-mutable structs that are then inherited by other non-mutable structs. I have posted a question about this to the XTypes RTF under ticket DDSXTY14-51.

PR #974, opened 2021-10-04
Fix use of non-existent paths when building as a subproject

robinlinden:
CMAKE_SOURCE_DIR always points to the top-level CMake project, so if
CycloneDDS was included as a gitmodule or pulled in using CMake
FetchContent, the CMake step would fail with errors like

```
CMake Error at cyclonedds/src/tools/idlc/CMakeLists.txt:52 (file):
file failed to open for reading (No such file or directory):
/home/robin/dds-project/src/ddsrt/include/getopt.h.in

CMake Error: File /home/robin/dds-project/src/ddsrt/src/getopt.c does not exist.
CMake Error at cyclonedds/src/tools/idlc/CMakeLists.txt:61 (configure_file):
configure_file Problem configuring file

CMake Error at cyclonedds/src/tools/idlc/CMakeLists.txt:95 (include):
include could not find load file:
/home/robin/dds-project/cmake/Modules/Generate.cmake
```

This commit replaces CMAKE_SOURCE_DIR with <PROJECT_NAME>_SOURCE_DIR, a
variable that always points to the source dir of the named project.

See:
https://cmake.org/cmake/help/v3.21/variable/CMAKE_SOURCE_DIR.html
https://cmake.org/cmake/help/v3.21/variable/PROJECT-NAME_SOURCE_DIR.html

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -11,7 +11,7 @@
 #
 set(CMAKE_INSTALL_EXAMPLESDIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}/examples")
 
-include("${CMAKE_SOURCE_DIR}/cmake/Modules/Generate.cmake")
+include("${CycloneDDS_SOURCE_DIR}/cmake/Modules/Generate.cmake")
 
 install(
   FILES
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
Thanks @robinlinden! I think I would've gone for `PROJECT_SOURCE_DIR` myself, but since this is available in at least 3.7 I don't see any reason to not use `CycloneDDS_SOURCE_DIR`. Much appreciated :+1:

robinlinden:
@k0ekk0ek That's fair. I normally give each "standalone" part of my projects its own `project()`-call, but in this case `PROJECT_SOURCE_DIR` absolutely would've worked.

Thank you for the great project! :)

PR #979, opened 2021-10-07
remove grammars from legal notice

alsora:
This PR removes grammars from the legal notice since it's not present anymore in cyclonedds.

Signed-off-by: Alberto Soragna <alberto.soragna@gmail.com>

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -35,10 +35,6 @@ libmd5-rfc (2002-04-13)
 
 * License: Zlib
 
-grammers-v4 (git SHA 6610e82ee235992f50e108cd59204f3bcd7128c1)
-
-* License: New BSD License
-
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Thanks @alsora! Good catch. Pretty sure this particular commit has nothing to do with the failing builds so we can ignore that, but I'm afraid I do have to insist on the ECA (even though it maybe a bit much for these types of changes). If you click the details link next to the failing eclipsefdn/eca check the process is pretty self explanatory.

k0ekk0ek:
You're sure the email address you used to register with Eclipse is the one you use for Git? It's either that or the service that provides the check is down.

alsora:
Thank you!
Now it passed. 
Indeed I only registered.

PR #989, opened 2021-10-14
IDLC command line option for default extensibility

dennis-adlink:
This adds a command line option to the C generator for IDLC to set the default extensibility for aggregated types. Aggregated types that do not have their extensibility explicitly set, get the default value. In case no default is provided, 'final' is used. Because this is different than the default in the XTypes spec (which is 'appendable'), a warning is shown in case there are types in the IDL that do not have an explicit extensibility set and no default is provided by the command line option. Enumerated types can also have an extensibility annotation, but that is currently not supported by the C generator.

This PR depends op #903, so only commit 044dfd2 is relevant. 

@reicheratwork the C++ back-end probably also needs to support for this option, I haven't implemented that as part of this PR. If we do, the implementation for this option should probably move from the C generator to idlc.c to make it a generic option for all back-ends.

### DIFF
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -11,7 +11,7 @@
 #
 
 function(IDLC_GENERATE)
-  set(one_value_keywords TARGET)
+  set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY)
   set(multi_value_keywords FILES FEATURES INCLUDES)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
@@ -20,11 +20,12 @@ function(IDLC_GENERATE)
     FILES ${IDLC_FILES}
     FEATURES ${IDLC_FEATURES}
     INCLUDES ${IDLC_INCLUDES}
+    DEFAULT_EXTENSIBILITY ${IDLC_DEFAULT_EXTENSIBILITY}
   )
 endfunction()
 
 function(IDLC_GENERATE_GENERIC)
-  set(one_value_keywords TARGET BACKEND)
+  set(one_value_keywords TARGET BACKEND DEFAULT_EXTENSIBILITY)
   set(multi_value_keywords FILES FEATURES INCLUDES SUFFIXES DEPENDS)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
@@ -87,6 +88,11 @@ function(IDLC_GENERATE_GENERIC)
     list(APPEND _depends ${_idlc_depends} ${IDLC_DEPENDS})
   else()
     set(_depends ${_idlc_depends})
+  endif()
+
+  if(IDLC_DEFAULT_EXTENSIBILITY)
+    set(_default_extensibility ${IDLC_DEFAULT_EXTENSIBILITY})
+    list(APPEND IDLC_ARGS "-x" ${_default_extensibility})
   endif()
 
   set(_dir ${CMAKE_CURRENT_BINARY_DIR})
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -12,7 +12,7 @@
 
 function(IDLC_GENERATE)
   set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY)
-  set(multi_value_keywords FILES FEATURES INCLUDES NO_WARN)
+  set(multi_value_keywords FILES FEATURES INCLUDES WARNINGS)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
 
@@ -20,14 +20,14 @@ function(IDLC_GENERATE)
     FILES ${IDLC_FILES}
     FEATURES ${IDLC_FEATURES}
     INCLUDES ${IDLC_INCLUDES}
-    NO_WARN ${IDLC_NOWARN}
+    WARNINGS ${IDLC_WARNINGS}
     DEFAULT_EXTENSIBILITY ${IDLC_DEFAULT_EXTENSIBILITY}
   )
 endfunction()
 
 function(IDLC_GENERATE_GENERIC)
   set(one_value_keywords TARGET BACKEND DEFAULT_EXTENSIBILITY)
-  set(multi_value_keywords FILES FEATURES INCLUDES NO_WARN SUFFIXES DEPENDS)
+  set(multi_value_keywords FILES FEATURES INCLUDES WARNINGS SUFFIXES DEPENDS)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
 
@@ -96,9 +96,9 @@ function(IDLC_GENERATE_GENERIC)
     list(APPEND IDLC_ARGS "-x" ${_default_extensibility})
   endif()
 
-  if(IDLC_NO_WARN)
-    foreach(_no_warn ${IDLC_NO_WARN})
-      list(APPEND IDLC_ARGS "-Wno-${_no_warn}")
+  if(IDLC_WARNINGS)
+    foreach(_warn ${IDLC_WARNINGS})
+      list(APPEND IDLC_ARGS "-W${_warn}")
     endforeach()
   endif()
 
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -12,7 +12,7 @@
 
 function(IDLC_GENERATE)
   set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY)
-  set(multi_value_keywords FILES FEATURES INCLUDES)
+  set(multi_value_keywords FILES FEATURES INCLUDES NO_WARN)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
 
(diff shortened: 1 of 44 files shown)


dennis-adlink:
> Wouldn't this be better implemented as a `#` pragma? When it's a command line option it doesn't carry over with the IDL right?

I think in this case a command line option is a better choice, because the idea is that an IDL file that worked with earlier versions of Cyclone does not need to be changed. That's also the reason for choosing 'final' as the current default, and not 'appendable' as it should be according to the XTypes spec. 

Adding a pragma for this would also give other challenges, like what the scope of the pragma option is (current file, or also propagated to included files, and how to handle conflicting pragmas). Another alternative would be using an (Cyclone specific) annotation on a module, like the default-nested, but this of course also has the same problem that IDL would need to be changed.

dennis-adlink:
> * If the default is currently "final", why explicitly pass that in the test code? Presumably this would be to guard against a possible future change to match the spec, but then the next question arises:
> * Do we care what the default extensibility of most of the topics used in tests is? I don't think so, and for those where it matters, would it not make more sense to annotate them properly?

The reason for adding this in the tests was to prevent that the 'unspecified extensibility' warning is shown when building. But you're right, in most cases it doesn't matter. I've now changed this, and only annotated the types where it matters, in cases where we don't want to have the (very limited) overhead of appendable types. 

Because initially I though it was a good idea to add the extensibility in all test IDL (to avoid the warning), and as I was going through all IDL, I've also replace all `pragma keylist` by `@key` (some cleaning-up I should have done in #904). 

Do we need to avoid these warnings by either adding the annotations or passing a default when idl-compiling for tests? Or remove the warning at all (although I think it's the best option to move towards using appendable as default)? 

> * Finally, I don't understand why the selected default extensibility needs to passed through in an additional argument in so many places. Can't we just set the extensibility on all types after parsing and then use it without worrying about the default and whether or not it was explicitly annotated?

Good point, I've refactored this in 014eff2

src/tools/idlc/src/idlc.c
```diff
@@ -358,6 +372,9 @@ static idl_retcode_t idlc_parse(void)
     ret = idl_parse(pstate);
     assert(ret != IDL_RETCODE_NEED_REFILL);
     if (ret == IDL_RETCODE_OK) {
+      if ((ret = set_type_extensibility()) != IDL_RETCODE_OK)
```
dennis-adlink, line 375: Note to self: move setting default extensibility to `idl_parse`, because this is also required for validations when parsing the idl (e.g. must-understand usage wrt extensibility)

dennis-adlink:
In the most recent (force) push, I've changed and added a few things:
- Added a `-Wno-...` option to IDLC, so that warnings can be disabled from the command line (also added a `NO_WARN` parameter to the cmake `idl_generate` function)
- Disable the 'implicit extensibility' warning when compiling IDL for tests, in case the extensibility is not relevant for the test case (and thus not set in the IDL)
- Introduced an `implicit` flag in `IDL_ANNOTATABLE` to indicate that an annotation value was set implicitly. This is needed when setting default extensibility on types, to skip types that already got an extensibility value, e.g. by inheriting from another type (and this flag could also be useful for other future cases)

dennis-adlink:
With this (force) push, I've added commit ddc229d that includes refactoring of passing configuration options to the IDL parser and generator, so that config that is only used in the generator is not in the `pstate` object. This moves the `generate_typeinfo_typemap` function pointer to a separate parameter of the `generate` function, which also means that the C++ and Python generators need to have this parameter.

thijsmie (approved):
Looks good to me, we'll have to merge a minor change to the python idl frontend too.

src/idl/include/idl/tree.h
```diff
@@ -207,7 +207,7 @@ enum idl_extensibility {
    the construct itself may be referenced. e.g. @default_nested annotations
    are only referenced by the annotated module, not by any subconstructs */
 #define IDL_ANNOTATABLE(type) \
-  struct { const struct idl_annotation_appl *annotation; type value; }
+  struct { const struct idl_annotation_appl *annotation; type value; bool implicit; }
```
eboasson, line 210: This makes the `implicit` flag apply to all annotations. Doesn't it make more sense to add this flag to the extensibility annotation only?

src/idl/src/tree.c
```diff
@@ -3502,10 +3518,64 @@ bool idl_is_topic_key(const void *node, bool keylist, const idl_path_t *path, ui
...
+bool idl_has_implicit_default_extensibility(const idl_node_t *node)
+{
+  assert(node);
+  for (; node; node = idl_next(node)) {
+    if (idl_mask(node) == IDL_MODULE) {
+      const idl_module_t *module = (const idl_module_t *)node;
+      return idl_has_implicit_default_extensibility(module->definitions);
+    } else if (idl_mask(node) == IDL_STRUCT) {
+      const idl_struct_t *_struct = (const idl_struct_t *)node;
+      if (!_struct->extensibility.annotation && !_struct->extensibility.implicit)
```
eboasson, line 3541: I find it surprising that for this function to return `true` (that is, "implicit"), the `implicit` field has to be false. There's got to be a better solution!

eboasson (commented):
I have a few questions, but overall it looks ok to me.

examples/shm_throughput/ShmThroughput.idl
```diff
@@ -1,146 +1,145 @@
...
-#pragma keylist DataType_524288
-
-struct DataType_1048576
-{
-  unsigned long long count;
-  unsigned long payloadsize;
-  octet payload[1048576 - 12];
-};
-#pragma keylist DataType_1048576
+  @final
```
reicheratwork, line 3: No key fields here or #pragma keylist, so shouldn't there be an @topic before each type which we want to make into a topic?

dennis-adlink: By default, types are considered topic types (the default of the `@default_nested` annotation is `false`), so there is no need to explicitly add `@topic`.

hansvanthag: It just doesn't feel right that 'by default' all types (that are not explicitly stated as being 'nested') are considered to be/become topics i.e. require descriptor- and code-generation. At least that wasn't the case in the pre-Xtypes era.
Apparently its also stated that an IDL-compiler 'shall' have the ability to 'overrule' that default_nested behavior at 'top-level', but that sounds like 'self-inflicted damage'. 
So summarizing, I think it would really makes sense (to avoid unnecessary descriptor/code-generation that might be 'huge' given some elaborate data-models) to have to explicitly specify which types are meant to be/become topics.

And in-case that's not achievable, perhaps 'our' IDL-compiler could 'by-default' overrule the default_nested definition to achieve the same/desirable behavior.

Also curious about your opinion @eboasson ...

eboasson: I think the desired behaviour is mostly covered by:

> 7.3.1.2.1.16 Explicit declaration of Topic types
> A type may be explicitly annotated to indicate the intent to use it as a top-level DDS Topic. Absent this annotation only non-nested types may be used as DDS Topics. The@topic annotation overrides that default behavior.

because a carefully structured set of IDL files will usually be fine under this rule. However if I'd have a library of common type definitions and then (say) one IDL file per topic each using a subset of the types in that common library, it'd still produce a mess. If I am right in that, I wouldn't mind a mode that says "if there are types annotated with @topic, only those types may be used as DDS Topics".

hansvanthag: indeed, but that i.m.o. sensible mode would imply that that library should be _explicitly_ stated to include only nested types as by default each type would be treated as a non-nested type i.e. a to-become-topic (and thus producesthat mess)

eboasson: I suppose that's what `@default_nested` is for ... and then you end up with the current behaviour being good enough.

hansvanthag: thats  fine .. just noticing that more people are placing question-marks with sensible default behavior, see https://issues.omg.org/issues/DDSXTY14-15

eboasson: According to that issue, the spec already says

> In addition to the above annotations, IDL compilers shall provide the means to change the default value for non-annotated top-level modules

which seems to allow you to do what you are looking for. I don't think IDLC supports that yet, but a possible implementation could be to have a command-line option for controlling the default of `@default_nested` very much in the same vein as what this PR does for the default extensibility.

src/idl/include/idl/tree.h
dennis-adlink: I think this flag could also be useful for other annotations, because other (e.g. `@id`, `@nested`) may also have an implicitly set values. So I've added code for setting this flag to `true` in these cases as well.

src/idl/src/tree.c
dennis-adlink: You're right, that's confusing! I've refactored this in 93447b3

cmake/Modules/Generate.cmake
```diff
@@ -12,8 +12,8 @@
 
 
 function(IDLC_GENERATE)
-  set(one_value_keywords TARGET)
-  set(multi_value_keywords FILES FEATURES)
+  set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY)
+  set(multi_value_keywords FILES FEATURES NO_WARN)
```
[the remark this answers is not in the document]
dennis-adlink: Fixed in d3362ce

src/idl/include/idl/tree.h
k0ekk0ek: The annotatable was written so that when there's no annotation, the value must be implicit. It makes for a tri-state, not sure if there's a reason that can't be used? Just curious, not a comment.

dennis-adlink:
In commit 861eef3 I've implemented a more elegant way to detect types with no extensibility set, without the need to add a `implicit` flag in `ANNOTATABLE`, thanks @k0ekk0ek!

examples/shm_throughput/ShmThroughput.idl
ichernev: Today I also read the `@nested` and `@topic` part in the IDL specification and it still doesn't make it clear how exactly the two interact. Listing all topic types (to me) seems more logical than listing all non-topic (`@nested`) types, but how that is controlled (idlc switch `--honor-topic / --honor-nested` possibly, but that seems weird).

PR #1000, opened 2021-10-25
Update serdata functions to handle sequences with shared memory

MatthiasKillat:
Now all types will be transferred with shared memory (if the QoS allow it).

This means that for dynamic types we serialize into shared memory (instead of the network buffer).

This requires the size of the serialization to be known, for which we now have the function
`sertype_get_serialized_size`. 

Currently this uses a placeholder implementation which actually serializes (so there is a lot of overhead). The plan is to replace it with a lightweight alternative which only computes the required size and is (almost?) independent of message size. This implementation is not part of this PR and will happen in a follow up issue.

@eboasson @sumanth-nirmal


src/core/ddsi/include/dds/ddsi/shm_types.h
```diff
@@ -0,0 +1,43 @@
...
+   uint32_t statusinfo;
+   uint32_t data_size;
+   unsigned char data_kind;
+   ddsi_keyhash_t keyhash;
+   iox_shm_data_state_t shm_data_state;
+};
+
+typedef struct iceoryx_header iceoryx_header_t;
+
+iceoryx_header_t *iceoryx_header_from_chunk(void *iox_chunk);
```
sumanth-nirmal, line 41: This probably is missing `DDS_EXPORT`, correct?

src/core/ddsi/include/dds/ddsi/shm_types.h
```diff
@@ -0,0 +1,43 @@
...
+#ifndef SHM_TYPES_H
+#define SHM_TYPES_H
+
+#include "dds/ddsi/q_protocol.h" /* for, e.g., SubmessageKind_t */
+#include "dds/ddsi/ddsi_tran.h"
+#include "dds/features.h"
+
+#include "dds/ddsi/ddsi_keyhash.h"
+#include "iceoryx_binding_c/chunk.h"
+
```
sumanth-nirmal, line 22: I think this header is also missing an ` extern “C”` block for C linkage, since this will also be used in the C++ code

src/core/ddsi/src/ddsi_sertype_default.c
```diff
@@ -205,6 +210,52 @@ static struct ddsi_sertype * sertype_default_derive_sertype (const struct ddsi_s
...
+
+  // but we do not have the serdata (we could construct it)
+  // TODO: which implementation to keep for now?
+#if 0
+  // TODO: this is off by 4 bytes (header?)
+  dds_ostream_t os;
+  dds_ostream_init(&os, 1024); // initial size chosen arbitrarily 
+  const struct ddsi_sertype_default *type_default = (const struct ddsi_sertype_default *)type;
+  dds_stream_write_sample(&os, sample, type_default);
+  return os.m_index;
```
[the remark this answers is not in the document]
MatthiasKillat: On second thought, it is not clear whether this should be written to the stream/serialized data part. I do not think so.

src/core/ddsi/include/dds/ddsi/shm_types.h
MatthiasKillat: Yes.

src/core/ddsi/src/ddsi_serdata_default.c
```diff
@@ -140,6 +140,27 @@ static bool serdata_default_eqkey_nokey (const struct ddsi_serdata *acmn, const
   return true;
 }
 
+#ifdef DDS_HAS_SHM
+static void serdata_default_free_iox_chunk(struct ddsi_serdata_default* d) {
```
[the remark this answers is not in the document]
MatthiasKillat: But most of the logic is to check the existence of the iceoryx chunk in serdata and then to lock and delete it. The call you propose would only do this but not work on serdata in any way.

I will introduce the call in ddsi and we can discuss whether we want to expose it in dds.

src/core/ddsi/include/dds/ddsi/ddsi_sertype.h
```diff
@@ -115,6 +115,23 @@ typedef void (*ddsi_sertype_free_samples_t) (const struct ddsi_sertype *d, void
 /* Serialized size for this type */
 typedef void (*ddsi_sertype_serialized_size_t) (const struct ddsi_sertype *d, size_t *dst_offset);
 
+/* Compute the serialized size based on the sertype information and the sample */
+// TODO: remove obsolete comments after discussion
```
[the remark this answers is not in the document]
MatthiasKillat: Interface clarified in review discussion.

sumanth-nirmal (changes requested):

src/core/ddsc/src/dds_write.c
```diff
@@ -579,6 +579,14 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
         // ret = DDS_RETCODE_OUT_OF_RESOURCES;
         // goto finalize_write;        
       } 
+    } else {
+      // The user already provided an iceoryx_chunk (by using the dds_loan API)
+      // We assume if we got the data from a loan it contains raw data
+      // (i.e. not serialized)
+      // This requires the user to adhere to the contract, we cannot enforce this
+      // with the given API.
+      iceoryx_header_t *iox_header = iceoryx_header_from_chunk(iox_chunk);
```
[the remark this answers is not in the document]
MatthiasKillat: True. Thanks

src/core/ddsc/src/dds_write.c
```diff
@@ -513,22 +550,43 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
...
       } else {
         // we failed to obtain a chunk, iceoryx transport is thus not available
-        // we will use the network path instead
+        // we cannot use the network path is now since due to the locators
+        // readers on the same machine would not get the data
+        // TODO: proper fallback logic to network path?
+
         iceoryx_available = false;
-      }
+        // TODO: activating this causes test failures
```
eboasson, line 578: Found it, or at least somewhat:
- the case where it needs to send via Iceoryx and also via the network ends up in the block starting at (new) line 620 (commented here with "X") and never sets `iox_chunk`. That in turn means the chunk never gets published, and after 8 (if I counted correctly) cases, it no longer can get a chunk from Iceoryx because there are too many chunks outstanding.
- the test then succeeds because `delivery_data_via_iceoryx` returns `false` if `iox_chunk` is a null pointer, and so the data gets delivered locally via Cyclone's native short-circuiting for local readers.

NB: This is not new in this PR, and clearly the test in `core/ddsi/tests/iceoryx.c` is too weak because it doesn't detect the underlying problem directly.

src/core/ddsc/src/dds_write.c
```diff
@@ -557,9 +615,12 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
 
     // do not serialize yet (may not need it if only using iceoryx or no readers)
     d = ddsi_serdata_from_loaned_sample (ddsi_wr->type, writekey ? SDK_KEY : SDK_DATA, data);
+        
   } else {
-    // serialize since we will need to send via network anyway
-    d = ddsi_serdata_from_sample (ddsi_wr->type, writekey ? SDK_KEY : SDK_DATA, data);
+    // serialize for network since we will need to send via network anyway
```
eboasson, line 620: X

src/core/ddsc/src/dds_write.c
```diff
@@ -513,22 +550,43 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
   thread_state_awake (ts1, &wr->m_entity.m_domain->gv);
 
   // 3. Check availability of iceoryx and reader status
-
   bool iceoryx_available = wr->m_iox_pub != NULL;
+  //bool has_fixed_size_type = wr->m_topic->m_stype->fixed_size;
```
eboasson, line 554: This can go

src/core/ddsc/src/dds_writer.c
```diff
@@ -284,15 +284,12 @@ const struct dds_entity_deriver dds_entity_deriver_writer = {
...
 #ifdef DDS_HAS_SHM
 #define DDS_WRITER_QOS_CHECK_FIELDS (QP_LIVELINESS|QP_DEADLINE|QP_RELIABILITY|QP_DURABILITY|QP_HISTORY)
-static bool dds_writer_support_shm(const struct ddsi_config* cfg, const dds_qos_t* qos, const struct dds_topic *tp)
+static bool dds_writer_support_shm(const struct ddsi_config* cfg, const dds_qos_t* qos)
 {
   if (NULL == cfg ||
       false == cfg->enable_shm)
     return false;
 
-  if (!tp->m_stype->fixed_size)
```
eboasson, line 293: Arguably, the condition for supporting transfer via Iceoryx is now `tp->m_stype->fixed_size || tp->m_stype->ops->get_serialized_size` (and the same for `dds_reader_supports_shm`).

@MatthiasKillat sorry for agreeing earlier that the topic is no longer needed ... although one could also argue that all `sertype` implementations simply must support `get_serialized_size` and `serialize_into`, that kinda doesn't work too well if the `plist` case then starts out with not supporting it.

eboasson (commented):
I figured out why returning `OUT_OF_RESOURCES` causes problems, and it is actually a bug that was introduced prior to this PR but that I think we may want to fix in this one. I'd like to know what you think, @MatthiasKillat.

Other than that, I just have a few minor comments.

src/core/ddsc/src/dds_write.c
```diff
@@ -513,22 +550,43 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
...
 
   if(iceoryx_available) {
     //note: whether the data was loaned cannot be determined in the non-iceoryx case currently
     if(!deregister_pub_loan(wr, data))
     {
-      void* chunk_data = create_iox_chunk(wr);
-      if(chunk_data) {
-        memcpy (chunk_data, data, wr->m_topic->m_stype->iox_size);
-        data = chunk_data;
+      size_t required_size = get_required_buffer_size(wr->m_topic, data);
```
eboasson, line 561: We should fail the operation with `BAD_PARAMETER` if the returned size is `SIZE_MAX` here.

src/core/ddsc/src/dds_write.c
```diff
@@ -513,30 +577,54 @@ dds_return_t dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstam
...
       } else {
         // we failed to obtain a chunk, iceoryx transport is thus not available
         // we will use the network path instead
+        // we cannot do this as the code is now (locators etc.)
+        // same machine nodes would not get the data?
+
         iceoryx_available = false;
-      }
+        // TODO: activating this causes test failures
+        //return DDS_RETCODE_OUT_OF_RESOURCES;
```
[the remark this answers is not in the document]
MatthiasKillat: Test is not able to deal with write failures. Disabled the fatal assert in this case. To be discussed. @eboasson

src/core/ddsc/src/dds_write.c
MatthiasKillat: I disabled this check in the test for now but I am not happy with it. Maybe I can fix this differently somehow from the description.

src/core/ddsc/src/dds_write.c
MatthiasKillat: Yes, but all these checks which almost always have the same result make me a sad panda. Will add it though.

src/core/ddsc/src/dds_writer.c
MatthiasKillat: I agree that this is better to check. I think you mean `&&` in the condition though.

src/core/ddsc/src/dds_write.c
MatthiasKillat: I now release the chunk if `deliver_data_via_iceoryx` fails but I still need to disable the check in the tests, otherwise it fails. It appears the problem is not fully understood. Unless the test uses the `dds_loan` API I do not see from the current code path where we may forget to either publish or release an iceoryx chunk.

@eboasson

MatthiasKillat: I think I fixed it. I also had to add more cleanup code in case of failures. Can probably be optimized somewhat, but the lack of RAII in C makes cleanup error-prone and cumbersome. Regardlesss, the problem was passing the iox_chunk in specific cases after `convert_serdata` (when it behaves as an identity). This function probably should not even exist but somehow it does. Unfortunately it makes the control flow even more convoluted.

src/core/ddsc/src/dds_write.c
```diff
@@ -449,8 +453,11 @@ static dds_return_t dds_writecdr_impl_common (struct writer *ddsi_wr, struct nn_
 
 #ifdef DDS_HAS_SHM
   // transfer ownership of an iceoryx chunk if it exists
-  d->iox_chunk = din->iox_chunk;
```
MatthiasKillat, line 452: This was causing the lost chunks. Maybe this should be part of the `convert_serdata``function instead but I decided against this in the past to not change anything in this function.

src/core/ddsc/src/dds_write.c
eboasson: That weird function does exist for a good reason: multi-sertype (in particular multi-language) support in a single process. E.g., if you have a single process written partly with a writer using the C++ API and a reader for the same topic using the C API, you need to convert in some manner. This does that by getting the serialised representation from the original `serdata` & deserializing for the reader.

What suprises me is that it'd be involved in the first place in this test because there's only a single language involved. Anyway, I'll study it now.

eboasson (approved):
@MatthiasKillat I totally agree with your observation that there are too many cases too deal with. It makes little sense to now decide that that should be taken care of prior to adding support for variable-length data, especially not given that it all bases seem to be covered after a pretty thorough review process.

#1017 adds the checks in the test to catch the problem with local delivery incorrectly bypass Iceoryx that your latest commit on the PR fixes. I'll merge it after this one.

Issue #897, opened 2021-07-29
IDLC C code generation ignores structure inheritance


eboasson:
That is, with this input:
```IDL
struct A { long x; };
struct B : struct A { long y; };
```
I would expect that `B`'s inheriting `A` would not be silently discarded when generating C code. I haven't been able to find an IDL4-to-C mapping specification on the OMG website — IDL4.0 is a mere 5 years old, I am no doubt being unreasonable in expecting one already — so perhaps it is as simple as a missing error message.

However, with a frontend and C++ backend that support it, it would probably be better to do a provisional mapping to C. I think something as simple as:
```C
struct A { int32_t x; };
struct B { struct A _parent; int32_t y; };
```
would take core it.

eboasson:
Should've thought of it (thanks @e-hndrks!): it is in the XTypes spec:

> 7.5.1.4.1.3 Other Programming Languages
> The language binding shall be generated as if an instance of the base type were the first member of the subtype with the name “parent,” as in the following equivalent IDL definition:
>```
> struct <struct_name> { <base_type_name> parent; // ... other members
> };
> ```

so it is actually defined and quite straightforward.

trittsv:
@eboasson is there any plan in which release this issue will be fixed?

dennis-adlink:
@trittsv I'm currently working on the implementation for structures (union inheritance will be implemented later). This is my work-in-progress version: https://github.com/dennis-adlink/cyclonedds/tree/idlc_inheritance

I think this generates the correct types and serializer instructions, but I've to do some more testing and check the CI results before creating a PR. But you could give this a try and see if it works for you (it is based on Cyclone master).

trittsv:
@dennis-adlink can this issue be closed due to the merge of #1011 ?

dennis-adlink:
@trittsv Yes, it can be closed, thanks. In case you run into any issue (related or unrelated to this one) or if you have further questions, please don't hesitate to open a new issue! 

@eboasson can you please close this one?

eboasson:
> @eboasson can you please close this one?

With pleausre!

PR #1011, opened 2021-11-02
Struct inheritance in idl C generator

dennis-adlink:
This PR adds support for struct inheritance to the C generator of the IDL compiler. As described in section 7.5.1.4.1.3 of the XTypes spec, in the generated types the inheritance is "as if an instance of the base type were the first member of the subtype with the name 'parent'". As a result, in the generated opcodes for the serializer the base type is also included as if it is a member of the subtype.

When using inheritance with mutable structs, the CDR stream serializer needs a member ID for the `parent` member. The hashed ID for the member name `parent` is used in the implementation in this PR, but I'm not sure if that the best approach. And this also needs a check that other members cannot use the name `parent` and can't have the same ID as the inheritance-parent member (only in the C backend, or in the generic parser?)

This PR resolved issue #897. Inheritance for union types is not implemented in this PR.

### DIFF
--- a/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
+++ b/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
@@ -98,6 +98,7 @@ enum dds_stream_opcode {
      e            = external: stored as external data (pointer) (DDS_OP_FLAG_EXT)
      f            = flags:
                     - key/not key (DDS_OP_FLAG_KEY)
+                    - base type member, used with EXT type (DDS_OP_FLAG_BASE)
      [offset]     = field offset from start of element in memory
      [elem-size]  = element size in memory (elem-size is only included in case 'external' flag is set)
      [max-size]   = string bound + 1
@@ -241,7 +242,8 @@ enum dds_stream_typecode_subtype {
 #define DDS_OP_FLAG_FP   (1u << 1) /* floating-point: applicable to {4,8}BY and arrays, sequences of them */
 #define DDS_OP_FLAG_SGN  (1u << 2) /* signed: applicable to {1,2,4,8}BY and arrays, sequences of them */
 #define DDS_OP_FLAG_MU   (1u << 3) /* must-understand flag */
-#define DDS_OP_FLAG_BASE (1u << 4) /* jump to base type, used with PLM in PL-CDR */
+#define DDS_OP_FLAG_BASE (1u << 4) /* jump to base type, used with PLM in mutable types and for
+                                      the TYPE_EXT 'parent' member in final and appendable types */
 
 /* Topic descriptor flag values */
 #define DDS_TOPIC_FLAGS_MASK                    0x3fffffff  /* The 2 most significant bits are used for type extensibility */
--- a/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
+++ b/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
@@ -148,8 +148,10 @@ enum dds_stream_opcode {
        where
          f           = flags:
                        - must-understand (DDS_OP_FLAG_MU)
+                       - jump to base type (DDS_OP_FLAG_BASE)
          [elem-insn] = (unsigned 16 bits) offset to instruction for element, from start of insn
-         [member id] = id for this member
+                        when FLAG_BASE is set, this is the offset of the PLM list of the base type
+         [member id] = id for this member (0 in case FLAG_BASE is set)
   */
   DDS_OP_PLM = 0x06 << 24,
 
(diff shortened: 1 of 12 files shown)


src/core/ddsc/include/dds/ddsc/dds_opcodes.h
```diff
@@ -236,9 +238,10 @@ enum dds_stream_typecode_subtype {
    and (3) the discriminator can be an integral type (or enumerated - here treated as equivalent).
    What it can't be is a floating-point type. So DEF and FP need never be set at the same time.
    There are only a few flag bits, so saving one is not such a bad idea. */
-#define DDS_OP_FLAG_FP  (1u << 1) /* floating-point: applicable to {4,8}BY and arrays, sequences of them */
-#define DDS_OP_FLAG_SGN (1u << 2) /* signed: applicable to {1,2,4,8}BY and arrays, sequences of them */
-#define DDS_OP_FLAG_MU  (1u << 3) /* must-understand flag, used with PLM in parameter list CDR */
+#define DDS_OP_FLAG_FP   (1u << 1) /* floating-point: applicable to {4,8}BY and arrays, sequences of them */
+#define DDS_OP_FLAG_SGN  (1u << 2) /* signed: applicable to {1,2,4,8}BY and arrays, sequences of them */
+#define DDS_OP_FLAG_MU   (1u << 3) /* must-understand flag */
+#define DDS_OP_FLAG_BASE (1u << 4) /* jump to base type, used with PLM in PL-CDR */
```
eboasson, line 244: I am ok with solving this problem by supporting fragmented member tables instead of inlining everything, at the very least because the alternative requires more changes to the IDL compiler. I do wonder why you chose to introduce a new flag on the `PLM` instruction, it seems to me like allowing a member table to include `JSR`s would do just as well. If that really is the case, I would find it more elegant, but I may be overlooking something.

(Changing it appears to be very little work — but again, I may be overlooking something!)

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1553,6 +1459,35 @@ static const uint32_t *dds_stream_read_delimited (dds_istream_t * __restrict is,
...
+      found one, because in many cases the members will be in the data sequentially */
+  while (!found && (insn = ops[ops_csr]) != DDS_OP_RTS)
+  {
+    assert (DDS_OP (insn) == DDS_OP_PLM);
+    uint32_t flags = DDS_PLM_FLAGS (insn);
+    const uint32_t *plm_ops = ops + ops_csr + DDS_OP_ADR_PLM (insn);
+    if (flags & DDS_OP_FLAG_BASE)
+    {
+      assert (DDS_OP (plm_ops[0]) == DDS_OP_PLC);
+      plm_ops++; /* skip PLC to go to first PLM from base type */
```
eboasson, line 1477: Why not take care of this off-by-1 in the generator?

eboasson (commented):
Thanks @dennis-adlink ! Looks good to me, but I do have a question/suggestion. Apart from that, I'll wait for the change to how you handle inheritance of "appendable" structs before merging, though.

src/core/ddsc/include/dds/ddsc/dds_opcodes.h
dennis-adlink: A reason for introducing this flag and not using JSR is that the code that handles the PLM list (e.g. skipping to end of the list in the serializer when reading PL-CDR) remains simple and does not need to change, it can safely do `ops += 2` and assume every 2nd op is a PLM or RTS. The same in IDLC, e.g. the `shift_plm_list_offsets` function. Although this could also be solved by adding a dummy (0) instruction after the JSR of course, and you're right that the changes required for using JSR in the list are straightforward. 

Another use of this flag is skipping the DHEADER for inheritance of appendable structs. When the flag is set on the ADR | EXT opcode, this instructs the serializer to skip the DLC instruction and continue adding the members from the base type as part of the current (derived) type (implemented in 69f4657)

Using the flag for both cases improves the readability of the generated opcodes imho, which can be useful for debugging, and there are not really drawbacks, except that it occupies an extra flag bit.

src/core/ddsi/src/ddsi_cdrstream.c
dennis-adlink: Again readability of the generated opcodes, because all offsets to a type are to the PLC/DLC instruction of the target type (when it's not a final type of course), and I preferred to keep it that way to make debugging easier. The same applies to appendable struct inheritance: the ADR | EXT could also point to the first instruction after the DLC. But on the other hand this will slightly impact runtime performance, so I'm also fine with handling this in the generator (in which case I could include a comment in the generated opcodes).

src/core/ddsc/include/dds/ddsc/dds_opcodes.h
eboasson: That makes sense! Thanks for explaining.

src/core/ddsi/src/ddsi_cdrstream.c
eboasson: I'm not worried about the run-time impact of the additional increment: with all the work needed to decode the operations and make sense of the data/generating the required format, the overhead really is mere noise.

I haven't been doing much debugging of this code, unlike you, so I'm not in a good position to judge the effect on the difficulty debugging it. I'm quite willing to accept your reasoning.

dennis-adlink:
The force push to d911d22 is a rebase on master to fix a few minor conflicts.

eboasson (approved):
Thanks @dennis-adlink !

Issue #977, opened 2021-10-06
IDL parser buffer not extended when parsing multi-line comment


dennis-adlink:
There is an issue in the parsing of multi-line comments in IDL files, which seems to be caused by the parser buffer not being extended while parsing a multi-line comment. 

To reproduce: create an IDL file that contains only a multi-line comment, with a length > 4041 characters (exclusing the /* and */). Compiling this file fails with an error message 'syntax error'.

reicheratwork:
This issue seems to occur whenever a buffer refill is requested while a comment is being read. The 4041 character size you see is caused by the chunk size of the buffer (default = 4096) and the paths of the files being parsed being prepended to the contents of the file in the buffer (which in your case seems to be ~55 chars).

This issue occurs when a comment is being parsed, and scanning the available buffer does not reveal the end of the comment marker `*/`.
The refill operation then moves new data into the buffer.
The part of the buffer being "dropped" to make place however ends at the current position of the cursor, which is inside the comment.
Then when the new attempt is made at parsing block of data is made, it begins reading inside the comment, and finds a terminating comment marker without the start, causing the error in the parser.

Working on a fix now.

PR #1012, opened 2021-11-02
Fix for issue #977

reicheratwork:
Add storing and reloading the parser state when a buffer refill
request is encountered while scanning. This will reset the parser
to the correct position for resizing/refilling the buffer.

### DIFF
--- a/src/idl/src/scanner.c
+++ b/src/idl/src/scanner.c
@@ -636,7 +636,9 @@ scan(idl_pstate_t *pstate, idl_lexeme_t *lex)
       pstate->scanner.state = IDL_SCAN_GRAMMAR;
     }
   } while (code == '\0');
-  move(pstate, lim);
+
+  if (code > 0)
+    move(pstate, lim);
   lex->limit = lim;
   lex->location.last = pstate->scanner.position;
 


k0ekk0ek:
Personally, I'd choose to not `move` the cursor if the `code` is negative (an error). Prepending line 639 with an `if` statement seems a little cleaner.

eboasson:
> Personally, I'd choose to not `move` the cursor if the `code` is negative (an error). Prepending line 639 with an `if` statement seems a little cleaner.

Good point. I only gave it a cursory look at what the code does that line, but it looks like it should work. @reicheratwork, I'd appreciate it if you looked into this suggestion.

reicheratwork:
> > Personally, I'd choose to not `move` the cursor if the `code` is negative (an error). Prepending line 639 with an `if` statement seems a little cleaner.
> 
> Good point. I only gave it a cursory look at what the code does that line, but it looks like it should work. @reicheratwork, I'd appreciate it if you looked into this suggestion.

No problem, just pushed that approach.

eboasson (approved):
It does the trick and I don't think it introduces any new problems after digging a bit into what `lex->limit` really means and how it is used.

I also believe this should be part of the test cases, which I've tried using string-based input. That experiment failed because of various assumptions in the test wrappers. You never get a "refill" code out of the scanner and it does seem to assume that the limit is where the 0-terminator of the string is; changing the flags so it does ask for a refill gets me into trouble with indicating the end of the input (I'm sure this is simply due to my not knowing the IDL compiler code base well enough).

I suspect some of the issues in handling the input buffer limit do carry over into input file handling but figuring all that out is going to take me more time than I want to spend on this PR. So I consider this fix "good enough for now".

PR #1015, opened 2021-11-04
Improve support for @external 

dennis-adlink:
This PR extends support for the @external annotation in the idl C generator and the CDR stream serializer. With these changes, the external annotation can be used on struct and union members. Applying the external annotation on collection element type is not supported yet.

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -354,6 +354,95 @@ static uint32_t get_type_size (enum dds_stream_typecode type)
   return (uint32_t)1 << ((uint32_t) type - 1);
 }
 
+static uint32_t get_arr_elem_size (uint32_t insn, const uint32_t * __restrict ops)
+{
+  uint32_t elem_sz;
+  switch (DDS_OP_SUBTYPE (insn))
+  {
+    case DDS_OP_VAL_1BY: case DDS_OP_VAL_2BY: case DDS_OP_VAL_4BY: case DDS_OP_VAL_8BY: case DDS_OP_VAL_ENU:
+      elem_sz = get_type_size (DDS_OP_SUBTYPE (insn));
+      break;
+    case DDS_OP_VAL_STR: case DDS_OP_VAL_BSP:
+      elem_sz = sizeof (char *);
+      break;
+    case DDS_OP_VAL_BST: case DDS_OP_VAL_SEQ: case DDS_OP_VAL_ARR: case DDS_OP_VAL_UNI: case DDS_OP_VAL_STU:
+      elem_sz = ops[4];
+      break;
+    case DDS_OP_VAL_EXT:
+      abort ();
+      break;
+  }
+  return elem_sz;
+}
+
+static uint32_t get_adr_type_size (uint32_t insn, const uint32_t * __restrict ops)
+{
+  uint32_t sz = 0;
+  switch (DDS_OP_TYPE (insn))
+  {
+    case DDS_OP_VAL_1BY: case DDS_OP_VAL_2BY: case DDS_OP_VAL_4BY: case DDS_OP_VAL_8BY: case DDS_OP_VAL_ENU:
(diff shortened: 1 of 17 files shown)


src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1284,6 +1374,15 @@ static const uint32_t *dds_stream_read_uni (dds_istream_t * __restrict is, char
 static inline const uint32_t *dds_stream_read_adr (uint32_t insn, dds_istream_t * __restrict is, char * __restrict data, const uint32_t * __restrict ops)
 {
   void *addr = data + ops[1];
+  if (op_type_external (insn) && DDS_OP_TYPE (insn) != DDS_OP_VAL_STR)
+  {
+    /* Allocate memory for @external struct member. This memory must be initialized
+        to 0, because the type may contain sequences that need to have 0 index/size */
+    uint32_t sz = get_adr_type_size (insn, ops);
+    *((char **) addr) = ddsrt_calloc (1, sz);
```
eboasson, line 1383: Doesn't this lead to a memory leak if one has a loop:
```C
T *x = 0;
dds_sample_info_t si;
while((n=dds_take(rd, &x, si, 1, 1))>0) { ... }
...
if (x) dds_return_loan(rd,&x,1);
```

I'm commenting here because the union case is already special-cased to deal with changing discriminants. (See `dds_stream_read_sample`.)

dennis-adlink: Thanks for finding this! I've fixed it in 198041c. The size of the allocated buffer for external members is a constant value independent of the data, so I think it can be safely re-used over multiple reads/takes (similar to string and sequence buffers).

src/tools/idlc/src/descriptor.c
```diff
@@ -177,6 +177,11 @@ stash_opcode(
   struct instruction inst = { OPCODE, { .opcode = { .code = code, .order = order } } };
   const struct alignment *alignment = NULL;
 
+  if (code | DDS_OP_FLAG_EXT) {
```
eboasson, line 180: `code | DDS_OP_FLAG_EXT` is always true. From the looks of it, the intent was `code & DDS_OP_FLAG_EXT`.

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -354,6 +354,96 @@ static uint32_t get_type_size (enum dds_stream_typecode type)
...
+      elem_sz = sizeof (char *);
+      break;
+    case DDS_OP_VAL_BST: case DDS_OP_VAL_SEQ: case DDS_OP_VAL_ARR: case DDS_OP_VAL_UNI: case DDS_OP_VAL_STU:
+      if (DDS_OP_TYPE (insn) == DDS_OP_VAL_ARR)
+        elem_sz = ops[4];
+      else
+        abort ();
+      break;
+    case DDS_OP_VAL_EXT:
+    default:
```
eboasson, line 375: I'd prefer to get rid of the `default` here (and in similar cases): then the compiler will warn if some cases are skipped.

eboasson (commented):
LM(ostly)GTM, I do have 2 small remarks and would appreciate it if you'd take a look at them.

I don't have all the details of the IDL compiler down cold, so I may have overlooked something, in particular, I am going on faith that the rules for determining whether some part of a data type is external are correct.

src/tools/idlc/src/descriptor.c
dennis-adlink: Yes, that was definitely wrong! I've fixed this and another issue in the rules to set NO_OPTIMIZE in commit 9c53626, and I've also added a test-case for this flag.

src/core/ddsi/src/ddsi_cdrstream.c
dennis-adlink: Good point, in this case there shouldn't be a `default`, fixed in 9c53626. In other cases, e.g. where only the `1BY` ... `ENU` values are included, it's better to leave out the other (7) values I think?

PR #1018, opened 2021-11-05
Fixes for coverity issues

reicheratwork:
Fixes for the following reported coverity issues:
- unchecked return values
- conditional check which will never be done
- data tainting through byte swapping

### DIFF
--- a/src/tools/idlc/src/descriptor.c
+++ b/src/tools/idlc/src/descriptor.c
@@ -858,11 +858,13 @@ emit_union(
 
     switch (((idl_union_t *)node)->extensibility.value) {
       case IDL_APPENDABLE:
-        stash_opcode(pstate, descriptor, &ctype->instructions, nop, DDS_OP_DLC, 0u);
+        if ((ret = stash_opcode(pstate, descriptor, &ctype->instructions, nop, DDS_OP_DLC, 0u)))
+          return ret;
         break;
       case IDL_MUTABLE:
         idl_error(pstate, idl_location(node), "Mutable unions are not supported yet");
-        // stash_opcode(pstate, descriptor, &ctype->instructions, nop, DDS_OP_PLC, 0u);
+        //if ((ret = stash_opcode(pstate, descriptor, &ctype->instructions, nop, DDS_OP_PLC, 0u)))
+        //  return ret;
         return IDL_RETCODE_UNSUPPORTED;
       case IDL_FINAL:
         break;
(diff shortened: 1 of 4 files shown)


src/tools/idlc/src/descriptor.c
```diff
@@ -1268,8 +1272,8 @@ emit_declarator(
       return ret;
 
     /* in case there is no revisit required (array has simple element type) we have
-       to close the mutable member immediately, otherwise close it when revisiting */
-    if (mutable_aggr_type_member && (!(ret & IDL_VISIT_REVISIT) || revisit)) {
+       to close the mutable member immediately */
+    if (mutable_aggr_type_member) {
```
dennis-adlink, line 1276: Although this change is not incorrect, I think a better fix would be to modify the check on line 1271 so that it only returns in case of an error:
```C
if ((ret = emit_array(pstate, revisit, path, node, user_data)) < 0) 
  return ret
```

Then the condition on line 1276 for closing the mutable member remains unchanged: the member is immediately closed in case the emit function does not set the `IDL_VISIT_REVISIT` bit, or it is closed in case of revisiting.

dennis-adlink (commented):
Looks good, only a minor remark regarding closing the mutable member for array types in `emit_declarator`, see the comment below.

src/tools/idlc/src/descriptor.c
dennis-adlink: Fixing the `if` on line 1271 as I suggested above reveals a bug on line 1280, this line should be: `return ret;`

src/tools/idlc/src/descriptor.c
```diff
@@ -1264,21 +1268,21 @@ emit_declarator(
...
-    if ((ret = emit_array(pstate, revisit, path, node, user_data)))
+    if ((ret = emit_array(pstate, revisit, path, node, user_data)) < 0)
       return ret;
 
     /* in case there is no revisit required (array has simple element type) we have
        to close the mutable member immediately, otherwise close it when revisiting */
     if (mutable_aggr_type_member && (!(ret & IDL_VISIT_REVISIT) || revisit)) {
-      if ((ret = close_mutable_member(pstate, descriptor, ctype)))
-        return ret;
+      ret = close_mutable_member(pstate, descriptor, ctype);
```
dennis-adlink, line 1277: It would be better to use a local `ret` variable here and not reuse `ret` that has the result of `emit_array`, so that the return value of `emit_array` (which can have instructions for the visit function) is returned from this function.

dennis-adlink (approved):

eboasson (approved):

PR #1024, opened 2021-11-09
Correct encoding in serialization to shared memory

reicheratwork:
This improvement was exposed by coverity issue 342212 and should fix
it as well
Add initialization of all dds_ostream_t members
Correct m_xcdr_version member initialization: this was not initialized
previously
Also the type information of the topic was passed, the writer type
information is used now, as this can be independant from the topic


eboasson (approved):
LGTM, but I only reviewed @dennis-adlink's work that introduces these changes, so I'd be much obliged if he could do a quick review as well.

dennis-adlink (approved):
LGTM! With these changes the XCDR version used to fill the iox chunk matches the version that is used to create the serdata wrapper in `ddsi_serdata_from_loaned_sample`, so the correct version (serdata->hdr.identifier) will be used for deserialization.

eboasson:
Thanks @reicheratwork and @dennis-adlink!

PR #1028, opened 2021-11-12
Annotation parsing for must_understand annotations

reicheratwork:
Add parsing for must_understand annotations
Improved key and optional annotations checks for conflicts with other annotations (like must_understand)


src/idl/tests/annotation.c
```diff
@@ -242,44 +244,98 @@ CU_Test(idl_annotation, default_literal)
...
+
+CU_Test(idl_annotation, key)
+{
+  key_test_t tests[] = {
+    {"struct s {\n"
+     "  @key char a;\n"
+     "  @key(TRUE) char b;\n"
+     "  @key(FALSE) char c;\n"
+     "  @must_understand @key char d;\n"
+     "  @optional(FALSE) @key char e;\n"
```
dennis-adlink, line 309: To avoid overlap in the `@optional`, `@key` and `@must-understand` tests with respect to annotation order, you could consider using the primary annotation for the test (`@key` in this case) as the first annotation and include all combinations of the other to as 2nd and 3rd. This ensures that all paths in the `annotate_...` functions are tested.

dennis-adlink (commented):
Looks good, a few comments below regarding annotation order.

src/idl/src/annotation.c
```diff
@@ -174,7 +174,7 @@ annotate_optional(
     idl_error(pstate, idl_location(annotation_appl),
       "@optional can only be assigned to members");
     return IDL_RETCODE_SEMANTIC_ERROR;
-  } else if (mem->key.value) {
+  } else if (value && mem->key.value) {
```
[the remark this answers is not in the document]
reicheratwork: You are correct.

It would be even better if the value (true/false) of the annotation would be read before this statement, since if it is false (which would mean the same as no `@optional` annotation), the @optional annotation will have no possible conflict with an `@key`, if any.

src/idl/tests/annotation.c
reicheratwork: I went with the following approach:
The first annotation to be parsed, and therefore to be set is the one that occurs first in the text.
So to check for the code paths checking for conflicting annotations in the `annotate_key` function, which in my eyes is the purpose of the test, I first set the conflicting annotation(s) `@optional` and `@must_understand` and then `@key` second.

dennis-adlink (approved):

eboasson (approved):
LGTM, too. Thanks @reicheratwork

PR #1031, opened 2021-11-15
Extend shared memory loan support

### LINKED ISSUE #1030: Extend shared memory loan support
1) Update the API to check whether shared memory and loaning is available (at readers an writers).
Note that this is not necessarily the same as we may have shared memory enabled now but only allow loaning for fixed size data types (otherwise we would need to ensure that the caller serialized the data into shared memory).

2) Consolidate the API to loan arbitrary on writer

3) Update DDS data allocator for the shared memory case?

MatthiasKillat:
Closes #1030


MatthiasKillat:
@eboasson @sumanth-nirmal Could you please have a look? I added a new public header for loan related calls dds_loan.h (to be discussed) to add the needed functionality. However, it is not easily possible (nor desirable) to move e.. writer related loan calls there due to existing dependencies (requires more refactoring for not much gain).

Personally I think we need to rethink the loaning and allocation API further, but this should not be done without further discussion or in this pull request. We need additional public API but it should not be in added to dds.h to separate it properly as some kind of extension/experimental API.

src/core/ddsc/include/dds/ddsc/dds_loan_api.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @returns loan available or not
+ */
+DDS_EXPORT bool dds_is_loan_available(const dds_entity_t entity);
+
+/** @brief Check if a shared memory is available to reader/writer.
+ *
+ * @note shared memory available and loan available have different meaning
+ * 1)We may not be able to use the loan API due to type constraints (not fixed)
+ * 2) We may also allow using the loan API but without shared memory
```
sumanth-nirmal, line 32: The note is confusing to me, this should say something in the lines that `dds_loan_sample` API is allowed to be used only when `dds_is_loan_available`, correct? or am I missing something here?

sumanth-nirmal (commented):
The changes look good to me except for the minor comments I have and do what they are supposed to do!

(These changed/updated APIs are used here https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/176)

src/core/ddsc/src/dds_loan.c
```diff
@@ -0,0 +1,242 @@
...
+  if (chunk) {
+    register_pub_loan(wr, chunk);
+    // NB: we set this since the user can use this chunk not only with write
+    // where we check whether it was loaned before.
+    // It is only possible to loan for fixed size types as of now.
+
+    // Unfortunate, since the chunk is not actually filled at this point.
+    // We should ensure that we cannot circumvent the write API
+    // (API redesign).
+    shm_set_data_state(chunk, IOX_CHUNK_CONTAINS_RAW_DATA);
```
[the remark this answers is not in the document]
eboasson: I think there is a good case for setting to `RAW_DATA` here: the interface is meant to be used for transferring the samples in the raw "in-memory" representation of the application, so it might as well be set accordingly. Alternatively, it could be set to `UNINITIALIZED` here and then overwritten with `RAW_DATA` in `dds_write`, but then one wonders: what's the point?

It all depends on how one interprets the "data state": does it encode only the representation, or does it also encode the validity of the contents? Both are fine; I have interpreted it as the former, and then setting it to `RAW_DATA` here is a sensible choice.

src/core/ddsi/include/dds/ddsi/shm_transport.h
```diff
@@ -12,16 +12,17 @@
 #ifndef DDS_SHM__TRANSPORT_H
```
eboasson, line 12: I'd prefer renaming it to `ddsi_shm_transport.h`. (Just like I fully intend to one day sit down and rename all the other files in that directory that don't start with `ddsi_`).

src/core/ddsi/include/dds/ddsi/shm_transport.h
```diff
@@ -12,16 +12,17 @@
...
-#include "dds/export.h"
-#include "iceoryx_binding_c/subscriber.h"
-#include "dds/ddsrt/sync.h"
-#include "dds/ddsi/q_protocol.h" /* for, e.g., SubmessageKind_t */
+#include "dds/ddsi/ddsi_config.h"
+#include "dds/ddsi/ddsi_keyhash.h"
 #include "dds/ddsi/ddsi_tran.h"
+#include "dds/ddsi/q_protocol.h" /* for, e.g., SubmessageKind_t */
+#include "dds/ddsrt/sync.h"
+#include "dds/export.h"
```
eboasson, line 20: I generally couldn't care less about the order of includes (which also means I don't get the point of reordering them here, or indeed in some of the other places). I know that there are some dependencies though, even when there shouldn't be. The one in particular that's scary is `dds/features.h`: I am not willing to guarantee it is really included everywhere that it matters (though I did try), so moving files ahead of that one in the order of includes requires some double-checking.

src/core/ddsi/include/dds/ddsi/shm_transport.h
```diff
@@ -71,6 +72,12 @@ DDS_EXPORT iceoryx_header_t *iceoryx_header_from_chunk(const void *iox_chunk);
 
 DDS_EXPORT void shm_set_loglevel(enum ddsi_shm_loglevel);
 
+DDS_EXPORT void *shm_create_chunk(iox_pub_t iox_pub, size_t size);
```
eboasson, line 75: I've tried to make sure all files that one can "reasonably expect" to be used by applications needing some special features consistently prefix all names with `dds_` or `ddsi_` (or uppercase; all depending on the circumstances; just compare `q_config.h` @ aba5e78760b9415d21beb8319b0af0c2b730ce9a with `ddsi_config.h` now). I know this `iox_`-, 'iceoryx_`- and `shm_`-prefixed stuff has been around for a while, but some of this slowly morphed from internal-only (where I don't care) to somewhat-public-API. Continuing down this path seems unwise, and indeed, I think the pre-existing symbols ought to be cleaned up.

I should've paid more attentive to that ...

Rename all that in the same PR as adding this new functionality doesn't seem like a great idea. That leaves us with: first fix all the existing names, then do this PR with updated names; or do this PR fitting the pattern, then do a PR updating all the names. Practical considerations make me willing to accept the latter.

src/core/ddsc/src/dds_loan.c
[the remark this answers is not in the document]
MatthiasKillat: The problem is that we get the chunk with `dds_loan_sample` and the user is supposed to fill it with a sample of a fixed size type. In `dds_write` we rely on the state to be correctly set. If it would ust be uninitialized here, the user would have to set the state and this is inconvenient and even more error-prone.

The only way to avoid this properly is having a function to fill the data to be called by the user. In C++ this would be a (variadic) emplace consctructor which also sets the state after the sample was intialized. I do not see this to be viable here, so the current state is the best alternative I came up with ...

src/core/ddsc/include/dds/ddsc/dds_loan_api.h
MatthiasKillat: Clarified notes.

src/core/ddsi/include/dds/ddsi/shm_transport.h
MatthiasKillat: Ok, at some point I thought there should be a separate module shm or similar, which builds upon ddsi and can be used to extend ddsi with shared memory support. This requires quite some refactoring (also getting rid ofmost of  the #ifdefs...) so is not  going to happen here.

src/core/ddsi/include/dds/ddsi/shm_transport.h
MatthiasKillat: Yes, it should not be the case that the order matters unless the headers rely on definitions to be included before the header itself is included which becomes unmaintainable fast.

The reorder is due to clang format, I must be careful with my clang format which reorders headers in lexicographic order if they are in the same block without empty lines. (ideally there should be a project specific format file).

src/core/ddsi/include/dds/ddsi/shm_transport.h
MatthiasKillat: I prefer the latter option and can prepare a renaming PR after this one, but as you say I would prefer not to mix it up with functionality changes.

MatthiasKillat:
@eboasson Adressed most of the issues, but not e.g. renaming of `shm_*` functions. This can be a follow-up.

src/core/ddsi/include/dds/ddsi/shm_transport.h
eboasson: That explains! ☺️ 

I've never been able to convince clang-format (nor uncrustify, for that matter) to do it the way I like, but if there's a way, it would definitely be good to at least have a project-specific file in the repo.

src/core/ddsc/src/dds__data_allocator.h
```diff
@@ -37,6 +38,7 @@ typedef struct dds_iox_allocator {
     iox_pub_t pub;
     iox_sub_t sub;
   } ref;
+  ddsrt_mutex_t mutex;
```
eboasson, line 41: I ran into one issue trying to build it locally on macOS with Iceoryx: the mutex makes it too large on macOS making the static assertion fail. The patch below takes care of that:
```diff
diff --git a/src/core/ddsc/include/dds/ddsc/dds_data_allocator.h b/src/core/ddsc/include/dds/ddsc/dds_data_allocator.h
index b8bfc535..123c3b8a 100644
--- a/src/core/ddsc/include/dds/ddsc/dds_data_allocator.h
+++ b/src/core/ddsc/include/dds/ddsc/dds_data_allocator.h
@@ -23,7 +23,9 @@
 extern "C" {
 #endif
 
-#define DDS_DATA_ALLOCATOR_MAX_SIZE (8 * sizeof (void *))
+// macOS' mutexes require quite a lot of space, but it is not quite enough
+// to make this system-dependent
+#define DDS_DATA_ALLOCATOR_MAX_SIZE (12 * sizeof (void *))
 
 typedef struct dds_data_allocator {
   dds_entity_t entity;
```

src/core/ddsc/include/dds/ddsc/dds_loan_api.h
```diff
@@ -0,0 +1,86 @@
...
+ * @param[out] sample the loaned sample
+ *
+ * @returns DDS_RETCODE_OK if successful, DDS_RETCODE_ERROR otherwise
+ */
+DDS_EXPORT dds_return_t dds_loan_sample(dds_entity_t writer, void **sample);
+
+#if defined(__cplusplus)
+}
+#endif
+#endif
```
eboasson, line 86: Clang warns about a missing newline for me ... (and given that there is still a minor issue, it probably makes sense to add the newline)

src/core/ddsc/include/dds/ddsc/dds_basic_types.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+typedef int32_t dds_entity_t;
+
+#define DDS_MIN_PSEUDO_HANDLE ((dds_entity_t)0x7fff0000)
+
+/** Special handle representing the entity corresponding to the CycloneDDS
+ * library itself */
+#define DDS_CYCLONEDDS_HANDLE ((dds_entity_t)(DDS_MIN_PSEUDO_HANDLE + 256))
+
+#endif /*_DDS_PUBLIC_TYPES_H_*/
```
eboasson, line 33: Clang warns about a missing newline for me ... (and given that there is still a minor issue, it probably makes sense to add the newline)

eboasson (commented):
LGTM, but building on macOS reveals that Apple has a different view: it warns about two missing newlines, and worse a static assertion fails because of the size of a mutex on macOS. I've marked them; I think the change in structure size works out for x86_64 macs as well, but I can't be bothered to go through the effort of actually verifying that.

If it weren't for static assertion, I'd fine with it as it is.

src/core/ddsc/src/dds_loan.c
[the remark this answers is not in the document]
sumanth-nirmal: Since this function is static and only used in `dds_loan_sample()`, this should be fine

src/core/ddsc/src/dds__data_allocator.h
MatthiasKillat: Done. It is kind of error-prone to do this if we want to implement this "base class" allocator interface. Should maybe have use the "first member of a derived is the base struct" pseudo-inheritance if we intend to keep the allocator anyway.

eboasson:
Thanks @MatthiasKillat !

PR #1032, opened 2021-11-15
Increase timeouts for some security tests

eboasson:
Based on a handful of CI runs, these tweaks appear to reduce the incidence of timeouts in the tests. The timeouts were originally chosen based on what worked in practice and bumping them a bit doesn't seem like it should introduce any new problems.

I have looked at changing it so each data point runs as a separate test with a separate timeout (that would then probably be the default 10s) but that's quite a project. That makes a few tweaks here and there worth a try.


eboasson:
Clearly that didn't work out ...

PR #1033, opened 2021-11-16
Fix new proxy topic race

eboasson:
The primary subject of this PR is that `new_proxy_topic` got a pointer to a topic definition by calling "lookup" (not the "_locked" version), which could result in allocating a new topic definition and adding it to the topic_defs hash table while initializing its refcount to 0.  The refcount would then later on be incremented (and outside the correct lock).

If a local topic with an identical definition were to be created and deleted in between creating the topic definition with refcount 0 in new_proxy_topic and the increment of the refcount, that local topic would reuse the definition.  Deleting the local topic would cause the refcount to drop to 0, and so the definition would be removed it from the hash table and freed.

The result is a dangling pointer in the proxy topic.

This requires both a local and a proxy topic with the same definition, because the local ones do the refcounting and hash table operations atomically, and all discovery work is strictly serialized in the dq.builtins thread.  So either by itself will not cause any trouble.

The secondary subject are some small housekeeping things that got triggered by the primary subject and that do not seem worth the effort of doing another PR.

### DIFF
--- a/src/core/ddsc/src/dds_handles.c
+++ b/src/core/ddsc/src/dds_handles.c
@@ -205,9 +205,7 @@ int32_t dds_handle_delete (struct dds_handle_link *link)
   assert ((cf & HDL_PINCOUNT_MASK) == 1u);
 #endif
   ddsrt_mutex_lock (&handles.lock);
-  int x = ddsrt_hh_remove (handles.ht, link);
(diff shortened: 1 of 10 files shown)


dennis-adlink (approved):
LGTM!

PR #1037, opened 2021-11-17
Fix for using uninitialized bool in topickind_qos_match_p_lock

dennis-adlink:
When matching endpoints, in the case the topic name or type name does not match and the remote endpoint does not include a type identifier, the bool variables that indicate if a type lookup request should be done are used uninitialized in `topickind_qos_match_p_lock`. This resulted in requesting type details for an empty type id.

This commit fixes the issue by initializing these boolean values to false in `qos_match_mask_p` before other checks are done.

This fixes the crash mentioned in https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/41


eboasson (approved):
Thanks you @dennis-adlink !

PR #1038, opened 2021-11-17
Don't try to use OpenSSL 1.0.2 on Windows x86 build

eboasson:
It looks like the use of OpenSSL 1.0.2 with the cross-build targetting
an older Visual Studio causes problems. No-one should be using OpenSSL
1.0.2 anymore, so if changing the build to OpenSSL 1.1 indeed makes the
problems go away, good.


reicheratwork:
Great, hopefully this fixes the issue.
I will be checking it on my side now!

PR #1039, opened 2021-11-18
Optional annotation support in IDLC and cdrstream serializer

dennis-adlink:
This PR adds support for the `@optional` annotation in IDL. This annotation can be used on struct members. An optional member is generated as a pointer type in the IDLC output, similar to external members. In plain and delimited CDR the field is preceded by a bool that indicates if the field is present (in pl-cdr the field is omitted if not present).

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -60,12 +60,13 @@
 #define dds_os_reserve4BO                             NAME_BYTE_ORDER(dds_os_reserve4)
 #define dds_os_reserve8BO                             NAME_BYTE_ORDER(dds_os_reserve8)
 #define dds_stream_write_stringBO                     NAME_BYTE_ORDER(dds_stream_write_string)
-#define dds_stream_writeBO                            NAME_BYTE_ORDER(dds_stream_write)
 #define dds_stream_write_seqBO                        NAME_BYTE_ORDER(dds_stream_write_seq)
 #define dds_stream_write_arrBO                        NAME_BYTE_ORDER(dds_stream_write_arr)
 #define write_union_discriminantBO                    NAME_BYTE_ORDER(write_union_discriminant)
 #define dds_stream_write_uniBO                        NAME_BYTE_ORDER(dds_stream_write_uni)
 #define dds_stream_writeBO                            NAME_BYTE_ORDER(dds_stream_write)
+#define dds_stream_write_implBO                       NAME_BYTE_ORDER(dds_stream_write_impl)
+#define dds_stream_write_adrBO                        NAME_BYTE_ORDER(dds_stream_write_adr)
 #define dds_stream_write_plBO                         NAME_BYTE_ORDER(dds_stream_write_pl)
 #define dds_stream_write_pl_memberlistBO              NAME_BYTE_ORDER(dds_stream_write_pl_memberlist)
 #define dds_stream_write_pl_memberBO                  NAME_BYTE_ORDER(dds_stream_write_pl_member)
@@ -89,11 +90,14 @@ struct key_off_info {
   const uint32_t *op_off;
 };
 
+static const uint32_t *dds_stream_skip_adr (uint32_t insn, const uint32_t * __restrict ops);
 static const uint32_t *dds_stream_skip_default (char * __restrict data, const uint32_t * __restrict ops);
 static const uint32_t *dds_stream_extract_key_from_data1 (dds_istream_t * __restrict is, dds_ostream_t * __restrict os, const uint32_t * __restrict ops,
   uint32_t n_keys, uint32_t * __restrict keys_remaining, const ddsi_sertype_default_desc_key_t * __restrict key, struct key_off_info * __restrict key_offs);
 static const uint32_t *dds_stream_extract_keyBE_from_data1 (dds_istream_t * __restrict is, dds_ostreamBE_t * __restrict os, const uint32_t * __restrict ops,
   uint32_t n_keys, uint32_t * __restrict keys_remaining, const ddsi_sertype_default_desc_key_t * __restrict key, struct key_off_info * __restrict key_offs);
+static const uint32_t *dds_stream_normalize1 (char * __restrict data, uint32_t * __restrict off, uint32_t size, bool bswap, uint32_t xcdr_version, const uint32_t * __restrict ops, bool is_mutable_member);
+static const uint32_t *dds_stream_read_impl (dds_istream_t * __restrict is, char * __restrict data, const uint32_t * __restrict ops, bool is_mutable_member);
 
 static void dds_ostream_grow (dds_ostream_t * __restrict st, uint32_t size)
 {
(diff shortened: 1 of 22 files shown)


dennis-adlink:
In commit 958d21d I've added a slightly changed version of the XCDR version checking in cdrstream (the force push was because I forgot to add the IDL file for the test). When of using `@optional` in a type, the minimal version will be XCDR2, because optional for XCDR1 is not supported by this serializer.

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -3181,6 +3258,17 @@ static bool dds_stream_print_sample1 (char * __restrict *buf, size_t * __restric
       case DDS_OP_ADR: {
         if (!delimited_sz || (is->m_index - delimited_offs < delimited_sz))
         {
+          if (op_type_optional (insn))
+          {
+            bool present = is_mutable_member ? true : dds_is_get1 (is);
```
eboasson, line 3263: I'd play around with creating the equivalent of `is_member_present` for the CDR stream — i.e., `!op_type_optional(insn) || is_mutable_member || dds_is_get1 (is)` — so that the logic here simplifies to `if (!is_member_present_cdr (is, insn)) { skip }`

Not sure if it really is beneficial, hence the "play around with" part.

src/tools/idlc/src/descriptor.c
```diff
@@ -1287,8 +1289,12 @@ emit_array(
        we're processing a struct type, and this can be used to determine if its
        an external member */
     idl_node_t *parent = idl_parent(node);
-    if (idl_is_struct(stype->node) && idl_is_external(parent))
-      opcode |= DDS_OP_FLAG_EXT;
+    if (idl_is_struct(stype->node)) {
+      if (idl_is_external(parent))
+        opcode |= DDS_OP_FLAG_EXT;
+      if (idl_is_optional(parent))
+        opcode |= DDS_OP_FLAG_OPT;
```
eboasson, line 1296: See some earlier comment, I am thinking life would be simpler if this were `opcode |= DDS_OP_FLAG_OPT | (is_string ? 0 : DDS_OP_FLAG_EXT)` if the type is not a string. (It should also be documented in the header file that things work this way.)

src/tools/idlc/src/descriptor.c
```diff
@@ -1287,8 +1289,12 @@ emit_array(
        we're processing a struct type, and this can be used to determine if its
        an external member */
     idl_node_t *parent = idl_parent(node);
-    if (idl_is_struct(stype->node) && idl_is_external(parent))
-      opcode |= DDS_OP_FLAG_EXT;
+    if (idl_is_struct(stype->node)) {
+      if (idl_is_external(parent))
```
eboasson, line 1293: I'd not mark a string as external. Maybe I am overlooking something, but it only seems like it complicates things.

src/tools/idlc/src/descriptor.c
```diff
@@ -1451,11 +1455,13 @@ emit_declarator(
...
     }
-    if (idl_is_struct(stype->node) && idl_is_external(parent))
-    {
-      opcode |= DDS_OP_FLAG_EXT;
-      /* For @external fields of type OP_TYPE_EXT include the size of the field to allow the serializer to allocate
-         memory for this field when deserializing. */
+    if (idl_is_struct(stype->node) && (idl_is_external(parent) || idl_is_optional(parent))) {
+      if (idl_is_external(parent))
+        opcode |= DDS_OP_FLAG_EXT;
+      if (idl_is_optional(parent))
```
eboasson, line 1461: See emit_array

eboasson (commented):
Thanks @dennis-adlink, I have a few comments, but nothing that would require significant effort to do anything about:
* I think there's one real bug in there (possibly leaking memory when overwriting a pointer with a null pointer when deserialising a sample where an optional field is absent);
* The vetting (`normalize`) of the bools for optional fields in non-PL XCDR should reject values other than 0 and 1;
* Looking at the code I think it would be better to play around a little bit differently with the "external" and "optional" flags so that the code simplifies a bit. (Basically, not set "external" for strings, and always set "external" for optional fields other than strings.)

Apart from that, LGTM!

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1348,10 +1388,25 @@ static const uint32_t *dds_stream_read_uni (dds_istream_t * __restrict is, char
   return ops;
 }
 
-static inline const uint32_t *dds_stream_read_adr (uint32_t insn, dds_istream_t * __restrict is, char * __restrict data, const uint32_t * __restrict ops)
+static inline const uint32_t *dds_stream_read_adr (uint32_t insn, dds_istream_t * __restrict is, char * __restrict data, const uint32_t * __restrict ops, bool is_mutable_member)
 {
   void *addr = data + ops[1];
-  if (op_type_external (insn) && DDS_OP_TYPE (insn) != DDS_OP_VAL_STR)
+  if (op_type_optional (insn))
```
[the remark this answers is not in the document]
dennis-adlink: You're right! Not adding the external flag for strings (unbounded string only) in combination with adding the flag for optional non-string members simplifies the serializer code a bit, I've changed this in d577bb2

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1348,10 +1388,25 @@ static const uint32_t *dds_stream_read_uni (dds_istream_t * __restrict is, char
...
+  {
+    /* mutable optional members are only included in PL-CDR2 if the
+       member is present, so present defaults to true */
+    bool present = is_mutable_member ? true : dds_is_get1 (is);
+    if (!present)
+    {
+      if (DDS_OP_TYPE (insn) != DDS_OP_VAL_STR)
+        *((char **) addr) = NULL;
```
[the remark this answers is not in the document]
dennis-adlink: Good point, the tests didn't cover that, so in addition to fixing the bug I've also added a test case. That revealed another (related) bug: when re-using a buffer for reading multiple samples of a type containing an optional string, the pointer was not set to `NULL` in case the member was not present in a sample after reading a sample in which the member is present. This is also fixed in d577bb2

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1696,6 +1751,15 @@ static bool normalize_uint32 (char * __restrict data, uint32_t * __restrict off,
   return true;
 }
 
+static bool read_and_normalize_uint8 (uint8_t * __restrict val, char * __restrict data, uint32_t * __restrict off, uint32_t size)
```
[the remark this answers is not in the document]
dennis-adlink: There was a check for a valid bool value in `stream_normalize_adr` in my previous commit, but I've moved that check and renamed the function to `read_and_normalize_bool` as suggested (as I don't think that the more generic `read_and_normalize_uint8` will be re-used).

eboasson (approved):
LJGMT - thanks @dennis-adlink!

PR #1040, opened 2021-11-19
Coverity fix

reicheratwork:
Fixes for coverity issues:
340813: untrusted loop bound
340817: infinite loop
342506: unchecked return value

### DIFF
--- a/src/core/ddsi/tests/sysdeps.c
+++ b/src/core/ddsi/tests/sysdeps.c
@@ -199,8 +199,12 @@ CU_Test (ddsi_sysdeps, log_stacktrace_other, .init = setup, .fini = teardown)
   CU_ASSERT_FATAL (rc == 0);
 
   ddsrt_mutex_lock (&arg.lock);
-  while (!arg.incallback)
+  /* coverity[loop_top: FALSE] */
+  /* coverity[exit_condition: FALSE] */
+  while (!arg.incallback) {
+    /* coverity[loop_bottom: FALSE] */
     ddsrt_cond_wait (&arg.cond, &arg.lock);
+  }
   ddsrt_mutex_unlock (&arg.lock);
 
   log_stacktrace (&logconfig, "other", tid);
--- a/src/ddsrt/include/dds/ddsrt/bswap.h
+++ b/src/ddsrt/include/dds/ddsrt/bswap.h
@@ -30,25 +30,21 @@ enum ddsrt_byte_order_selector {
 
 DDS_INLINE_EXPORT inline uint16_t ddsrt_bswap2u (uint16_t x)
 {
-  /* coverity[byte_swapping]*/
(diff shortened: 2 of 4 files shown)


src/ddsrt/include/dds/ddsrt/bswap.h
```diff
@@ -28,41 +28,41 @@ enum ddsrt_byte_order_selector {
   DDSRT_BOSEL_LE,
 };
 
+// coverity[ -tainted_data_return ]
```
eboasson, line 31: The root of the problem seems to me that Coverity considers byte-swapping a reason to treat the result as tainted, even though it is tainted if and only if the input was tainted. Often the heuristic works out, but sometimes we have to deal with byte-swapped trusted (= locally generated) data, and this annotations seems to throw out the valuable warnings as well.

I am not aware of a way to solve that problem short of having two sets of functions and annotating only one of them, but: (1) duplicating code like that is almost insane, and (2) it would do nothing to prevent accidentally using the one that strips the "taintedness" from the result (because of C's weak type system ...)

The only alternative that I know is explicitly marking the cases where it warns as "false positives" in the GUI. While annoying, I think that could be the better option. What do you think?

src/core/ddsi/tests/sysdeps.c
```diff
@@ -199,7 +199,10 @@ CU_Test (ddsi_sysdeps, log_stacktrace_other, .init = setup, .fini = teardown)
   CU_ASSERT_FATAL (rc == 0);
 
   ddsrt_mutex_lock (&arg.lock);
+  /* coverity[loop_top: FALSE] */
+  /* coverity[exit_condition: FALSE] */
   while (!arg.incallback)
+    /* coverity[loop_bottom: FALSE] */
```
eboasson, line 205: I would prefer adding { } here. I love my dangling statements, but not when they aren't immediately below the conditional.

eboasson (commented):
Thanks @reicheratwork, I think the changes are perfectly reasonable but there is a style issue and I also have question on whether it really is the best approach to annotate the byte swapping functions like this.

src/ddsrt/include/dds/ddsrt/bswap.h
reicheratwork: I agree with both your points, though then it is more a "treating the symptoms rather than the root cause" approach, which I personally try to avoid whereever possible.

Though in this case I think your arguments for not causing false negatives in coverity weigh heavier than not solving the root cause of the reported issue.

src/core/ddsi/tests/sysdeps.c
reicheratwork: No problem

eboasson (approved):
Thanks as always @reicheratwork . The test failures on CI clearly having nothing to do with this.

PR #1044, opened 2021-11-23
Fix for coverity issue #343303

reicheratwork:
Returning result of call to descriptor_fini which was being discarded

### DIFF
--- a/src/tools/idlc/src/descriptor.c
+++ b/src/tools/idlc/src/descriptor.c
@@ -2282,7 +2282,7 @@ ctype_fini(struct constructed_type *ctype)
 
 idl_retcode_t generate_descriptor(const idl_pstate_t *pstate, struct generator *generator, const idl_node_t *node);
 
-idl_retcode_t
+void
 descriptor_fini(struct descriptor *descriptor)
 {
 #if defined(_MSC_VER)
@@ -2302,7 +2302,6 @@ descriptor_fini(struct descriptor *descriptor)
 #if defined(_MSC_VER)
 #pragma warning(pop)
 #endif
-  return IDL_RETCODE_OK;
 }
 
 idl_retcode_t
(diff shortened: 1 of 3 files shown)


src/tools/idlc/src/descriptor.c
```diff
@@ -2365,8 +2365,8 @@ generate_descriptor(
   if (print_descriptor(generator->source.handle, &descriptor) < 0)
     { ret = IDL_RETCODE_NO_MEMORY; goto err_print; }
 
+  ret = descriptor_fini(&descriptor);
```
eboasson, line 2368: I'd expect `descriptor_fini` not to be necessary when `generate_descriptor_impl` returns an error, but for it to be necessary if any of the other functions fail. Moving it here means it only gets called if the descriptor was generated successfully.

Looking at the code of `generate_descriptor_impl`, it seems to be written under the assumption that `descriptor_fini` will be called if it fails. That implies the original code was correct (even if it failed to look at the return value). That seems likely to be a mistake.

From what I can tell `descriptor_fini` never fails (not so surprising for a desctructor) and I think the coverity issue would therefore best be fixed by changing its return type to `void` instead. Furthermore, I think the code overall would improve  if `generate_descriptor_impl` would clean up on error, seemingly as simple as calling `descriptor_fini` on error, and the call to `descriptor_fini` to end up between `err_print` and `err_gen`.

reicheratwork: You are correct about the sequencing of the descriptor_fini function, though I am not sure about changing the return value of `descriptor_fini`, since this is "changing the interface".

Maybe this, and the other improvements should be in their own PR?

eboasson: It is an internal interface, it is strictly the C backend and it is only used for testing. So I think fixing the interface is the sensible thing to do.

This https://github.com/eboasson/cyclonedds/tree/pr1044 is what I had in mind. If you agree, would you mind updating this PR?

reicheratwork: No problem, I was just unsure about possible side-effects of removing the return code or that it could be a deviation from the concept that all these implementation functions should return an `idl_retcode_t`.

eboasson:
Thanks @reicheratwork 🙂

PR #1050, opened 2021-11-26
Fix for coverity issue #343617

reicheratwork:
Suppress false positive for tainting data in dds_stream_normalize1
caused by byteswapping data in function peek_and_normalize_uint32

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -2231,6 +2231,7 @@ static const uint32_t *dds_stream_normalize1 (char * __restrict data, uint32_t *
       }
       case DDS_OP_PLC: {
         assert (xcdr_version == CDR_ENC_VERSION_2);
+        /* coverity[tainted_data_argument: FALSE] */
         if ((ops = stream_normalize_pl (data, off, size, bswap, xcdr_version, ops)) == NULL)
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Thanks for looking into this one @reicheratwork but I think in this case suppressing the warning is not correct. I happen to have been staring at a similar one and noticed that `stream_normalize_pl` manage to slip its lack of correct error propagation through the review cracks. It *may* be that this causes the Coverity warning, or may be something else, but something is amiss for sure.

As these `normalize` functions take care of determining well-formedness of the CDR and byte swapping it if necessary, this is the place where random bytes from the network get laundered. It is therefore the ideal place for attacks and so this code needs to check *everything*. Coverity's "tainted" checks are often obnoxious because they rely on heuristics, but they do provide an additional check that input validation is indeed performed — making them rather valuable.

See https://github.com/eclipse-cyclonedds/cyclonedds/pull/1049 for some indication of the direction I think we need to go for fixing (with the caveat that it may be too strict for some configurations of the data representation QoS setting and encoding modes).

reicheratwork:
@eboasson I think that your commit eacdbf3 in PR #1049  will fix most of the issues with "unverified" or "tainted" data in practice.

Though I do not know whether this will suppress all the warnings generated in Coverity. We can only see....

PR #1052, opened 2021-11-29
Bugfix in extract key from data

dennis-adlink:
This fixes a bug in extract key from data in the cdrstream serializer: after storing `ops` and source offset for a key field, skip the field in `ops` and input stream.


eboasson (approved):
👍🏻

PR #1062, opened 2021-12-06
Add ubsan

eboasson:


k0ekk0ek:
LGTM. Out of curiosity, did you try to enable it on the Windows+gcc build? That may show interesting things too...

eboasson:
> LGTM. Out of curiosity, did you try to enable it on the Windows+gcc build? That may show interesting things too...

It never even occurred to me that it might be supported on Windows, but good point. I'll give it a go.

k0ekk0ek:
libubsan is probably required by default(?) This [post](https://stackoverflow.com/questions/55480333/clang-8-with-mingw-w64-how-do-i-use-address-ub-sanitizers) maybe interesting. Adding a flag on `MINGW` builds is easy enough, but I'll leave the decision to you. Eventually, adding even more [sanitizers](https://clang.llvm.org/docs/index.html) might be good(?) Not all are compatible, but e.g. the thread sanitizer could show races in the stress tests? (Just some random thoughts :slightly_smiling_face:)

eboasson:
Yes, it needs `libusan`. I am not actually too worried about not enabling it on Windows because we are fortunate enough to have very little Windows-specific code.

Enabling more of them will is a definitely something valuable, with `tsan` the most important one. Unfortunately, `tsan` also turns out to be somewhat difficult to deal with for the concurrent bits. Most likely some real issues are hiding among the false positives.

PR #1064, opened 2021-12-07
Include the dev documents in the Sphinx documentation

thijsmie:
This is an experiment that includes the markdown documents in the dev/ folder in the Sphinx documentation build. This also allows them to appear on the website. While I was at it I also included options.md.

### DIFF
--- a/.azure/publish-documentation.yml
+++ b/.azure/publish-documentation.yml
@@ -43,7 +43,7 @@ steps:
     name: cache_pip
   - bash: |
       set -e -x
-      pip install sphinx breathe exhale sphinx-rtd-theme --user --upgrade
+      pip install sphinx breathe exhale sphinx-rtd-theme myst-parser --user --upgrade
     name: install_sphinx
   - bash: |
       set -e -x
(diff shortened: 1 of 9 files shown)


k0ekk0ek:
I agree that it would be good to have the design documentation somewhere more accessible, but the documents in dev are mostly braindumps (the ones I've written anyway), so this is definitely a start, but I'm unsure if we aren't providing too much unstructured information(?) I think the end-result should be a contribution guide or something(?) Then having a landing page that references the individual documents and somewhat guides new contributors is best(?) Also, at least the typetree document is out of date and publishing that does more harm than good. Back in the day, the idea was to reuse the IDL representation internally too, this idea was dropped and @dennis-adlink implemented an optimized form, while the tree structure in the IDL compiler was updated to be easily usable in backends (at least as user-friendly as I good get it).

Anyway, I think we can integrate this, but we should have a plan to follow up and improve the state, but that's just my 0.02$.

thijsmie:
This is overtaken by other documentation work.

PR #1070, opened 2021-12-14
idlc: implement generic idlc_generate() function

mtudan:
This commit extends current implementation of the idlc_generate() function
to be more generic and include some missing options like language support
and additional include paths. For example, this function should now be
used in the C++ repo instead of implementing almost the same thing there
as well.

Signed-off-by: Matija Tudan <tudan.matija@gmail.com>

### DIFF
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -10,17 +10,30 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 
-
 function(IDLC_GENERATE)
   set(one_value_keywords TARGET)
-  set(multi_value_keywords FILES FEATURES)
+  set(multi_value_keywords FILES FEATURES INCLUDES)
   cmake_parse_arguments(
     IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
 
-  if (CMAKE_CROSSCOMPILING)
+  idlc_generate_generic(TARGET ${IDLC_TARGET}
+    FILES ${IDLC_FILES}
+    FEATURES ${IDLC_FEATURES}
+    INCLUDES ${IDLC_INCLUDES}
+  )
+endfunction()
+
+function(IDLC_GENERATE_GENERIC)
+  set(one_value_keywords TARGET BACKEND)
+  set(multi_value_keywords FILES FEATURES INCLUDES SUFFIXES DEPENDS)
+  cmake_parse_arguments(
+    IDLC "" "${one_value_keywords}" "${multi_value_keywords}" "" ${ARGN})
+
+  # find idlc binary
+  if(CMAKE_CROSSCOMPILING)
     find_program(_idlc_executable idlc NO_CMAKE_FIND_ROOT_PATH REQUIRED)
 
-    if (_idlc_executable)
+    if(_idlc_executable)
       set(_idlc_depends "")
     else()
       message(FATAL_ERROR "Cannot find idlc executable")
@@ -57,6 +70,25 @@ function(IDLC_GENERATE)
     list(APPEND IDLC_ARGS "-f" ${_feature})
   endforeach()
 
+  # add directories to include search list
+  if(IDLC_INCLUDES)
+    foreach(_dir ${IDLC_INCLUDES})
+      list(APPEND IDLC_INCLUDE_DIRS "-I" ${_dir})
+    endforeach()
+  endif()
+
+  # generate using which language (defaults to c)?
+  if(IDLC_BACKEND)
+    string(APPEND _language "-l" ${IDLC_BACKEND})
+  endif()
+
+  # set dependencies
+  if(IDLC_DEPENDS)
+    list(APPEND _depends ${_idlc_depends} ${IDLC_DEPENDS})
+  else()
+    set(_depends ${_idlc_depends})
+  endif()
+
   set(_dir ${CMAKE_CURRENT_BINARY_DIR})
   set(_target ${IDLC_TARGET})
   foreach(_file ${IDLC_FILES})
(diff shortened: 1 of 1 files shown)


mtudan:
@eboasson @k0ekk0ek FYI: C++ repo PR: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/192.

cmake/Modules/Generate.cmake
```diff
@@ -30,6 +30,22 @@ function(IDLC_GENERATE)
     set(_idlc_depends CycloneDDS::idlc)
   endif()
 
+  # find idlcxx shared library
+  if(IDLC_CXX_SUPPORT)
```
eboasson, line 34: Thanks @mtudan, this would work but it would pull in knowledge of the C++ binding into the core repository, and we've tried our best to avoid doing that so far. Looking at the changes, it seems to me that it would be possible to avoid that by passing in two different arguments and removing the `CXX_SUPPORT` one: the backend library and a list of generated suffixes. For example something like:
```
set(one_value_keywords TARGET BACKEND)
set(multi_value_kewords FILES FEATURES INCLUDES SUFFIXES)
```
Where the defaults for the SUFFIXES would be ".c .h" and nothing for BACKEND. Actually, if I'd try it, I'd make a `IDLC_GENERATE_GENERIC` for that, and then let all language bindings use to that, including the C one. Then the C++ version could do
```
    if(CMAKE_CROSSCOMPILING)
       find_library(_idlcxx_shared_lib cycloneddsidlcxx NO_CMAKE_FIND_ROOT_PATH REQUIRED)

       if(_idlcxx_shared_lib)
         set(_idlcxx_depends "")
       else()
         message(FATAL_ERROR "Cannot find idlcxx shared library")
       endif()
     else()
       set(_idlcxx_shared_lib "$<TARGET_FILE:CycloneDDS-CXX::idlcxx>")
       set(_idlcxx_depends CycloneDDS-CXX::idlcxx)
     endif()

    IDLC_GENERATE_GENERIC(... BACKEND ${_idlcxx_shared_lib} SUFFIXES .hpp)
```

That seems to cover the bits where there is now something dependent on whether or not C++ is being targeted.

What do you think?

mtudan: @eboasson I think all of your points are valid, e.g. I wasn't aware of the effort put to avoid the knowledge of the C++ binding into the core repository. I did changes according to your suggestions, if there is still something you think it can be done better don't hesitate to contact me.

eboasson (commented):
Thanks @mtudan! I like it, I am not entirely sure about the need for distinguishing between "source" and "header" (it seems like they are piled together in the end, but perhaps not because of the quoting in `add_custom_target` at the very end).

So what do you think: is my hunch that there is no need to distinguish between them is right? If it is, a couple of lines can be deleted and gets to be very simple and straightforward; and if it isn't, I'll be happy to merge it as-is.

mtudan:
@eboasson you're referencing to this snippet below?
``` # set source suffixes (defaults to .c and .h)
  if(IDLC_SUFFIXES)
    foreach(_s ${IDLC_SUFFIXES})
      string(FIND ${_s} ".c" _spos)
      string(FIND ${_s} ".h" _hpos)
      if(_spos GREATER_EQUAL 0)
        set(_ssuffix ${_s})
      elseif(_hpos GREATER_EQUAL 0)
        set(_hsuffix ${_s})
      endif()
    endforeach()
  else()
    set(_hsuffix ".h")
    set(_ssuffix ".c")
  endif()
```

eboasson:
Yes, 'cos then that gets used to build up the _source/_header and the lists of them, which then gets used for
```
OUTPUT   ${_source} ${_header}
```
and for
```
add_custom_target("${_target}_generate" DEPENDS "${_sources}" "${_headers}")
```
I think the "OUTPUT" thing doesn't care about what's what, nor about the order. So the only bit I wonder about is:
because there the quotes make it look like DEPENDS takes two arguments (sources and headers), rather than simply a list of dependencies.

If for `add_custom_target` it is simply a list of dependencies, then I don't think there is a need for distinguishing between the different types of output generated and it could become (in perl-style because it is so much more compact than CMake): `@output = map { "$name.$_" } for @suffixes`

mtudan:
But the real reason why that mess is needed it’s because we need to skip generating `.cpp` files in case of the `idlcxx_generate()` function - so to make sure only `.hpp` is generated.

mtudan:
Well maybe we can go with much simpler snippet:
```
  if(IDLC_SUFFIXES)
    list(GET IDLC_SUFFIXES 0 _hsuffix)
    list(GET IDLC_SUFFIXES 1 _ssuffix)
  else()
    set(_hsuffix ".h")
    set(_ssuffix ".c")
  endif()
```
assuming that `idlcxx_generate()` will set suffixes as `SUFFIXES .hpp .cpp` and every other implementation will follow that. Current implementation is a bit complicated, but it allows to swap that for example. And the reason behind all of that is that generating `.cpp` files is currently not implemented in `idlcxx_generate()` function.

thijsmie:
I think sources and headers should indeed be unified as @eboasson suggests, there is no practical difference in the generate function and many languages (such as python) don't even have separate headers and sources.

mtudan:
Won't that be an issue since `.cpp` files are currently not generated (only `.hpp`) in case of C++?

eboasson:
Hi @mtudan, this has sat here for far too long ... so the answer to the question is that all that's needed is to give a list of files to CMake. It knows how to deal with `.c` vs `.h` files, so for C that works fine. If for C++ there's only a `.hpp` file generated, well that's fine too.

I did want to try before making any unfounded claims ... https://github.com/eboasson/cyclonedds/commit/ab3e157b2a9b5250f30eeb262e68bc696bb23265 is a small commit on top of (a rebased version of) this PR and seems to work. I'm no CMake expert, so if someone who knows CMake a bit better could comment, that'd be appreciated.

(My not knowing CMake all that well and the web having many different versions of the "current idiomatic CMake" makes me always feel a bit uncomfortable deciding whether a PR uses CMake "properly" or not ... That's just too bad, and I am sorry that your PR likely suffered from that as well.)

eboasson (approved):
Thanks @mtudan!

PR #1072, opened 2021-12-16
Restructure documentation

thijsmie:
This is a WIP of restructuring the docs and reviewing all doxygen docstrings in the public API.


eboasson (commented):
Wonderful @thijsmie! I admit haven't properly checked the output, not even checked the color of the background, perhaps it is some peachy gradient from plum to whatever the color is you get from averaging colours over all living crocodiles skins' 😁 Perhaps you can give me a hint as to how to best review such things 🙂

I had three nitpicks on a small mistake and two bits of clarification that stood out to me. Obviously there are many opportunities for improving the content beyond these three, but that would involve adding real content, and that's not the purpose of this pull request. Not in my opinion anyway.

thijsmie:
I'd recommend you have a peek at what is generated, here a small recipe, since I have switched the preferred theme of the documentation:
```shell
$ pip install --user sphinx breathe piccolo_theme
$ cmake -DBUILD_DOCS ..
$ make docs
$ cd docs/docs
$ python -m http.server
```

eboasson (approved):
☺️

PR #1073, opened 2021-12-16
Drop messages with incorrect entity kinds in header

### LINKED ISSUE #1065: Question about assert in Debug build mode
I want to build cyclonedds in Debug mode and run some examples.But I was failed,it will crash in some assertion.
I use cmake command like this,and run hello world example
```
cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_EXAMPLES=ON ..

bin/HelloworldPublisher

=== [Publisher]  Waiting for a reader to be discovered ...
HelloworldPublisher: /home/youxinxin/repos/cyclonedds/src/core/ddsi/src/ddsi_entity_index.c:404: entidx_lookup_proxy_writer_guid: Assertion `is_writer_entityid (guid->entityid)' failed.
[1]    13408 abort (core dumped)  bin/HelloworldPublisher
```
Everything ok if build in release mode,so is this assertion always be checked failed?

eboasson:
And then some, because I spotted another few things, tried to add fuzzer for the RTPS message parsing and found a number of things that I convinced myself are not actually causing problems but are not as they should be:
* Signed left shift (pretty sure the actual behaviour was fine on supported platforms).
* Possibility of signed integer overflow (an almost-certainly malformed `HEARTBEAT` that states no data is available, but doesn't introduce any new issues).
* Misaligned accesses if submessages, inline QoS or data is misaligned: the spec explicitly states the submessages must be aligned to multiples of 4 bytes relative to the start of the RTPS message, but it doesn't state this about inline QoS or data. I think it is missing in the spec, we'll see if disallowing ones not aligned to 4 bytes is going to cause trouble. 
* There was an assert somewhere in the internal representation that the payload offset is a multiple of 4, but without an actual reason to do so.
* Possibility of overflow on fragment numbers while validating a `DATA_FRAG` header.

As well as some compiler warnings that popped up with an updated build environment.

@dennis-adlink do you have time for a quick sanity check whether OSS-Fuzz's build scripts are happy with this? I don't have a machine that is sanctioned by Google ...

Fixes #1065 — @yooouxin perhaps you would like to check?


yooouxin:
@eboasson I tested on my machine,it works well!

eboasson:
Thanks for checking @yooouxin !

dennis-adlink:
@eboasson sorry for my (too) late reply, and you probably already found out from the CI results, but I can confirm that this change works fine with the oss-fuzz build scripts.

Issue #1060, opened 2021-12-04
abort() in dds_stream_countops1 for deeply-nested message


ErikAtApex:
Hi folks!

We recently received [an interesting report](https://gitlab.com/ApexAI/performance_test/-/issues/134) in the Apex.AI performance_test repository. It seems that Cyclone DDS is having trouble with a specific deeply-nested message.

- [`Struct32k`](https://gitlab.com/ApexAI/performance_test/-/blob/master/performance_test/msg/Struct32k.idl) is the message in question
- A `Struct32k` contains 8x [`Struct4k`](https://gitlab.com/ApexAI/performance_test/-/blob/master/performance_test/msg/Struct4k.idl) members
- A `Struct4k` contains 16x [`Struct256`](https://gitlab.com/ApexAI/performance_test/-/blob/master/performance_test/msg/Struct256.idl) members
- A `Struct256` contains 16x [`Struct16`](https://gitlab.com/ApexAI/performance_test/-/blob/master/performance_test/msg/Struct16.idl) members

When running the tool with `Struct32k`, the `abort` is generated when the first message is published. Strangely, the `Struct4k`, `Struct256`, and `Struct16` messages all work fine. Do you have any ideas why the `Struct32k` message specifically would cause this particular `abort`?

If you wish to observe this yourself, it can be easily reproduced:

```
git clone git@gitlab.com:ApexAI/performance_test.git
cd performance_test
docker build . -f dockerfiles/Dockerfile.CycloneDDS -t performance_test_cyclone_dds
docker run -it performance_test_cyclone_dds bash
/opt/performance_test/lib/performance_test/perf_test -c CycloneDDS -m Struct32k
```

This Dockerfile is using the default Cyclone DDS in ROS 2 rolling, but I have also tried with 0.8.1, with the same results.

Thanks a ton!

eboasson:
@ErikAtApex it is the IDL compiler writing out an incomplete serializer (see the end of the generated output), I haven't (yet) tried to figure out why this happens. If it had written it out, the result would have been a horrific array of roughly 64k integers, because it enumerates *all* the primitive members explicitly.

In `master`, because of all the work for XTypes, changes got merged in where each type is written out once, and that works fine. However, I'm afraid that backporting those changes into the 0.8 branch defeats the purpose of having a release branch ...

Some observations:
[...]

PR #1077, opened 2021-12-21
Detect overflow in idlc when stashing instructions

dennis-adlink:
Add a check in the IDLC C generator so that it shows an error message in case the maximum number of serializer instructions is reached. Because of 16-bits offset fields in the instructions, the maximum number of instructions that can be used is limited to 65k.

This check prevents getting incomplete instructions in the generated code without any error or warning, as reported in #1060.


eboasson (approved):
Thanks @dennis-adlink!

PR #1078, opened 2021-12-22
Change OSS-Fuzz build script to fix coverage issue

dennis-adlink:
A few changes in the OSS-Fuzz build script that will hopefully fix the segfault that occurs in the undefined behavior sanitizer when generating code coverage reports. Although on my dev machine (where I reproduced the issue) the coverage reports are generated successfully with these changes in, I'm not sure if using the unmodified rpath for shared libs will also work on CI.


eboasson (approved):
Thanks @dennis-adlink ! Let's see what happens 🙂

PR #1080, opened 2021-12-23
Add prototype event framework based on select

k0ekk0ek:
This (draft) PR adds a prototype event framework implementation to replace `os_sockWaitset` from the *ddsi* sources. I have some code laying around to use *epoll* on Linux (and Windows through [wepoll](https://github.com/piscisaureus/wepoll)) and *kqueue* for macOS and FreeBSD. This also implements netlink events for Linux, Windows, macOS and FreeBSD. I threw in a temporary `netlink` tool to demonstrate those work as intended (basically what #471 intended to do too), but this also adds interface up/down and ipv4/ipv6 added/deleted events without requiring an extra thread (*recv* thread can handle those as well?). Anyway, it requires some polishing still, but the basics are there and I'm hoping to get some feedback on the design, etc. Also, @eboasson, now is probably a good time to issue a CQ for wepoll and (probably required) CMake's CheckTypeSize module. Both are BSD licensed, so that shouldn't be much of a problem(?)


k0ekk0ek:
Oh, btw, I should add that I intend to use the CheckTypeAlignment module to do type-punning for the eventlist, so I don't have to expose the types that wepoll introduces. I also think that we can use the mechanism to hide platform implementation headers for other sources. I also intend to reorder ddsrt slightly so I can use ddsrt within the idl library etc. A WiP that I'm hacking together in some lost hours and spare time can be found [here](https://github.com/k0ekk0ek/sys).

k0ekk0ek:
Right, so this PR now also includes generating the utility functions used in `libidl` from `ddsrt` sources, ditching in overely complex feature checking in `ddsrt`.

k0ekk0ek:
The little `netlink` application that I've added as an example works with both `select` and `epoll` on Windows, I'm hoping the same goes for Linux. I'm pretty sure I've tested everything with `select` on Linux, Windows, macOS and FreeBSD, but forgot to mention that part in the comments :slightly_smiling_face:

k0ekk0ek:
Right, so there's one or two compilation warnings but I have verified the netlink tool works on all supported platforms, which makes this a nice base I guess(?) Anyway, moving on to writing test cases now and then hopefully soonish get to replace stuff in ddsi's connection code.

k0ekk0ek:
Ok, rebased on #1086. Updated the eventlist implementation and added some tests for that. I should add more tests for the event loop itself, but I'll focus on implementing this in ddsi first.

k0ekk0ek:
Tests all pass, so I think this is fine UDP wise(?) Not sure about TCP connections yet as I removed the listener thread, so I'll have to go and test that some more. There's a couple of small warnings (that's why the builds fail), but I think this is nearly there. Ran tests and verified netlink events on all platforms and with different polling mechanisms.

k0ekk0ek:
The latest push makes TCP "work" again.

Flova:
I am interested in the functionality of fallback interfaces in cyclone dds and #1518 linked this PR in this context. Are there any future plans to continue development here?

Issue #1079, opened 2021-12-23
Assertion failure in IDLC C backend for typedefs of arrays


eboasson:
Looks good! I also agree with the refactoring in the code for handling forward declarations. When doing some additional manual testing, I found an issue with array-typedefs (which can be nasty as I found out myself when implementing XTypes related changes for #904 in IDLC). 

The IDL fragment below is causing a failing assert in `descriptor.c:404`:

```
struct a {
    octet a1;
};

typedef a b[2];

struct c {
    b c1;
};
```

_Originally posted by @dennis-adlink in https://github.com/eclipse-cyclonedds/cyclonedds/pull/1068#pullrequestreview-838855618_

eboasson:
Confirmed fixed. Thanks @dpotman !

PR #1082, opened 2021-12-23
Fix idl_strip for typedefs to array

dennis-adlink:
This changes the `idl_strip` function so that it only resolves forward declarations and alias types, and not arrays. This also changes the `idl_is_alias` function back the behavior before PR #1068, which allowed array to be an alias. This, in combination with changes in the flags passed in calls to `idl_strip` from `emit_array`, fixes issue #1079

### DIFF
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -548,8 +548,8 @@ IDL_EXPORT void *idl_iterate(const void *root, const void *node);
 
 IDL_EXPORT void *idl_unalias(const void *node);
 
-#define IDL_STRIP_ALIASES (1u<<0) /**< expose base type of aliase(s) */
-#define IDL_STRIP_ARRAYS (1u<<1) /**< expose base type of array(s) */
+#define IDL_STRIP_ALIASES (1u<<0) /**< expose base type of aliase(s) for a non-array type */
+#define IDL_STRIP_ALIASES_ARRAY (1u<<1) /**< expose base type of aliase(s) for array type */
 #define IDL_STRIP_FORWARD (1u<<2) /**< expose definition */
 IDL_EXPORT void *idl_strip(const void *node, uint32_t flags);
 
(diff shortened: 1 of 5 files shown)


k0ekk0ek:
@dennis-adlink, this may fix a problem. Kudos for finding it! But is does discard some IDL language concepts and it is therefore I have a problem with these changes. An `array` is not an `alias`. An `alias` is a different name for the exact same thing. In IDL `typedef long someint` means using `someint` synonymous to `long`, whereas that is specifically not true for `typedef long someints[2]`. But alas, this is already in...

eboasson:
@k0ekk0ek that was a deliberate decision: the changes to `idl_strip` broke the existing code and blocked everything. So merging this was simple damage control.

That doesn't mean I disagree with you that the representation of types in the IDL compiler needs fixing. Its internal type representation follows the grammar far too strictly, leading to this type of problems. E.g., a struct has members and how they are mapped to declarators/declarator lists in the source is irrelevant, and more relevant to this issue, an array of size N of type T is simply a type itself, and should be represented as such internally rather than mixing it with names. Parsing the source into a simple representation of types and names referencing types will really simplify things, but it will also affect all front-ends and so should be done in a coordinated manner.

dennis-adlink:
@k0ekk0ek When restoring that lines in `idl_is_alias` I realized that this was conflicting with having aliases as a synonym for a type (and you had good reasons for adding these lines), but I couldn't find a better work-around to fix this in short-term. 

For a better fix to this problem, we could consider handling arrays similar to sequences, so make a `idl_node` for an array that has the element type as its type-spec (I think you also proposed this solution in a discussion we had some time ago)? This approach also matches the way array types are handled in XTypes type objects, and this way we avoid the exceptions for arrays in the C and type object generators. 

Something else I just realized, is that the changes in `idl_unalias` in #1068 may have impact on the cxx generator, because the logic in this function changed a bit. I think we can rewrite the calls to `idl_unalias` to `idl_strip`, which behaves the same as `idl_unalias` did before if passing the right flags.

PR #1093, opened 2022-01-06
Fix the copying of builtin/custom annotations in type objects

thijsmie:
In a few places where switches between the Cyclone XTypes representation and the spec representation happen the ann_custom and ann_builtin fields were not properly considered. They are tricky because they are optional but also again contain optional fields. I believe I have now ensured that NULL's and values are always properly carried over. This could use a proper testcase, however, the ann_custom and ann_builtin stuff is not used anywhere as of yet.

### DIFF
--- a/src/core/ddsi/src/ddsi_typewrap.c
+++ b/src/core/ddsi/src/ddsi_typewrap.c
@@ -435,7 +435,7 @@ static void set_member_detail (struct xt_member_detail *dst, const DDS_XTypes_Co
     dst->annotations.ann_builtin = NULL;
   }
 
-  if (src->ann_builtin) {
+  if (src->ann_custom) {
     dst->annotations.ann_custom = ddsrt_calloc(1, sizeof(DDS_XTypes_AppliedAnnotationSeq));
     DDS_XTypes_AppliedAnnotationSeq_copy (dst->annotations.ann_custom, src->ann_custom);
   } else {
@@ -902,7 +902,7 @@ static void get_type_detail (DDS_XTypes_CompleteTypeDetail *dst, const struct xt
     dst->ann_builtin = NULL;
   }
 
-  if (src->annotations.ann_builtin) {
+  if (src->annotations.ann_custom) {
     dst->ann_custom = ddsrt_calloc(1, sizeof(DDS_XTypes_AppliedAnnotationSeq));
     DDS_XTypes_AppliedAnnotationSeq_copy (dst->ann_custom, src->annotations.ann_custom);
   } else {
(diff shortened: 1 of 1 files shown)


src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -423,6 +423,26 @@ static void xt_lbounds_dup (struct DDS_XTypes_LBoundSeq *dst, const struct DDS_X
...
+{
+  ddsrt_strlcpy (dst->name, src->name, sizeof (dst->name));
+  if (src->ann_builtin) {
+    dst->annotations.ann_builtin = ddsrt_calloc(1, sizeof(struct DDS_XTypes_AppliedBuiltinMemberAnnotations));
+    DDS_XTypes_AppliedBuiltinMemberAnnotations_copy (dst->annotations.ann_builtin, src->ann_builtin);
+  } else {
+    dst->annotations.ann_builtin = NULL;
+  }
+
+  if (src->ann_builtin) {
```
eboasson, line 438: Shouldn't that be `ann_custom`?

src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -858,25 +873,55 @@ static void DDS_XTypes_AppliedBuiltinMemberAnnotations_copy (struct DDS_XTypes_A
...
-  DDS_XTypes_AppliedBuiltinTypeAnnotations_copy (dst->ann_builtin, src->annotations.ann_builtin);
-  DDS_XTypes_AppliedAnnotationSeq_copy (dst->ann_custom, src->annotations.ann_custom);
+  if (src->annotations.ann_builtin) {
+    dst->ann_builtin = ddsrt_calloc(1, sizeof(struct DDS_XTypes_AppliedBuiltinTypeAnnotations));
+    DDS_XTypes_AppliedBuiltinTypeAnnotations_copy (dst->ann_builtin, src->annotations.ann_builtin);
+  } else {
+    dst->ann_builtin = NULL;
+  }
+
+  if (src->annotations.ann_builtin) {
```
eboasson, line 905: Shouldn't this be `ann_custom`?

src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -858,25 +873,55 @@ static void DDS_XTypes_AppliedBuiltinMemberAnnotations_copy (struct DDS_XTypes_A
...
-  DDS_XTypes_AppliedBuiltinMemberAnnotations_copy (dst->ann_builtin, src->annotations.ann_builtin);
-  DDS_XTypes_AppliedAnnotationSeq_copy (dst->ann_custom, src->annotations.ann_custom);
+  if (src->annotations.ann_builtin) {
+    dst->ann_builtin = ddsrt_calloc(1, sizeof(struct DDS_XTypes_AppliedBuiltinMemberAnnotations));
+    DDS_XTypes_AppliedBuiltinMemberAnnotations_copy (dst->ann_builtin, src->annotations.ann_builtin);
+  } else {
+    dst->ann_builtin = NULL;
+  }
+
+  if (src->annotations.ann_builtin) {
```
eboasson, line 923: Shouldn't this be `ann_custom`?

eboasson (commented):
LGTM except for a few apparent typos. I'll take a Python testcase that hits this as sufficient evidence for now, especially if these fixes unblock merging the xtypes support in `cyclonedds-python`.

thijsmie:
Good catch on the typos, that is what I get for copy-pasting.

eboasson:
Fix confirmed locally with the python tests. That's good enough for me.

PR #1095, opened 2022-01-06
Fix package.xml/CMakeList.txt version mismatch and bump to 0.8.2

### LINKED ISSUE #1089: Use package.xml versioning
It would be nice if the `package.xml` semantic version was bumped with every change. Without being diligent on this, ROS users get nonsensical debian package versions like these:
```
$ apt policy ros-galactic-cyclonedds

ros-galactic-cyclonedds:
  Installed: 0.8.0-5focal.20210608.002038
  Candidate: 0.8.0-6focal.20211217.210911
  Version table:
     0.8.0-6focal.20211217.210911 500
        500 http://packages.ros.org/ros2/ubuntu focal/main amd64 Packages
 *** 0.8.0-5focal.20210608.002038 100
        100 /var/lib/dpkg/status
```
where both packages are `0.8.0`, but with different build numbers, even though `0.8.0-6` has features that `0.8.0-5` does not, like `dds_is_loan_available`.

eboasson:
Backport from `master`. Bumping the version so we can do a new release and fix #1089.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -15,6 +15,8 @@
 #
 
 steps:
+  - bash: bash versioncheck.bash
+    name: versioncheck
   - task: UsePythonVersion@0
     inputs:
       versionSpec: '3.8'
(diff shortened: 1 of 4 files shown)


eboasson:
ROS 2 CI run checks out, as expected.

PR #1096, opened 2022-01-07
Free'ing non-owned pointer on normalize failure

thijsmie:
You only run into this if you make a mistake elsewhere, but still, `ddsi_typeinfo_deser` should not free the data pointer in a serdata if it did not make its own copy to byteswap and the stream normalize fails.


eboasson (approved):
😱 

Regardless of whether you found it by inspection or through fault-injection, I am grateful for this fix!

PR #1098, opened 2022-01-10
Fix handling unsupported types in opcode generation

reicheratwork:
Replaced calling of abort() in cases of unsupported types in opcode
generation with more descriptive error codes and messages


eboasson (approved):
LGTM!

PR #1099, opened 2022-01-10
Validate CDR in ddsi_typemap_deser and fix sanitize issue

dennis-adlink:
Similar to `ddsi_typeinfo_deser`, the function `ddsi_typemap_deser` should validate the CDR by calling the normalize function also in case there is no byte swapping. And fixing an undefined behavior sanitizer error, that showed up in CI (unrelated to the changes in the other commit in this PR).


eboasson (approved):
Thanks @dennis-adlink !

PR #1100, opened 2022-01-10
Make sequence numbers unsigned

eboasson:
There are some issues handling sequence number 2**63-1 (the maximum sequence number allowed by the spec), representing them as unsigned numbers fixes those issues.

The first few commits deal with some cleanups to prepare the code base by removing the few cases where -1 was used and by consistently using the `seqno_t` type for sequence numbers.

The next-to-last commit is the actual transformation. In addition to making the representation unsigned, it also wraps it in a struct to guarantee that all arithmetical and relational operations turn into compile-time errors. This ensures that every potentially problematic case must be inspected. It eliminates the risk of overlooking a usage, it doesn't guarantee that the inspection and manual modification is correct.

The final commit is a (mostly mechanical) rewrite to remove the wrapper.

### DIFF
--- a/src/core/ddsc/src/dds_whc.c
+++ b/src/core/ddsc/src/dds_whc.c
@@ -220,7 +220,7 @@ static uint32_t whc_node_hash (const void *vn)
   /* we hash the lower 32 bits, on the assumption that with 4 billion
    samples in between there won't be significant correlation */
   const uint64_t c = UINT64_C (16292676669999574021);
-  const uint32_t x = (uint32_t) n->seq.v;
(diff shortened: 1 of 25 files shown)


dennis-adlink (approved):
LGTM! I've reviewed the changes in detail and didn't find any issue, except for wrong indenting in q_xevent.c:767 and line 776

eboasson:
Rebased and updated to fix indentation problem spotted by @dennis-adlink. Also double-checked that `master` hasn't introduced any new uses of a sequence number that need updating. (That I can't remember merging anything in the past two weeks that would qualify as such is insufficient reason not to check!)

PR #1101, opened 2022-01-10
Derive iceoryx settings from qos

### LINKED ISSUE #894: Set shared memory subscriber options based on the reader QoS settings
Currently the shared memory subscriber options (of iceoryx) are set by config file and afterwards it is determined whether the reader QoS are compatible with the options.

This happens in dds_reader.c in the function `dds_reader_support_shm`. 
There the check comparing `sub_history_req` with `qos->history.depth` appears to be wrong leading to us not being able to use e.g. Keep All with a certain history depth n and sub_history_req < n. 

E.g. if the `SubscriberHistoryRequest` is set to 9 and we have set the QoS to `Keep All 10` the communication will not use shared memory but the network, which seems unintended. 

Potentially we want to use `SubScriberQueueSize` instead but the issue of how the options are related to the QoS is more complex and we should properly write them down here.

In general we should change the overall concept to determine the iceoryx options on a per reader basis based on the QoS settings provided to the reader. This also allows to have different shared memory options for individual subscribers, something which is currently not possible.

### Required Steps

1. Clarify relations between the iceoryx config parameters and QoS settings
2. Check if the current behavior is not intentional and should be fixed.
3. Clarify the scope of these changes and the details of the intended behavior. We could also start with a bugfix (if needed) before changing the overall concept of setting the options based on the QoS.

MatthiasKillat:
Closes #894

### DIFF
--- a/src/core/ddsc/src/dds_reader.c
+++ b/src/core/ddsc/src/dds_reader.c
@@ -482,19 +482,13 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
     return false;
   }
 
-  // uint32_t sub_history_req = cfg->sub_history_request;
-  // MAKI: need the max from iceoryx here (available on master in the API but
-  // not in release_1.0)
-  const int32_t max_sub_history_req = 16;
-
   return (NULL != qos &&
           DDS_READER_QOS_CHECK_FIELDS ==
               (qos->present & DDS_READER_QOS_CHECK_FIELDS) &&
           DDS_LIVELINESS_AUTOMATIC == qos->liveliness.kind &&
           DDS_INFINITY == qos->deadline.deadline &&
           DDS_DURABILITY_VOLATILE == qos->durability.kind &&
-          DDS_HISTORY_KEEP_LAST == qos->history.kind &&
-          max_sub_history_req >= qos->history.depth);
+          DDS_HISTORY_KEEP_LAST == qos->history.kind);
 }
 #endif
 
@@ -664,20 +658,35 @@ static dds_entity_t dds_create_reader_int (dds_entity_t participant_or_subscribe
 
     iox_sub_options_t opts;
     iox_sub_options_init(&opts);
-
-    // ICEORYX TODO: handle failure (how should the system behave if resources are insufficient?)
+    
     iox_sub_storage_extension_init(&rd->m_iox_sub_stor);
 
     assert (rqos->durability.kind == DDS_DURABILITY_VOLATILE);
-    // opts.queueCapacity = rd->m_entity.m_domain->gv.config.sub_queue_capacity;
-    opts.queueCapacity = (uint64_t)rqos->history.depth;
 
-    // MAKI: if this is always zero we need no
-    // capacity on the writer side ...
-    // opts.historyRequest = 0;
-    // TODO: will those mechanism used by iceoryx and cyclonedds to get data
-    // at late joining readers interfere?
-    opts.historyRequest = (uint64_t)rqos->history.depth;
+    // TODO: need the max from iceoryx here,
+    // available in the API of master but not in release_1.0, will be available
+    // in the upcoming release 2.0 (planned March 2022)
+    const int32_t max_sub_queue_capacity = 256; // value from iceoryx 1.0
+
+    // NB: We may lose data after history.depth many samples are received (if we
+    // are not taking them fast enough from the iceoryx queue and move them in
+    // the reader history cache), but this is valid behavior for volatile.
+    // It may still lead to undesired behavior as the queues are filled very
+    // fast if data is published as fast as possible.
+    // NB: If the history depth is larger than the queue capacity, we still use
+    // shared memory but limit the queueCapacity accordingly (otherwise iceoryx
+    // emits a warning and limits it itself)
+
+    if (rqos->history.depth <= max_sub_queue_capacity) {
+      opts.queueCapacity = (uint64_t)rqos->history.depth;
+    } else {
+      opts.queueCapacity = (uint64_t)max_sub_queue_capacity;
+    }
+
+    // NB: since currently we only support the volatile reader case we will
+    // never request historical data
+    opts.historyRequest = 0;
+    
     rd->m_iox_sub = iox_sub_init(&rd->m_iox_sub_stor.storage, gv->config.iceoryx_service, rd->m_topic->m_stype->type_name, rd->m_topic->m_name, &opts);
     shm_monitor_attach_reader(&rd->m_entity.m_domain->m_shm_monitor, rd);
 
--- a/src/core/ddsc/src/dds_reader.c
+++ b/src/core/ddsc/src/dds_reader.c
@@ -482,15 +482,19 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
     return false;
   }
 
-  uint32_t sub_history_req = cfg->sub_history_request;
+  // uint32_t sub_history_req = cfg->sub_history_request;
+  // MAKI: need the max from iceoryx here (available on master in the API but
+  // not in release_1.0)
+  const int32_t max_sub_history_req = 16;
 
   return (NULL != qos &&
-    DDS_READER_QOS_CHECK_FIELDS == (qos->present & DDS_READER_QOS_CHECK_FIELDS) &&
-    DDS_LIVELINESS_AUTOMATIC == qos->liveliness.kind &&
-    DDS_INFINITY == qos->deadline.deadline &&
-    DDS_DURABILITY_VOLATILE == qos->durability.kind &&
-    DDS_HISTORY_KEEP_LAST == qos->history.kind &&
-    (int)sub_history_req >= (int)qos->history.depth);
+          DDS_READER_QOS_CHECK_FIELDS ==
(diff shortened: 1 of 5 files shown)


src/core/ddsc/src/dds_reader.c
```diff
@@ -482,15 +482,19 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
     return false;
   }
 
-  uint32_t sub_history_req = cfg->sub_history_request;
+  // uint32_t sub_history_req = cfg->sub_history_request;
+  // MAKI: need the max from iceoryx here (available on master in the API but
+  // not in release_1.0)
+  const int32_t max_sub_history_req = 16;
```
eboasson, line 488: Perhaps it is possible to use `#if ...` to check whether the Iceoryx version used is new enough to provide the max value?

src/core/ddsc/src/dds_reader.c
```diff
@@ -665,8 +669,15 @@ static dds_entity_t dds_create_reader_int (dds_entity_t participant_or_subscribe
...
-    opts.historyRequest = 0;
+    // opts.queueCapacity = rd->m_entity.m_domain->gv.config.sub_queue_capacity;
+    opts.queueCapacity = (uint64_t)rqos->history.depth;
+
+    // MAKI: if this is always zero we need no
+    // capacity on the writer side ...
+    // opts.historyRequest = 0;
+    // TODO: will those mechanism used by iceoryx and cyclonedds to get data
+    // at late joining readers interfere?
+    opts.historyRequest = (uint64_t)rqos->history.depth;
```
eboasson, line 680: A late joiner only needs (or: is allowed to get, depending on how you interpret things) historical data if its durability kind ≠ `VOLATILE`. So I think it should be 0 for volatile readers.

As an aside: in DDS the amount of historical data available is determined by the writer (for `TRANSIENT_LOCAL`) or the topic (for `TRANSIENT` and `PERSISTENT` — those are NIY so don't matter), and for `TRANSIENT_LOCAL` at least, the reader is force-fed with more than anything. Naturally, you end up with at most `rqos->history.depth` samples/instance, so using this for an Iceoryx history request seems reasonable.

And also beware of the `assert (rqos->durability.kind == DDS_DURABILITY_VOLATILE)`.

src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h
```diff
@@ -1706,19 +1706,7 @@ static struct cfgelem shmem_cfgelems[] = {
       "<p>If you don't want to see any log from shared memory, use <i>off</i> to disable log message.</p>"),
     VALUES(
       "off","fatal","error","warn","info","debug","verbose"
-    )),
```
eboasson, line 1709: Rather than deleting them outright, I'd prefer to make them deprecated. This is pretty easy: just do `DEPRECATED("SubQueueCapacity")` (etc.) and replace the `pf_int` by `pf_nop` so it is no longer is printed in the configuration dump at the beginning of a trace file.

It is also possible to use `uf_nop` instead of `uf_natint` to completely ignore the value provided in the configuration (it would then no longer detect garbage where an integer is expected, but that's ok). And hopefully removing the `MEMBER` entry works, but I'm not 100% sure about that ... if not, it requires but a small change so if it is worth trying.

eboasson (commented):
@MatthiasKillat I know it is just a draft for now, but I had a look anyway and there are some small things I thought might be good to note now, rather than wait.

src/core/ddsc/src/dds_reader.c
MatthiasKillat: Allowing non-volatile readers (e.g. extending to transient local) is not part of this pull request since there are other issues to be resolved first.

I wonder how useful this assert was anyway, if it has to be ensured at all times that we need a volatile reader. As of now, nothing is checked in this regard in a release build.

src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h
MatthiasKillat: Deprecated them, but I am not sure whether this is the proper way. In any case, they are not used internally anymore.

src/core/ddsc/src/dds_reader.c
MatthiasKillat: Changed the logic on reader and writer side to essentially not care anymore for those parameters. It was pointless before to have iceoryx buffer data at the writer side (history capacity) but never requesting this data anyway in the volatile case (sub history request = 0).

Now those are always zero for the volatile reader case (the only case currently supported, but this is to be extended in another issue #965). However, we no limit the subscriber queue size to the history depth. This saves memory under load but leads to data being lost earlier than before (still within spec). If we do not want this we should always set it to the maximum available in iceoryx or leave it to the configuration as it was.
Note that the history request and history capacity still make no sense as they are used and need to be removed or the semantics changed.

src/core/ddsc/src/dds_reader.c
MatthiasKillat: I see no way to check for the version with the headers we currently have in `iceoryx_binding_c`. A header that allows something like this exists for the C++ library, `iceoryx_versions.hpp` but not the C library. I will create an issue for this in the next release but then this point is moot since the values are available with the API.

MatthiasKillat:
@eboasson Ready for another review, but there may still be semantic issues requiring discussion. 

Note that #965 will be based on this PR, the main problem is how to deal with transient local readers using iceoryx. I still am not sure how a proper solution would look like, even with iceoryx changes. But one problem at a time...

src/core/ddsc/src/dds_reader.c
```diff
@@ -480,17 +480,26 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
...
+  // NB: since we only support DDS_DURABILITY_VOLATILE which does not allow receiving of historical data, we do not
+  // need to care about the iceoryx subscriber history request
 
-  uint32_t sub_history_req = cfg->sub_history_request;
+  // NB: if the maximum iceoryx queue size is smaller than the history depth of the reader, 
+  // we cannot fulfill the contract (we will start losing data earlier than allowed)
+  // and hence will not use shared memory
+  // TODO: need the max from iceoryx here (available in the API of master but not in release_1.0)
+  //       constant added for clarity
+  const int32_t max_sub_queue_size = 256;
```
MatthiasKillat, line 493: comment that we can get it in 2.0

src/core/ddsc/src/dds_reader.c
```diff
@@ -480,17 +480,26 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
...
-    DDS_DURABILITY_VOLATILE == qos->durability.kind &&
-    DDS_HISTORY_KEEP_LAST == qos->history.kind &&
-    (int)sub_history_req >= (int)qos->history.depth);
+          DDS_READER_QOS_CHECK_FIELDS ==
+              (qos->present & DDS_READER_QOS_CHECK_FIELDS) &&
+          DDS_LIVELINESS_AUTOMATIC == qos->liveliness.kind &&
+          DDS_INFINITY == qos->deadline.deadline &&
+          DDS_DURABILITY_VOLATILE == qos->durability.kind &&
+          DDS_HISTORY_KEEP_LAST == qos->history.kind &&          
+          max_sub_queue_size >= qos->history.depth);
```
MatthiasKillat, line 502: should not check here, but saturate to max at the writer

src/core/ddsc/src/dds_reader.c
```diff
@@ -660,13 +669,24 @@ static dds_entity_t dds_create_reader_int (dds_entity_t participant_or_subscribe
...
+    
+    // NB: Conversion is ok as history.depth is > 0
+    // If the value is larger than the max supported by iceoryx this would saturate in the iox_pub_init,
+    // but currently we do not use iceoryx in this case.
+
+    // NB: We may lose data after history.depth many samples are received (if we are not taking them fast enough from iceoryx and move them in the reader history cache), but this is valid behavior
+    // It may still lead to undesired behavior as the queues are filled very fast 
+    // if data is published as fast as possible
+    // Essentially this is another level of caching/buffering data.
+    opts.queueCapacity = (uint64_t)rqos->history.depth;
```
MatthiasKillat, line 685: min(max_possible, depth) (depth)
depends on reliability as well when we should driop data

for keyed topics this may no be enough to guarantee delivery up to a certain history depth but this cannot be solved with the iceoryx queue model. 

Assume we have a topic with two keys, to guarantee delivery of up to n elements for each key it does not suffice to have a queue of size 2n since we cannot control that iceoryx pushes samples into the queue for each key fairly/evenly.

MatthiasKillat:
@eboasson I applied the changes as discussed. For the upcoming support for transient local we will have to reconsider how to use the iceoryx parameters internally (historyCapacity etc.). But as it is the configuration options are gone (but still allowed to be in a configuration) and we should have no regression compared to the default config settings before.

eboasson (approved):
Thanks @MatthiasKillat — I must have overlooked a mail from GitHub, sorry about that!

MatthiasKillat:
No worries, will continue on enabling other transient local QoS but need to sort something out in iceoryx in parallel. It is good to have this to base the functionality on.

PR #1102, opened 2022-01-11
Must-understand annotation for struct members

dennis-adlink:
This adds support for the must-understand annotation in the IDL C generator and cdrstream serializer. Must-understand is only supported on members of structs with extensibility mutable, because the XCDR2 data representations for final and appendable types do not provide a way to indicate that a member is must-understand in the CDR data.

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -2317,7 +2317,7 @@ static const uint32_t *stream_normalize_pl (char * __restrict data, uint32_t * _
     if (!read_and_normalize_uint32 (&em_hdr, data, off, size1, bswap))
       return NULL;
     uint32_t lc = EMHEADER_LENGTH_CODE (em_hdr), m_id = EMHEADER_MEMBERID (em_hdr), msz;
-    bool must_understand = (em_hdr & EMHEADER_FLAG_MASK) & EMHEADER_FLAG_MUSTUNDERSTAND;
+    bool must_understand = em_hdr & EMHEADER_FLAG_MUSTUNDERSTAND;
     switch (lc)
     {
       case LENGTH_CODE_1B: case LENGTH_CODE_2B: case LENGTH_CODE_4B: case LENGTH_CODE_8B:
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -2317,6 +2317,7 @@ static const uint32_t *stream_normalize_pl (char * __restrict data, uint32_t * _
     if (!read_and_normalize_uint32 (&em_hdr, data, off, size1, bswap))
       return NULL;
     uint32_t lc = EMHEADER_LENGTH_CODE (em_hdr), m_id = EMHEADER_MEMBERID (em_hdr), msz;
+    bool must_understand = (em_hdr & EMHEADER_FLAG_MASK) & EMHEADER_FLAG_MUSTUNDERSTAND;
     switch (lc)
     {
       case LENGTH_CODE_1B: case LENGTH_CODE_2B: case LENGTH_CODE_4B: case LENGTH_CODE_8B:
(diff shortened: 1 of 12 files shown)


src/tools/idlc/src/descriptor.c
```diff
@@ -1548,42 +1553,53 @@ static int print_opcode(FILE *fp, const struct instruction *inst)
...
-    case DDS_OP_VAL_SEQ: vec[len++] = " | DDS_OP_TYPE_SEQ"; break;
-    case DDS_OP_VAL_ARR: vec[len++] = " | DDS_OP_TYPE_ARR"; break;
-    case DDS_OP_VAL_UNI: vec[len++] = " | DDS_OP_TYPE_UNI"; break;
-    case DDS_OP_VAL_STU: vec[len++] = " | DDS_OP_TYPE_STU"; break;
-    case DDS_OP_VAL_ENU: vec[len++] = " | DDS_OP_TYPE_ENU"; break;
-    case DDS_OP_VAL_EXT: vec[len++] = " | DDS_OP_TYPE_EXT"; break;
+  if (opcode == DDS_OP_ADR || opcode == DDS_OP_JEQ4) {
+    if (inst->data.opcode.code & DDS_OP_FLAG_EXT)
+      vec[len++] = " | DDS_OP_FLAG_EXT";
+    switch (DDS_OP_TYPE(inst->data.opcode.code)) {
```
eboasson, line 1559: I may be mistaken here, but shouldn't this be done for `DDS_OP_JEQ` as well?

If am *not* mistaken, `DDS_OP_JEQ` is never generated by IDLC and merely exists for backwards compatibility. Assuming that's indeed the case, then I think it would be best to eliminate (or `assert(0)`) the `DDS_OP_JEQ` cases in IDLC.

src/tools/idlc/src/descriptor.c
```diff
@@ -1548,42 +1553,53 @@ static int print_opcode(FILE *fp, const struct instruction *inst)
...
+    /* Must-understand flag for mutable types set on PLM instruction */
+    if (DDS_PLM_FLAGS(inst->data.opcode.code) & DDS_OP_FLAG_MU)
+      vec[len++] = " | (DDS_OP_FLAG_MU << 16)";
+  }
 
   if (opcode == DDS_OP_JEQ || opcode == DDS_OP_JEQ4 || opcode == DDS_OP_PLM) {
     /* lower 16 bits contain an offset */
     idl_snprintf(buf, sizeof(buf), " | %u", (uint16_t) DDS_OP_JUMP (inst->data.opcode.code));
     vec[len++] = buf;
   } else {
```
eboasson, line 1599: If this is turned into an `else if (opcode == DDS_OP_ADR)` neither the assert below nor the `goto print`s in the first switch statement of the function appear to be needed any longer. It is a matter of taste whether removing the `goto print`s would improve things, but I suspect it would, simply by eliminating the question: "why are these treated differently?".

eboasson (commented):
In what it actually does, it looks good to me, but I did notice a few things that I think should get a second look because of they triggered some investigation during review that I think could have been avoided.

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -2317,6 +2317,7 @@ static const uint32_t *stream_normalize_pl (char * __restrict data, uint32_t * _
     if (!read_and_normalize_uint32 (&em_hdr, data, off, size1, bswap))
       return NULL;
     uint32_t lc = EMHEADER_LENGTH_CODE (em_hdr), m_id = EMHEADER_MEMBERID (em_hdr), msz;
+    bool must_understand = (em_hdr & EMHEADER_FLAG_MASK) & EMHEADER_FLAG_MUSTUNDERSTAND;
```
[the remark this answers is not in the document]
dennis-adlink: Fixed in c741607

src/tools/idlc/src/descriptor.c
dennis-adlink: You're right, `DDS_OP_JEQ` is not generated and still exists for backwards compatibility with topic descriptors compiled by earlier versions of IDLC. I've removed the code for JEQ from descriptor.c and added a remark in dds_opcodes.h that the instruction is deprecated.

src/tools/idlc/src/descriptor.c
dennis-adlink: Good point, the `goto` also bothered me. I've refactored this a bit in c741607

eboasson (approved):
LGTM, thanks @dennis-adlink !

PR #1106, opened 2022-01-17
Fix issues exposed by linking to C++ binding

reicheratwork:
- Changed usage of the reserved C++ keyword "template" as a function
  parameter to "keyobject"
- Removed unnessary forward declaration of struct struct ddsi_typeid_t
  which caused conflicts with the "real" typedef


eboasson (approved):
Thanks @reicheratwork !

PR #1107, opened 2022-01-17
Fix rexmit duplicate crash

eboasson:
The retransmit queue tracks messages by writer GUID, sequence number and
fragment number. If a duplicate is presented for insertion in the queue,
it first tries to merge it, enqueueing it only if that fails.  The cases
where it fails are exceedingly rare, but if it does, it would end up
trying to insert a second entry with the same key in an AVL tree that
doesn't allow duplicates, leading to a crash (an assertion in a debug
build, a null pointer dereference in a release one).

This commit marks a duplicate message such that it is not entered in the
table at all.


dennis-adlink (approved):
LGTM

PR #1108, opened 2022-01-17
Set must-understand flag on key members

dennis-adlink:
This adds setting the must-understand (MU) flag on key members in the cdrstream serializer instructions and in the type object that is generated for aggregated types (so that it conforms to the XTypes spec section 7.2.2.4.4.4.6 and 7.2.2.4.4.4.8). Some refactoring is also included, to move the MU flag in the serializer instructions from the mutable member's PLM instruction to its ADR instruction, to keep the KEY and MU flags together. Commit b3273e35373e4ad510d5dd9642df54074c362cd7 is included to fix a few issues found by Coverity in #1102. 

In the python fuzzer tests, the lines for setting the MU flags in type object for keys/discriminators are currently commented out. @thijsmie could you open a PR for the Python binding to enable these lines?


eboasson (approved):
LGTM!

PR #1109, opened 2022-01-18
Fix failing assert in writer throttling for built-in type lookup writers

dennis-adlink:
The built-in type lookup writers are keep-all writers, because the topic is keyless (using DDS-RPC request and reply type). This means that the throttling will occur in case the WHC limits are reached. But the function `throttle_writer` asserts that the writer is not a built-in writer: throttling the type lookup writer may force the thread to go asleep, and because these requests are done during QoS matching, this can cause other problems. This PR fixes this issue by setting the WHC watermarks to a high value, so that no throttling will occur.


eboasson (approved):
Thanks @dennis-adlink !

PR #1111, opened 2022-01-19
Fix get_typelookup_writer racing new_participant

eboasson:
A new (local) participant is made visible in the entity index before
creating the built-in writers, to avoid complications in the matching.
This is generally fine, they don't get used until later on anyway,
except in those cases where one tries to use a built-in writer by simply
scanning all the participants.  And this is what the
get_typelookup_writer does.

The result is that calling get_builtin_writer may find that the
requested writer doesn't exist (yet, in this case), causing it to fall
back to the "privileged participant" without that "privileged
participant" being set for the participant.  This causes an assertion
failure.

This commit replaces a boolean value tracking the progress of deleting
the participant to a enum that allows one to mark also check progress in
initialization.  Any participant not yet ready is then simply skipped,
avoiding the race.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_entity.h
+++ b/src/core/ddsi/include/dds/ddsi/q_entity.h
@@ -242,6 +242,13 @@ struct avail_entityid_set {
   struct inverse_uint32_set x;
 };
 
+enum participant_state {
+  PARTICIPANT_STATE_INITIALIZING,
+  PARTICIPANT_STATE_OPERATIONAL,
+  PARTICIPANT_STATE_DELETE_STARTED,
+  PARTICIPANT_STATE_DELETING_BUILTINS
+};
+
 struct participant
 {
   struct entity_common e;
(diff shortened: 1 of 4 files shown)


dennis-adlink (approved):
LGTM, only thing I noticed is that in `insert_sample_in_whc` the mutex `pp->refc_lock` is not held when checking the value of `pp->state` in the assert. But the writer lock is held, so the pp can't be deleted at that point and `pp->state` cannot change?

eboasson:
> LGTM, only thing I noticed is that in `insert_sample_in_whc` the mutex `pp->refc_lock` is not held when checking the value of `pp->state` in the assert. 

Oops.

> But the writer lock is held, so the pp can't be deleted at that point and `pp->state` cannot change?

I think so, but I ought to double-check that. That makes it an old bug and I don't want to introduce new problems by hastily adding a lock/unlock pair for an assertion ... So I'll merge this one as is. Any further fixes on that front can be done in a separate PR.

PR #1114, opened 2022-01-20
Fix windows docs build

thijsmie:
Sometimes tools are misled by windows path conventions, and who can blame them, because it confuses me too sometimes. In any case: the windows docs build was messing up because cmake inserts a UNIX style path into the python config, that is then compared to the build directory, which then concludes that C:/something/something/subdir is not a subdirectory of C:\something\something.

I could go through the trouble of trying to make CMAKE output a windows native path on windows, but I decided not to bother and lean on the excellent python pathlib to solve this for me, by converting to a python Path instance and then back to string.

### DIFF
--- a/docs/manual/conf.py
+++ b/docs/manual/conf.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


eboasson (approved):
Thanks!

PR #1115, opened 2022-01-20
Bounded sequences

dennis-adlink:
This PR adds the implementation of bounded sequences in IDLC and the cdrstream serializer. A new serializer instruction `BSQ` is introduced, that has a bound parameter and is handled similar to the existing `SEQ` instruction.

Serialization of a sample that has a sequence-type member containing more elements than the specified bound will fail. For incoming data, deserialization (`dds_stream_normalize`) also fails in case a sequence has an element count that is larger than the bound set in the type; the sample will be dropped in that case.


dennis-adlink:
Yes, we'll definitely implement this option, and the other options from the type consistency enforcement QoS policy. We decided not to include it (except for _force_type_validation_) in the first set of PR's for XTypes support, because we assumed it not to be used widely, but you can expect a PR with support for the other options soonish.

PR #1116, opened 2022-01-20
Consistency of type consistency QoS serialization

eboasson:
This brings the serialization of the type consistency QoS in line with
the deserialization of it, which, I believe is consistent with the
XTypes 1.3 spec:
```
  @extensibility(APPENDABLE) @nested
  struct TypeConsistencyEnforcementQosPolicy {
    TypeConsistencyKind kind;
    boolean ignore_sequence_bounds;
    boolean ignore_string_bounds;
    boolean ignore_member_names;
    boolean prevent_type_widening;
    boolean force_type_validation;
  };
```

### DIFF
--- a/src/core/ddsi/src/ddsi_plist.c
+++ b/src/core/ddsi/src/ddsi_plist.c
@@ -578,11 +578,11 @@ static dds_return_t ser_type_consistency (struct nn_xmsg *xmsg, nn_parameterid_t
   const uint16_t kind = ddsrt_toBO2u (bo, (uint16_t) x->kind);
   memcpy (p, &kind, 2);
   size_t offs = sizeof (kind);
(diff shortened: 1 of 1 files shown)


dennis-adlink (approved):
Thanks for fixing this! When writing the `ser` code I probably copy-pasted a few lines from `deser` without checking the array index. We could consider moving line 553 to after the if-else block, that would improve the readability of the `deser` code a bit?

eboasson:
> We could consider moving line 553 to after the if-else block, that would improve the readability of the deser code a bit?

It might help, but I am not sure that it really would make much of a difference. Perhaps my judgement is clouded by my trying to clear a bit of a backlog ... Oh well, we can always rewrite it for clarity if we find it really makes a difference!

PR #1118, opened 2022-01-24
Fix typo

asd142513:


eboasson:
Thanks for making the effort of creating a PR but I do wonder: why bother for a single innocent typo in an internal comment that doesn't affect interpretation? There is no way we can hope to fix all such comments one PR at the time.

It'd be different if it fixed loads of typos, or if it fixed a typo in function documentation that really affects the interpretation.

I'm tempted to simply close this as "won't fix" but there's always the possibility that I am overlooking some consideration. I'd therefore prefer to give you a chance to explain the reasoning behind this PR (or indeed for anyone else comment).

asd142513:
I'm sorry if this PR bothered you, I just found the typo and fixed it, nothing else.

PR #1120, opened 2022-01-25
Fix alignment for reading collections with 64 bits element type

dennis-adlink:
This fixes an issue in the cdrstream deserializer, the alignment of elements of collections with 64 bits element type. When reading data from a XCDR2 stream containing a sequence or array that has a 64 bits element type, the max-alignment for xcdr2 (4 bytes) was not taken into account.

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -240,7 +240,7 @@ static uint32_t dds_is_peek4 (dds_istream_t * __restrict s)
 
 static uint64_t dds_is_get8 (dds_istream_t * __restrict s)
 {
-  dds_cdr_alignto (s, s->m_xcdr_version == CDR_ENC_VERSION_2 ? 4 : 8);
+  dds_cdr_alignto (s, xcdr_max_align (s->m_xcdr_version, 8));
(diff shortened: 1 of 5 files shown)


eboasson (commented):
Thanks @dennis-adlink, but could you do me a favour and double-check the other places as well? I just did a quick search through the sources and I suspect that `dds_stream_write_keyBO_impl` has the same problem. So there might be other still ...

Also, in most places the code now uses `xcdr_max_align`, but there are also a few remaining ones, like `dds_stream_write_arrBO` that deal with it in a slightly different fashion. I think it would make sense to consistently do it via `xcdr_max_align`.

Actually, I am tempted to suggest making all the functions performing alignment take a `struct align { uint32_t a; }` and make `xcdr_max_align` the only function returning a `struct align`. That makes it much harder to miss a case, but I am not sure it would get compiled out completely on all platforms, and it certainly would be painful if it doesn't!

dennis-adlink:
Thanks for reviewing! I checked other cases, but clearly missed that one in `ddsi_cdrstream_keys.part.c` I've fixed it and pushed a new commit. I'll do the refactoring (using `struct align`) in a new PR.

eboasson (approved):
Thanks @dennis-adlink! (And looking forward to hearing more about my musing on using the type system to reduce the risk of this happening again in the future 🙂)

PR #1125, opened 2022-01-28
Fix export of SHM_SUPPORT_IS_AVAILABLE property

k0ekk0ek:
I missed this one. Causes cyclonedds-cxx build to fail. Verified locally.


eboasson (approved):
You beat to me to it (#1126)!

PR #1127, opened 2022-01-28
Add versioncheck to tests

k0ekk0ek:
Replace versioncheck.bash by versioncheck.cmake and add it to tests executed with CTest instead. That way, should we ever need to move to a separate CI system, there's no dependency on Bash and the test is executed by default.


thijsmie:
This was superseeded by a pre-commit hook implementation.

PR #1128, opened 2022-01-28
Publish test results in Azure Pipelines

k0ekk0ek:
What it says in the title.


thijsmie:
An alternative was merged, thank you @k0ekk0ek for the original work.

PR #1131, opened 2022-01-31
Fixing build warnings in idlc xtests

dennis-adlink:
This fixes some strcpy build-warnings in the idlc xtests on Windows.


eboasson (approved):
Thanks!

PR #1132, opened 2022-01-31
Fix data representation in RTPS encapsulation identifier

dennis-adlink:
The RTPS encapsulation identifier should correspond to the encoding version and data representation of the outermost object of the type used for a topic. This commit fixes the issue that Cyclone was using an incorrect value for this identifier when writing data for a topic with an appendable or mutable top-level type.

The Cyclone deserializer uses the data representation as provided in the serializer instructions for that type, but did not check that the encapsulation identifier in the RTPS submessage indicates the same data representation. This commit also introduces that check.

### DIFF
--- a/src/core/ddsi/src/ddsi_serdata_default.c
+++ b/src/core/ddsi/src/ddsi_serdata_default.c
@@ -164,7 +164,7 @@ static void serdata_default_init(struct ddsi_serdata_default *d, const struct dd
   d->fixed = false;
 #endif
   if (xcdr_version != CDR_ENC_VERSION_UNDEF)
-    d->hdr.identifier = ddsi_sertype_get_native_encoding_identifier (xcdr_version, tp->encoding_format);
+    d->hdr.identifier = ddsi_sertype_get_native_enc_identifier (xcdr_version, tp->encoding_format);
   else
     d->hdr.identifier = 0;
   d->hdr.options = 0;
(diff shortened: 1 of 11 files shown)


src/core/ddsi/src/ddsi_serdata_default.c
```diff
@@ -333,7 +333,11 @@ static struct ddsi_serdata_default *serdata_default_from_ser_common (const struc
   const bool needs_bswap = !CDR_ENC_IS_NATIVE (d->hdr.identifier);
   d->hdr.identifier = CDR_ENC_TO_NATIVE (d->hdr.identifier);
   const uint32_t pad = ddsrt_fromBE2u (d->hdr.options) & 2;
-  const uint32_t xcdr_version = get_xcdr_version (d->hdr.identifier);
+  const uint32_t xcdr_version = ddsi_sertype_enc_id_xcdr_version (d->hdr.identifier);
+  const uint32_t encoding_format = ddsi_sertype_enc_id_enc_format (d->hdr.identifier);
+  if (encoding_format != tp->encoding_format)
+    return NULL;
```
eboasson, line 339: That leaks `d` ...

src/core/ddsi/src/ddsi_serdata_default.c
```diff
@@ -391,7 +395,11 @@ static struct ddsi_serdata_default *serdata_default_from_ser_iov_common (const s
...
```
eboasson, line 401: Leaks `d` on failure.

src/core/ddsi/src/ddsi_sertype.c
```diff
@@ -274,6 +274,22 @@ uint32_t get_xcdr_version (uint16_t cdr_identifier)
...
+  {
+    case CDR_LE: case CDR_BE:
+    case CDR2_LE: case CDR2_BE:
+      return CDR_ENC_FORMAT_PLAIN;
+    case D_CDR2_LE: case D_CDR2_BE:
+      return CDR_ENC_FORMAT_DELIMITED;
+    case PL_CDR2_LE: case PL_CDR2_BE:
+      return CDR_ENC_FORMAT_PL;
+    default:
+      abort ();
```
eboasson, line 289: I am probably being paranoid here, but I just want to confirm that `cdr_identifier` is not taken from external input and cannot ever be anything other than the covered cases.

eboasson (commented):
Looks good to me for the most part, but it needs a small fix still (plus a question because I want to confirm something).

src/core/ddsi/src/ddsi_sertype.c
dennis-adlink: Good point! There was an assert on a valid CDR encapsulation identifier, but this should have proper error handling because the value is from external input (and I think was not verified before?). I've refactored these 2 functions a bit in commit b783323

eboasson (approved):

PR #1136, opened 2022-02-01
Set OpenSSL API compatibility to 1.1.1

eboasson:
This suppresses deprecation warnings from OpenSSL 3.0. I think this is a sensible fix for the 0.8.x release branch, but if we can avoid it for master, then that would be my preference as it really is not a proper fix to just suppress deprecation warnings.


eboasson:
@clalancette with a local build with OpenSSL 3.0 this seems to do the trick. If it works for you, I'm inclined to merge this into the `releases/0.8.x` branch. For the master branch I hope we will have time to fix it properly for the 0.9 (probably) release later this quarter.

clalancette:
Confirmed that it fixes the problems for me on my Jammy container.  Just for fun, here is a run on ROS 2 CI using Jammy and this PR:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16057/)

clalancette:
CI looks good.  The other warnings are from mypy (we'll need to look into that), and similar deprecation warnings from Fast-DDS (which I'll open a separate bug about).

eboasson:
Thanks for confirming!

PR #1137, opened 2022-02-02
Data representation qos refactoring

dennis-adlink:
Some minor refactoring of the implementation of the data representation qos: in discovery data, consider an empty list of values as 'not present', so that the default (`XCDR1`) will be merged in the qos, and the qos matching code can safely assume that the data representation qos is always present and has at least 1 value.

Also add a check in `deser_data_representation` to ensure that the buffer has sufficient bytes remaining to the provided data representation value count, to avoid allocating a large amount of memory (e.g. in case of a malicious node that send a large count).


dennis-adlink (commented):
@eboasson I agree with your changes in create reader/writer for the data representation default, and thanks for fixing the compiler warning!

PR #1139, opened 2022-02-07
Fix deprecated use of idlc_generate

dpotman:
To support the deprecated use of `idlc_generate` (without using keyword based arguments), unparsed arguments should be passed to `idlc_generate_generic`


eboasson (approved):
LGTM! Thanks @dpotman!

PR #1140, opened 2022-02-07
Fix optional collections of unbounded strings

dpotman:
Fix issue in the IDL compiler C generator for optional collections of unbounded strings: only members of type unbounded string should not get the EXT flag (because that's already a pointer type), but members of type collection-of-unbounded-string should get the flag.


eboasson (approved):
Thanks @dpotman !

PR #1149, opened 2022-02-15
Extend support for enumeration type

dpotman:
Improves support for the enumeration data type in idlc and the cdrstream serializer:
- Support the @bit_bound annotation to set the CDR size of enum type
- Introduce the `ENU` opcode for the serializer, so that max-value checking for enums can be done. In case the values for enum literals are not consecutive, the serializer will not check for invalid values that are within the 0..max range. In case non-consecutive values are used, idlc will show a warning (which can be disabled by passing the `-Wno-enum-consecutive` command line option)
- Add a check in idlc so that extensibility 'mutable' cannot be set on enumerated types (enum and bitmask)
- Fix a bug in the enum literal value check wrt bit-bound in idlc
- Add tests that check the assignability rules for enums with final and appendable extensibility (and fixed some bugs found when running these tests)
- Add key extraction calls to the idlc xtests, and add a test for enum types in arrays, sequences, typedefs, etc.


src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -376,20 +388,33 @@ static void dds_os_put_bytes_aligned (dds_ostream_t * __restrict os, const void
...
-  assert (type == DDS_OP_VAL_1BY || type == DDS_OP_VAL_2BY || type == DDS_OP_VAL_4BY || type == DDS_OP_VAL_8BY || type == DDS_OP_VAL_ENU);
-  if (type == DDS_OP_VAL_ENU)
-    return 4;
+  assert (type == DDS_OP_VAL_1BY || type == DDS_OP_VAL_2BY || type == DDS_OP_VAL_4BY || type == DDS_OP_VAL_8BY);
   return (uint32_t)1 << ((uint32_t) type - 1);
 }
 
+static enum enum_cdr_sz get_enum_cdr_size (uint32_t flags)
+{
+  uint32_t flag_value = ((flags & DDS_OP_FLAG_SZ2) ? 2 : 0) + ((flags & DDS_OP_FLAG_SZ1) ? 1 : 0);
```
eboasson, line 397: I think using log2(size) (see earlier) indeed makes sense, then this becomes 1 << ((flags >> something) & 0x3) (provided there is also a static assert that the two flag bits are adjacent; alternatively solved by defining, e.g., MASK and SHIFT instead of SZ2 and SZ2)

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1961,26 +2126,65 @@ static bool normalize_primarray (char * __restrict data, uint32_t * __restrict o
...
-  else
-  {
-    for (uint32_t i = 0; i < num; i++)
-      if (xs[i] > max)
+    case ENUM_CDR_SZ_16: {
+      if ((*off = check_align_prim_many (*off, size, 1, num)) == UINT32_MAX)
+        return false;
+      uint16_t * const xs = (uint16_t *) (data + *off);
+      if (bswap)
+      {
```
eboasson, line 2149: A total nitpicky suggestion: squeeze the braces a bit: I generally like it this way and I appreciate your efforts, but here the information density becomes a bit low ... No obligation but if you happen to feel the same way ...

eboasson (commented):
It looks ok, but I think it would end up simpler if the two size bits for the enum were mapped differently. Perhaps you had a good reason for doing it this, or perhaps I am being overly optimistic on the effects of the change, so I'd be interested to know what you think.

src/core/ddsc/include/dds/ddsc/dds_opcodes.h
```diff
@@ -264,6 +266,10 @@ enum dds_stream_typecode_subtype {
 #define DDS_OP_FLAG_OPT  (1u << 5) /* optional flag, used with struct members. For non-string types,
                                       an optional member also gets the FLAG_EXT, see above. */
 
+#define DDS_OP_FLAG_SZ1  (1u << 6) /* Enum storage size -- SZ2, SZ1: 00 = 4 bytes, 10 = 2 bytes, 01 = 1 byte, 11 = unused */
```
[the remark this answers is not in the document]
dpotman: You're right, my idea was to have no size flags in the opcodes for the default enum size, but I prefer the approach you're suggesting below, that makes the serializer code a bit cleaner. So I've refactored the flags code a bit in [d0c8ce5](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1149/commits/d0c8ce5bbe632c3473da6b1b0051853e04b20b30)

src/core/ddsi/src/ddsi_cdrstream.c
dpotman: I've combined the swap and non-swap versions into a single line, which saves some lines of code and is still readable I think.

PR #1152, opened 2022-02-16
Implement interface priority

thijsmie:
This allows you to specify the priority of each interface in the config file. Some care was taken to not disturb any existing configuration out there: any old config files should always result in the same behaviour as before this change. This is not quite complete:
 - [x] Move AssumeMulticastCapable/MulticastRecvNetworkInterfaceAddresses/PreferMulticast to this and possibly deprecate the old style?
 - [x] Why an int32_t for priority? Well you could add a listen-only interface with 'priority=-10000' or something? It was motivated by allowing me to leave the `cost` calculation fully unchanged by adding `base cost = - priority`.
 - [x] Some docs
 - [x] Need to regenerate options.md, cyclonedds.rnc
 - [x] Testcase
 - [ ] This pointer redirection: `locs->locs[lidx].conn->m_interf->priority` makes me feel queasy, can we do better?


eboasson (approved):
Given that my minor suggestions passed the review and got merged in, I suppose there's now nothing left but to merge it. Thanks @thijsmie !

tonynajjar:
I didn't spend too much time looking into it but maybe this PR breaks the xml schema? 
image

thijsmie:
Hi @tonynajjar,

This PR deprecates the NetworkInterfaceAddress attribute, and deprecated elements are removed from the xml schema. You are free to continue to use it for a little bit, but it can be removed in future releases. The new style is using the `General/Interfaces` element.

tonynajjar:
Thank you for the clarifications :) Will update the config.

PR #1154, opened 2022-02-16
Add value checking for boolean type in cdrstream serializer

dpotman:
Note: this PR builds on #1149 and includes the commits from that PR. When #1149 is merged (after I've fixed the issues from the review) I'll rebase the changes for this PR. 

This adds a serializer opcode for boolean type in idlc and cdrstream, to enable checking for valid value for boolean types when reading and writing sample data and keys. 

Commit 1c45bfce8b00dcffbd4ca4cd636871494a3e9bb8 contains a fix for a bug that I found when writing a test-case for this PR. This fixes the `sizeof()` that is included in the serializer instructions for the element size of a typedef of an array of sequences.

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -67,6 +67,8 @@
 #define dds_stream_write_stringBO                     NAME_BYTE_ORDER(dds_stream_write_string)
 #define dds_stream_write_seqBO                        NAME_BYTE_ORDER(dds_stream_write_seq)
 #define dds_stream_write_arrBO                        NAME_BYTE_ORDER(dds_stream_write_arr)
+#define dds_stream_write_bool_valueBO                 NAME_BYTE_ORDER(dds_stream_write_bool_value)
+#define dds_stream_write_bool_arrBO                   NAME_BYTE_ORDER(dds_stream_write_bool_arr)
 #define dds_stream_write_enum_valueBO                 NAME_BYTE_ORDER(dds_stream_write_enum_value)
 #define dds_stream_write_enum_arrBO                   NAME_BYTE_ORDER(dds_stream_write_enum_arr)
 #define dds_stream_write_union_discriminantBO         NAME_BYTE_ORDER(dds_stream_write_union_discriminant)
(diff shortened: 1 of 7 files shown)


src/core/ddsi/src/ddsi_cdrstream_write.part.c
```diff
@@ -43,6 +51,16 @@ static void dds_stream_write_stringBO (DDS_OSTREAM_T * __restrict os, const char
     dds_os_put1BO (os, 0);
 }
 
+static bool dds_stream_write_bool_arrBO (DDS_OSTREAM_T * __restrict os, const uint8_t * __restrict addr, uint32_t num)
```
eboasson, line 54: It is a bit sad that it ends up like this when one'd really want to a `memcpy`. It might actually not be that much worse to do it like this so let's leave it for now.

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -895,10 +900,10 @@ static bool insn_key_ok_p (uint32_t insn)
 static uint32_t read_union_discriminant (dds_istream_t * __restrict is, uint32_t insn)
 {
   enum dds_stream_typecode type = DDS_OP_SUBTYPE (insn);
-  assert (type == DDS_OP_VAL_1BY || type == DDS_OP_VAL_2BY || type == DDS_OP_VAL_4BY || type == DDS_OP_VAL_ENU);
+  assert (is_primitive_type (type) || type == DDS_OP_VAL_ENU);
```
eboasson, line 903: The introduction of `is_primitive_type` is very useful, I think the code would further benefit from also adding `is_primitive_or_enum_type`.

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -1369,7 +1374,7 @@ static const uint32_t *dds_stream_read_seq (dds_istream_t * __restrict is, char
   dds_sequence_t * const seq = (dds_sequence_t *) addr;
   const enum dds_stream_typecode subtype = DDS_OP_SUBTYPE (insn);
   uint32_t bound_op = seq_is_bounded (DDS_OP_TYPE (insn)) ? 1 : 0;
-  if (subtype > DDS_OP_VAL_8BY && is->m_xcdr_version == CDR_ENC_VERSION_2)
+  if (!is_primitive_type (subtype) && is->m_xcdr_version == CDR_ENC_VERSION_2)
```
eboasson, line 1377: Like the primitive-or-enum check, this one also occurs fairly often, and it seems like it could be replaced by a `is_dheader_needed` function. If it works for at least the vast majority of the cases, it think it's worth it.

eboasson (commented):
LGTM, some of the improvements you made me think it might be a good idea to do some more of the same kind. I am ok with merging it as-is but I would like your opinion before I do that.

src/core/ddsi/src/ddsi_cdrstream.c
dpotman: Good suggestion, I've added both `is_primitive_or_enum_type` and `is_dheader_needed`

eboasson (approved):
LGTM, there is a trivial build error on some platforms (`is_primitive_or_enum_type` defined-but-not-used) that I've fixed it already. I'll push the fixed version to your PR before merging.

PR #1155, opened 2022-02-16
Turn bison into an optional dependency

thijsmie:
This allows CycloneDDS to build without bison by tracking the generated parser.c in the repository. A hash check on parser.y should prevent that files get out of sync if you forget to check in the updated parser.y without updating parser.c.


eboasson (commented):
Thanks @thijsmie 🙂 I like this, but there's been something nagging me in the back of my mind, and I finally figured out what it is: if I read this correctly, you only find out if you forgot to check in `parser.c` and `parser.h` when you re-run CMake, and only if bison isn't present. The Windows CI would pick that up with the change in CI scripting in this PR, and so hopefully would the sharp-eyed committers (🤓🤔).

Still it seems to make more sense to (1) always check that things match, also when bison is available, and (2) to make it really hard to commit only the `.y` file, which is what I suppose git commit hooks are for.

I kinda like the idea in https://stackoverflow.com/questions/3284292/can-a-git-hook-automatically-add-files-to-the-commit of auto-committing the generated files, but if it is possible to simply block committing the `.y` file without the corresponding updated `.c` and `.h` files, that'd be good enough, too. And it seems less risky. Something I don't know is whether this can be done in a portable manner — can one count on, e.g., `bash` being available for the scripting?

thijsmie:
I agree a pre-commit hook is the better solution here, and it would definitely be useful in a couple of other places (version check, options.md generation etc). The portablity of bash is questionable, combined with the fact that you then have to ask developers to 'install' the git hooks when they clone the repository. Maybe this is my chance to introduce _even more python?_ into the mix? There is this really nice [pre-commit framework](https://pre-commit.com/) that we can use to introduce some commit automation (I use it to prevent myself from making commits without JIRA ticket number on ospl).

eboasson:
So basically, the commit hooks are niceties for developers and always checking during CI is not only sensible but necessary. That's fine. And I am all for the niceties. Go for it! 😀

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    name: bison-parser-hashcheck
+    language: python
+    entry: python hooks/must-match-hash.py
+    files: 'src/idl/src/parser\.y$'
+  - id: checked-in-together
+    name: bison-parser-all-files-checked
+    language: python
+    entry: python hooks/checked-in-together.py
+    pass_filenames: false
+    args: ['src/idl/src/parser.c', 'src/idl/src/parser.h', 'src/idl/src/parser.y', 'src/idl/src/parser.y.hash']
```
eboasson, line 14: Nitpicking a bit here, but what if `parser.y` changes in a way that doesn't affect the output (e.g., fixing a comment)? I reckon that'll be a rare event, so as long as there is a good way around it, that'd be fine.

src/idl/src/parser.h
```diff
@@ -34,8 +34,8 @@
 /* Undocumented macros, especially those whose name start with YY_,
    are private implementation details.  Do not rely on them.  */
 
-#ifndef YY_IDL_YY_SRC_PARSER_H_INCLUDED
-# define YY_IDL_YY_SRC_PARSER_H_INCLUDED
+#ifndef YY_IDL_YY_HOME_TMIEDEMA_PROJECTS_WORKSPACE_CYCLONEDDS_CYCLONEDDS_SRC_IDL_SRC_PARSER_H_INCLUDED
```
eboasson, line 37: Just checking you don't mind your local paths to be included

eboasson (commented):
LGTM, just two questions ...

src/idl/src/parser.h
thijsmie: I've switched the bison target to a relative path which make sure the path is not visible any longer.

.pre-commit-config.yaml
thijsmie: Solved by appending the parser.y hash to parser.{c,h} in a comment at cmake configure time.

PR #1156, opened 2022-02-16
Annotations

reicheratwork:
Add support for the following annotations in the IDL parser:
min
max
range
unit
try_construct
data_representation
And also added these to the xtypes discovery information
Also added unittests for all

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_typewrap.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_typewrap.h
@@ -29,6 +29,11 @@ extern "C" {
 
 #define XT_FLAG_EXTENSIBILITY_MASK  0x7
 
+#define DDS_XTypes_TRY_CONSTRUCT_INVALID 0
+#define DDS_XTypes_TRY_CONSTRUCT_DISCARD DDS_XTypes_TRY_CONSTRUCT1
+#define DDS_XTypes_TRY_CONSTRUCT_USE_DEFAULT DDS_XTypes_TRY_CONSTRUCT2
+#define DDS_XTypes_TRY_CONSTRUCT_TRIM (DDS_XTypes_TRY_CONSTRUCT1 | DDS_XTypes_TRY_CONSTRUCT2)
+
 struct xt_type;
 
 typedef struct ddsi_typeid ddsi_typeid_t;
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -302,6 +302,8 @@ struct idl_member {
   IDL_ANNOTATABLE(bool) external;
   IDL_ANNOTATABLE(bool) must_understand;
   IDL_ANNOTATABLE(idl_try_construct_t) try_construct;
+  IDL_ANNOTATABLE(const idl_literal_t*) min;
+  IDL_ANNOTATABLE(const idl_literal_t*) max;
   IDL_ANNOTATABLE(const idl_literal_t*) value;
 };
 
@@ -375,6 +377,8 @@ struct idl_case {
   idl_declarator_t *declarator;
   IDL_ANNOTATABLE(bool) external;
   IDL_ANNOTATABLE(idl_try_construct_t) try_construct;
+  IDL_ANNOTATABLE(const idl_literal_t*) min;
+  IDL_ANNOTATABLE(const idl_literal_t*) max;
 };
 
 typedef struct idl_switch_type_spec idl_switch_type_spec_t;
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -200,6 +200,13 @@ enum idl_try_construct {
   IDL_TRIM
 };
 
+typedef uint32_t allowable_data_representations_t;
+typedef enum {
+  IDL_DATAREPRESENTATION_FLAG_XCDR1 = 0x1 << 0,
+  IDL_DATAREPRESENTATION_FLAG_XML   = 0x1 << 1,
+  IDL_DATAREPRESENTATION_FLAG_XCDR2 = 0x1 << 2
+} idl_data_representation_flags_t;
+
 /* most types have convenience members for information that is shared between
    generators or makes sense to calculate in advance. e.g. the field
    identifier for struct members, which can be assigned through @id, @hashid,
@@ -280,6 +287,7 @@ struct idl_module {
   /* metadata */
   IDL_ANNOTATABLE(bool) default_nested;
   IDL_ANNOTATABLE(idl_autoid_t) autoid;
+  IDL_ANNOTATABLE(allowable_data_representations_t) data_representation;
 };
 
 typedef struct idl_declarator idl_declarator_t;
@@ -351,6 +359,7 @@ struct idl_struct {
      single boolean */
   IDL_ANNOTATABLE(bool) nested;
   IDL_ANNOTATABLE(idl_extensibility_t) extensibility;
+  IDL_ANNOTATABLE(allowable_data_representations_t) data_representation;
 };
 
 typedef struct idl_forward idl_forward_t;
@@ -407,6 +416,7 @@ struct idl_union {
   IDL_ANNOTATABLE(bool) nested; /**< if type is nested or a topic */
   IDL_ANNOTATABLE(idl_extensibility_t) extensibility;
   IDL_ANNOTATABLE(idl_autoid_t) autoid;
+  IDL_ANNOTATABLE(allowable_data_representations_t) data_representation;
 };
 
 typedef struct idl_enumerator idl_enumerator_t;
@@ -552,6 +562,8 @@ IDL_EXPORT const idl_name_t *idl_name(const void *node);
 IDL_EXPORT uint32_t idl_array_size(const void *node);
 IDL_EXPORT uint32_t idl_bound(const void *node);
 IDL_EXPORT const idl_literal_t *idl_default_value(const void *node);
+IDL_EXPORT bool idl_requires_xtypes_functionality(const void *node);
+IDL_EXPORT allowable_data_representations_t supported_data_representations(const void *node);
 IDL_EXPORT uint32_t idl_enum_max_value(const void *node);
 
 /* navigation */
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -193,6 +193,13 @@ enum idl_extensibility {
   IDL_MUTABLE
 };
 
+typedef enum idl_try_construct idl_try_construct_t;
+enum idl_try_construct {
+  IDL_DISCARD,
+  IDL_USE_DEFAULT,
+  IDL_TRIM
+};
+
 /* most types have convenience members for information that is shared between
    generators or makes sense to calculate in advance. e.g. the field
    identifier for struct members, which can be assigned through @id, @hashid,
@@ -294,6 +301,7 @@ struct idl_member {
   IDL_ANNOTATABLE(bool) optional;
   IDL_ANNOTATABLE(bool) external;
   IDL_ANNOTATABLE(bool) must_understand;
+  IDL_ANNOTATABLE(idl_try_construct_t) try_construct;
   IDL_ANNOTATABLE(const idl_literal_t*) value;
 };
 
@@ -366,6 +374,7 @@ struct idl_case {
   idl_type_spec_t *type_spec;
   idl_declarator_t *declarator;
   IDL_ANNOTATABLE(bool) external;
+  IDL_ANNOTATABLE(idl_try_construct_t) try_construct;
 };
 
 typedef struct idl_switch_type_spec idl_switch_type_spec_t;
--- a/src/idl/include/idl/tree.h
+++ b/src/idl/include/idl/tree.h
@@ -305,6 +305,7 @@ struct idl_member {
   IDL_ANNOTATABLE(const idl_literal_t*) min;
   IDL_ANNOTATABLE(const idl_literal_t*) max;
   IDL_ANNOTATABLE(const idl_literal_t*) value;
+  IDL_ANNOTATABLE(const char *) unit;
 };
 
 /* types can inherit from and extend other types (interfaces, values and
@@ -379,6 +380,7 @@ struct idl_case {
   IDL_ANNOTATABLE(idl_try_construct_t) try_construct;
   IDL_ANNOTATABLE(const idl_literal_t*) min;
   IDL_ANNOTATABLE(const idl_literal_t*) max;
+  IDL_ANNOTATABLE(const char *) unit;
 };
 
 typedef struct idl_switch_type_spec idl_switch_type_spec_t;
--- a/src/idl/src/annotation.c
+++ b/src/idl/src/annotation.c
@@ -510,6 +510,230 @@ annotate_must_understand(
   return IDL_RETCODE_OK;
 }
 
+static idl_floatval_t idl_arithmetic_to_double(const idl_literal_t *lit) {
+  if (idl_type(lit) & IDL_INTEGER_TYPE) {
+    idl_intval_t val = idl_intval(lit);
+    if (IDL_UNSIGNED & val.type)
+      return (idl_floatval_t)val.value.ullng;
+    else
+      return (idl_floatval_t)val.value.llng;
+  } else if (idl_type(lit) & IDL_FLOATING_PT_TYPE) {
+    return idl_floatval(lit);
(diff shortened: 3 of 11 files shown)


dpotman (commented):
Generally looking ok, but a few things need a fix. In addition to the comments above, a few  remarks:
- Using this IDL fragment results in an error message: `@data_representation(XCDR2) struct a { long f1; };`. Something goes wrong when parsing the bitmask-type annotation parameter?
- `@try_construct` and `@min/max/range` on a typedef'ed type does not work
- Since we haven't implemented `@data_representation` in the C (and C++?) back-end, I think it would be nice to have idlc showing a warning in case this annotation is used, to inform the user that it currently won't have any effect. The same goes for `@try_construct` as long as that is not implemented. 
- The application of a `@data_representation` annotation on a type should be stored in the type's type object (I think it should go in `CompleteStructType.CompleteStructHeader.CompleteTypeDetail.AppliedAnnotationSeq`). But this is a more general problem, some other annotations (auto_id, default, hash_id) are also not stored yet. So we can also decide to make an issue for this and don't fix it as part of this PR.

src/idl/include/idl/tree.h
```diff
@@ -193,6 +193,20 @@ enum idl_extensibility {
...
 };
 
+typedef enum idl_try_construct idl_try_construct_t;
+enum idl_try_construct {
+  IDL_DISCARD,
+  IDL_USE_DEFAULT,
+  IDL_TRIM
+};
+
+typedef uint64_t allowable_data_representations_t;
```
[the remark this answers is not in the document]
reicheratwork: Good catch!

src/idl/include/idl/tree.h
```diff
@@ -193,6 +193,20 @@ enum idl_extensibility {
...
+typedef enum {
+  IDL_DATAREPRESENTATION_FLAG_XCDR1 = 0x1 << 0,
+  IDL_DATAREPRESENTATION_FLAG_XML   = 0x1 << 1,
+  IDL_DATAREPRESENTATION_FLAG_XCDR2 = 0x1 << 2
+} idl_data_representation_flags_t;
```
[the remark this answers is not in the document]
reicheratwork: This links to your 3rd point in your main comment: these flags are the C equivalents of the bitmask positions in the @data_representation annotation.
As no current backend implementation currently implements support for this annotation, they are not used anywhere (yet). Though I am working on support for data representation in C++, which needs to wait on this PR to be merged before I can make that into its own PR.
Maybe think of it as a "forward declaration" ?

src/core/ddsi/include/dds/ddsi/ddsi_xt_typeinfo.h
```diff
@@ -147,6 +147,10 @@ typedef struct DDS_XTypes_TypeObjectHashId
 typedef uint16_t DDS_XTypes_MemberFlag;
 #define DDS_XTypes_TRY_CONSTRUCT1 (1 << 0)
 #define DDS_XTypes_TRY_CONSTRUCT2 (1 << 1)
+#define DDS_XTypes_TRY_CONSTRUCT_INVALID 0
+#define DDS_XTypes_TRY_CONSTRUCT_DISCARD DDS_XTypes_TRY_CONSTRUCT1
+#define DDS_XTypes_TRY_CONSTRUCT_USE_DEFAULT DDS_XTypes_TRY_CONSTRUCT2
+#define DDS_XTypes_TRY_CONSTRUCT_TRIM DDS_XTypes_TRY_CONSTRUCT1 | DDS_XTypes_TRY_CONSTRUCT2
```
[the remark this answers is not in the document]
reicheratwork: Okay.

src/tools/idlc/src/descriptor_type_meta.c
```diff
@@ -577,48 +577,157 @@ get_type_spec_typeids(
...
 }
 
+static idl_retcode_t
+set_xtypes_annotation_parameter_value(
+  struct DDS_XTypes_AnnotationParameterValue *val,
+  const idl_literal_t *lit)
+{
+  switch (idl_type(lit)) {
+    case IDL_BOOL:
+      val->_d = 0;
```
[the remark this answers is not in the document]
reicheratwork: I think you are mixing up two things here, we are switching over the type of an idl_literal_t, so the cases switched over will be the IDL types, not the DDS_XTypes, which have different values.

src/idl/src/annotation.c
```diff
@@ -715,6 +715,79 @@ annotate_position(
...
+    /*USE_DEFAULT is the annotation default, NOT the default
+      try_construct action in absence of an annotation, that
+      is DISCARD*/
+    *try_construct_p = IDL_USE_DEFAULT;
+  } else if (idl_strcasecmp(str, "discard") == 0) {
+    *try_construct_p = IDL_DISCARD;
+  } else if (idl_strcasecmp(str, "trim") == 0) {
+    /*TRIM can only be used on bounded sequences/strings*/
+    if (idl_is_string(ts)) {
+      if (0 == ((const idl_string_t*)ts)->maximum) {
```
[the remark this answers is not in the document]
reicheratwork: Will do.

src/tools/idlc/src/descriptor_type_meta.c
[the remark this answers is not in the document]
dpotman: Yes, the `IDL_BOOL` is correct, but my comment is about `val->_d = 0`, this `0` should be `TK_BOOLEAN`

src/idl/src/annotation.c
```diff
@@ -502,6 +502,236 @@ annotate_must_understand(
...
+
+  const idl_literal_t *lit = (const idl_literal_t *)const_expr;
+  if (param_type == IDL_OCTET || (param_type & IDL_INTEGER_TYPE)) {
+    //check whether range of value falls inside type
+    idl_intval_t intval = idl_intval(lit);
+    bool is_unsigned = intval.type & IDL_UNSIGNED;
+    uint64_t uval = intval.value.ullng;
+    int64_t sval = intval.value.llng;
+
+    if (!is_unsigned && sval < 0 && field_type & IDL_UNSIGNED) {
```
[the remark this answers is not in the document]
reicheratwork: This is only done on integer/octet types, due to the check on line 557.

src/tools/idlc/src/descriptor_type_meta.c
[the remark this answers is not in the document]
reicheratwork: Ah, right!

Seems the mix up went both ways 👍

src/idl/src/annotation.c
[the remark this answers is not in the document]
dpotman: The check on 557 is for `param_type`, not `field_type`

src/idl/src/annotation.c
```diff
@@ -502,6 +502,236 @@ annotate_must_understand(
...
+                      *max_lit = NULL;
+  if (idl_is_member(node)) {
+    min_lit = ((idl_member_t*)node)->min.value;
+    max_lit = ((idl_member_t*)node)->max.value;
+  } else if (idl_is_case(node)) {
+    min_lit = ((idl_case_t*)node)->min.value;
+    max_lit = ((idl_case_t*)node)->max.value;
+  }
+
+  if (idl_arithmetic_to_double(min_lit) > idl_arithmetic_to_double(max_lit)) {
```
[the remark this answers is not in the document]
reicheratwork: Will do.

src/idl/src/annotation.c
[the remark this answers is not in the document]
reicheratwork: It will now also check whether the field type is not a floating point type.

src/idl/src/expression.c
```diff
@@ -908,6 +908,11 @@ compare_int(const idl_const_expr_t *lhs, const idl_const_expr_t *rhs)
   }
 }
 
+idl_floatval_t
+idl_floatval(const idl_const_expr_t *const_expr) {
```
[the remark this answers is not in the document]
reicheratwork: This was made just make the const-expr -> floating point conversion function follow the same pattern as the int conversion function.
The intval function is used in multiple places so refactoring it out is not really desirable I think.
The floatval function is used in just a single place, so I will just implement it inline.

src/core/ddsi/include/dds/ddsi/ddsi_typewrap.h
```diff
@@ -29,6 +29,11 @@ extern "C" {
 
 #define XT_FLAG_EXTENSIBILITY_MASK  0x7
 
+#define DDS_XTypes_TRY_CONSTRUCT_INVALID 0
+#define DDS_XTypes_TRY_CONSTRUCT_DISCARD DDS_XTypes_TRY_CONSTRUCT1
+#define DDS_XTypes_TRY_CONSTRUCT_USE_DEFAULT DDS_XTypes_TRY_CONSTRUCT2
+#define DDS_XTypes_TRY_CONSTRUCT_TRIM DDS_XTypes_TRY_CONSTRUCT1 | DDS_XTypes_TRY_CONSTRUCT2
```
eboasson, line 35: Defining the macro like this is really asking for trouble ...

```suggestion
#define DDS_XTypes_TRY_CONSTRUCT_TRIM (DDS_XTypes_TRY_CONSTRUCT1 | DDS_XTypes_TRY_CONSTRUCT2)
```

src/idl/src/annotation.c
```diff
@@ -510,6 +510,299 @@ annotate_must_understand(
...
+    idl_error(pstate, idl_location(param),
+      "attempting to overwrite existing limit value");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
+  if (!(field_type & IDL_INTEGER_TYPE) && !(field_type & IDL_FLOATING_PT_TYPE) && field_type != IDL_OCTET) {
+    idl_error(pstate, idl_location(param),
+      "@max/@min/@range field type mismatch: attempt to set limits on non-arithmetic field");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  } else if (param_type & IDL_FLOATING_PT_TYPE && field_type & IDL_INTEGER_TYPE) {
```
eboasson, line 573: I personally prefer parenthesising bitwise operators when used as operands to boolean operators, but this isn't incorrect. I just couldn't help noticing that in line 580 this does get done ...

```suggestion
  } else if ((param_type & IDL_FLOATING_PT_TYPE) && (field_type & IDL_INTEGER_TYPE)) {
```

src/idl/src/annotation.c
```diff
@@ -510,6 +510,299 @@ annotate_must_understand(
...
+
+  const idl_literal_t *lit = (const idl_literal_t *)const_expr;
+  if (param_type == IDL_OCTET || (param_type & IDL_INTEGER_TYPE)) {
+    //check whether range of value falls inside type
+    idl_intval_t intval = idl_intval(lit);
+    bool is_unsigned = intval.type & IDL_UNSIGNED;
+    uint64_t uval = intval.value.ullng;
+    int64_t sval = intval.value.llng;
+
+    if (!is_unsigned && sval < 0 && field_type & IDL_UNSIGNED && (field_type & IDL_FLOATING_PT_TYPE) == 0) {
```
eboasson, line 587: ... and while we're at it, for consistency with the other places (in this PR anyway 🙂) it seems more logical to use a logical-not rather than equals-to-zero

```suggestion
    if (!is_unsigned && sval < 0 && (field_type & IDL_UNSIGNED) && !(field_type & IDL_FLOATING_PT_TYPE)) {
```

(and yes, this is nitpicking; I wouldn't have done it if this were all, but the non-parenthesised `DDS_XTypes_TRY_CONSTRUCT_TRIM` means I have to do another round anyway)

src/idl/src/annotation.c
```diff
@@ -510,6 +510,299 @@ annotate_must_understand(
...
+      ((idl_case_t*)node)->max.value = lit;
+    } else {
+      ((idl_case_t*)node)->min.annotation = parent;
+      ((idl_case_t*)node)->min.value = lit;
+    }
+    min_lit = ((idl_case_t*)node)->min.value;
+    max_lit = ((idl_case_t*)node)->max.value;
+  }
+
+  if (min_lit && max_lit && idl_arithmetic_to_double(min_lit) > idl_arithmetic_to_double(max_lit)) {
```
eboasson, line 686: All the checking is done with such great care (much appreciated), but then unfortunately there is this ... double-precision floating-point has a mantissa of 53 bits and so you get loss of precision when converting 64-bit integers of magnitude > 2^53 (or thereabouts ...) to floating-point.

Try it with 2^64-1 and 2^64-2, for example: they convert to the same floating-point number.

```C
#include <stdint.h>
#include <stdio.h>
#include <inttypes.h>
int main(int argc,char **argv){
  uint64_t a, b;
  sscanf(argv[1],"%"SCNu64,&a);
  sscanf(argv[2],"%"SCNu64,&b);
  printf("%a > %a = %d\n",(double)a,(double)b,(double)a>(double)b);
  return 0;
}
```

src/idl/src/annotation.c
```diff
@@ -510,6 +510,299 @@ annotate_must_understand(
...
+    ((idl_member_t*)node)->unit.value = name;
+  } else if (idl_is_case(node)) {
+    ((idl_case_t*)node)->unit.annotation = annotation_appl;
+    ((idl_case_t*)node)->unit.value = name;
+  } else {
+    idl_error(pstate, idl_location(annotation_appl),
+      "@unit can only be assigned to members and cases");
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
```
eboasson, line 762: If the min/max annotations need to check for the field having been annotated before already, shouldn't this one do so, too?

src/idl/src/annotation.c
```diff
@@ -510,6 +510,299 @@ annotate_must_understand(
...
+  if (idl_requires_xtypes_functionality(node))
+    val &= ~((allowable_data_representations_t)IDL_DATAREPRESENTATION_FLAG_XCDR1);
+
+  if (0 == val) {
+    idl_error(pstate, idl_location(annotation_appl),
+      "no viable datarepresentations for type %s", idl_name(node)->identifier);
+    return IDL_RETCODE_SEMANTIC_ERROR;
+  }
+
```
eboasson, line 787: If the min/max annotations need to check for the field having been annotated before already, shouldn't this one do so, too?

src/idl/src/annotation.c
```diff
@@ -722,6 +1015,67 @@ annotate_position(
...
+  idl_annotation_appl_t *annotation_appl,
+  idl_node_t *node)
+{
+  assert(annotation_appl);
+
+  const char *str = NULL;
+  const idl_type_spec_t *ts = NULL;
+  const idl_annotation_appl_t **annotation_appl_p = NULL;
+  idl_try_construct_t *try_construct_p = NULL;
+
```
eboasson, line 1030: If the min/max annotations need to check for the field having been annotated before already, shouldn't this one do so, too?

eboasson (commented):
Just a little bit of nitpicking on an otherwise fine PR. Perhaps you can have a look?

src/core/ddsi/include/dds/ddsi/ddsi_typewrap.h
reicheratwork: You're correct that this may cause issues when combined with other operators.
I will go over the #defines in this PR and check that this is fixed.

src/idl/src/annotation.c
reicheratwork: This occurs because the two parameters are parsed separately, whether in the same annotation like @range, or with @min and @max, and one can be read as an int, and the other as a float.
I will put in a small check that will give a warning if you compare integers and floats in a manner that loses resolution.

src/idl/src/annotation.c
reicheratwork: I will fix this.

src/idl/src/annotation.c
reicheratwork: I will fix this.

src/idl/src/annotation.c
reicheratwork: The way I understand the parser works, this is caught before the annotation contents are parsed in the annotation.c:dedup function and you will get a "Incompatible reapplication of @$name_of_annotation$" error.

src/idl/src/annotation.c
reicheratwork: Same as the previous comment.

eboasson (approved):
Thanks @reicheratwork !

PR #1167, opened 2022-02-28
Type consistency enforcement qos

dpotman:
This adds the implementation for the type consistency enforcement qos policy: 
- Add support for `ignore_sequence_bounds`, `ignore_string_bounds`, `ignore_member_names`, `prevent_type_widening`
and the type consistency kind (allow/disallow type coercion) options that can be set using the type consistency enforcement qos policy
- Commit 56cc62067f13264706b7d329e95962e8cff98376 removes the assignability check from sertype interface: this initial implementation will not work for all intended use cases (specific non-xtypes based implementations), and might be replaced by a new generic interface in a future release. For now the assignability check is called directly from the qos matching function. 
- Fixes for a bug and memory leaks in the assignability check
- Add option to disable type meta-data generation in the `idlc_generation` cmake function

### DIFF
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -11,26 +11,32 @@
 #
 
 function(IDLC_GENERATE)
+  set(options NO_TYPE_INFO)
   set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY)
   set(multi_value_keywords FILES FEATURES INCLUDES WARNINGS)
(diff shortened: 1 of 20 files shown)


eboasson (approved):
LGTM, I found the change in default type consistency QoS and the change in the validation code for it a surprise, but it turns out that this brings it in line with the spec. Thanks @dpotman !

(I'll review the companion PRs for the C++ and Python bindings before merging all three of them.)

Issue #352, opened 2019-12-09
How to handle multiple topics in single application efficiently


sjames:
Hi,
I'm a newcomer to DDS and was looking for a C implementation and happily stumbled upon this project.

The examples are trivial and publish only a single topic.  A real-world application will publish many more topics. 

What is the recommended pattern to deal with subscribing to several topics? I won't want to have to poll multiple readers continuously.  Is there a more reactive way of doing this with poll/select?

Thanks and regards,
Soj

k0ekk0ek:
Hi @sjames. Glad you like it so far! :slightly_smiling_face: DDS offer two ways to deal with this. You can use waitsets, which behave like poll/select or you can use listeners, which allow you to use more of an event based way to read the data. The roundtrip example shows how to use either one.

sjames:
Thank you for the quick answer.  I will try out waitsets and listeners.
Cheers!
Sj

PR #1169, opened 2022-03-01
Add APPEND_PROJECT_NAME_TO_INCLUDEDIR

sloretz:
The first commit is a bugfix that makes sure installed targets use `CMAKE_INSTALL_INCLUDEDIR` instead of hard coding `include`.

The second commit is an enhancement supporting https://github.com/ros2/ros2/issues/1150.  It adds a CMake option defaulting to `ON` that installs headers to `include/CycloneDDS` instead of `include`.

When using ROS 2 and Colcon it's possible to have two versions of a package installed: typically one from the debian packages (like`ros-[rosdistro]-cyclonedds`) and another in a from-source workspace. In Colcon terms this is called "overriding a package", and the `/opt/ros/[rosdistro]` folder would be called the underlay while the from-source workspace would be called the overlay. A user would do this when they want to try a newer version of a package without building everything from source.

When two or more packages install their headers to the same directory in an underlay, it becomes possible for packages in the overlay to accidentally find the headers of an overridden package in the underlay. This can cause build failures or undefined behavior at runtime depending on the differences between the headers in the overridden and overriding packages. The exact conditions this happens are a little complicated, so I'll skip writing it out here. In ROS 2 we're solving the issue by making every package install its headers to a unique folder using `${PROJECT_NAME}`. That's what the second commit does.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,15 @@ include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 include(CTest)
 
+option(APPEND_PROJECT_NAME_TO_INCLUDEDIR
+  "When ON headers are installed to a folder ending with an extra ${PROJECT_NAME}. \
+  This avoids include directory search order issues when overriding this package
+  from a merged catkin, ament, or colcon workspace." ON)
+
+if(APPEND_PROJECT_NAME_TO_INCLUDEDIR)
+  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}")
+endif()
+
 if(APPLE)
   set(CMAKE_INSTALL_RPATH "@loader_path/../${CMAKE_INSTALL_LIBDIR}")
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,7 +266,7 @@ include(CTest)
 option(APPEND_PROJECT_NAME_TO_INCLUDEDIR
   "When ON headers are installed to a folder ending with an extra ${PROJECT_NAME}. \
   This avoids include directory search order issues when overriding this package
-  from a merged catkin, ament, or colcon workspace." ON)
+  from a merged catkin, ament, or colcon workspace." OFF)
 
 if(APPEND_PROJECT_NAME_TO_INCLUDEDIR)
   set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}")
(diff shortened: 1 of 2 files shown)


sloretz:
I see this test timed out on OSX 10.15, but it seems unrelated to this PR. Is it known to be flaky?

```
499/992 Test #352: CUnit_ddsc_domain_torture_explicit .................................................***Timeout  10.37 sec


     CUnit - A unit testing framework for C - Version 2.1-3
     http://cunit.sourceforge.net/
```

sloretz:
Friendly ping. Any thoughts on this change, or on how it was done in eclipse-iceoryx/iceoryx#1194?

eboasson (commented):
Sorry for the abysmally slow response ...

The only problem I have with it is defaulting the option to `ON`. The more I see of CMake the lower my opinion of it, and by moving the include files like this, life becomes a lot harder for those sensible enough to avoid CMake — and those people exist, because I already ran into the consequences of the corresponding change in Iceoryx.

The plan is furthermore to do some API improvements and rename the include directory from `dds` to `cyclonedds` in the process, which is gives me the room to do backwards compatibility hacks in the old `dds/dds.h`. I don't want to rename directories twice in a row. If the default behaviour changes, that becomes an issue.

Would it not be possible to default to `OFF` and set `APPEND_PROJECT_NAME_TO_INCLUDEDIR` in the ROS 2 build system? In the ROS 2 environment, everyone uses CMake and colcon, and so this change wouldn't really affect anyone there.

eboasson:
> I see this test timed out on OSX 10.15, but it seems unrelated to this PR. Is it known to be flaky?
> 
> ```
> 499/992 Test #352: CUnit_ddsc_domain_torture_explicit .................................................***Timeout  10.37 sec
> 
> 
>      CUnit - A unit testing framework for C - Version 2.1-3
>      http://cunit.sourceforge.net/
> ```

There are a number of tests that occasionally timeout, on some platforms more often than on others. I've looked into that a number of times and so far the evidence seems to point to background load on the CI system, more than a problem in our code or tests.

sloretz:
> Would it not be possible to default to OFF and set APPEND_PROJECT_NAME_TO_INCLUDEDIR in the ROS 2 build system? In the ROS 2 environment, everyone uses CMake and colcon, and so this change wouldn't really affect anyone there.

Rebased and made the default value `OFF` in https://github.com/eclipse-cyclonedds/cyclonedds/pull/1169/commits/fc9ce80e06bae2c0382222dc2a4ea8d3234fc056

>  moving the include files like this, life becomes a lot harder for those sensible enough to avoid CMake

It looks like CycloneDDS provides a pkg-config file https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/PkgConfig.pc.in . That should make it easier to use with other build systems (on Linux).

eboasson (approved):
Thanks @sloretz !

PR #1179, opened 2022-03-07
Data representation annotation

dpotman:
This PR adds support for the data representation annotation and includes some refactoring and fixes for parsing and handling this annotation in IDLC:
- 159a1ed3d3a5f59f37265567db1940cc3834b741 adds a `requires_xcdr2` field to struct and union nodes in the IDLC node tree, so that the allowed data representations and required xcdr version are stored independently. The function `idl_allowable_data_representations` (renamed from `supported_data_representations`) now returns all representations set using the annotation, not only the supported ones. To filter out unsupported representation, a backend can use the `requires_xcdr2` flag, that is set on aggregated types. 
- In 2994ed57893300a3f2b4d83d49807e0556a8780a the code in IDLC C-backend to set the allowed data representations in the topic descriptor is added
- Commit 74e78c4ef1ab941120dafce8d2020d60f8100085 adds the implementation for using the data representation restrictions from IDL annotations in DDSI, and included a test case for combinations of annotation and qos values for data representation.

### DIFF
--- a/src/idl/tests/annotation.c
+++ b/src/idl/tests/annotation.c
@@ -1389,8 +1389,8 @@ test_rep(
 
   rep_test_t *test = (rep_test_t*)user_data;
 
-  allowable_data_representations_t result, required;
-    result = idl_supported_data_representations(node),
+  allowable_data_representations_t
+    result = idl_allowable_data_representations(node),
     required = test->reps[test->i++];
   if (idl_requires_xcdr2(node))
     result &= ~((allowable_data_representations_t)IDL_DATAREPRESENTATION_FLAG_XCDR1);
--- a/src/idl/tests/annotation.c
+++ b/src/idl/tests/annotation.c
@@ -1,4 +1,5 @@
 /*
+ * Copyright(c) 2022 ZettaScale Technology and others
  * Copyright(c) 2020 ADLINK Technology Limited and others
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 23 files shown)


src/idl/tests/annotation.c
```diff
@@ -1387,10 +1389,13 @@ test_rep(
...
   rep_test_t *test = (rep_test_t*)user_data;
 
-  allowable_data_representations_t
-    result = supported_data_representations(node),
+  allowable_data_representations_t result, required;
+    result = idl_supported_data_representations(node),
     required = test->reps[test->i++];
-  CU_ASSERT_EQUAL(required,result);
+  if (idl_requires_xcdr2(node))
+    result &= ~((allowable_data_representations_t)IDL_DATAREPRESENTATION_FLAG_XCDR1);
```
eboasson, line 1396: This is intriguing: it is almost as if there is a mix-up between "allowed" and "supported". Clearly the old "supported" was changed to "allowed", but then you'd expect that writing `@data_representation(XCDR1|XCDR2)` should result in both being "allowed", regardless of whether `XCDR1` is actually supported for the type.

So I think the test tries mixes two different things: whether or not the type requires `XCDR2` or not, and whether or not the allowed representations match. Wouldn't it then be better to make that explicit in the test expectation?

eboasson (commented):
Thanks @dpotman! The annoying thing is that I have a nitpick on a test ... If I am misreading it, or if it would delay things quite significantly to change it, then I'll merge it as is.

src/idl/tests/annotation.c
dpotman: You're right, this is confusing! In [3c9d757](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1179/commits/3c9d75774b3d1addcabf6bc7795edd6f9cbb6725) I've removed the tests for `idl_requires_xcdr2` from the data representation test and added a new test for that. I've also excluded ([ddc4e34](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1179/commits/ddc4e34c7fefe0933d4a1f8a4f9615d2bfb6dbd0)) appendable enums and bitmasks in the requires-xcdr2 check, as there is no reason to force xcdr2 for that

eboasson (approved):
Thanks!

PR #1180, opened 2022-03-07
Dds #965 shared memory support with transient local qos

### LINKED ISSUE #965: Enable shared memory support for transient local readers and writers

MatthiasKillat:
Resolves #965

### DIFF
--- a/src/core/ddsc/src/dds_reader.c
+++ b/src/core/ddsc/src/dds_reader.c
@@ -518,19 +518,21 @@ static iox_sub_options_t create_iox_sub_options(const dds_qos_t* qos) {
     opts.queueCapacity = (uint64_t)qos->history.depth;
   } else {
     opts.queueCapacity = max_sub_queue_capacity;
+    if (qos->reliability.kind == DDS_RELIABILITY_RELIABLE) {
+      opts.queueFullPolicy = QueueFullPolicy_BLOCK_PRODUCER;
+    }
   }
 
-  if(qos->reliability.kind == DDS_RELIABILITY_RELIABLE) {
-    opts.queueFullPolicy = QueueFullPolicy_BLOCK_PRODUCER; 
-  }
-
-  if(qos->durability.kind == DDS_DURABILITY_VOLATILE) {
+  // with BEST EFFORT DDS requires that no historical
+  // data is received (regardless of durability)
+  if(qos->reliability.kind == DDS_RELIABILITY_BEST_EFFORT ||
+     qos->durability.kind == DDS_DURABILITY_VOLATILE) {
     opts.historyRequest = 0;
   } else {
-    // Transient Local and stronger
+    // TRANSIENT LOCAL and stronger
     opts.historyRequest = (uint64_t) qos->history.depth;
-    // if the publisher cannot guarantee support for this request we
-    // will not be connected by iceoryx
+    // if the publisher does not support historicial data
+    // it will not be connected by iceoryx
     opts.requirePublisherHistorySupport = true;
   }
 
--- a/src/core/ddsc/src/dds_reader.c
+++ b/src/core/ddsc/src/dds_reader.c
@@ -491,6 +491,45 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
           DDS_DURABILITY_VOLATILE == qos->durability.kind &&
           DDS_HISTORY_KEEP_LAST == qos->history.kind);
 }
+
+static iox_sub_options_t create_iox_sub_options(const dds_qos_t* qos) {
+
+  iox_sub_options_t opts;
+  iox_sub_options_init(&opts);
+
+  const uint32_t max_sub_queue_capacity = iox_cfg_max_subscriber_queue_capacity();
+
+  // NB: We may lose data after history.depth many samples are received (if we
+  // are not taking them fast enough from the iceoryx queue and move them in
+  // the reader history cache), but this is valid behavior for volatile.
+  // It may still lead to undesired behavior as the queues are filled very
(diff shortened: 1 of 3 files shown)


src/core/ddsc/src/dds_writer.c
```diff
@@ -298,15 +298,50 @@ static bool dds_writer_support_shm(const struct ddsi_config* cfg, const dds_qos_
...
+
+  if(qos->reliability.kind == DDS_RELIABILITY_RELIABLE) {
+    opts.subscriberTooSlowPolicy = ConsumerTooSlowPolicy_WAIT_FOR_CONSUMER;
+  }
+
+  if(qos->durability.kind == DDS_DURABILITY_VOLATILE) {
+    opts.historyCapacity = 0;
+  } else {
+    // Transient Local and stronger
+    if (qos->durability_service.history.kind == DDS_HISTORY_KEEP_LAST) {
```
MatthiasKillat, line 337: Not sure whether the logic is completely sound here. It is hard to test in https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/238 as it seems it is not possible to create a DataWriter with non-default values for durability_service.history.kind (despite using `TopicQoS`).

I am unsure whether the ctor of `DataWriterQos` from `TopicQoS` there should even work in general, but apparently it compiles ... I did not really look into it.

MatthiasKillat:
@eboasson ready for review, but I am not sure whether durability_service.history.depth is used correctly and everything behaves as expected. Testing in https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/238 proves difficult as it seems that I cannot set this value with the QoS.

src/core/ddsc/src/dds_reader.c
```diff
@@ -483,13 +483,60 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
...
+  return (DDS_READER_QOS_CHECK_FIELDS ==
               (qos->present & DDS_READER_QOS_CHECK_FIELDS) &&
           DDS_LIVELINESS_AUTOMATIC == qos->liveliness.kind &&
-          DDS_INFINITY == qos->deadline.deadline &&
-          DDS_DURABILITY_VOLATILE == qos->durability.kind &&
-          DDS_HISTORY_KEEP_LAST == qos->history.kind);
+          DDS_INFINITY == qos->deadline.deadline);
+}
+
+static iox_sub_options_t create_iox_sub_options(const dds_qos_t* qos) {
```
MatthiasKillat, line 501: iox options are now created in this function according to the DDS QoS.

The mapping can be tricky in edge cases.

src/core/ddsc/src/dds_writer.c
eboasson: I think I know why that is: the spec doesn't list `DurabilityService` as a QoS on the Topic, but that must be an oversight because it is an essential part of the definition of "transient" behaviour. OpenSplice corrected that, but incorrectly dropped it from the DataWriter. Cyclone's C++ binding came from OpenSplice and I suppose nobody noticed.

What should work is setting the `DurabilityService` QoS on the topic, because it looks to me like that gets propagated by the C++ binding to the Cyclone core, and then the DataWriter QoS should then inherit from it. If it doesn't propagate, then I am tempted to call that a bug.

eboasson (commented):
LGTM, I suppose I can't fault you for not doing tests in C (extending the existing tests is tricky business ...) and https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/238/files looks to not be quite done yet because of the "durability service" QoS complications.

MatthiasKillat:
@eboasson Testing in C++ is a bandaid fix as Testing in C is difficult due to the framework used. It is much easier to set up proper tests in GTest (especially with proper test fixture and typed tests). We should add some tests in C as well, but I would prefer not to in this pull request.

MatthiasKillat:
@eboasson If you mean creating topic QoS and writer QoS from them like so https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/238#discussion_r824783756, then this does not appear to work (I observe this when I add some output or debug in the C code).

However, the tests pass there (with this branch ApexAI:dds-#965-shared-memory-support-with-transient-local-qos, not with the current master of course). It is just that it never uses non-default values.

src/core/ddsc/src/dds_writer.c
MatthiasKillat: If it the history depth is set on the topic QoS it appears to propagate correctly.

MatthiasKillat:
@eboasson @budrus @sumanth-nirmal Changed to not block iceoryx publishers with Reliable QoS

src/core/ddsi/src/ddsi_cdrstream.c
```diff
@@ -434,6 +434,9 @@ static uint32_t get_collection_elem_size (uint32_t insn, const uint32_t * __rest
     case DDS_OP_VAL_EXT:
       abort ();
       break;
+    default:
+      // unexpected value
+      abort ();
```
MatthiasKillat, line 439: Stealth fix of warning to appease ROS 2 but also to avoid returning an uninitialized value in the default case (can this happen?). We could also initialize the value with zero upon declaration but this a) is likley not optimized away in all other cases and b) returns still a zero in all non-handled cases (default, is this logically acceptable?)

eboasson (approved):
Looks good to me in combination with https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/238. (The test failures there are obviously related to the changes here, and locally it all works as expected.) Thanks @MatthiasKillat !

PR #1182, opened 2022-03-08
Fix bug in cdrstream extract key skip union

dpotman:
In `dds_stream_extract_key_from_data_skip_union` the wrong `ops` pointer was passed to the `skip_subtype` function, which causes the key extraction to fail in case of a non-key union member with a selected case that has an enum type with size > 8 bits.


eboasson (approved):
Thanks @dpotman !

PR #1188, opened 2022-03-09
Set default allowed data representations to XCDR1 and XCDR2

dpotman:
Set the default for allowed data representations in `ddsi_sertype_init` to XCDR1 and XCDR2. Additional restrictions from annotations in IDL and limitations for certain types (e.g. extensibility non-final) can override this (`ddsi_sertype_default_init`)

### DIFF
--- a/src/core/ddsi/src/ddsi_sertype.c
+++ b/src/core/ddsi/src/ddsi_sertype.c
@@ -180,7 +180,7 @@ void ddsi_sertype_init_flags (struct ddsi_sertype *tp, const char *type_name, co
   tp->typekind_no_key = (flags & DDSI_SERTYPE_FLAG_TOPICKIND_NO_KEY) ? 1u : 0u;
   tp->request_keyhash = (flags & DDSI_SERTYPE_FLAG_REQUEST_KEYHASH) ? 1u : 0u;
   tp->fixed_size = (flags & DDSI_SERTYPE_FLAG_FIXED_SIZE) ? 1u : 0u;
-  tp->allowed_data_representation = DDS_DATA_REPRESENTATION_FLAG_XCDR1;
+  tp->allowed_data_representation = DDS_DATA_REPRESENTATION_RESTRICT_DEFAULT;
   tp->base_sertype = NULL;
   tp->wrapped_sertopic = NULL;
 #ifdef DDS_HAS_SHM


eboasson (approved):

src/core/ddsi/src/ddsi_sertype.c
```diff
@@ -180,7 +180,7 @@ void ddsi_sertype_init_flags (struct ddsi_sertype *tp, const char *type_name, co
   tp->typekind_no_key = (flags & DDSI_SERTYPE_FLAG_TOPICKIND_NO_KEY) ? 1u : 0u;
   tp->request_keyhash = (flags & DDSI_SERTYPE_FLAG_REQUEST_KEYHASH) ? 1u : 0u;
   tp->fixed_size = (flags & DDSI_SERTYPE_FLAG_FIXED_SIZE) ? 1u : 0u;
-  tp->allowed_data_representation = DDS_DATA_REPRESENTATION_FLAG_XCDR1;
+  tp->allowed_data_representation = DDS_DATA_REPRESENTATION_RESTRICT_DEFAULT;
```
eboasson, line 183: This causes (I think) a rather important change in what gets advertised during discovery, for example it seems like it would result in the Cyclone  RMW layer to start advertising "XCDR1 or XCDR2", where that one is XCDR1-only. That makes this seem like a less-than-desirable change.

The problem is of course that it is hard-coded. The `ddsi_sertype_init_flags` function got added because the flags needed to be initialised ... adding a function that also allows initialising the `allowed_data_representation` field would be the logical extension that would avoid causing changes in behaviour for users. It would also allow the C/C++/Python bindings to initialise it via this "constructor", rather than having to override it.

Adding yet another function for initialising this is not so great, but I think that it would be the better option.

dpotman: You're right that using this default causes RMW to advertise `XCDR2`, but that problem was already introduced in commit 5382ece652164f31fbb6484c2ef4fab75026f8e8. That defaults the minimal XCDR version for a sertype to `XCDR1`, so `XCDR2` is also allowed, which is undesired unless a sertype implementation explicitly includes `XCDR2` in the allowed representations.  The reason that I introduced the change from `XCDR1` to `DDS_DATA_REPRESENTATION_RESTRICT_DEFAULT` was that Python fuzzer tests were failing, but that should be fixed by setting the representations here https://github.com/eclipse-cyclonedds/cyclonedds-python/blob/master/clayer/pysertype.c#L905, @thijsmie will create a PR for this. 

I agree that the sertype initialisers needs some refactoring, but I think it is better to do that with the planned sertype and serdata interface refactoring.

Issue #1198, opened 2022-03-14
CMake install fail for pubsub when building static library. 


icucode:
CMake configuration enforce the use of CMAKE_INSTALL_RPATH even when building static library.

thijsmie:
Pubsub has been removed in favour of the python tool with the same name.

PR #1199, opened 2022-03-14
Fix CMake install fail when building static library #1198

icucode:


CMakeLists.txt
```diff
@@ -263,10 +263,12 @@ include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 include(CTest)
 
-if(APPLE)
-  set(CMAKE_INSTALL_RPATH "@loader_path/../${CMAKE_INSTALL_LIBDIR}")
-else()
-  set(CMAKE_INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}")
+if(BUILD_SHARED)
```
eboasson, line 266: This looks very logical to me, but the CI seems to disagree ...

PR #1203, opened 2022-03-15
Bugfix in IDLC type meta-data generator for array-typedefs

dpotman:
Fix for type-id-not-found error in IDLC type meta-data generation when using multiple levels of typedefs with arrays in IDL.

### DIFF
--- a/src/tools/idlc/tests/type_meta.c
+++ b/src/tools/idlc/tests/type_meta.c
@@ -643,6 +643,61 @@ static DDS_XTypes_TypeObject *get_typeobj12 (void)
     });
 }
 
+static DDS_XTypes_TypeObject *get_typeobj13 (void)
+{
+  DDS_XTypes_TypeIdentifier ti_f1;
+
+  /* f1 type identifier */
+  {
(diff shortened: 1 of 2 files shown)


src/tools/idlc/tests/type_meta.c
```diff
@@ -643,6 +643,62 @@ static DDS_XTypes_TypeObject *get_typeobj12 (void)
...
+{
+  DDS_XTypes_TypeIdentifier ti_f1;
+
+  /* f1 type identifier */
+  {
+    DDS_XTypes_TypeIdentifier *ti_long = calloc (1, sizeof (*ti_long));
+    ti_long->_d = DDS_XTypes_TK_INT32;
+
+    /* typedef long td_arr[3] */
+    DDS_XTypes_TypeIdentifier *ti_alias_arr = calloc (1, sizeof (*ti_alias_arr));
```
eboasson, line 656: This one leaks according the address sanitizer. Otherwise, it LGTM.

dpotman: I've force-pushed a fix for this, the type identifier can be stored directly in the type object of the outer typedef.

PR #1204, opened 2022-03-15
Add a breakpoint for rejected CDR

eboasson:
This adds a "normalize_error" function that is called at the point where
ddsi_stream_normalize rejects an invalid CDR blob, which makes analyzing
the exact reason for rejecting an input a lot easier.


dpotman:
LGTM, only the `skip_..` functions could return `NULL` without reaching the error functions, but that can only occur in case of invalid serializer instructions (a type that's not in `enum dds_stream_typecode`), and that's not very likely.

PR #1209, opened 2022-03-17
Exclude iceoryx from windows build

### LINKED ISSUE #1208: Exlude iceoryx binding from windows build
I recently noticed that the iceoryx C-Binding is also included when building on Windows target.
In the previous release the iceoryx binding was excluded from the Windows build because iceoryx faced a linker issue. 
See: https://github.com/eclipse-cyclonedds/cyclonedds/blob/releases/0.8.x/src/CMakeLists.txt#L71
Therefore iceoryx is currently only building as static library in windows.
It is addressed here: https://github.com/eclipse-iceoryx/iceoryx/issues/846

Since the windows target is currently not properly tested I would propose to exclude the iceoryx binding from the windows build for ROS 2, like we did in last year.

dkroenke:
Signed-off-by: Dietrich Krönke <dietrich.kroenke@apex.ai>

@eboasson This excludes iceoryx from the windows build

- Closes #1208

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -40,7 +40,13 @@ if(ENABLE_SOURCE_SPECIFIC_MULTICAST)
 endif()
 
 # Prefer iceoryx integration but do not require it.
-set(ENABLE_SHM "AUTO" CACHE STRING "Enable shared memory support")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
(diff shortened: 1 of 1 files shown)


eboasson (approved):
It looks good to me, and it also seems sensible to black-list Windows rather than to white-list Linux/macOS/QNX (like it did some time back) because there won't be any Iceoryx packages on other platforms anyway.

PR #1210, opened 2022-03-17
Bit bound validation

reicheratwork:
Additional validation of bit bounds of bitmasks and enums

This validation was previously only triggered on the @bit_bound annotation being present, now it is done when the tree is completed parsing.


eboasson (approved):
Thanks @reicheratwork !

PR #1211, opened 2022-03-17
Fix uninitialized variable warning in cdrstream

dpotman:


eboasson (approved):
Thanks @dpotman !

PR #1215, opened 2022-03-18
Exclude iceoryx tests from colcon

dkroenke:
Signed-off-by: Dietrich Krönke <dietrich.kroenke@apex.ai>


eboasson (approved):
Thanks, @dkroenke ! It is not ideal, but it does solve the problem of running the test suite on the ROS 2 test farm, and I verified the tests still run on our own CI. (The one test failure has nothing to do with this, it is one of those CI/virtualised platforms misbehave in mysterious ways that I have come to accept as a fact of life ...)

PR #1216, opened 2022-03-18
Set CMake CMP0075 policy to NEW

eboasson:
CMP0075 concerns a change in behaviour in the combination of
check_include_files and CMAKE_REQUIRED_LIBRARIES. This combination
occurs on Windows, and in this case both the old and the new behaviour
are acceptable.

We support CMake versions older than the one where the policy was
introduced, so select the new behaviour only if the policy actually
exists.

Note: needs backport to `releases/0.9.x` branch because of the ROS 2 build farm.


eboasson:
Confirmed to work as intended by ROS 2 CI: [Build Status](https://ci.ros2.org/job/ci_windows/16734/)

PR #1218, opened 2022-03-18
Improve "oneliner" a bit

eboasson:
* print time stamps
* print wait duration
* allow setting QoS on participant, subscriber, publisher
* erase dependent entities to avoid an error on termination after explicitly
  deleting, e.g., a subscriber containing a reader
* make it the forced expiry of a lease and publication of SPDP messages
  in "deaf" and "hearing" optional


dpotman (approved):
LGTM

PR #1222, opened 2022-03-22
Fix CMake _confgen invocation

eboasson:
It turns out add_custom_command uses "ARGS" to separate the command from
its arguments, but add_custom_target "COMMAND" doesn't.


thijsmie (approved):
Yup, missed this while shuffling between the two. 

cmake_meme

PR #1224, opened 2022-03-23
Fixes for resolving types and type dependencies

dpotman:
Bugfix in resolving type dependencies: in case type info from proxy endpoint was received before creating local topic using the same top-level type, the type object for the local type dependencies was not added in the type library


eboasson (approved):
Thanks @dpotman !

PR #1225, opened 2022-03-23
Entity names

thijsmie:
This adds entity names to the user-accessible QoS and an auto-namer, enabled via configuration.


docs/manual/options.md
```diff
@@ -267,7 +267,7 @@ The default value is: "".
 
 
 ### //CycloneDDS/Domain/General
```
eboasson, line 269: I ❤️ your improvements to the documentation generation!

src/core/ddsc/src/dds_reader.c
```diff
@@ -622,6 +622,19 @@ static dds_entity_t dds_create_reader_int (dds_entity_t participant_or_subscribe
   if (tp->m_ktopic->qos)
```
eboasson, line 622: This is a good reason why the entity name shouldn't be set on the QoS. The
```C
  if (sub->m_entity.m_qos)
    ddsi_xqos_mergein_missing (rqos, sub->m_entity.m_qos, ~(uint64_t)0);
```
above (which GitHub doesn't allow me attach on comment on, silly GitHub) I think means it would copy the entity name from the subscriber.

Using `~QP_ENTITY_NAME` instead of `~(uint64_t)0` for the `ddsi_xqos_mergein_missing` means it doesn't copy the entity name. That should solve it.

src/core/ddsi/src/ddsi_config.c
```diff
@@ -1177,6 +1184,30 @@ static enum update_result uf_tracingOutputFileName (struct cfgst *cfgst, UNUSED_
...
+    memcpy((ddsrt_md5_byte_t*)elem, buf, 16);
+    memcpy(((ddsrt_md5_byte_t*)elem) + 16, buf, 16);
+  }
+  return URES_SUCCESS;
+}
+
+static void pf_random_seed (struct cfgst *cfgst, void *parent, struct cfgelem const * const cfgelem, uint32_t sources)
+{
+  ddsrt_prng_seed_t * const seed = cfg_address (cfgst, parent, cfgelem);
+  cfg_logelem (cfgst, sources, "%"PRIu32" %"PRIu32" %"PRIu32" %"PRIu32" %"PRIu32" %"PRIu32" %"PRIu32" %"PRIu32"",
```
eboasson, line 1207: So if I configure the seed to be "triskaidecafobie" here it'll print `5722008f 768a9428 306e67eb e5014210 5722008f 768a9428 306e67eb e5014210` on a big-endian machine and `8f002257 28948a76 eb676e30 104201e5 8f002257 28948a76 eb676e30 104201e5` on a little-endian machine (if I didn't get it wrong).

I have no idea whether the two result in the same sequence (I don't know the Mersenne Twister that well), and it would be difficult to reproduce it based on the configuration dump. But arguably the former matters not and the latter could even be considered a feature.

Just noting it in case some of this is important to you but you forgot it earlier 😀

eboasson (commented):
Just a few nitpicks 😀

src/core/ddsc/src/dds__qos.h
```diff
@@ -24,35 +24,37 @@ extern "C" {
    QP_DEADLINE | QP_LATENCY_BUDGET | QP_OWNERSHIP | QP_LIVELINESS |     \
    QP_RELIABILITY | QP_TRANSPORT_PRIORITY | QP_LIFESPAN |               \
    QP_DESTINATION_ORDER | QP_HISTORY | QP_RESOURCE_LIMITS |             \
-   QP_DATA_REPRESENTATION)
+   QP_DATA_REPRESENTATION | QP_ENTITY_NAME)
```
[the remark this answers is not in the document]
thijsmie: I also thought about it and decided to allow it and just not fill it by default, even with autonaming. If a user wants a separate name for a topic when logging that is possible this way, its not in the way of anyone to have it this way.

src/core/ddsi/src/ddsi_config.c
thijsmie: I did indeed consider that using the same seed on little or big endian machines would most likely result in different names. However, going to the effort of ensuring equivalent outcomes between hosts is actually what you do *not* want in all cases that I can come up with. In fact, I am still on the fence about something we previously discussed: should we *always* include the hostname in the seed? What about *process/executable name*? Maybe some config booleans, e.g. `<EntityAutoNaming seed="hexakosioihexekontahexafobie" mixin_host="true" mixin_executable="true">fancy</EntityAutoNaming>`

eboasson: Ah, it appears you run into scary integers later than I do if start enumerating them from, let's say, 0. (If I'd say 14, there'd the possibility I'd never run into a scary one anymore, so that'd cheating, no? 🙂)

It is probably a good idea to make mixing in host name and executable information an option. That way we can keep sitting on the fence.

thijsmie:
I think it should be good now, I had to update a test (configuration with random seed is not deterministic). The one failure seems a random failure of gcov, probably unrelated? Maybe a rerun of that job..

eboasson (approved):
LGTM now, thanks @thijsmie ! And indeed, the failure is caused by a gcov-related problem that should be solved now thanks to @dpotman's PR #1229.

PR #1227, opened 2022-03-25
Fixes for issues reported by coverity

reicheratwork:
This fixes a number of issues reported by coverity:
Uninitialized variable use: 349351, 349352 and 349353
Passing large entities by value: 350274
Memory leak: 350493, 350494


eboasson (commented):
Thanks as always @reicheratwork for tidying up the Coverity issues list. There are some details where I think the changes you propose are not the right ones, but for many of them I do agree. I also have something lying around in the same area, so what I'll do is create a new PR where I cherry-pick the commits from this PR that I agree with and fiddle a bit with the other details.

PR #1228, opened 2022-03-28
Check type lookup request/reply type

dpotman:
As the getDependencies service operation for the type lookup service is not supported yet, the code that handles the requests and replies should check that the message is a getTypes request or reponse.


eboasson (approved):
Good catch! Thanks @dpotman !

PR #1230, opened 2022-03-29
Coverity fixes (replaces #1227)

eboasson:
With thanks to @reicheratwork, some of whose fixes I cherry-picked from #1227.

### DIFF
--- a/src/core/ddsi/src/ddsi_ownip.c
+++ b/src/core/ddsi/src/ddsi_ownip.c
@@ -273,7 +273,11 @@ static bool gather_interfaces (struct ddsi_domaingv * const gv, size_t *n_interf
   ddsrt_freeifaddrs (ifa_root);
   assert ((*n_interfaces > 0) == (*maxq_count > 0));
   if (*n_interfaces == 0)
+  {
     GVERROR ("failed to find interfaces for \"%s\"\n", gv->m_factory->m_typename);
+    ddsrt_free (*maxq_list);
+    ddsrt_free (*interfaces);
+  }
   return (*n_interfaces > 0);
 }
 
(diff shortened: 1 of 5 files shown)


src/core/ddsi/src/ddsi_ownip.c
```diff
@@ -251,22 +260,35 @@ static bool gather_interfaces (struct ddsi_domaingv * const gv, size_t *n_interf
...
     {
+      size_t *new_maxq_list;
+      struct nn_interface *new_interfaces;
       max_interfaces *= 2;
-      *maxq_list = ddsrt_realloc (*maxq_list, max_interfaces * sizeof (**maxq_list));
-      *interfaces = ddsrt_realloc (*interfaces, max_interfaces * sizeof (**interfaces));
+      if ((new_maxq_list = ddsrt_realloc (*maxq_list, max_interfaces * sizeof (**maxq_list))) == NULL ||
+          (new_interfaces = ddsrt_realloc (*interfaces, max_interfaces * sizeof (**interfaces))) == NULL)
+      {
+        goto fail;
```
dpotman, line 269: Nit-picking: in case the first realloc succeeds and the second fails, this leaks `new_maxq_list`, and `*maxq_list` shouldn't be freed.

dpotman (commented):
LGTM! Only a detail in the out-of-memory handling in `gather_interfaces`, see comment below.

src/core/ddsi/src/ddsi_ownip.c
eboasson: I wouldn't call that nit-picking! Thanks for spotting it 🙂

dpotman (approved):

PR #1231, opened 2022-03-29
Do not use shared memory with ignorelocal QoS

MatthiasKillat:
As seen in some ROS tests, local readers can receive data despite ignorelocal if iceoryx is used (due to iceoryx not being aware of this QoS). This cannot be fixed without changing iceoryx, hence we disable shared memory use in this case which results in consistent behavior if this `ignorelocal` is true.

@eboasson @sumanth-nirmal @dkroenke


src/core/ddsc/src/dds_reader.c
```diff
@@ -490,7 +490,11 @@ static bool dds_reader_support_shm(const struct ddsi_config* cfg, const dds_qos_
   if(!(qos->durability.kind == DDS_DURABILITY_VOLATILE ||
     qos->durability.kind == DDS_DURABILITY_TRANSIENT_LOCAL)) {
     return false;
-  }  
+  }
+
+  if(qos->present & QP_CYCLONE_IGNORELOCAL && qos->ignorelocal.value != DDS_IGNORELOCAL_NONE) {
```
eboasson, line 495: Shouldn't this also be checked for the writing side?

MatthiasKillat: I do not think so. From my understanding ignorelocal implies: the reader cannot use shared memory if it is set to ignore local publishers (as they might send something with iceoryx which it would receive).

I am not sure what the meaning would be on writer side: ignore any local readers? Then we should indeed check it on the writer side as well (and if it iset to ignore local readers, it will not use iceoryx). But I think it is sufficient if the reader basically tells the writer that it wants to be ignored.

Ideally iceoryx could deal with this (by setting a flag on its subscriber), but it cannot without creating further issues.

MatthiasKillat:
I am trying to fix/generalize the local delivery logic if iceoryx is involved but this causes a lot of issues (corner case semantics and potentially undesired side-effects I cannot yet determine).

MatthiasKillat:
@eboasson If fixed/improved the local delivery logic with iceoryx, but at the expense of almost-code duplication. There is refactoring potential but it is not straightforward.

src/core/ddsc/src/dds_reader.c
MatthiasKillat: Add on writer side as well.

src/core/ddsi/include/dds/ddsi/q_entity.h
```diff
@@ -425,6 +425,9 @@ struct reader
 #ifdef DDS_HAS_SECURITY
   struct reader_sec_attributes *sec_attr;
 #endif
```
MatthiasKillat, line 427: move up to other bitfield

src/core/ddsi/src/ddsi_deliver_locally.c
```diff
@@ -267,3 +267,116 @@ dds_return_t deliver_locally_allinsync (struct ddsi_domaingv *gv, struct entity_
...
+  type_sample_cache_init(&tsc);
+  if (!source_entity_locked)
+    ddsrt_mutex_lock(&source_entity->lock);
+  rd = ops->first_reader(gv->entity_index, source_entity, &it);
+  if (rd != NULL)
+    EETRACE(source_entity, " =>");
+  while (rd != NULL) {
+    if (rd->has_iceoryx) {
+      rd = ops->next_reader(gv->entity_index, &it);
+      continue; // skip non iceoryx readers
```
MatthiasKillat, line 289: skip iceoryx readers actually

src/core/ddsi/src/ddsi_deliver_locally.c
```diff
@@ -267,3 +267,116 @@ dds_return_t deliver_locally_allinsync (struct ddsi_domaingv *gv, struct entity_
...
+
+static dds_return_t deliver_locally_fastpath_except_iceoryx(
+    struct ddsi_domaingv *gv, struct entity_common *source_entity,
+    bool source_entity_locked, struct local_reader_ary *fastpath_rdary,
+    const struct ddsi_writer_info *wrinfo,
+    const struct deliver_locally_ops *__restrict ops, void *vsourceinfo) {
+  struct reader **const rdary = fastpath_rdary->rdary;
+  uint32_t i = 0;
+  while (rdary[i]) {
+    if (rdary[i]->has_iceoryx) {
```
MatthiasKillat, line 319: do this while constructing the array

src/core/ddsi/src/ddsi_deliver_locally.c
```diff
@@ -267,3 +267,116 @@ dds_return_t deliver_locally_allinsync (struct ddsi_domaingv *gv, struct entity_
...
+    struct ddsi_serdata *payload;
+    struct ddsi_tkmap_instance *tk;
+    if ((payload = ops->makesample(&tk, gv, type, vsourceinfo)) == NULL) {
+      /* malformed payload: skip all readers with the same type */
+      while (rdary[++i] && rdary[i]->type == type)
+        ; /* do nothing */
+    } else {
+      do {
+        if (rdary[i]->has_iceoryx) {
+          continue;
```
MatthiasKillat, line 333: increment i here

src/core/ddsi/src/q_entity.c
```diff
@@ -2288,7 +2290,13 @@ static void writer_add_local_connection (struct writer *wr, struct reader *rd)
             PGUID (wr->e.guid), PGUID (rd->e.guid));
   m->rd_guid = rd->e.guid;
   ddsrt_avl_insert_ipath (&wr_local_readers_treedef, &wr->local_readers, m, &path);
-  local_reader_ary_insert (&wr->rdary, rd);
+
+#ifdef DDS_HAS_SHM
+  if (!wr->has_iceoryx || !rd->has_iceoryx)
+    local_reader_ary_insert(&wr->rdary, rd);
+#else
+  local_reader_ary_insert(&wr->rdary, rd);
```
MatthiasKillat, line 2298: I know that technically this #elseif is not needed if we exclude the branch from the #ifdef but I think this way intention and formatting is more clear.

src/core/ddsi/src/q_entity.c
```diff
@@ -385,7 +385,9 @@ static void local_reader_ary_remove (struct local_reader_ary *x, struct reader *
   for (i = 0; i < x->n_readers; i++)
     if (x->rdary[i] == rd)
       break;
-  assert (i < x->n_readers);
+  if (i >= x->n_readers) {
+    return; // rd not found, nothing to do
```
MatthiasKillat, line 389: Could also be done with #ifdefs at the call site, but I think it is safer this way.

src/core/ddsc/src/dds_write.c
```diff
@@ -193,27 +193,28 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
-      // iox_pub_publish_chunk takes ownership, storing a null pointer here doesn't
-      // preclude the existence of race conditions on this, but it certainly improves
-      // the chances of detecting them
-
-      iox_pub_publish_chunk (wr->m_iox_pub, d->iox_chunk);
-      d->iox_chunk = NULL;
-      return true; //we published the chunk
-    }
-    return false; // we did not publish the chunk
+  if (wr && wr->m_iox_pub != NULL && d->iox_chunk != NULL) {
```
MatthiasKillat, line 196: Moved the check for `wr` here, rest is only reformatting (reduce indent whitespace).

src/core/ddsc/src/dds_write.c
```diff
@@ -193,27 +193,28 @@ static dds_return_t deliver_locally (struct writer *wr, struct ddsi_serdata *pay
...
-      // iox_pub_publish_chunk takes ownership, storing a null pointer here doesn't
-      // preclude the existence of race conditions on this, but it certainly improves
-      // the chances of detecting them
-
-      iox_pub_publish_chunk (wr->m_iox_pub, d->iox_chunk);
-      d->iox_chunk = NULL;
-      return true; //we published the chunk
-    }
-    return false; // we did not publish the chunk
+  if (wr && wr->m_iox_pub && d->iox_chunk) {
```
MatthiasKillat, line 196: Moved `wr` check here, and reformatted incorrect indent. No further logic changes.

MatthiasKillat:
@eboasson Removed code duplication and changed it as discussed to not include iceoryx readers into the local reader array used by the fast path logic.

MatthiasKillat:
@eboasson Still trying to figure out the test failures with the changes, hence it is not quite ready to review. The test that failed before this commit appeared unrelated, but the tests failing now are due to some problem with the changes.

MatthiasKillat:
Commit "optimize local delivery" 687f231e67aff9dbe52e16cc454c0cf2b25abaee appears to pass the CI but does not yet exclude iceoryx readers in the fast path array.

MatthiasKillat:
@eboasson I cannot really understand what in the transition of commits from the previous to [](https://github.com/MatthiasKillat)@MatthiasKillat
[Ignore local iceoryx readers in fastpath](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1231/commits/0f7f72b84dc64fa0c6811deb4bcdc10f601d6d96) is breaking.

Note that the code in write could be changed somewhat to not consider `deliver_locally_except_iceoryx`,

```
bool delivered = deliver_data_via_iceoryx(wr, d);
  if (ret == DDS_RETCODE_OK) {

    if (delivered) {
      // delivers to all local non-iceoryx readers since
      // only those are in the fastpath if DDS_HAS_SHM is true
      ret = deliver_locally_except_iceoryx(ddsi_wr, d, tk);
    } else {
      ret = deliver_locally(ddsi_wr, d, tk);
    }
  }
```
But semantically there should be no difference if all iceoryx readers are removed from the array. Unless there is a side effect or something else entirely.

It is an option to drop the commit and live with the inefficiency/duplication of the previous one.

MatthiasKillat:
@eboasson I integrated your changes from https://github.com/eboasson/cyclonedds/tree/prohibit-shared-memory-with-ignorelocal-qos but we may need some more commits. I squashed some of the commits, I took 5 commits, 2 WiPs and the Oopsies). I kept the two larger commits separate.

Not sure we need more for this issue or whether we want to abandon this pull request completely in favor of the branch I took the commits from.

src/core/ddsc/src/dds_write.c
```diff
@@ -382,233 +410,188 @@ static bool fill_iox_chunk(dds_writer *wr, const void *sample, void *iox_chunk,
...
-      iox_pub_release_chunk(wr->m_iox_pub, d->iox_chunk);
-      d->iox_chunk = NULL;
-      ret = DDS_RETCODE_ERROR;
+    // deliver via iceoryx and handle local non-iceoryx readers
+
+    // NB: there may be local non-iceoryx readers due to QoS like
+    // Keep Last not being supported with iceoryx, these are handled without
+    // iceoryx as otherwise they would get no data
+
+    if (deliver_locally_to_non_iox(ddsi_wr, d) == DDS_RETCODE_OK) {
```
eboasson, line 529: I think this is not the correct solution to the problem, because we first do our best to avoid touching the content if it came from a chunk (that is, Iceoryx owns the chunk), and then here we forward the pointer to subscriptions that Iceoryx is not aware of. I expect that in:
```
r1 = create reader via iceoryx (e.g., keep-last 1)
r2 = create reader not via iceoryx (e.g. keep-all)
w = create writer via iceoryx
x = borrow sample from w ; fill x ; write w, x
take r1
y = borrow sample from w ; fill y
take r2
```
r2 could see "y" where it should've seen "x". Perhaps my understanding of Iceoryx's buffer management is a little simplistic and it involves more work, but it seems like something of this sort could happen.

I propose changing the condition for `use_only_iceoryx` instead (actually also ensuring that it still means what the name suggests). That should be fairly easy: we just have to additionally require `wr->rdary.n_readers == 0`.

eboasson (commented):
Thanks for all the work @MatthiasKillat! I am in agreement with almost everything, it is just that I think there is a problem in the one edge case you discovered and addressed in the last commit. Fortunately, I think the fix is small. What do you think?

src/core/ddsc/src/dds_write.c
MatthiasKillat: Yes, I think you are right. It appears we need to restrict this further with this condition. I will change the latest commit.

MatthiasKillat: @eboasson Fixed, basically as you say. I was hoping to avoid the serialization of the mixed case we now use instead. In theory it would be possible but a headache.

Assuming we have all readers R = I + L with + disjoint union, I iceoryx readers, and L other local readers (basically those in the array). If I = IL + INL (iceoryx local and non-local). We could deliver the chunk as follows.

INL: chunk, via iceoryx, RHC will have a chunk
IL: chunk, via iceoryx, RHC will have a chunk
L: copy(!) chunk to RHC (instead of serialization and whatnot), RHC will have a copy

The IL case can be improved by an iceoryx change I will work on next:
move chunk directly to RHC (no copy, no iceoryx queuing etc.). But this is out of scope here.

MatthiasKillat:
@eboasson Changed the last commit in a way you suggested. Should be ready for review again. Sorry for the late reply, I was traveling last week.

MatthiasKillat:
@eboasson Not sure what is up with the Windows build but looks unrelated.

eboasson:
@MatthiasKillat 

> Changed the last commit in a way you suggested. Should be ready for review again. Sorry for the late reply, I was traveling last week.

I can offer the same apology ... (though it wasn't so much travelling as being occupied with non-work things).

src/core/ddsc/src/dds_write.c
```diff
@@ -382,233 +396,192 @@ static bool fill_iox_chunk(dds_writer *wr, const void *sample, void *iox_chunk,
...
+  // Furthermore, all readers that ignore local publishers will not use iceoryx.
+  // We will never use only(!) iceoryx if there is any local reader in L.
+  // We will serialize the data in this case and deliver it mixed, i.e.
+  // partially with iceoryx as required by the QoS and type. The readers in L
+  // will get the data via the local delivery mechanism (fast path or slow
+  // path).
+  const bool use_only_iceoryx =
+      no_network_readers &&
+      ddsi_wr->xqos->durability.kind == DDS_DURABILITY_VOLATILE &&
+      ddsi_wr->rdary.n_readers == 0;
```
eboasson, line 474: I am a little bit worried about this access outside all locks, but at the same time I suspect it will actually work fine in practice. If unlocked access is indeed fine, it should become an `ddsrt_atomic_uint32_t`, but if you then look at the definition, loading it really is nothing special on any machine.

Perhaps I should sleep on this detail for a night ...

eboasson (commented):
LGTM, except for one detail that worries me a bit (I've added a comment there).

Another thing is that I strongly prefer merging things into `master` and then backport to the release branch. I think retargetting it to master is pretty easy, so I can sort that out myself.

Before merging this work in the release branch, I'll need to do a ROS 2 CI run, but I know how to start that.

src/core/ddsc/src/dds_write.c
MatthiasKillat: I somewhat agree. I usually use another style of programming where the locks are tightly coupled to the structures they are supposed to protect (otherwise it gets unmaintainable fast). This is easier in languages that allow encapsulation though. Then it can be made impossible to use wrongly (without lock).

In this case we should think about a dedicated lock for this whole fast-path structure, unless we can argue it is only modified/read under writer lock anyway. This `n_readers` should only be accessed under lock then, by some `get` function. I can change it like this.

MatthiasKillat:
@eboasson We can retarget master no problem and I agree with this. This whole odyssey started with the intention of a small fix which got more complicated than anticipated, that is why the target is the release branch.

I can change this if you prefer.

src/core/ddsc/src/dds_write.c
eboasson: If you time to do that, that'd be great!

By the way, there is a lock: https://github.com/eclipse-cyclonedds/cyclonedds/blob/3e5a4ab093cf18e9e7a6887d6b0a00c82b7b1240/src/core/ddsi/include/dds/ddsi/q_entity.h#L233

eboasson:
> @eboasson We can retarget master no problem and I agree with this. This whole odyssey started with the intention of a small fix which got more complicated than anticipated, that is why the target is the release branch.
> 
> I can change this if you prefer.

Again, that'd be really with my recovering from having had some time off ... !

MatthiasKillat:
@eboasson Regarding retargeting to master: shall I do this? (may cause conflicts, I did not check yet). Not sure about the workflow. Usually I would think targeting release and backporting is the way to go for small changes. It mainly depends on the difference between release and master. If master already contains experimental new features we need two PRs anyway.

eboasson (approved):
I went over it another time and can't think of anything else. Thank you @MatthiasKillat !

eboasson:
> @eboasson Regarding retargeting to master: shall I do this? (may cause conflicts, I did not check yet). Not sure about the workflow. Usually I would think targeting release and backporting to master is the way to go for small changes. It mainly depends on the difference between release and master. If master already contains experimental new features we need two PRs anyway.

I tried it locally, and it is free of conflicts, so I think I can just do it right here on GitHub

PR #1233, opened 2022-03-30
Fix Coverity issues

dpotman:
Fixing some issues reported by Coverity


src/core/ddsi/src/ddsi_security_omg.c
```diff
@@ -2291,6 +2289,9 @@ static bool q_omg_security_register_remote_writer_match(struct proxy_writer *pwr
   bool send_tokens = false;
   bool allowed = false;
 
+  if (!sc)
+    return false;
+
   if ((proxypp_match = get_pp_proxypp_match_if_authenticated(pp, proxypp, pwr->e.guid.entityid)) == NULL)
```
eboasson, line 2295: I believe that for `proxypp_match` to be non-null, `sc` has to be non-null, because it would otherwise never have been able to locate that match. But I agree that adding a check is sensible, given that there is an obvious way to handle the case.

eboasson (approved):

PR #1235, opened 2022-03-31
Fix Android port reference links

LanderU:
Signed-off-by: LanderU <lander.usategui@gmail.com>


LanderU:
I've tested the instructions with Android 12 and they also work changing API level to 31 and ABI to x86_64.

eboasson:
> I've tested the instructions with Android 12 and they also work changing API level to 31 and ABI to x86_64.

Nice, thanks!

eboasson (approved):

PR #1239, opened 2022-04-01
Redesign the type definition of "ddsrt_ifaddrs"

aohwang:
With the consideration of other transports whose address is not sockaddr, redesign the type definition of "ddsrt_ifaddrs".

### DIFF
--- a/src/core/ddsi/src/ddsi_ownip.c
+++ b/src/core/ddsi/src/ddsi_ownip.c
@@ -133,7 +133,7 @@ static enum maybe_add_interface_result maybe_add_interface (struct ddsi_domaingv
   if ((ifa->flags & IFF_UP) == 0) {
     GVLOG (DDS_LC_CONFIG, " (interface down)");
     return MAI_IGNORED;
-  } else if (ddsrt_sockaddr_isunspecified(ifa->addr)) {
+  } else if (ddsrt_sockaddr_isunspecified(ifa->ifaddr.ipaddr.addr)) {
     GVLOG (DDS_LC_CONFIG, " (address unspecified)");
     return MAI_IGNORED;
   }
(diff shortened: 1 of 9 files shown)


eboasson (commented):
Thanks @aohwang, it is definitely good to move away from `struct sockaddr` here. I do wonder:

It now simply maps everything to `ipaddr`, but in reality it is currently using IPv4, IPv6 and Ethernet (so 6-byte MAC addresses). It seems better to distinguish between those rather than to map them all on the same thing.

It is very well possible that you were planning to do this in a follow-up, but looking just at this PR that isn't immediately clear.

eboasson:
Just saw your message on Gitter @aohwang , good of you to ping me!

I'm holding two conflicting opinions in mind — [perhaps I should be happy about that? 🤔](https://quoteinvestigator.com/2020/01/05/intelligence/) — on this. The one says: the `struct sockaddr` abstraction exists precisely to handle all these different formats. I've never done any bluetooth programming myself, but a bit of googling suggests that there various `sockaddr` variants for it. So then why handle IP or Ethernet differently from Bluetooth?

The other is that `sockaddr`s are relatively unpleasant things to work with and it really would be much nicer to have proper definitions and not have `sockaddr` in the `ddsrt` interface. But then we need to define our own abstraction, and it is going to be identical to `sockaddr` for all intents and purposes, something like:
```C
enum ddsrt_address_family { DDSRT_IP4, DDSRT_IP6, DDSRT_ETHERNET, DDSRT_BLUETOOTH };
struct ddsrt_address {
  enum ddsrt_address_family af;
  union {
    unsigned char ip4[4];
    unsigned char ip6[16];
    unsigned char ethernet[6];
    unsigned char bluetooth[whatever];
  } u;
};
```
This option feels a bit like re-abstracting addresses just for the sake of it ...

Still, I think the consistency and precise typing of the second option would be better than using `sockaddr` for some and something else for some others, and it would eliminate the very few places where `sockaddr` makes an appearance in the "generic" code.

The alternative of using `ddsi_locator_t` everywhere is not the right one, I think: we have to deal with the distinction between UDP/TCP in the locators but when we're dealing with a network interface we can't make that distinction, it is just an IP address and it can be used for both UDP and TCP (if both are supported by the platform).

In short, I like what you're proposing to do (cleaning up the definition of `ddsrt_ifaddrs`), I just think it is never going to succeed without completely eliminating `struct sockaddr` from there.

thijsmie:
I have to agree with @eboasson's opinion here. There is a cleanup here that could improve the code, this isn't it quite yet. The change also doesn't introduce any new features or make things more readable so I am going to close this now. However, we'll definitely give this another look when we are inevitably forced to redesign this particular structure.

PR #1242, opened 2022-04-13
Fixes for XTypes interoperability

dpotman:
This fixes a number of issues that were found when testing interoperability for XTypes functionality:
- The IDL for `TypeLookup_Reply` in the XTypes spec has a typo (OMG issue [DDSXTY14-39](https://issues.omg.org/issues/DDSXTY14-39)) - reply type should be using `DDS_RPC_ReplyHeader` instead of `DDS_RPC_RequestHeader`. 
- Use the proxy participant guid in the `instanceName` field in a type lookup request
- The types reused from DDS RPC in the type lookup request and response need extensibility final, other types should have the default extensibility (appendable) - OMG issue DDSXTY14-30
- Default value for the try_construct flag in struct and union members and collection element type should be `01` (discard) and not `00` (invalid).
- Accept data from a non-cyclone remote writer before the reliable handshake is completed because other implementations rely on this, e.g. for the type lookup service

Note: this PR is based on #1171, only commits starting from 9dddcacd234c0dd2e28ce47bb76a9b43cdfdeedf are part of this PR

### DIFF
--- a/src/core/ddsc/src/dds_domain.c
+++ b/src/core/ddsc/src/dds_domain.c
@@ -489,7 +489,8 @@ static dds_return_t wait_for_type_resolved(
   }
   ddsrt_mutex_unlock (&gv->typelib_lock);
 
-  if (!ddsi_tl_request_type (gv, ddsi_type_id, include_deps))
+  // FIXME: provide proxy pp guid
+  if (!ddsi_tl_request_type (gv, ddsi_type_id, NULL, include_deps))
   {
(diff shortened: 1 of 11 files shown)


eboasson (approved):
LGTM! Too bad I picked some nits in #1171, else I could've merged all of this right now 🙂

dpotman:
The force-push rebases this PR on the commits I added to #1171, and fixes the try-construct flags for the union discriminator in type objects.

eboasson:
@dpotman looks like it may need another rebase

PR #1253, opened 2022-04-20
Use only functions that are aware of loaned samples when dealing with…

sjames:
loaned pointers

Loaned samples are just pointers to the iceoryx buffers. CDDS allows
bindings to wrap the sample in their own structure.  A binding needs
to know when the sample pointer is a loaned pointer so it doesn't
attempt to dereference it as a custom sample structure.

When iceorx is available on a reader we route the calls through the
ddsi_serdata_from_loaned_sample. The binding will know to handle
it properly.


thijsmie:
Hi @sjames!
Sorry to leave this for so long! This has been taken over by other events in the meanwhile, which were also brewing in the background by the time you submitted this PR. @reicheratwork did the hard work of restructuring the way we do shared memory transfer and loans and we hope to get to merging that sooner rather than later. Thank you for making the effort of making this fix though!

PR #1255, opened 2022-04-20
Remove wrong inline in ddsi_cdrstream.c

vectorsli:
The function is declared as export in the header. Defining the implementation as `inline` breaks building a static library.

### DIFF
--- a/src/core/ddsi/src/ddsi_cdrstream.c
+++ b/src/core/ddsi/src/ddsi_cdrstream.c
@@ -1282,7 +1282,7 @@ static inline bool dds_stream_write_bitmask_arr (dds_ostream_t * __restrict os, 
   return dds_stream_write_bitmask_arrLE ((dds_ostreamLE_t *) os, insn, addr, num, bits_h, bits_l);
 }
 
-inline const uint32_t *dds_stream_write (dds_ostream_t * __restrict os, const char * __restrict data, const uint32_t * __restrict ops)
+const uint32_t *dds_stream_write (dds_ostream_t * __restrict os, const char * __restrict data, const uint32_t * __restrict ops)
 {
   return dds_stream_writeLE ((dds_ostreamLE_t *) os, data, ops);
 }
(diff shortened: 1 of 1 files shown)


eboasson (commented):
That makes me wonder how it can ever work in non-static builds!

Anyway, you've only fixed the little-endian case, there is an identical definition for big-endian machines (line 1346 or thereabouts). Furthermore, it looks to me that `dds_stream_write` has the same problem, it is probably simply that it isn't (currently) being called from elsewhere.

If you update it so that all 4 definitions lose the `inline` it covers all obvious cases and I'll be happy to merge it.

vectorsli:
For shared builds, the header defines the symbol to be exported by the DLL. This seems to overwrite the `inline` declaration. For static builds, the function is declared as `extern "C"` but somehow it's still discarded by the compiler.

I'll update the other places as well.

For context: I'm trying to write a proper conan recipe for `cyclonedds` and for that static builds are necessary.
To support static builds I had to make some changes (remove PDB installation, remove LTO and more).
Specifically the PDB installation and LTO could easily be hidden behind a CMake `option` so they could be disabled without the need for patching. Is that something I should contribute, too?

eboasson:
> For shared builds, the header defines the symbol to be exported by the DLL. This seems to overwrite the `inline` declaration. For static builds, the function is declared as `extern "C"` but somehow it's still discarded by the compiler.

Ah, I see, yes, that makes some sense. There's always been something "funny" with making inlines external, for example it turned out that we used to do didn't sit nicely with MinGW, and that's for the cases where everything lined up.

> I'll update the other places as well.

Great, thanks!

> For context: I'm trying to write a proper conan recipe for `cyclonedds` and for that static builds are necessary. To support static builds I had to make some changes (remove PDB installation, remove LTO and more). Specifically the PDB installation and LTO could easily be hidden behind a CMake `option` so they could be disabled without the need for patching. Is that something I should contribute, too?

I think a `cyclonedds` conan package would be really neat, and it is in everyone's interest if it doesn't require patching CMake files, so, yes, it seems to me making those optional would be a nice improvement and a PR would be welcome.

eboasson (approved):
LGTM!

PR #1258, opened 2022-04-21
fix zero pointer problem when we use tcp mode.

johnflj:
Signed-off-by: john.fan <john.fan@autocore.ai>

### DIFF
--- a/src/core/ddsi/src/ddsi_tcp.c
+++ b/src/core/ddsi/src/ddsi_tcp.c
@@ -805,6 +805,10 @@ static dds_return_t ddsi_tcp_create_conn (ddsi_tran_conn_t *conn_out, struct dds
   struct ddsi_tran_factory_tcp * const fact = (struct ddsi_tran_factory_tcp *) fact_cmn;
   (void) qos;
   (void) port;
+  struct ddsi_domaingv const * const gv = fact->fact.gv;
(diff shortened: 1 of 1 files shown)


johnflj:
hi @eboasson I did it.

eboasson:
@johnflj thanks for the paperwork and also for checking the difference in test results. It is much appreciated!

It looks like that failure is a spurious one. We do suffer a bit from some test flakiness, mostly (but not exclusively 😢) apparently caused by the CI environment. We run multiple tests in parallel with generally fairly tight timeouts, to speed things up and to have a better chance of catching "interesting" events. Running them in parallel gives more varied timing and makes them interact at the DDS discovery level. But it does get interesting when some stress test that creates/deletes hundreds of readers or writers screws up timing or packet delivery over the loopback interface for some other test.

Generally this works well, but there are the occasional failures. I have seen traces that I can really only explain by assuming a process didn't get CPU time for a number of seconds, so some of these failures are (almost) certainly caused by the cloud services themselves.

I *could* just run the tests one at a time. That would most likely eliminate all flakiness, the ROS 2 CI does that and there hasn't been a false positive there as far as I can recall. But I like it getting exercised a little bit better, and so far anyway, I have accepted that it means I regularly need to look carefully at CI results that report failures ...

PR #1259, opened 2022-04-21
Add cmake options to facilitate static builds

vectorsli:
Building a static version is relevant when creating a conan package.
More changes might be needed, we'll see.


eboasson (approved):
Thanks @vectorsli !

PR #1263, opened 2022-04-22
(idlc) Add options for output directory and path reconstruction / prefix

### LINKED ISSUE #801: IDL compiler option to specify the output directory for the generated code

### LINKED ISSUE #1220: idlc does not have an option to respect directory structure (feature?)
I'm working on a project that makes use of a large collection of generated IDL files. 
The structure of the files is fairly complex with 4-5 layers of sub-directories. Many of the IDL files have `#include` headers for dependencies to other structs, modules, etc. within this structure
```
Base
---- Sub1
-------- Sub1A
------------ Sub1AA
---------------- Sub1AAA.idl
------------ Sub1AB
---------------- Sub1ABA.idl
------------ Sub1AC
---------------- Sub1ACA.idl
-------- Sub1B
-------- Sub1C
-------- ....
---- Sub2
---- Sub3
---- Sub4
---- ....
```
When using `idlcxx_generate` with cmake, all the files are just thrown into the cmake bin directory and don't retain any of the directory information, **but** the `include` directives are preserved which leads to compilation errors.

## Example

**TL;DR** You have to manually recreate the directory structure or use an external script, I want a flag for idlc to create a directory structure matching the path from a user-defined root. I'll do the PR myself if it's something you're open to

I have this directory structure with IDL files
image
### IDL code
https://gist.github.com/ahoenerBE/812a9baaeef2c63c652357bb728f59da

### idlcxx_generate output
```
file(GLOB_RECURSE idl_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} idl/**/*.idl)

idlcxx_generate(TARGET test_idls
        FILES ${idl_files}
        INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}/idl
        FEATURES case-sensitive
        DEFAULT_EXTENSIBILITY appendable
        WARNINGS no-implicit-extensibility
        )
```
image

Inside the files, the idl `#includes` have been translated correctly to the right output headers, but they reference directories that don't exist
image
```
FAILED: CMakeFiles/ddscxxHelloworldPublisher.dir/Item.cpp.o 
/usr/bin/c++  -I/home/drew/CLionProjects/CycloneTests/cmake-build-debug -isystem /home/drew/CLionProjects/cyclonedds-cxx/install/include/ddscxx -isystem /home/drew/CLionProjects/cyclonedds/install/include -g -std=gnu++17 -MD -MT CMakeFiles/ddscxxHelloworldPublisher.dir/Item.cpp.o -MF CMakeFiles/ddscxxHelloworldPublisher.dir/Item.cpp.o.d -o CMakeFiles/ddscxxHelloworldPublisher.dir/Item.cpp.o -c /home/drew/CLionProjects/CycloneTests/cmake-build-debug/Item.cpp
In file included from /home/drew/CLionProjects/CycloneTests/cmake-build-debug/Item.cpp:9:
/home/drew/CLionProjects/CycloneTests/cmake-build-debug/Item.hpp:12:10: fatal error: Base/Inventory/ItemCategory.hpp: No such file or directory
   12 | #include "Base/Inventory/ItemCategory.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Either using a script or manually recreating the directory structure allows the program to compile
image

## Request
Could there be an idlc flag to recreate a directory structure starting at a certain root, like the include path flag?
If so, would you be open to a PR for it, I'd like to have it integrated in the tool if I could

ahoenerBE:
Updated IDLC to allow for an output directory flag, as well as the option to specify how to preserve a directory structure.
* Added -d command line option to specify an output directory which may or may not exist.
* Added -b command line option to specify a base path for a directory structure.
* When using -b, any directories from <base_dir> to <file> will be reconstructed in the output directory. If no output directory is specified, this will result in files being compiled in-place.
* Necessary directories will be constructed automatically to hold the output files. 

This PR uses a bit of code from @k0ekk0ek's PR for creating the directory structure (`idl_mkpath`).

Also updated Generate.cmake to allow the idlc_generate function to make use of the new options.
* -d is added as a default option to the `add_custom_command` call but is not available to the outside caller since the output should usually be the output directory.
* Prefix directory can be specified with `BASE_DIR` argument, cmake will do the same path calculation that idlc does to determine where the output files go.

Fixes #801
Fixes #922  
Fixes #947 
Fixes #1220

### DIFF
--- a/src/idl/src/file.c
+++ b/src/idl/src/file.c
@@ -414,3 +414,158 @@ idl_retcode_t idl_mkpath(const char *path)
   err_full_path:
   return ret;
 }
+
+/**
+ * \verbatim
+ *  Given inputs path, output_dir, base_dir, out_ext, constructs the correct output path for the file.
+ *  Unless specified otherwise, will also create intermediate directories to allow for immediate access
+ *  to the file on return.
+ *
+ *  If the input file is absolute, the output path will always be in the form of
+ *  cwd/file.ext OR output_dir/file.ext
+ *  File is derived from the end of `path`, any directories `a, b, c, etc.` are derived from rel. path difference
+ *  between `path` and `base_dir`.
+ *
+ *  If the input path is relative, output path will be in the form of
+ *  a/b/c/file.ext OR output_dir/a/b/c/file.ext, given an input path like a/b/c/file.idl
+ *
+ *  If base_dir is provided, output_path will take the form of
+ *  b/c/file.ext OR output_dir/b/c/file.ext given
+ *  input path /a/b/c/file.idl and a base path /a
+ *  Memory will be allocated to ensure `path` is absolute, to use idl_relative_path and determine the common path.
+ *  This will return an error if the path to the base_dir requires us to go up the tree,
+ *  which could lead to unwanted behavior.
+ *  (e.g. path=/a/b/c/d/file.idl, base_path=/a/b/c/f => ../d/file.idl)
+ * \endverbatim
+ *
+ * @param path          Path (abs / rel) to the input file. Non NULL
+ * @param output_dir    Path (abs / rel) to the desired output directory. Can be NULL
+ * @param base_dir      Absolute path to the "base directory" of a tree. Can be NULL
+ * @param out_ext       Desired output extension. Can be NULL
+ * @param out_ptr       Pointer to output str. Non NULL
+ * @param skip_mkpath   Allows the function to skip making the output path. Used for testing, recommended False
+ * @return retcode
+ */
+idl_retcode_t idl_generate_out_file(const char *path, const char *output_dir, const char *base_dir, const char *out_ext, char ** out_ptr, int skip_mkpath) {
+  assert(out_ptr);
+
+  idl_retcode_t ret = IDL_RETCODE_NO_MEMORY;
+  const char *sepr, *ext, *file;
+  char empty[1] = { '\0' };
+  char *dir = NULL, *basename = NULL, *abs_file_path = NULL;
+  char* rel_path = NULL;
+  char *output_path = NULL;
+
+  if(base_dir && !idl_isabsolute(base_dir)) {
+    ret = IDL_RETCODE_BAD_PARAMETER;
+    goto err_dir;
+  }
+
+  sepr = ext = NULL;
+  for (const char *ptr = path; ptr[0]; ptr++) {
+    if (idl_isseparator((unsigned char)ptr[0]) && ptr[1] != '\0')
+      sepr = ptr;
+    else if (ptr[0] == '.')
+      ext = ptr;
+  }
+
+  file = sepr ? sepr + 1 : path;
+  if (idl_isabsolute(path) || !sepr)
+    dir = empty;
+  else if (!(dir = idl_strndup(path, (size_t)(sepr-path))))
+    goto err_dir;
+  if (!(basename = idl_strndup(file, ext ? (size_t)(ext-file) : strlen(file))))
+    goto err_basename;
+
+  /* replace backslashes by forward slashes */
+  for (char *ptr = dir; *ptr; ptr++) {
+    if (*ptr == '\\')
+      *ptr = '/';
+  }
+
+  if(base_dir) {
+    char* old_dir = dir;
+
+    // We need an absolute path to use idl_relative_path so grab the absolute path of our input file
+    // if we were provided relative paths
+    if(!idl_isabsolute(path)) {
+      if((idl_normalize_path(path, &abs_file_path)) < 0) {
+        goto err_rel_path;
+      }
+    } else {
+      if(!(abs_file_path = idl_strdup(path))) {
+        goto err_rel_path;
+      }
+    }
+
+    // If we run into a path error we can still recover and use existing dir
+    if(idl_relative_path(base_dir, abs_file_path, &rel_path) == IDL_RETCODE_NO_MEMORY)
+      goto err_rel_path;
+    // If root is not a parent of file path, or there was an error resolving rel path then return an error
+#if _WIN32
+    int is_reverse_path = (rel_path && (strncmp(rel_path, "..\\", 3) == 0));
+#else
+    int is_reverse_path = (rel_path && (strncmp(rel_path, "../", 3) == 0));
+#endif
+    if(is_reverse_path){
+      ret = IDL_RETCODE_BAD_PARAMETER;
+      goto err_rel_path;
+    }
+
+    // Extract the common directory to dir, free the old_dir pointer
+    dir = NULL;
+    size_t print_len = rel_path == NULL ? 0 : strlen(rel_path);
+    if (file) {
+      size_t file_len = sepr == NULL ? 0 : strlen(sepr);
+      print_len = print_len >= file_len ? print_len - file_len : 0;
+    }
+    if (!(dir = idl_strndup(rel_path, print_len))) {
+      goto err_rel_path;
+    }
+    if (old_dir && old_dir[0] != '\0')
+      free(old_dir);
+  }
+
+  sepr = dir[0] == '\0' ? "" : "/";
+  if(output_dir && output_dir[0] != '\0') {
+    if(idl_asprintf(&output_path, "%s%s%s", output_dir, sepr, dir) < 0)
+      goto err_rel_path;
+    sepr = "/";
+  } else {
+    if(!(output_path = idl_strdup(dir)))
+      goto err_rel_path;
+  }
+
+  // Allow skipping of path generation for unit testing
+  if(!skip_mkpath) {
+    if(idl_mkpath(output_path) < 0)
+      goto err_mkpath;
+  }
+
+  if (idl_asprintf(out_ptr, "%s%s%s%s%s",
+                   output_path,
+                   sepr,
+                   basename,
+                   out_ext ? "." : "",
+                   out_ext ? out_ext : ""
+                   ) < 0) {
+    goto err_outpath;
+  }
+  ret = IDL_RETCODE_OK;
+
+err_outpath:
+err_mkpath:
+  if(output_path)
+    free(output_path);
+err_rel_path:
+  if(rel_path)
+    free(rel_path);
+  if(abs_file_path)
+    free(abs_file_path);
+  free(basename);
+err_basename:
+  if (dir && dir != empty)
+    free(dir);
+err_dir:
+  return ret;
+}
--- a/src/idl/src/file.c
+++ b/src/idl/src/file.c
@@ -363,3 +363,54 @@ idl_retcode_t idl_relative_path(const char *base, const char *path, char **relpa
   *relpathp = rel;
   return IDL_RETCODE_OK;
 }
+
+#if _WIN32
+static inline idl_retcode_t idl_mkdir(const char *pathname, int mode) {
+  (void) mode;
+  return _mkdir(pathname);
+}
+#else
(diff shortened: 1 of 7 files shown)


ahoenerBE:
Apologies for all the force pushes, had a slight issue with the Generate.cmake and then I forgot to sign off my commits

src/tools/idlc/src/generator.c
```diff
@@ -395,12 +396,49 @@ idlc_generate(const idl_pstate_t *pstate, const idlc_generator_config_t *config)
...
   generator.path = file;
 
+  if(config->base_dir) {
+    char* rel_path = NULL;
+    char* old_dir = dir;
+    // If we run into a path error we can still recover and use existing dir
+    if(idl_relative_path(config->base_dir, pstate->sources->path->name, &rel_path) == IDLC_NO_MEMORY)
+      goto err_mkpath;
+    // If root is not a parent of file path, or there was an error resolving rel path
+    // TODO maybe find a common directory between the two and build from there?
```
eboasson, line 406: Regarding the TODO: I generally leaning towards being strict and predictable. Doing something smart using the common directory seems like it would be a likely cause of unexpected behaviours, and that's not nice when one is modifying the file system.

src/tools/idlc/src/generator.c
```diff
@@ -395,12 +396,49 @@ idlc_generate(const idl_pstate_t *pstate, const idlc_generator_config_t *config)
...
+    if(rel_path == NULL || strncmp(rel_path, "..", 2) == 0)
```
eboasson, line 407: Should it not be: `rel_path` starts with `..` followed by a directory separator? It is quite the edge case that someone would seriously create a directory named `...` and I dare say it is unlikely anyone would run into it by accident!

src/tools/idlc/src/generator.c
```diff
@@ -395,12 +396,49 @@ idlc_generate(const idl_pstate_t *pstate, const idlc_generator_config_t *config)
...
+      goto use_dir_path;
+    dir = NULL;
+    size_t print_len = strlen(rel_path);
+    if(sep)
+      print_len -= strlen(sep);
+    if(!(dir = idl_strndup(rel_path, print_len))){
+      dir = old_dir;
```
eboasson, line 414: `idl_strndup` failing would mean memory exhaustion, I think falling back to `old_dir` (whatever that may be exactly in this context — see my earlier remark on my not knowing the IDLC code base intimately) here won't solve anything.

src/tools/idlc/src/generator.c
```diff
@@ -395,12 +396,49 @@ idlc_generate(const idl_pstate_t *pstate, const idlc_generator_config_t *config)
...
+      goto use_dir_path;
+    }
+    if(old_dir && old_dir[0] != '\0')
```
eboasson, line 417: I'd appreciate it if you would help me understand why the `old_dir[0]!=0` is needed.

src/tools/idlc/src/generator.c
```diff
@@ -395,12 +396,49 @@ idlc_generate(const idl_pstate_t *pstate, const idlc_generator_config_t *config)
...
+use_dir_path:
+    if(rel_path)
+      free(rel_path);
+  }
+
   sep = dir[0] == '\0' ? "" : "/";
-  if (idl_asprintf(&generator.header.path, "%s%s%s.h", dir, sep, basename) < 0)
+  if(config->output_dir && config->output_dir[0] != '\0') {
+    if(idl_asprintf(&output_path, "%s%s%s", config->output_dir, sep, dir) < 0)
+      goto skip_output_dir;
```
eboasson, line 427: Why would it fail, other than out-of-memory? I'd bail out and avoid the somewhat inelegant `goto` in the process.

src/tools/idlc/src/idlc.c
```diff
@@ -544,6 +546,15 @@ static const idlc_option_t *compopts[] = {
...
     "no-eof-newline, no-enum-consecutive. " },
+  &(idlc_option_t){
+    IDLC_STRING, { .string = &config.output_dir }, 'd', "", "<directory>",
+    "Set the output directory for compiled files." },
+  &(idlc_option_t){
+    IDLC_STRING, { .string = &config.base_dir }, 'b', "", "<directory>",
+    "Enable directory reconstruction starting from <base_dir> "
+    "Attempts to recreate directory structure matching input. "
+    "Use without the -d option will compile IDL files in-place. "
+    "Fail-silent if root_dir is not a parent of input" },
```
eboasson, line 557: See above — there may be a perfectly valid reason for preferring this behaviour, perhaps is obvious as guarding against possible undetectable edge cases and TOCTTOU issues if someone is moving files at the same time. My inclination is that it would be best to consider it an error.

eboasson (commented):
This is great, thanks @ahoenerBE ! Yes, I do have some comments, but as you can see they are a mixture of nitpicking and musings and the odd error path that I believe is incorrect. All of that is minor stuff that I am sure won't take much effort to resolve.

@reicheratwork: would you be so kind as to look at what it takes to also support this feature the C++ backend? It looks to me that would be a pretty straightforward port, but I haven't looked at the details of that at all.

src/idl/src/file.c
```diff
@@ -363,3 +363,57 @@ idl_retcode_t idl_relative_path(const char *base, const char *path, char **relpa
...
+      /* skip ahead to first separator to skip drive */;
+
+    while (*ptr)
+    {
+      for (++ptr; *ptr && *ptr != '/' && *ptr != sep; ptr++)
+        /* search for next segment */;
+      assert(ptr[-1] != '/' && ptr[-1] != sep);
+      chr = *ptr;
+      *ptr = '\0';
+      if ((ret = idl_mkdir(full_path, 0755)) == -1 && errno != EEXIST)
```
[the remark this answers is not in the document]
ahoenerBE: `0755` was what the other PR I referenced used so I just included it. I think you're right that `0777` would work better

reicheratwork:
@ahoenerBE : You managed to steal some of my thunder :), I was also working on something in this direction, but I was not 100% ready to create a PR out of it.

I do have some questions/remarks:

1. Currently there is no way of using this functionality in other backends for idlc without duplicating a lot of the code, like if you want to generate C++ files with extensions .cpp/hpp. Maybe it would be a good idea to have an "output file generation function" or the like, which takes the input file path, an output file extension and an optional output directory as parameters, and generate the output file from that. And then have this function also accessible for other backends.
2. Currently the -d option is already "reserved" (even though the code is commented out on `idlc.c:503-509`) for displaying debug information. So we will either have to use a different option for the output directory (-o comes to mind) or use a different option for the displaying the debug information. @eboasson your suggestion on this?
3. There currently are no unittests for this functionality. This requires the output path generation functionality to be more accesible from the outside. This also points towards the first point I made, as that would also then make other backends' lives easier.

Maybe you could have a look at what I put together on my branch [reicheratwork:output-directory](https://github.com/reicheratwork/cyclonedds/tree/output-directory) and see if we can synthesize something with the best of both of our ideas.

ahoenerBE:
@reicheratwork I agree, it does seem like a lot of duplicated code to maintain. 
* I think a combo between what you have and what I have could easily be done, a lot of the functionality is already there to create paths because of the prefix feature. I hope you don't mind if I steal/rework a bit of your unit test code since you were kind enough to do it already :)
* I'll wait for @eboasson on the `-o` flag input, I completely missed that `-d` was reserved.

eboasson:
Regarding `-o` vs `-d` for the output directory: I don't actually care all that much, but I suppose `-o` is an equally good mnemonic for "output directory", whereas I can't think of one that's much better than `-d` for debugging settings. It is also the closest in spirit to what `gcc` does, so there's a reasonable chance it will help people to guess correctly when they are too lazy to check the documentation.

So let's do `-o` for the output directory.

src/tools/idlc/src/idlc.c
ahoenerBE: I guess that makes sense. My thought process was that it shouldn't be allowed to create directories backwards from the "base folder", since that defeats the purpose of having preserved directory structure. They should be in the same tree.
But if that fails they could still compile in the working/output dir if they fail that check? I guess writing it out now it makes more sense just to not allow it.
Do you agree that it should fail with a potential reverse path though?

reicheratwork:
> @reicheratwork I agree, it does seem like a lot of duplicated code to maintain.
> 
> * I think a combo between what you have and what I have could easily be done, a lot of the functionality is already there to create paths because of the prefix feature. I hope you don't mind if I steal/rework a bit of your unit test code since you were kind enough to do it already :)
> * I'll wait for @eboasson on the `-o` flag input, I completely missed that `-d` was reserved.

@ahoenerBE  go right ahead and take whatever you want, i'd rather have that it finds some use in adding to your PR than it just fading away...

ahoenerBE:
Fixed up the requested changes, moved the output directory flag from `-d` to `-o`.
Added some unit tests based on @reicheratwork branch

reicheratwork (approved):
Looks very good to me!

This should make it much easier to integrate the output file generation in other backends and keep everything straight between them and the base CycloneDDS.

eboasson:
CI on Windows is failing because `pip install --upgrade pip` now fails on Windows. #1286 fixes that, so once that one is in, I'll kick off the CI again just to be sure this PR doesn't cause any trouble on Windows.

eboasson:
Bummer, it does break on Windows:
```
 101/1268 Test  #104: CUnit_idl_file_out_file_generation .............................................................***Failed    0.14 sec


     CUnit - A unit testing framework for C - Version 2.1-3
     http://cunit.sourceforge.net/


Suite idl_file, Test out_file_generation had failures:
    1. D:\a\1\s\src\idl\tests\file.c:346  - CU_ASSERT_STRING_EQUAL(out,test.expected_out)
    2. D:\a\1\s\src\idl\tests\file.c:343  - CU_ASSERT_EQUAL(ret,test.ret)
    3. D:\a\1\s\src\idl\tests\file.c:346  - CU_ASSERT_STRING_EQUAL(out,test.expected_out)
    4. D:\a\1\s\src\idl\tests\file.c:343  - CU_ASSERT_EQUAL(ret,test.ret)
    5. D:\a\1\s\src\idl\tests\file.c:346  - CU_ASSERT_STRING_EQUAL(out,test.expected_out)

Run Summary:    Type  Total    Ran Passed Failed Inactive
              suites     13      1    n/a      0       12
               tests    120      1      0      1        6
             asserts     38     38     33      5      n/a

Elapsed time =    0.002 seconds
```
(I haven't at all looked at the details, just copy-pasting from the relevant CI run, which isn't the one linked to by GitHub anymore.)

ahoenerBE:
@eboasson bummer, is there any way I can see which tests failed? I don't have a windows machine. Should I try a VM?

ahoenerBE:
Got it, I kind of forgot that windows uses backslashes in the function, but the unit test caught it.
Just missed it because I don't use Windows oops

eboasson (approved):
LGTM, thank you very much @ahoenerBE!

PR #1267, opened 2022-04-26
Delete deprecated shared memory settings

eboasson:


thijsmie:
Removing deprecated settings post-0.9.0 seems like a reasonable strategy. In fact, the more we can responsibly prune the better. LGTM!

PR #1269, opened 2022-05-02
Fix manual liveliness with infinite lease duration

dpotman:
This fixes the issue that the lease renewal for writers incorrectly checks for a non-null writer lease, which is not initialized in case the lease duration is infinite. Using a manual liveliness policy with an infinite lease on a writer duration caused a segfault. Similar to the implementation for proxy writers, the renewal code should check for a non-infinite lease duration.

Fixes https://github.com/ros2/rmw_cyclonedds/issues/390#issuecomment-1111871668

### DIFF
--- a/src/core/ddsi/src/q_entity.c
+++ b/src/core/ddsi/src/q_entity.c
@@ -4005,6 +4005,10 @@ static dds_return_t new_writer_guid (struct writer **wr_out, const struct ddsi_g
       }
     }
   }
+  else
+  {
+    wr->lease = NULL;
+  }
 
   return 0;
 }


src/core/ddsi/src/q_entity.c
```diff
@@ -3688,7 +3688,7 @@ void writer_set_alive_may_unlock (struct writer *wr, bool notify)
   {
     if (wr->xqos->liveliness.kind == DDS_LIVELINESS_MANUAL_BY_PARTICIPANT)
       participant_add_wr_lease_locked (wr->c.pp, wr);
-    else if (wr->xqos->liveliness.kind == DDS_LIVELINESS_MANUAL_BY_TOPIC)
+    else if (wr->xqos->liveliness.kind == DDS_LIVELINESS_MANUAL_BY_TOPIC && wr->lease_duration != NULL)
```
eboasson, line 3691: I did a bit of digging and I came up with the following table:
```
QK xqos.liveliness.kind (A, MbP, MbT)
QD xqos.liveliness.lease_duration
D lease_duration
L lease
RENEW lease to renew

QK    QD    D      L      D is stored where?   L is stored where?      RENEW
A     d     ≠null  undef  pp->ldur_auto_wr     -                       -
A     ∞     null   undef  -                    -                       -
MbP   d     ≠null  ≠null  -                    pp->leaseheap_man       minl_man
MbP   ∞     null   undef  -                    -                       minl_man*
MbT   d     ≠null  ≠null  -                    global lease heap       L
MbT   ∞     null   undef  -                    -                       -

"undef" means: not actually initialised to null, which is a problem for (MbT, ∞) cos it checks for L≠null
checking for D≠null does indeed fix that, so does initializing L

"minl_man*" in RENEW: one would think minl_man can’t be null if (MbP, d ≠ ∞)
```
If I got that right, this additional check for `lease_duration != NULL` is superfluous.

src/core/ddsi/src/q_entity.c
```diff
@@ -4005,6 +4005,10 @@ static dds_return_t new_writer_guid (struct writer **wr_out, const struct ddsi_g
       }
     }
   }
+  else
```
eboasson, line 4008: This takes care of the "undef"s in the table above, except for the (A,d) case. That was always harmless, but I do think it wiser to make sure that `wr->lease` is initialised to a null pointer when not used.

Initialising `wr->lease` to a null pointer may have been the intent in `new_writer_guid_common_init`, but I have a slight preference for doing it here like you did, because it keeps setting the `lease` pointer in a single place. One could easily argue the other way as well.

src/core/ddsi/src/q_receive.c
```diff
@@ -2321,7 +2321,7 @@ static void handle_regular (struct receiver_state *rst, ddsrt_etime_t tnow, stru
     lease_renew (lease, tnow);
   if ((lease = ddsrt_atomic_ldvoidp (&pwr->c.proxypp->minl_man)) != NULL && renew_manbypp_lease)
     lease_renew (lease, tnow);
-  if (pwr->lease && pwr->c.xqos->liveliness.kind == DDS_LIVELINESS_MANUAL_BY_TOPIC)
+  if (pwr->c.xqos->liveliness.lease_duration != DDS_INFINITY && pwr->c.xqos->liveliness.kind == DDS_LIVELINESS_MANUAL_BY_TOPIC)
```
eboasson, line 2324: Whether checking the lease duration (here in the QoS, elsewhere using the `lease_duration` field) against ∞ instead of the lease against a null pointer is clearer I don't know ... In any case, it is not going to make a difference. This is better for consistency, but at the same time, it relies more heavily on keeping track of how the various fields relate to each other (see table) and that's a downside.

src/core/ddsi/src/q_entity.c
dpotman: You're right, the check for lease duration ≠ infinite makes this check superfluous.

dpotman:
@eboasson I've updated this PR so that it only adds initialization of `wr->lease` to `NULL` in case of an infinite lease duration. My idea for the other changes in this PR was to make the checks consistent for writers and proxy writers (both checking for a non-infinite lease duration), but I agree that some refactoring of this code will make it more readable (and may result in slightly improved performance). I'll copy your comments to a separate issue.

eboasson (approved):
Thanks!

PR #1271, opened 2022-05-09
fix idlc segfault while handling symlinks

anhlinh123:
If the idl file is a sym link, idlc will crash. The reason is that while the input (idl) file is normalized and symlink dereferenced (by idlcpp), some idlc code still keep the symlink (without dereferencing), and simply check the file by strcmp.
The fix replaces strcmp by a helper function which checks if two paths are actually pointing to the same file.
Be noted that I'm not a Windows guy, so I don't know how to do that on Windows.

### DIFF
--- a/src/idl/src/directive.c
+++ b/src/idl/src/directive.c
@@ -16,6 +16,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if SYS_FAMILY == SYS_UNIX
+#include <sys/stat.h>
+#endif
+
 #include "idl/processor.h"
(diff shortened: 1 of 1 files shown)


eboasson:
Hi @anhlinh123, I am very sorry that I overlooked your PR and that I only noticed it today when finally looking into #1399. You found a problem and donated a fix — that's definitely not something that should have been ignored.

I am closing it now because it appears #1324 fixed the problem as a side effect. If we got that wrong and the problem still exists, please reopen it. I'll be sure to notice it if you do!

Issue #1268, opened 2022-04-27
QNX build error


trittsv:
Hello,

i am trying to compile cyclonedds for QNX 7.1 with qcc 8.3 libc++ and got the following build error:

May this error is releated with ipv6? (Can i disable ipv6?)

```
[2022-04-27T07:03:35.983Z]  >> Using specified default base profile: x86_64-qnx-neutrino-7.1_qcc-8.3-libc++
[2022-04-27T07:03:35.983Z] 
[2022-04-27T07:03:35.983Z]  >> **************************************************
[2022-04-27T07:03:35.983Z] 
[2022-04-27T07:03:35.983Z] ****************************************************************************************************
[2022-04-27T07:03:35.983Z] +-------------------------------------------------+
[2022-04-27T07:03:35.983Z] | Profile                                         |
[2022-04-27T07:03:35.983Z] |-------------------------------------------------|
[2022-04-27T07:03:35.983Z] | include(x86_64-qnx-neutrino-7.1_qcc-8.3-libc++) |
[2022-04-27T07:03:35.983Z] |                                                 |
[2022-04-27T07:03:35.983Z] | [settings]                                      |
[2022-04-27T07:03:35.983Z] |                                                 |
[2022-04-27T07:03:35.983Z] | [options]                                       |
[2022-04-27T07:03:35.983Z] | BUILD_IDLC=False                                |
[2022-04-27T07:03:35.983Z] | BUILD_EXAMPLES=False                            |
[2022-04-27T07:03:35.984Z] | BUILD_TESTING=False                             |
[2022-04-27T07:03:35.984Z] | [env]                                           |
[2022-04-27T07:03:35.984Z] |                                                 |
[2022-04-27T07:03:35.984Z] | [build_requires]                                |
[2022-04-27T07:03:35.984Z] +-------------------------------------------------+
[2022-04-27T07:03:35.984Z] 
[2022-04-27T07:03:35.984Z] [conan_create]
[2022-04-27T07:03:35.984Z] 
[2022-04-27T07:03:35.984Z]  >> Calling 'conan create'
[2022-04-27T07:03:35.984Z] +--------------------+-------------------------------------------------------------------------------------+
[2022-04-27T07:03:35.984Z] | Configuration      | value                                                                               |
[2022-04-27T07:03:35.984Z] |--------------------+-------------------------------------------------------------------------------------|
[2022-04-27T07:03:35.984Z] | name               | cyclonedds                                                                          |
[...]

PR #1272, opened 2022-05-09
optional ipv6 support

trittsv:
Make ipv6 support optional to be able to build for limited targets like qnx 7.0.0 and 7.1.0.

ref #1268

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ There are some configuration options specified using CMake defines in addition t
 * `-DENABLE_DEADLINE_MISSED=NO`: to exclude support for finite deadline QoS settings
 * `-DENABLE_TYPE_DISCOVERY=YES`: to include support for type discovery and checking type compatibility (likely to become enabled by default in the future)
 * `-DENABLE_TOPIC_DISCOVERY=YES`: to include support for topic discovery (requires `-DENABLE_TYPE_DISCOVERY=YES`; somewhat likely to become enabled by default in the future)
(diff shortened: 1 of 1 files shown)


trittsv:
check for `struct sockaddr_in6` was not enough on qnx 7.0.0 to determine if ipv6 is supported.

Hint: For qnx 7.0.0 i had to disable `ENABLE_SOURCE_SPECIFIC_MULTICAST` too.

eboasson (commented):
Looks good to me, but there are two things. The first is a formality: the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php), anyone contributing anything to Cyclone DDS needs to sign the ECA, such are the rules of the Eclipse Foundation.

Secondly, I recently added some configuration options to the README (though I seem to have overlooked `ENABLE_SOURCE_SPECIFIC_MULTICAST`). I think it would be worthwhile adding these two options. It would also probably be good to add a note that disabling these may be needed on QNX.

Would you be willing to do that?

trittsv:
@eboasson yeah i also noticed that ```ENABLE_SOURCE_SPECIFIC_MULTICAST``` is missing in the README but i luckily found it by searching through the cmake files.
I will fix this along with your other suggestions in this Pull-Request.
Just have to figure out how to sign the ECA but i guess its not that big of a deal.

eboasson (approved):
Thanks @trittsv !

PR #1278, opened 2022-05-12
Update to 0.9.1

eboasson:
This a clean cherry-pick from a handful of commits from master, plus a version number bump:

* A typing mistake in a return value, confusing a `dds_return_t` with a `bool`, hidden by C's lovely — not! — type system, caused it never reject invalid "type information" in a discovery message
* Apparently just to be nasty, the XTypes spec has the default setting for the `try_construct` encoded in the type objects as `0b01`, and forbids the expected `0b00` generated by Cyclone's IDL compiler (Cyclone accepts both and treats the same, and in this release interoperability with other implementations is very limited regardless of this detail)
* Fix for the crash with manual-by-topic with lease duration ∞
* Fix for a crash with TCP use
* Fixes for QNX (optional IPv6, fix for `mutex_unlock` that should have been `mutex_destroy`)
* Fix for an incorrect handling of array bounds in type object interpretation


eboasson:
ROS 2 CI for Humble

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16687/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11274/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17088/)

thijsmie (approved):
Verified interoperability, looks good to me!

PR #1281, opened 2022-05-16
Participant properties

thijsmie:
This PR adds several properties to a participant upon creation. Implemented here are:
 * *__ProcessName*
 * *__Pid*
 * *__Hostname*
 * *__NetworkAddresses*
 * *__DebugMonitor*

This also contains a small fix for a bug in the security plugin that was unearthed by the addition of these properties, as well as a slight rework of some security tests that relied on knowing the exact structure of the properties QoS policy (but it now depends on runtime values with the added properties).


src/core/ddsc/src/dds_serdata_builtintopic.c
```diff
@@ -99,6 +106,53 @@ static struct ddsi_serdata_builtintopic *serdata_builtin_new(const struct ddsi_s
...
+  if (!arg->buf) return;
+
+  ddsi_xlocator_to_string (buf, sizeof(buf), n);
+  const size_t nsize = strlen(buf) + (arg->first ? 0 : 1);
+
+  if (nsize > arg->buf_size - arg->buf_pos - 1) {
+    arg->buf_size += 4 * nsize;
+    arg->buf = ddsrt_realloc(arg->buf, arg->buf_size);
+
+    if (!arg->buf) return;
```
eboasson, line 130: That's a memory a leak (`realloc` is really annoying ... I fell into the same trap the other day and had to be rescued by @dpotman ...)

src/core/ddsi/src/ddsi_plist.c
```diff
@@ -3615,6 +3615,42 @@ dds_qos_t * ddsi_xqos_dup (const dds_qos_t *src)
   return dst;
 }
 
+bool ddsi_xqos_add_property_if_unset (dds_qos_t *q, bool propagate, const char *name, const char *value)
```
eboasson, line 3618: This is really similar to `dds_qset_prop`, the difference being that this one doesn't replace the value. Perhaps it would make more sense to generalize `dds_qset_prop` instead?

src/core/ddsi/src/q_init.c
```diff
@@ -2039,6 +2046,15 @@ int rtps_start (struct ddsi_domaingv *gv)
       rtps_stop (gv);
       return -1;
     }
+    else {
+      ddsi_locator_t loc;
+      char buf[DDSI_LOCSTRLEN];
+
+      if (get_debug_monitor_locator(gv->debmon, &loc)) {
+        ddsi_xqos_add_property_if_unset(&gv->default_local_plist_pp.qos, true, DDS_BUILTIN_TOPIC_PARTICIPANT_DEBUG_MONITOR, \
```
eboasson, line 2054: I think I would classify that backslash as a stray one ...

eboasson (commented):
Thanks @thijsmie 🙂 I noticed a few details worth another thought during a hasty review. I'm sure you know what it is like when you do things hastily: you make more mistakes of all possible kinds ... so please be kind ... 🙂

src/core/ddsi/src/ddsi_plist.c
thijsmie: I don't know about this one. All the dds_qset/get ops are nicely contained in one place, if I generalise something I don't know what the logical place is to put the definition and implementation in a way that makes the function accessible from both dds and ddsi. Are you okay with leaving this semi-duplicate implementation or do you have a suggestion where to put the definitions for the generalised one?

eboasson (approved):
👍🏻

PR #1283, opened 2022-05-19
Add guide for interpreting the tracing output

eboasson:
This is only a first attempt. Some bits are still missing, and some no doubt be confusing, but it has been a very long time now where there has been no information on this whatsoever, and that's presumably worse than having at least documentation in place.


thijsmie (approved):
This looks great! I think it is great we add this, but how about you also make a PR that adds it [here](https://github.com/eclipse-cyclonedds/eclipse-cyclonedds.github.io/tree/master/content/guides) so it ends up on the website? 😇

PR #1286, opened 2022-05-25
Use "python -m pip" instead of "pip" on Azure

eboasson:
This way updating pip itself works again on Windows platforms.


thijsmie (approved):
I'm approving under the assumption that the CI will not yield any problems. I don't see how this change would cause any, but it is still a gamble :game_die:

PR #1291, opened 2022-05-31
Fix GCC 12 analyzer warnings

eboasson:
GCC 12 has a much improved static analyzer, and this PR takes care of most of the warnings it gives after enabling all of them.

There are still a few false positives in IDLC (generating idlc.c, generating type objects) that I have suppressed locally here, and there are some issues left in some test cases. So it is not quite ready for prime-time yet.

Another problem is that Azure only just added an Ubuntu 22.04 image, that Ubuntu 22.04 doesn't yet include gcc 12, and that the package repository of Ubuntu currently gives you gcc 12.0.1 instead of 12.1.0, and apparently a lot improved in the static analyzer between those two.

I think the changes are worth considering, even if perhaps it is still too early to actually switch platforms and enable all warnings from the static analyzer.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,12 +23,12 @@ jobs:
     vmImage: $(image)
   strategy:
     matrix:
-      'Ubuntu 20.04 LTS with GCC 10 (Debug, x86_64)':
-        image: ubuntu-20.04
+      'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64)':
+        image: ubuntu-22.04
         analyzer: on
         # not enabling "undefined" because of some false positives
         sanitizer: address
-        cc: gcc-10
+        cc: gcc-12
         coverage: on
       'Ubuntu 20.04 LTS with GCC 10 (Debug, x86_64, Iceoryx)':
         image: ubuntu-20.04
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,6 +23,10 @@ jobs:
     vmImage: $(image)
   strategy:
     matrix:
+      #gcc-12.2 is the first version of gcc's static analyzer that accepts this code
+      #without -Wno-analyzer-malloc-leak without any false positives.  Ubuntu is far
+      #behind in gcc versions, and so falling back to macOS is the most pragmatic
+      #option at this time.
       #'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64)':
       #  image: ubuntu-22.04
       #  analyzer: on
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,13 +23,13 @@ jobs:
     vmImage: $(image)
   strategy:
     matrix:
-      'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64)':
-        image: ubuntu-22.04
-        analyzer: on
-        # not enabling "undefined" because of some false positives
-        sanitizer: address
-        cc: gcc-12
-        coverage: on
+      #'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64)':
+      #  image: ubuntu-22.04
+      #  analyzer: on
+      #  # not enabling "undefined" because of some false positives
+      #  sanitizer: address
+      #  cc: gcc-12
+      #  coverage: on
       'Ubuntu 20.04 LTS with GCC 10 (Debug, x86_64, Iceoryx)':
         image: ubuntu-20.04
         #analyzer: on  # disabled for now because of warnings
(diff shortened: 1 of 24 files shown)


dpotman (approved):
LGTM!

eboasson:
@dpotman perhaps you would be willing to review this again? I had to do some work to rebase on current master, and so it is probably wise if another review is done.

src/idl/src/vector.c
```diff
@@ -0,0 +1,106 @@
...
+#include <stdint.h>
+#include <assert.h>
+
+#include "idl/vector.h"
+
+bool idl_boxed_vector_init (struct idl_boxed_vector *v)
+{
+  v->n = 0;
+  v->cap = 1;
+  v->xs = malloc (v->cap * sizeof (*v->xs));
```
dpotman, line 25: Use `idl_malloc`?

src/idl/src/vector.c
```diff
@@ -0,0 +1,106 @@
...
+  return (v->xs != NULL);
+}
+
+bool idl_boxed_vector_append (struct idl_boxed_vector *v, void *x)
+{
+  if (v->n == v->cap)
+  {
+    assert (v->cap <= SIZE_MAX / 2);
+    const size_t cap1 = 2 * v->cap;
+    void **xs1 = realloc (v->xs, cap1 * sizeof (*xs1));
```
dpotman, line 35: `idl_realloc`?

azure-pipelines.yml
```diff
@@ -23,13 +23,13 @@ jobs:
...
   strategy:
     matrix:
-      'Ubuntu 20.04 LTS with GCC 10 (Debug, x86_64)':
-        image: ubuntu-20.04
-        analyzer: on
-        # not enabling "undefined" because of some false positives
-        sanitizer: address
-        cc: gcc-10
-        coverage: on
+      #'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64)':
```
dpotman, line 26: Maybe add a comment here that explains why this is disabled?

src/idl/src/keylist.c
```diff
@@ -16,39 +16,171 @@
...
+  return boxed_vector_append_with_failure_action (&kcs->key_containers, kc, key_container_free_wrapper);
+}
+
+static const struct key_container *key_containers_first (const struct key_containers *kcs, struct key_containers_iter *it) {
+  return idl_boxed_vector_first_c (&kcs->key_containers, &it->x);
+}
+static const struct key_container *key_containers_next (struct key_containers_iter *it) {
+  return idl_boxed_vector_next_c (&it->x);
+}
+static struct key_container *key_containers_first_nc (struct key_containers *kcs, struct key_containers_iter *it) {
```
dpotman, line 153: Used suffix `_nc` not consistent with `key_container_first_c` above.

src/idl/src/keylist.c
```diff
@@ -255,26 +327,26 @@ static idl_retcode_t get_keylist_key_paths(idl_pstate_t *pstate, void *list, str
...
-  size_t nkc = 0;
+  struct key_containers *key_containers;
+  if ((key_containers = key_containers_new ()) == NULL)
+    return IDL_RETCODE_NO_MEMORY;
 
   assert(pstate);
-  if ((ret = get_keylist_key_paths(pstate, pstate->root, &key_containers, &nkc) != IDL_RETCODE_OK))
+  if ((ret = get_keylist_key_paths (pstate, pstate->root, key_containers) != IDL_RETCODE_OK))
     goto err_create_key;
+  sort_parent_paths (key_containers);
```
dpotman, line 337: `has_conflicting_keys` requires the parent paths to be sorted, so wouldn't it be better to move this call to that function?

dpotman (commented):
LGTM, the boxed vector makes the keylist implementation much easier to read! A few minor issues below.

src/idl/src/keylist.c
eboasson: Agree

azure-pipelines.yml
eboasson: Agree

src/idl/src/vector.c
eboasson: Agree

src/idl/src/keylist.c
eboasson: The rationale for doing it this way was that this allows `has_conflicting_keys` to take a `const` input. The desire for that has to do with the name: `has_...` to me suggests a predicate, not a procedure that (potentially) modifies its input. So either we leave it this way, or we have to come up with another for `has_conflicting_keys`. How would you feel about `check_for_conflicting_keys`?

dpotman: Good point, I'm fine with the `check_` variant, or just a comment on the current `has_` function that it requires sorted input.

eboasson: Done!

@dpotman If you're happy with the updates I did based on your comments, then, after nearly half a year, I'll finally merge this PR 🙂

dpotman: Yes, I am, this PR can be merged!

eboasson:
Rebase because of merging #1483, fortunately a rather trivial rebase (only two places where I had to add/change a prefix). I'm my view that means @dpotman's approval still applies 🙂.

PR #1292, opened 2022-06-02
Topic descriptor align and optimize flag

dpotman:
This PR contains a number of changes in the topic descriptor that is generated by IDLC:
- All supported compilers have support for the C11 `_Alignof` keyword, so we can rely on this for the alignment used in the topic descriptor. This replaces the custom implementation of alignment calculation in idlc, that in certain cases reported the wrong alignment when using external or optional members.
- The flag `DDS_TOPIC_NO_OPTIMIZE` in the topic descriptor is now deprecated, and replaced by an unconditional call to `dds_stream_check_optimize` when creating the sertype_default
- Fix a bug in the ops count in idlc

Commit b3fc6065938c0682f6effb1aaaebba6d93f2b04f contains the re-generated code for the xtypes typeinfo/typemap/typelookup, so that `dds_alignof` is used (and the number of ops in correct).


src/core/ddsi/src/q_radmin.c
```diff
@@ -684,7 +685,7 @@ void *nn_rmsg_alloc (struct nn_rmsg *rmsg, uint32_t size)
   RMSGTRACE ("rmsg_alloc(%p, %"PRIu32" => %"PRIu32")\n", (void *) rmsg, size, size8P);
   ASSERT_RBUFPOOL_OWNER (rbuf->rbufpool);
   ASSERT_RMSG_UNCOMMITTED (rmsg);
-  assert ((chunk->u.size % ALIGNOF_RMSG) == 0);
```
eboasson, line 687: See comment at removal of `ALIGNOF_RMSG`

eboasson (commented):
Thanks @dpotman, just one tiny little thing prevents me from merging it ... I can easily fix it if you are busy

PR #1293, opened 2022-06-02
Bugfix in xtypes wrapper

dpotman:
This fixes two issues in the xtypes type wrapper implementation: 
- Set the member name hash also in the member details in a complete typeobject. As a result of this bugfix, in the `xtypes.resolve_dep_type` test both complete and minimal types need to be modified to (intentionally) avoid a match and force type resolving. 
- Fix the clean-up of applied member annotations in `ddsi_xt_type_fini`

### DIFF
--- a/src/core/ddsc/tests/xtypes.c
+++ b/src/core/ddsc/tests/xtypes.c
@@ -55,7 +55,7 @@ static dds_entity_t g_subscriber2 = 0;
 
 typedef void (*sample_init) (void *s);
 typedef void (*sample_check) (void *s1, void *s2);
-typedef void (*typeobj_modify) (dds_sequence_DDS_XTypes_TypeIdentifierTypeObjectPair *type_id_obj_seq);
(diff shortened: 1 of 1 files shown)


src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -454,11 +454,23 @@ static void xt_lbounds_dup (struct DDS_XTypes_LBoundSeq *dst, const struct DDS_X
   dst->_buffer = ddsrt_memdup (src->_buffer, dst->_length * sizeof (*dst->_buffer));
 }
 
+static void get_namehash (DDS_XTypes_NameHash name_hash, const char *name)
+{
+  /* FIXME: multi byte utf8 chars? */
```
eboasson, line 459: If one doesn't need to interpret them, then handling as simple sequences of bytes is ok, so I think we need not worry about that

eboasson (approved):

eboasson:
I think the CI failures are unrelated to the change, do you agree?

PR #1306, opened 2022-06-13
Second attempt at fixing OSS-Fuzz builds

eboasson:


thijsmie (approved):
Yup, looks good to me!

PR #1307, opened 2022-06-14
Update "perftest" script configuration settings

### LINKED ISSUE #1303: Perftest between containers
Hi, I'm trying to run the perftest between two identical containers that both cloned and build cyclonedds, but I don´t get it to work. As the script assumes, I can ssh from one container into the other without passwords.

If I understand correctly, the script searches for bin/ddsperf which is in cyclonedds/build/bin/ so I moved perftest to cyclonedds/build/ to run it like this `./perftest root@172.17.0.3` in one container where 172.17.0.3 is the address in the other container. This runs something but everything fails. Then I saw the `-d` option that you can use to locate the remote DIR that contains /bin/ddsperf/ so I use `./perftest -d ../cyclonedds/build root@172.17.0.3` with the same failing result. 

Am I doing something stupid? Or is this maybe not even possible. Any suggestion will be really welcome. Thanks a lot.

eboasson:
This script was overlooked when updating the configuration settings on two points:

* Moving the default lease duration to the Discovery section
* Improving the way socket buffer sizes get specified

The old settings are still handled, but it shouldn't trigger any deprecation warnings.  The new way of specifying socket receive buffer sizes also distinguishes between what to ask for and what the require. This should prevent an error on many Linux systems in the default configuration.

Fixes #1303 I believe. @JesseRengers would you be willing to check this?

### DIFF
--- a/examples/perfscript/perftest
+++ b/examples/perfscript/perftest
@@ -309,10 +309,12 @@ cat >>$cfg <<EOF
 EOF
 fi
 cat >>$cfg <<EOF
+    <Discovery>
+      <LeaseDuration>2s</LeaseDuration>
+    </Discovery>
     <Internal>
       <SynchronousDeliveryPriorityThreshold>\${async:-0}</SynchronousDeliveryPriorityThreshold>
-      <LeaseDuration>2s</LeaseDuration>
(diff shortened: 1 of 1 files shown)


eboasson:
@JesseRengers thanks for checking!

PR #1321, opened 2022-06-29
Fix memory leak in type lookup request

dpotman:
This fixes a memory leak in the type lookup request message creation, in case there are no types to resolve.


eboasson (approved):
LGTM!

Issue #1319, opened 2022-06-28
Idlc fails when handling nested includes with files in different folders


ceyxiang:
Idlc will return with code 1 when processing an idl file that includes a idl file in another folder with nested includes.

Example case:
```
Directory layout
base_folder/Folder1/File1.idl
base_folder/Folder1/Folder2/File2.idl
base_folder/Folder1/Folder2/File3.idl

File1.idl
#include "Folder2/File2.idl"

File2.idl
#include "File3.idl"
```

When running idlc on File1.idl, nothing will be printed or generated. From what I have traced, it will fail at [directive.c L150](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/idl/src/directive.c#L150) as the `dir->path` would be `<Absolute Path to base_folder>/base_folder/Folder1/Folder2/Folder2/File2.idl` (duplicate `Folder2` in path) From my understanding it comes about when returning from an included file. 

Changing [this](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/idl/src/directive.c#L129) line to `if (dir->flags & (START_OF_FILE) && !idl_isabsolute(dir->path)) {` will fix the issue if idlc is called from `Folder1`. But if idlc is called from `base_folder` (with the argument to idlc as `Folder1/File1.idl`), this fix will fail as `dir->path` will now be `<Absolute Path to base_folder>/base_folder/Folder2/File2.idl` (Missing `Folder1` in path).

A solution that seems to fix this properly would be to change the preprocessor to output the absolute paths to the included files instead of their relative paths. Changing [these 2 lines](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/tools/idlpp/src/system.c#L4007-L4008) to just `name = file->full_fname;` seems to fix the issue properly.

As I am not familiar with the idlc code, I would like someone's opinion on whether the fix would potentially cause any other issues.

dpotman:
Thanks for reporting this issue! We're currently looking into this. Your proposed solution to add the absolute path in the linemarkers might have impact on the includes in generated sources, and I'd prefer to solve this issue in the IDL library, e,g, by fixing the conversion to an absolute path in `push_line`, or by changing the way the file is popped from the stack when handling `RETURN_TO_FILE`. I'll keep you updated on this.

dpotman:
[...]

PR #1324, opened 2022-07-04
Fix issue in IDLC related to nested include files

dpotman:
This fixes the issue in IDLC that when using input files that have nested include files, the conversion of the relative to absolute path in IDLC is incorrect when popping sources files from the stack in the `push_line` function. This issue resulted in no output from IDLC, but it also didn't report an error.

This commit changes the way files are popped from the stack based on the processor filemakers: it assumes that for every `START_OF_FILE` there is a corresponding `RETURN_TO_FILE` to return to the parent file, and therefore there is no need to convert the relative path to an absolute path when processing a `RETURN_TO_FILE` marker.

### DIFF
--- a/src/idl/src/directive.c
+++ b/src/idl/src/directive.c
@@ -155,13 +155,13 @@ static idl_retcode_t push_line(idl_pstate_t *pstate, struct line *dir)
     dir->path = norm;
     assert(dir->file);
 
-    /* reuse normalized filename if include is absolute */
     if (idl_isabsolute(dir->file)) {
+      /* reuse normalized filename if include is absolute */
       free(dir->file);
       if (!(dir->file = idl_strdup(dir->path)))
         return IDL_RETCODE_NO_MEMORY;
-    /* use original filename by default */
     } else {
+      /* use original filename by default */
       (void)idl_untaint_path(dir->file);
     }
 
--- a/src/idl/src/directive.c
+++ b/src/idl/src/directive.c
@@ -122,7 +122,7 @@ static idl_retcode_t push_line(idl_pstate_t *pstate, struct line *dir)
 {
   idl_retcode_t ret;
 
-  if (dir->flags & (START_OF_FILE|RETURN_TO_FILE)) {
+  if (dir->flags & START_OF_FILE) {
     char *norm = NULL;
     const idl_source_t *src = pstate->scanner.position.source;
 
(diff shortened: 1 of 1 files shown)


src/idl/src/directive.c
```diff
@@ -147,39 +147,32 @@ static idl_retcode_t push_line(idl_pstate_t *pstate, struct line *dir)
...
-        idl_error(pstate, idl_location(dir),
-          "Invalid line marker, file '%s' not on include stack", dir->path);
-        return IDL_RETCODE_SEMANTIC_ERROR;
-      }
+    /* reuse normalized filename if include is absolute */
+    if (idl_isabsolute(dir->file)) {
+      free(dir->file);
+      if (!(dir->file = idl_strdup(dir->path)))
+        return IDL_RETCODE_NO_MEMORY;
+    /* use original filename by default */
```
eboasson, line 163: I’m looking at the diff, so I may be wrong, but the indentation looks off to me here

eboasson (commented):
Apart from one nitpick … LGTM, thanks @dpotman for diving in this and fixing it!

Even if I had nothing to comment, I’d still wait a short while before merging it, hopefully those who have run into issues handling include paths can double-check that this works for them as well. I’m not going to wait for very long 🙃

dpotman:
@ceyxiang would you be willing to share the IDL you used for testing this? I have also set up the same folder structure with three IDL files, and I've added a minimal top-level type (`@topic struct type1 { type2 f1; };`) to `File1.idl` and similar IDL to `File2.idl` and `File3.idl`, so that a type descriptor and C-types are generated for `type1`. I can reproduce the error from issue #1319 without the changes in this PR. With the changes applied, I can successfully generate C and C++ output for this scenario.

src/idl/src/directive.c
dpotman: The indentation was not really off, it applies to the `else` block and matches indentation of that line. But I agree that it looks better when moved to inside the block, so I've fixed that in [e652aeb](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1324/commits/e652aeb95f5a4bfde8923b89e592aaa2f77f91fc).

ceyxiang:
Sure! Here are the idl files I am using for testing [idlc_files.zip](https://github.com/eclipse-cyclonedds/cyclonedds/files/9051246/idlc_files.zip). It could potentially be a flag I did not enable in cmake or something that could be causing this issue if it works fine on your end. 

I tested building with the absolute paths and the above files work fine

dpotman:
Thanks @ceyxiang! The crash on this IDL appeared to be a bug in the type meta-data generation for the `int8` type, which I fixed in PR #1329. With both PRs applied, it should be working and idlc generates the expected output for these IDL files.

ceyxiang:
Thank you for working on the fix!

PR #1329, opened 2022-07-08
Bugfix in idl meta data generation for int8

dpotman:
This fixes the type meta data generation for int8 IDL data type,
which was not mapped in the function to get a fully descriptive
type identifier.


eboasson (approved):
LGTM, these indeed seem to be the only places where it was missing.

Incidentally, this is exactly the reason why I prefer the contortions needed to avoid default cases in switches over enums ...

PR #1331, opened 2022-07-11
Fix assignability check for unbounded key members

dpotman:
This fixes the rules in the assignability check for bounds of string, sequence and map typed key members, so that the value 0 for unbounded is taken into account.


eboasson (approved):
Thanks!

PR #1336, opened 2022-07-15
Add a 'presence_required' option to network interfaces

### LINKED ISSUE #1330: Optional network interfaces
We are using cyclonedds on a robot that is sometimes used with an external network interface and sometimes runs without it. It would therefore be nice to be able to define network interfaces as optional to avoid the error "does not match an available interface" on startup when an interface is currently not connected. This way, it would not be necessary to change the configuration when an interface is connected or disconnected.

thijsmie:
Fixes #1330, allow users  to make interface presence optional, as requested by @timonegk.


src/core/ddsi/src/ddsi_config.c
```diff
@@ -695,6 +695,8 @@ static int if_network_interfaces(struct cfgst *cfgst, void *parent, struct cfgel
   new->cfg.automatic = false;
   new->cfg.name = NULL;
   new->cfg.address = NULL;
+  new->cfg.presence_required = true;
```
eboasson, line 698: The values to which the fields are initialised — and so for most fields, whether they are initialised at all — here are irrelevant because all the missing values are set from the default value specified in configuration table.

Strings do need to be initialised to a null pointer because it'll try to free them in the configuration parsing error handling paths.

Doesn't make setting the values here incorrect, but I thought it worth pointing out.

eboasson (approved):
LGTM, thanks @thijsmie ! I just have one remark, it is hardly worth the effort to address it ... (Approved pending CI)

timonegk:
I just tested the patch and it works for me, thank you. The only thing is that there is still the error message "...: does not match an available interface". It should probably be removed or changed to an informational log level.

thijsmie:
Some small tweaks so it gives a warning when an optional interface is missing instead of an error. I also found that the XML parser is not really happy when the config options are 'bool' and not 'int' typed so swapped that around. Next I also made sure that the 'presence_required' option now doesn't also makes it ignore mistakes in the configuration file.

thijsmie:
Fixed the merge conflicts. Is this okay to merge @eboasson?

eboasson:
Assuming the CI is happy, then yes, go ahead!

thijsmie:
One failure on CI, but it looks like Azure borked there and it is unrelated.

PR #1341, opened 2022-07-19
Fix Coverity issues introduced by typeobject → serializer code

eboasson:
Two were really much ado about nothing (yes, an unsigned value is always ≥ 0, and a suspicious memset that turned out to be correct), the other was slightly more interesting because some issues in handling an out-of-memory condition that was otherwise handled quite diligently.

I've opted to refactor the code in question a bit, splitting it into multiple functions so that the error handling is slightly more compact.


eboasson (commented):
Apparently no forthcoming reviews ... double-checked it, I really do think it is best to merge them.

thijsmie (approved):
LGTM! I appreciate the split into smaller functions 😄

PR #1344, opened 2022-07-21
PoC C program for a dynamic subscription

eboasson:
Waits for a publication of a given topic name to show up, then takes a subscription to that topic and prints the samples it receives as JSON.

Caveats:
* The JSON printer makes a lot of optimistic assumptions about the contents and doesn't support all types.
* It is quite likely that the interpretation of the type object is incorrect to begin with. But it is good enough for a start.

Other things:
* Extends the "hello world" type a bit to make it more interesting.
* Fixes some bugs in computing minimal type information given *only* complete type information. It is not the actual computation computation that was the problem, but at switching from complete type identifiers to minimal type identifiers in the type dependencies. Fixing the root cause would mean fixing the spec ...
* Changes the signature of adding entries to a hash table, the old one collided with assumptions made by the gcc/clang static analyzers, and changing it was nearly free.


thijsmie (approved):
Nice example! Good to see you need more lines of code in C then you need in python to do this stuff :wink:

PR #1348, opened 2022-07-26
Drop macOS 10 on CI, make musl libc build

thijsmie:
Sorry for combining these two small fixes, but I figured the PR is still small enough to review :smile:

Azure Pipelines is dropping macOS 10 very soon, I've already noticed a dropoff in build machine availability and more failing tests on timing, presumably due to overcrowdedness of the still available ones. This switches all builds over to macOS 11.

I am also building python wheels to run on Alpine-based containers with musl libc where I ran into the fact that `strtoull_l` is not defined. The maintainers of musl claim that a `__MUSL__` is never needed :confounded: so they don't provide one so I had to hack one together.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -73,19 +73,12 @@ jobs:
         topic_discovery: off
         idlc_testing: off # temporary dislabled because of passing -t option to idlc in this test for recursive types
         cc: clang-10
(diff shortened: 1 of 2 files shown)


eboasson (approved):
LGTM!

I agree with dropping the macOS 10.12 target. It was there for an old release of ROS 2, so that is no longer relevant.

PR #1350, opened 2022-07-29
Retrieval of historical data after partial disconnect

eboasson:
When:
- A transient-local reader/writer pair exists with a network in between, and for some reason the writer's participant's lease expires at the reader's but not the other way around (so a "partial" or "asymmetric" disconnect);
- All the data published by the writer was acknowledged and no new data is being published;
- After rediscovery of the writer by the reader, not all historical data arrives on the first NACK and the heartbeat sent by writer in its response to the first NACK is lost; then
- The reader waits forever and retrieval of the historical data stalls.

Note that this problem exists both for application transient-local data and for discovery data.

This commit restores Cyclone's old behaviour (pre-63b1a7179b5 or thereabouts) of spontaneously sending NACKs at a low rate if some data is known to be missing.

The alternative (or perhaps both should be done!) is to make the writer start sending heartbeats when it detects a reader that suffered such an asymmetrical disconnection.  It can detect this because it suddenly starts receiving requests for retransmission of already acknowledged data.

Letting the writer send heartbeats would be entirely in line with the specification, spontaneously sending retransmit requests is not. Nothing in the specification guarantees that the writer will send heartbeats in this case, and so there is no interoperable solution that adheres to the specification.

The other commits are a regression test and some changes to make that one simpler to write, as well as a slight lengthening of the interval for spontaneously sending NACKs.

### DIFF
--- a/docs/manual/options.md
+++ b/docs/manual/options.md
@@ -502,7 +502,7 @@ This setting controls the interval with which a reader will continue NACK'ing mi
 
 Valid values are finite durations with an explicit unit or the keyword 'inf' for infinity. Recognised units: ns, us, ms, s, min, hr, day.
 
-The default value is: "1 s".
+The default value is: "3 s".
 
 
 #### //CycloneDDS/Domain/Internal/BuiltinEndpointSet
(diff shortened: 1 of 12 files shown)


dpotman (approved):
LGTM! And nice usage of the oneliner test (good that you added some comments there, the syntax is somewhat hard to read 😄)

PR #1351, opened 2022-07-29
by-source-timestamp destination order and a silly crash

### LINKED ISSUE #754: The condition is always 'false'
https://github.com/eclipse-cyclonedds/cyclonedds/blob/3108aff3ee8c8b4cdb418441d1fe0925de3ef273/src/core/ddsc/src/dds_rhc_default.c#L1068

so `inst->wr_guid = wrinfo->guid;` will never be executed.

eboasson:
It turns out that #754 uncovered a nasty bug that makes the by-source-timestamp destination order work fine with 2 writers but not with 3 or more ...

Fixes #754. Thanks @secondage for spotting it and reporting it!

In the process of figuring it out and building a regression test, I ran into a crash when calling `dds_take` before the library is initialized (that means the reader handle is by definition invalid, and it should return error). That turns out to be caused by calling `lookup_thread_state` before making sure the library is initialized. This led to a test case and some trivial fixes. These are obviously unrelated to the subject matter, but they are so straightforward that I figured I could just include them here.

### DIFF
--- a/src/core/ddsc/src/dds_instance.c
+++ b/src/core/ddsc/src/dds_instance.c
@@ -66,7 +66,6 @@ static void dds_instance_remove (struct dds_domain *dom, const dds_writer *write
 
 dds_return_t dds_register_instance (dds_entity_t writer, dds_instance_handle_t *handle, const void *data)
 {
-  struct thread_state * const thrst = lookup_thread_state ();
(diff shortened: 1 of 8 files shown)


thijsmie (approved):
LGTM

PR #1352, opened 2022-08-01
Start at last_seq on first HB is Cyclone-specific

eboasson:
Cyclone DDS is careful to perform a three-way handshake to have a
well-defined starting point for reliability (and hence for whether
samples can be lost despite it being reliable and the two sides being
connected all the time).

That requires dropping data until a heartbeat has been received and
starting at the highest sequence number advertised by the writer.  In
principle, this is interoperable with other implementations, but it
turns out that for the type discovery protocol (at least for OpenDDS)
this can result in dropping the first request, which (at least for
OpenDDS) stalls type discovery and prevents matching.

Hence b6dadb3439d82412b2a7492ce31017960fbb3beb changed it to accept data
from other DDS implementations before a heartbeat is received.  This
means there may be gaps in the data sequence on starting up the reliable
session if the writer is using another DDS implementation.  If this is a
problem, just use Cyclone DDS.

What did not change is that it expects to start at sequence number 1,
and it is the first heartbeat that changes that expectation.  So if,
from that other DDS implementation, the first data arrives before the
first heartbeat and it has sequence numbers > 1, it will now be buffered
and (likely) made available on receipt of the heartbeat.

The processing of the first heartbeat remained unchanged.  This has two
problematic consequences:

* It jumps forward to the latest advertised sequence number (correct for
  a Cyclone peer but arguably wrong in the light of accepting the first
  data that comes in from another peer), and

* There may be data ready for delivery where the code asserts that there
  cannot be any.  The assert in itself fortunately disappears in a
  release build (so there is no crash), but it does leak some memory.

This commit changes the handling of the first heartbeat.  In the case of
a Cyclone writer, nothing changes.  In the case of a non-Cyclone writer,
it will now simply deliver whatever was received already (in line with
the behaviour prior to adding the handshake).

### DIFF
--- a/src/core/ddsc/tests/builtin_topics.c
+++ b/src/core/ddsc/tests/builtin_topics.c
@@ -10,9 +10,11 @@
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
 #include "dds/dds.h"
+#include "dds/ddsrt/environ.h"
 #include "dds__reader.h"
 #include "test_common.h"
 
+static dds_entity_t g_domain      = 0;
 static dds_entity_t g_participant = 0;
 static dds_entity_t g_subscriber  = 0;
 static dds_entity_t g_publisher   = 0;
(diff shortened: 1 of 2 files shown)


thijsmie:
This looks good to me, although it is hard for me to estimate the impact of this change. If you've tested it with other DDS implementations and this is a helpful change you have my approval to merge :+1:

eboasson:
I actually ran into the assertion with OpenDDS ... The referenced commit quite clearly changes the possible states of the message reordering mechanism prior to receiving the first heartbeat: before it, nothing would ever be accepted and hence the buffers would necessarily be empty and nothing could possibly be ready for delivery (the assert in question). Following it, for a non-Cyclone writer there can be data buffered away. It follows that the assertion (in the now conditional bit) is incorrect for non-Cyclone writers. I am quite distraught by having overlooked that ...

But there is a real change in here: the sequence number used for moving the state forward: before it used the highest available sequence number in all cases; now, for non-Cyclone writers, it uses the first available sequence number. This is in line with what data is being accepted and it matches behaviour from before the wait-for-first-heartbeat behaviour was introduced.

So in a sense, it really restores the historical behaviour when interoperating. That seemed an ok choice, even if I had a nagging little bit of doubt whether that historical behaviour was correct to begin with: and that's for the combination of a non-Cyclone DDS transient-local writer and a volatile reader.

I am pretty sure I can and should do a little experiment with that. It has been in this state for a couple of months, another day won't hurt, so I'll do that prior to merging this. Thanks @thijsmie for making me considering the reasoning another time 😀

eboasson:
024662b changes it to only remove the assert, the real change is that it reduces some code duplication: there was indeed something "interesting" in the combination of a transient-local writer and a volatile reader.

The other two commits are because one of the CI runs failed with a crash on in a built-in topic test consistent with the test's not handling invalid samples and running it in parallel with other tests while allowing it to discover those other tests. They now run in isolation in the customary manner for the (few) cases that need it.

dpotman (approved):
LGTM

PR #1353, opened 2022-08-04
macOS: if Iceoryx used, add library dir to rpath

eboasson:
The CMAKE_INSTALL_RPATH set in the top-level CMakeLists.txt causes the
rpath for the Iceoryx library to disappear on installation, but not
setting it means the build directories are used in the rpaths of the
installed Cyclone library, which also doesn't work.

This adds the directory where the iceoryx library is found to the rpath
used for installing.  I suspect there are better ways, but this should
be better than nothing.


thijsmie (approved):
While not a great fix, I think it is indeed better than nothing. If we come up with something better in the future we can remove this again.

PR #1355, opened 2022-08-04
Bump minimum CMake version to 3.16

### LINKED ISSUE #1354: Required CMake version should be bumped (maybe)
The following line uses a CMake 3.12 feature, [EXPORT_PROPERTIES](https://cmake.org/cmake/help/latest/prop_tgt/EXPORT_PROPERTIES.html):

https://github.com/eclipse-cyclonedds/cyclonedds/blob/21fe67fb243d3a660b7608038802f459ceaf3e78/src/core/CMakeLists.txt#L89-L91

As a result, when building and doing `make install` with CMake 3.10 , the `CycloneDDSTargets.cmake` won't contain certain properties f.e. that type/topic discovery are supported. 

In turn, a project that uses those properties , like [cyclonedds-cxx](https://github.com/eclipse-cyclonedds/cyclonedds-cxx), won't know that those properties are actually available. 

I discovered this because the current master does not build without Type/Topic discovery, see [Build without type and topic discovery](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/300) in that project.

I understand #550 was merged to allow building with default Ubuntu 18.04 CMake, so i'm not sure what's the best solution here is:

- Lift baseline to Ubuntu 20.04 Cmake 3.16, it's fairly easy to install a newer CMake
- add a CMake version check for install 
- add property manually using CMake 3.10? not sure if its possible at all

eboasson:
Support for EXPORT_PROPERTIES was added in CMake 3.12, so either the
minimum version required needs to be bumped, or this feature needs to be
emulated.  Ubuntu 22.04 ships with CMake 3.16, is over two years old now
and also the minimum required for ROS 2 Foxy, so that seems like a
reasonable new minimum version.

Fixes #1354

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,12 +9,8 @@
 #
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.16)
 project(CycloneDDS VERSION 0.10.0 LANGUAGES C)
-if(CMAKE_VERSION VERSION_LESS 3.12)
-  # GENERATE_EXPORT_HEADER requires a C++ compiler up to version 3.12
-  enable_language(CXX)
-endif()
 
 # Set a default build type if none was specified
 set(default_build_type "RelWithDebInfo")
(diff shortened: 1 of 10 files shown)


stan-guer:
Issue seems to be same as #935

eboasson:
> Build failed due to same issue as #935

I agree, and, I am happy to report, I just figured out the root cause for that crash 🥳

I also just wanted to leave it a little while for observant people to comment before merging it, but I think for a trivial thing like this enough time has passed. Merging this one (and the C++ companion one).

PR #1363, opened 2022-08-10
Add validation for enum extensibility in type object

dpotman:
The DDS XTypes spec states that enums can be `@final` or `@appendable` (section 7.4.1, table 12), however a comment in the IDL for the TypeObject says that the enumerated type flag is unused. We decided to solve this problem by ignoring the "unused" comment in the IDL and encode the extensibility in the TypeObject. This also means that the TypeObject validation should check for this, which is fixed in this PR (also for aggregated types).


eboasson (approved):
Good catch! LGTM, thanks @dpotman  🙂

PR #1364, opened 2022-08-10
Fix for requesting type dependencies for incomplete type lookup response

dpotman:
When processing a type lookup response results in resolving a top-level type for an endpoint, but one or more dependent types remain unresolved (e.g. because these dependencies were not included in the `type_information` for the endpoint and therefore not requested yet), the endpoint matching is not triggered (which is reasonable, because it will fail) but there is also no request sent for the additional types required for matching.

This PR fixes this problem by re-triggering the matching for all endpoints that are using this type, regardless of the resolved 
state of its dependencies. The matching will trigger a type lookup request in case any dependencies need to be resolved. This fix is a temporary solution for this problem, and will be replaced by a proper solution that also takes `find_topic` calls into account (which could currently also run into this issue).


eboasson (approved):
LGTM, thanks @dpotman !

PR #1372, opened 2022-08-18
Split q_entity.c into logical parts and prefix with ddsi

dpotman:
This splits `q_entity.c` into logical parts for (proxy) participants, topics and (proxy) endpoints. The header file is also split up into the corresponding parts. As a result of placing this code into multiple files, more entity related functions are non-static and thus exported from the shared library. To reduce the risk of name clashes, all exported symbols related to DDSI entities are prefixed with `ddsi_`.

Commit [ca7330c](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1372/commits/ca7330c486209e3613f07719016e1b38ff1655fa) in this PR removes some unused code for representing proxy subscriber and publisher in DDSI, and [957bf9d](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1372/commits/957bf9d68315178ff6b25310fb39e508db97be55) is included to fix a static analyzer warning in CI.


eboasson (approved):
This isn't the easiest PR to give a thorough review, but I think I nonetheless managed to do a good job of verifying that it really only moves and renames stuff (long live a bit of perl hacking & diffing!).

In short, approved! With a great many thanks for doing this work 😀

PR #1373, opened 2022-08-18
Fix crash dropping gap from reorder buffer

eboasson:
If the following arrives from a remote writer expecting sequence number K and with a reorder buffer limited to storing at most N sequence numbers:
```
  GAP #(K+N+1) ; DATA #(K+N) .. DATA #(K+1)
```
then the last message (which can't be delivered yet because it was expecting #K) will force the GAP with sequence number #(K+N+1) out. Updating the number of discarded bytes then results in a null pointer dereference because GAPs don't store sample info.

Other commits in the PR are a test case and moving some code from `rtps_stop` to `rtps_fini` where it makes more sense.


dpotman (approved):
LGTM!

PR #1374, opened 2022-08-18
Set default symbol visibility to hidden

dpotman:
This adds the `-fvisibility=hidden` option to the GCC/Clang compile flags, so that only declarations marked with `DDS_EXPORT` are exported from the shared library. 

Note that this commit is pushed on top of the commits from PR #1372, so only commit aab6976bd508cf02a6fbb2daa20108c2e5cf4424 is relevant here.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -17,6 +17,8 @@ else()
   add_library(ddsc)
 endif()
 
+set_property(TARGET ddsc PROPERTY C_VISIBILITY_PRESET hidden)
+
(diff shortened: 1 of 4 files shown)


eboasson (approved):
LGTM!

I think we could clean up the exported symbols further by remove all the `DDS_EXPORT`s for functions only used in tests, and then simply export *everything* when `BUILD_TESTING` is set. That in turn would require verifying that everything we want to be exported in a non-test build is indeed exported (the entire stable API + some evolving stuff). This seems to me to be something that would better be done in in a future PR.

Issue #1356, opened 2022-08-07
Use-after-free on providing DCPSTopic historical data


eboasson:
Seen in CI (some irrelevant details edited out):

```
ERROR: AddressSanitizer: heap-use-after-free on address 0x60400016bbf0
READ of size 1 at 0x60400016bbf0 thread T0
WARNING: Can't read from symbolizer at fd 19
 #0 bwhc_sample_iter_borrow_next_proxy_topic dds_whc_builtintopic.c:86
 #1 bwhc_sample_iter_borrow_next dds_whc_builtintopic.c:207
 #2 whc_sample_iter_borrow_next q_whc.h:119
 #3 deliver_historical_data q_entity.c:2250
 #4 writer_add_local_connection q_entity.c:2300
 #5 connect_writer_with_reader q_entity.c:2997
 #6 connect_writer_with_reader_wrapper q_entity.c:3029
 #7 generic_do_match_connect q_entity.c:3062
 #8 generic_do_match q_entity.c:3146
 #9 match_reader_with_local_writers q_entity.c:3194
#10 new_reader_guid q_entity.c:4549
#11 new_reader q_entity.c:4575
#12 dds_create_reader_int dds_reader.c:700
#13 dds_create_reader dds_reader.c:830
#14 CU_Test_ddsc_topic_discovery_remote_topics topic_discovery.c:110

0x60400016bbf0 is located 32 bytes inside of \
  40-byte region [0x60400016bbd0,0x60400016bbf8)
freed by thread T2 here:
#1 ddsrt_free heap.c:89
#2 gc_delete_proxy_topic q_entity.c:6012
#3 gcreq_queue_thread q_gc.c:183
```

The presumed cause is that `gc_delete_proxy_topic` updates the list of proxy topics that
`bwhc_sample_iter_borrow_next_proxy_topic` is iterating over, but the updates are racing
the iteration.  There is a similar race between adding a proxy topic to the list, but it
prepends them and would be published into the reader directly (as non-historical data) and
so, despite it being very much incorrect, can often happen without any symptoms.

"Presumed cause" instead of root cause is because I can see that this race condition can
produce exactly the reported symptom, but it hasn't been checked against the test case
behaviour -- and the test case doesn't run in isolation and so it would involve checking
against all test cases that happened to be running at the same time.  Under those
conditions, an analysis that explains the symptoms is good enough.

The analysis assumes a proxy participant with a few proxy topics {T,U,V,W} which some
[...]

PR #1378, opened 2022-08-23
Fix use-after-free on providing proxy topic historical data

dpotman:
This fixes issue #1356, a race condition between iterating proxy topics for providing historical data and (gc) deletion of proxy topics. This commit replaces the linked-list implementation for storing proxy topics in proxy participants by an AVL storage, and by looking up the next proxy topic by entityid instead of keeping a pointer to the previous proxy topic, the lookup is also safe in case proxy topics are deleted from the tree during the iteration.

Commit 55c04a7508244949db6798eae50d01b0470c6069 increases the time-out for a specific topic discovery test. By debugging a recorded session for this test, the cause for the time-out appeared to be that under high load (other tests running in parallel) that causes high packet loss, the ack-nacks and heartbeats for the built-in topic reader/writer are not received in time and the cause the test to fail.

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/q_whc.h
+++ b/src/core/ddsi/include/dds/ddsi/q_whc.h
@@ -45,7 +45,7 @@ struct whc_state {
    an iter on the stack without specifying an implementation. If future changes or
    implementations require more, these can be adjusted.  An implementation should check
    things fit at compile time. */
-#define WHC_SAMPLE_ITER_SIZE (7 * sizeof(void *))
+#define WHC_SAMPLE_ITER_SIZE (8 * sizeof(void *))
 struct whc_sample_iter_base {
   struct whc *whc;
 };
--- a/src/core/ddsi/include/dds/ddsi/q_whc.h
+++ b/src/core/ddsi/include/dds/ddsi/q_whc.h
@@ -45,7 +45,7 @@ struct whc_state {
    an iter on the stack without specifying an implementation. If future changes or
    implementations require more, these can be adjusted.  An implementation should check
    things fit at compile time. */
-#define WHC_SAMPLE_ITER_SIZE (9 * sizeof(void *))
+#define WHC_SAMPLE_ITER_SIZE (7 * sizeof(void *))
 struct whc_sample_iter_base {
   struct whc *whc;
 };
--- a/src/core/ddsi/src/ddsi_entity.c
+++ b/src/core/ddsi/src/ddsi_entity.c
@@ -42,7 +42,8 @@ int ddsi_compare_guid (const void *va, const void *vb)
 
 int ddsi_compare_entityid (const void *va, const void *vb)
 {
-  return memcmp (va, vb, sizeof (ddsi_entityid_t));
+  const ddsi_entityid_t *a = va, *b = vb;
+  return (a->u == b->u) ? 0 : ((a->u < b->u) ? -1 : 1);
 }
 
 bool ddsi_is_null_guid (const ddsi_guid_t *guid)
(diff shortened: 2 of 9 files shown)


src/core/ddsi/src/ddsi_proxy_participant.c
```diff
@@ -534,7 +540,8 @@ void ddsi_unref_proxy_participant (struct ddsi_proxy_participant *proxypp, struc
 #ifdef DDS_HAS_TOPIC_DISCOVERY
     /* Last unref is called from gc_delete_proxy_participant, which is added to the gc queue after all proxy
        topic gc tasks. So we can safely assert that the proxy topic list is empty at this point */
-    assert (proxy_topic_list_count (&proxypp->topics) == 0);
+    assert (ddsrt_avl_is_empty (&proxypp->topics));
+    ddsrt_avl_free (&ddsi_proxypp_proxytp_treedef, &proxypp->topics, 0);
```
eboasson, line 544: This call to `ddsrt_avl_free` is not really going to do anything given that it is empty — but it is also not wrong, and it is not worth the effort of removing it.

src/core/ddsi/src/ddsi_entity.c
```diff
@@ -40,6 +40,11 @@ int ddsi_compare_guid (const void *va, const void *vb)
   return memcmp (va, vb, sizeof (ddsi_guid_t));
 }
 
+int ddsi_compare_entityid (const void *va, const void *vb)
+{
+  return memcmp (va, vb, sizeof (ddsi_entityid_t));
```
eboasson, line 45: I'm curious what the performance difference between a `memcmp` and `return (a->u==b->u)?0:(a->u<b->u)?-1:1` would be ... https://godbolt.org/z/rfe6aahb1 suggests on 64-bit ARM with clang the difference will be really small, but on x64 with gcc (and who cares about x64 anyway? 😁), it is a call to `memcmp` vs a few handful instructions.

The difference will likely be meaningless compared to all the other things that happen, but if you don't mind making this change, then I would like it ever so slightly better, because I do prefer the other way.

src/core/ddsi/include/dds/ddsi/q_whc.h
```diff
@@ -45,7 +45,7 @@ struct whc_state {
    an iter on the stack without specifying an implementation. If future changes or
    implementations require more, these can be adjusted.  An implementation should check
    things fit at compile time. */
-#define WHC_SAMPLE_ITER_SIZE (9 * sizeof(void *))
+#define WHC_SAMPLE_ITER_SIZE (7 * sizeof(void *))
```
eboasson, line 48: I do remember changing it from 8 to 9, and I think that was it failed to build on some 32-bit platforms with the introduction of the proxy topic iteration — some combination of platform and build flags that just happened not to run on the CI.

Obviously just tracking an entity id is going to require less state than the list iterator, so I'm not surprised it can shrink a bit. But now I am a little nervous about the `7` where it used to be `8`, perhaps you could double-check that the one case that forced it to grow to `9` is not affected by this?

eboasson (commented):
Thanks @dpotman ! A few nitpicks, perhaps even really just one. Would you mind taking a look and seeing what you make of them?

src/core/ddsi/include/dds/ddsi/q_whc.h
dpotman: Good catch, that was PR #1295, and I've checked it with a local x86 build, that fails with `7`, so I'll change it to `8` (which matches with the change in `bwhc_iter` member from `proxy_topic_list_iter_t` to `ddsi_entityid_t`, one pointer is removed from the type).

src/core/ddsi/src/ddsi_entity.c
dpotman: Fixed in [45353d8](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1378/commits/45353d8bdbd47230f17555794c945c3678780302)

eboasson (approved):
Thanks @dpotman !

PR #1379, opened 2022-08-23
Suppress Coverity false positives

dpotman:
Add some Coverity annotations to mark issues false positives or intentional


eboasson (approved):
Thank you @dpotman! 😀

PR #1382, opened 2022-08-24
Fix coverity issues 352882 and 352883

dpotman:

### DIFF
--- a/src/idl/src/file.c
+++ b/src/idl/src/file.c
@@ -522,7 +522,7 @@ idl_retcode_t idl_generate_out_file(const char *path, const char *output_dir, co
     if (!(dir = idl_strndup(rel_path, print_len))) {
       goto err_rel_path;
     }
-    if (old_dir && old_dir[0] != '\0')
+    if (old_dir != empty)
       free(old_dir);
   }
 


src/idl/src/file.c
```diff
@@ -522,8 +522,10 @@ idl_retcode_t idl_generate_out_file(const char *path, const char *output_dir, co
     if (!(dir = idl_strndup(rel_path, print_len))) {
       goto err_rel_path;
     }
-    if (old_dir && old_dir[0] != '\0')
+    if (old_dir != empty) {
+      // coverity[incorrect_free:FALSE]
```
eboasson, line 526: I'd rather not annotate it until we *know* Coverity flags it incorrectly. It may well decide the issue is fixed, like I do after looking at the change :)

eboasson (commented):
Thanks @dpotman, the fix looks good to me, but I'm wondering about the necessity of the annotation given the fix.

src/idl/src/file.c
dpotman: Yes, you're right, this change might fix the other issue as well. I've remove the Coverity annotation and force pushed a new commit

eboasson (approved):
Thanks!

Issue #1383, opened 2022-08-24
[address bug] Heap-use-after-free bug in binary idlc.


Yhcrown:
**Hi, developers of cyclonedds:**
Thank you very much for your attention to my previous issues and fixes. 
In the last few days, I have found two additional bugs, one of them is the **heap-use-after-free** which may has a security risk.

**input file:**
https://github.com/Yhcrown/fuzzing-crashes/blob/main/validated/idlc/206
**output:**
```
=================================================================
==85416==ERROR: AddressSanitizer: heap-use-after-free on address 0x60e0000014b0 at pc 0x7f4e29032fe4 bp 0x7fff0bce6ca0 sp 0x7fff0bce6c98
READ of size 4 at 0x60e0000014b0 thread T0
    #0 0x7f4e29032fe3 in idl_reference_node /workspace/test/cyclonedds/src/idl/src/tree.c:48:37
    #1 0x7f4e2904c744 in idl_yypush_parse /workspace/test/cyclonedds/src/idl/src/parser.y:1050:24
    #2 0x7f4e29022214 in parse_grammar /workspace/test/cyclonedds/src/idl/src/processor.c:327:11
    #3 0x7f4e29022214 in idl_parse /workspace/test/cyclonedds/src/idl/src/processor.c:535:15
    #4 0x555c65b57730 in idlc_parse /workspace/test/cyclonedds/src/tools/idlc/src/idlc.c:372:11
    #5 0x555c65b56c84 in main /workspace/test/cyclonedds/src/tools/idlc/src/idlc.c:703:16
    #6 0x7f4e28916d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #7 0x7f4e28916e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #8 0x555c65a988e4 in _start (/workspace/test/cyclonedds/bin/idlc+0x438e4) (BuildId: e526cf43a1a1020257bfd1c9437645cac70cf01e)

0x60e0000014b0 is located 80 bytes inside of 152-byte region [0x60e000001460,0x60e0000014f8)
freed by thread T0 here:
[...]

Issue #1368, opened 2022-08-12
[address bug]Some Heap-Buffer-Overflow crashes with asan in binary idlc


Yhcrown:
**Hi, developers of cyclonedds:**
Recently, I tested the binary idlc instrumented with ASAN by our fuzzing tool. It incurred some crashes with the following error information. 
I found three heap-buffer-overflow bug with different execution stack:

For crash input:
https://github.com/Yhcrown/fuzzing-crashes/blob/main/validated/idlc-new/0
error output:
```
=================================================================
==46256==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6310000007ff at pc 0x563bd10ef878 bp 0x7fff20ec4c00 sp 0x7fff20ec4bf8
READ of size 1 at 0x6310000007ff thread T0
    #0 0x563bd10ef877 in get_line /workspace/test/cyclonedds/src/tools/idlpp/src/support.c:1960:13
    #1 0x563bd10e9e89 in parse_line /workspace/test/cyclonedds/src/tools/idlpp/src/support.c:1653:15
    #2 0x563bd10e6c6a in get_ch /workspace/test/cyclonedds/src/tools/idlpp/src/support.c:1580:13
    #3 0x563bd10de4a9 in mcpp_main /workspace/test/cyclonedds/src/tools/idlpp/src/main.c:637:17
    #4 0x563bd10de4a9 in mcpp_lib_main /workspace/test/cyclonedds/src/tools/idlpp/src/main.c:430:5
    #5 0x563bd10b9579 in idlc_parse /workspace/test/cyclonedds/src/tools/idlc/src/idlc.c:331:9
    #6 0x563bd10b8c84 in main /workspace/test/cyclonedds/src/tools/idlc/src/idlc.c:703:16
    #7 0x7f94631d3d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #8 0x7f94631d3e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #9 0x563bd0ffa8e4 in _start (/workspace/test/cyclonedds/bin/idlc+0x438e4) (BuildId: e526cf43a1a1020257bfd1c9437645cac70cf01e)

[...]

PR #1384, opened 2022-08-25
Additional fixes for idlc fuzzer tests

dpotman:
This PR has a fix for the issue caused by input `7` in #1368, and fixes the bug that was triggered by input `92` in #1383.


eboasson (approved):
LGTM, thanks @dpotman!

PR #1386, opened 2022-08-25
Add coverity annotation for false positive issue 356274

dpotman:

### DIFF
--- a/src/idl/src/processor.c
+++ b/src/idl/src/processor.c
@@ -256,6 +256,7 @@ idl_log(
     return;
 
   off = ((size_t)cnt > sizeof(buf)) ? sizeof(buf) : (size_t)cnt;
+  // coverity[overrun-local:FALSE]
   cnt = vsnprintf(buf+off, sizeof(buf)-off, fmt, ap);
 
   if (cnt == -1)


src/idl/src/processor.c
```diff
@@ -256,6 +256,7 @@ idl_log(
     return;
 
   off = ((size_t)cnt > sizeof(buf)) ? sizeof(buf) : (size_t)cnt;
+  // coverity[overrun-local:FALSE]
   cnt = vsnprintf(buf+off, sizeof(buf)-off, fmt, ap);
```
eboasson, line 260: I do agree with the assessment that it is a false positive, but I did have to double check the definition of `snprintf` and I realised the assessment does involve an assumption 😱

The spec says: "If n is zero, nothing is written, and s may be a null pointer." — so, in the edge case that `off == sizeof(buf)`, nothing will be written. The address `buf + sizeof(buf)` is out of bounds, but the spec explicitly allows this particular out-of-bounds address to be computed, so that, too doesn't invoke undefined behaviour.

That leaves one detail: whether `snprintf` *reads* from the buffer ... and that isn't prohibited by the spec. I don't think it will, but once you realise there's an assumption, you just know that the assumption will find a way to break something, somewhere at some inconvenient time.

But I'm willing to take the risk, because the spec also doesn't require that `snprintf` does not read some other random piece of memory, so the same assumption applies to any use of it, including those that are "obviously correct".

eboasson (approved):
LGTM!

PR #1387, opened 2022-08-25
Fix for uninitialized variables in typeinfo generation in idlc

reicheratwork:
Members in the idl_typeinfo_typemap_t container would not be
initialized in a number of code paths, this would cause issues
with the pointers to typeinfo and typemap blobs being derefenced
while pointing to god knows what
Fixed this by making the initialization of the idl_typeinfo_typemap_t
members happen in all paths, not assuming that they are initialized

### DIFF
--- a/src/tools/idlc/src/descriptor_type_meta.c
+++ b/src/tools/idlc/src/descriptor_type_meta.c
@@ -1752,33 +1752,48 @@ generate_type_meta_ser (
   const idl_node_t *node,
   idl_typeinfo_typemap_t *result)
 {
+  if (NULL == result || NULL == pstate || NULL == node)
+    return IDL_RETCODE_BAD_PARAMETER;
+
   struct DDS_XTypes_TypeInformation type_information;
   dds_ostream_t os_typeinfo;
   dds_ostream_t os_typemap;
   idl_retcode_t rc;
+  bool typeinfo_generated = false;
 
-  if ((rc = generate_type_meta_ser_impl (pstate, node, &type_information, &os_typeinfo, &os_typemap)))
-    return rc;
+  if (IDL_RETCODE_OK == (rc = generate_type_meta_ser_impl (pstate, node, &type_information, &os_typeinfo, &os_typemap))) {
+    typeinfo_generated = true;
+    result->typeinfo_size = os_typeinfo.m_index;
+    if ((result->typeinfo = malloc (result->typeinfo_size)) != NULL)
+      memcpy (result->typeinfo, os_typeinfo.m_buffer, result->typeinfo_size);
+    else
+      rc = IDL_RETCODE_NO_MEMORY;
 
-  result->typeinfo = NULL;
-  result->typemap = NULL;
-  result->typeinfo_size = os_typeinfo.m_index;
-  result->typemap_size = os_typemap.m_index;
-  if ((result->typeinfo = malloc (result->typeinfo_size)) == NULL) {
-    rc = IDL_RETCODE_NO_MEMORY;
-    goto err_nomem;
+    result->typemap_size = os_typemap.m_index;
+    if ((result->typemap = malloc (result->typemap_size)) != NULL)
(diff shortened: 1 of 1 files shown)


src/tools/idlc/src/descriptor_type_meta.c
```diff
@@ -1752,33 +1752,48 @@ generate_type_meta_ser (
   const idl_node_t *node,
   idl_typeinfo_typemap_t *result)
 {
```
eboasson, line 1754: Assuming `result` points to uninitialised memory (perfectly reasonable for an out parameter) ...

src/tools/idlc/src/descriptor_type_meta.c
```diff
@@ -1752,33 +1752,48 @@ generate_type_meta_ser (
...
+
   struct DDS_XTypes_TypeInformation type_information;
   dds_ostream_t os_typeinfo;
   dds_ostream_t os_typemap;
   idl_retcode_t rc;
+  bool typeinfo_generated = false;
 
-  if ((rc = generate_type_meta_ser_impl (pstate, node, &type_information, &os_typeinfo, &os_typemap)))
-    return rc;
+  if (IDL_RETCODE_OK == (rc = generate_type_meta_ser_impl (pstate, node, &type_information, &os_typeinfo, &os_typemap))) {
```
eboasson, line 1764: ... and `generate_type_meta_ser_impl` returns an error ...

src/tools/idlc/src/descriptor_type_meta.c
```diff
@@ -1752,33 +1752,48 @@ generate_type_meta_ser (
...
+      memcpy (result->typemap, os_typemap.m_buffer, result->typemap_size);
+    else
+      rc = IDL_RETCODE_NO_MEMORY;
   }
-  memcpy (result->typeinfo, os_typeinfo.m_buffer, result->typeinfo_size);
-  if ((result->typemap = malloc (result->typemap_size)) == NULL) {
-    rc = IDL_RETCODE_NO_MEMORY;
-    goto err_nomem;
+
+  if (IDL_RETCODE_OK != rc) {
```
eboasson, line 1779: ... then this condition is true and `result` still points to unintialised memory ...

src/tools/idlc/src/descriptor_type_meta.c
```diff
@@ -1752,33 +1752,48 @@ generate_type_meta_ser (
...
+    if (result->typeinfo)
+      free (result->typeinfo);
```
eboasson, line 1781: and you may well up passing garbage into `free` here or a few lines down ...

eboasson (commented):
Thanks for looking into the issue @reicheratwork, but this is (a subtly broken) fix for a problem that I think should be addressed elsewhere. The common two patterns are that if a function returns an error, its outputs are undefined or guaranteed to be untouched, and that the system's state is unchanged. Of course it is safe to implement "undefined" as initialised properly, but that's of little use if it isn't done consistently throughout the code base ...

Reviewing this PR does tell me that the old `generate_type_meta_ser` is broken because a failure of the 2nd `malloc` will result in a memory leak (because the callers can't free the memory). Your changes would fix that.

Unfortunately, now there is an error path that results in free'ing undefined memory in this function ...

In any case I would prefer to stick to the rule that on error, the outputs are not to be trusted and no resources will have been leaked. So that means this PR is the wrong fix and instead a small one for a potential leak is needed here, and another small one is needed to prevent the calls to `free` at https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/3c5cdd31d760106a84c26afd927824c942d23897/src/idlcxx/src/traits.c#L171 from happening on failure.

eboasson:
I think these two should cover it:

https://github.com/eclipse-cyclonedds/cyclonedds/pull/1389
https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/312

Would you be willing to review them?

PR #1389, opened 2022-08-26
IDLC: fix leak on OOM in generating type info/map

eboasson:

### DIFF
--- a/src/tools/idlc/src/descriptor_type_meta.c
+++ b/src/tools/idlc/src/descriptor_type_meta.c
@@ -1770,6 +1770,7 @@ generate_type_meta_ser (
   }
   memcpy (result->typeinfo, os_typeinfo.m_buffer, result->typeinfo_size);
   if ((result->typemap = malloc (result->typemap_size)) == NULL) {
+    free (result->typeinfo);
     rc = IDL_RETCODE_NO_MEMORY;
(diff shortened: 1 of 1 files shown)


reicheratwork:
One thing that I think would be an improvement is moving the zeroing of the outparameter `result`'s members before the call to `generate_type_meta_ser_impl`, since that would guarantee that the consuming end of the `generate_type_meta_ser` function in other implementations always has these pointers correctly set, even if the call to `generate_type_meta_ser_impl` does not set them.

eboasson:
> One thing that I think would be an improvement is moving the zeroing of the outparameter `result`'s members before the call to `generate_type_meta_ser_impl`, since that would guarantee that the consuming end of the `generate_type_meta_ser` function in other implementations always has these pointers correctly set, even if the call to `generate_type_meta_ser_impl` does not set them.

I considered doing that, but decided not to because I am of the opinion that always promising that any output parameters make sense even when function returns an error is ultimately an unnecessary complication in the function's implementation and that it will only reduce error handling complexity in the easy cases anyway.

Since this PR only fixes a single and presumably rather unlikely path to a memory leak, without any changes to the contract, I think it is best to merge this PR. The discussion on the value of output parameters on error is one we can have independently of this one detail 🙂

Thanks for reviewing it!

PR #1392, opened 2022-08-30
Fix idlc typedef for scoped name

dpotman:
This fixes a bug in IDLC that causes the compilation to fail when the IDL contains a typedef with a scoped name as type definition:

```idl
module m1 {
  struct s1 { long a; };
};
typedef m1::s1 m2;
```


eboasson (approved):
Thanks @dpotman !

PR #1396, opened 2022-09-01
Specifying interfaces in network partitions

eboasson:
This PR adds an "interface" attribute to the NetworkPartition configuration element, allowing easier configuration of the subset of network interfaces on which data is expected by readers mapping to that network partition. Without these changes, one must specify the host's IP addresses corresponding to those interfaces, which is very impractical and error-prone.

The other commits are a bit of refactoring/cleaning up in the affected code.

Creating as a draft because I think it is useful as is for those who happen to be specifying unicast addresses in the `NetworkPartition`s today, but this PR really needs to also update the configuration guide and include some tests for these features. (The old code was working but without tests, the new code has changes that are substantial enough to not allow "grandfathering" it in.)


src/core/ddsi/include/dds/ddsi/ddsi_ownip.h
```diff
@@ -24,7 +24,7 @@ extern "C" {
 
 struct ddsi_domaingv;
 
-struct nn_interface {
+struct ddsi_interface {
```
dpotman, line 27: Maybe add `nw` or `network` in the name?

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -314,13 +314,17 @@ static int ddsi_raweth_is_ssm_mcaddr (const struct ddsi_tran_factory *tran, cons
...
   (void) ninterf;
   (void) interf;
-  *interf_idx = 0;
-  return DNAR_LOCAL;
+  if (interf_idx)
+    *interf_idx = 0;
+  if (memcmp (interf[0].loc.address, loc->address, sizeof (loc->address)) == 0)
+    return DNAR_SELF;
+  else
+    return DNAR_LOCAL;
```
dpotman, line 327: Nit-picking: `loc` and `interf` are not unused args any more

src/core/ddsi/src/ddsi_nwpart.c
```diff
@@ -0,0 +1,375 @@
...
+
+static int convert_network_partition_interfaces (struct ddsi_domaingv *gv, uint32_t port_data_uc)
+{
+  struct nwpart_iter npit;
+  nwpart_iter_init (&npit, gv, ": interface name");
+  struct ddsi_config_networkpartition_listelem *np;
+  while ((np = nwpart_iter_next (&npit)) != NULL)
+  {
+    char *copy = ddsrt_strdup (np->interface_names), *cursor = copy, *tok;
+    while ((tok = ddsrt_strsep (&cursor, ",")) != NULL)
```
dpotman, line 335: Is whitespace already stripped out of `np->interface_names`? If not, `strcmp` on line 339 might fail because of this?

src/core/ddsi/include/dds/ddsi/ddsi_nwpart.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef DDS_DDSI_NWPART_H
+#define DDS_DDSI_NWPART_H
+
+#include <stdint.h>
+#include "dds/features.h"
+#include "dds/ddsrt/attributes.h"
+
```
dpotman, line 18: `extern "C"` is missing

dpotman (commented):
Looks good! Just a few minor remarks, and as you mentioned it would be nice to have at least a test case that checks some basics of specifying the address and interface of network partitions.

src/core/ddsi/src/ddsi_nwpart.c
eboasson: I figured it would be acceptable to treat all white space in the names as significant and that there really was no reason to allow empty entries. I'm open to changing it, it just seems like a needless complication in the code.

eboasson:
Still have to do the tests, but I think otherwise it is starting to look halfway decent.

src/core/ddsi/src/ddsi_nwpart.c
dpotman: I was triggered by the call to `strspn` in `convert_network_partition_addresses_one` that uses `", \t"`, but accepting leading/trailing white space in a name sounds reasonable, so I'm fine with the current code.

eboasson:
@dpotman if you'd be willing to give it another look now that I have added some tests (and fixed a few things that seemed missing previously). Clearly the "DO NOT MERGE" commit is supposed not to be merged because I'd much rather merge #1397 first.

Once #1397 is merged, this one has had another look at it and I've squashed some of the commits, this is ready for merging, I'd say.

dpotman (approved):
I've reviewed the fixes and changes, LGTM!

eboasson:
Thanks @dpotman !

PR #1397, opened 2022-09-05
Reduce the number of exported symbols in shared library

dpotman:
This PR limits the number of symbols that is exported from the Cyclone shared library and includes some related cleaning-up and improvements:
- Set the default visibility option (for gcc and clang) to 'hidden' so that only functions that have the `DDS_EXPORT` macro are exported (only when build option `BUILD_TESTING` is set to `off`)
- Remove export macros for functions that are exported only for testing purposes; when building tests is enabled, all functions are exported (default on Linux/Mac, on Windows using the CMAKE option `WINDOWS_EXPORT_ALL_SYMBOLS`). 
- All exported symbols are prefixed with `dds_`, `ddsi_` or `ddsrt_`. 
- A test is added that includes calls to all exported functions, to validate that the linker can resolve these symbols (this test is enabled in a newly added CI build, that also checks the consistency between this test and the symbols exposed by the shared object). 
- Deprecated functions in the public API that have been deprecated long enough are now removed.


src/core/ddsc/src/dds_sertype_builtintopic.c
```diff
@@ -24,7 +24,7 @@
 #include "dds/ddsi/ddsi_typelookup.h"
 #include "dds__serdata_builtintopic.h"
 
-/* FIXME: sertopic /= ddstopic so a lot of stuff needs to be moved here from dds_topic.c and the free function needs to be implemented properly */
+/* FIXME: sertype /= ddstopic so a lot of stuff needs to be moved here from dds_topic.c and the free function needs to be implemented properly */
```
eboasson, line 27: This looks like a weird leftover from something. A FIXME that doesn't make sense is best removed.

src/core/ddsc/tests/filter.c
```diff
@@ -363,113 +363,6 @@ static bool filter_long1_eq_1 (const void *vsample)
...
-  struct exp exp = {
-    .n = 1, .xs = (const Space_Type1[]) {
-      {1,0,0}
-    },
-  };
-  checkdata (rd, &exp, "rd");
-  dds_delete (dp);
-}
-
-CU_Test (ddsc_filter, get)
```
eboasson, line 403: This one shouldn't go away in its entirety: `dds_get_topic_filter_extended` is still around and this seems to cover some cases not covered elsewhere. Perhaps you could double-check?

src/core/ddsi/include/dds/ddsi/ddsi_cdrstream.h
```diff
@@ -71,28 +71,26 @@ DDS_EXPORT void dds_stream_read_sample (dds_istream_t * __restrict is, void * __
...
-DDS_EXPORT void dds_stream_extract_keyBE_from_key (dds_istream_t * __restrict is, dds_ostreamBE_t * __restrict os, const struct ddsi_sertype_default * __restrict type);
+size_t dds_stream_check_optimize (const struct ddsi_sertype_default_desc * __restrict desc, uint32_t xcdr_version);
+void dds_istream_from_serdata_default (dds_istream_t * __restrict s, const struct ddsi_serdata_default * __restrict d);
+void dds_ostream_from_serdata_default (dds_ostream_t * __restrict s, const struct ddsi_serdata_default * __restrict d);
+void dds_ostream_add_to_serdata_default (dds_ostream_t * __restrict s, struct ddsi_serdata_default ** __restrict d);
+
+void dds_stream_write_key (dds_ostream_t * __restrict os, const char * __restrict sample, const struct ddsi_sertype_default * __restrict type);
+void dds_stream_write_keyBE (dds_ostreamBE_t * __restrict os, const char * __restrict sample, const struct ddsi_sertype_default * __restrict type);
+bool dds_stream_extract_key_from_data (dds_istream_t * __restrict is, dds_ostream_t * __restrict os, const struct ddsi_sertype_default * __restrict type);
```
eboasson, line 81: I have a suspicion this one may eventually prove very useful for, e.g., the Python binding. But I am in agreement with removing the `DDS_EXPORT` now, we can re-add it later if indeed it proves needed.

src/core/ddsi/src/ddsi_typelib.c
```diff
@@ -132,15 +132,13 @@ void ddsi_typeinfo_fini (ddsi_typeinfo_t *typeinfo)
 
 const ddsi_typeid_t *ddsi_typeinfo_minimal_typeid (const ddsi_typeinfo_t *typeinfo)
 {
-  DDSRT_STATIC_ASSERT (offsetof (struct ddsi_typeid, x) == 0);
```
eboasson, line 135: Why did you remove these? I might be that a comment is missing somewhere, but IIRC and also looking around a bit, the idea was that ensure converting a `ddsi_typeid_t *` pointer into a `DDS_XTypes_TypeIdentifier *` works, and that this is just a sanity check. (Also in ddsi_typewrap.c)

Maybe something changed, or I am overlooking something. That's quite possible, too!

eboasson (commented):
Thank you @dpotman! I am delighted that soon we will no longer be exporting tons of internal symbols ☺️

That said, I did manage — with quite some effort — to find a few nits to pick. If you look at them, I'm sure you will agree it is practically a "LGTM!"

There are a few further things I want to check (or to be certain have been checked):
- examples (dynsub, in particular, you probably tried it)
- OSS-Fuzz (I suspect it no longer builds without also setting BUILD_TESTING)
- C++, Python bindings (pretty sure you checked them)
- RMW layer (same)

When all this checks out (and with a second look at the few comments), then I think it will be good to go.

src/core/ddsi/src/ddsi_typelib.c
dpotman: I noticed that a few static asserts were missing, so I've added them all in [ddsi_xt_impl.h](https://github.com/eclipse-cyclonedds/cyclonedds/blob/26e670521cc78f872181a8b9369e3563ffc8baa5/src/core/ddsi/include/dds/ddsi/ddsi_xt_impl.h#L290)

src/core/ddsc/tests/filter.c
dpotman: I think that all that was tested in `ddsc_filter.get` with respect to `dds_get_topic_filter_extended` is covered in the test `ddsc_filter.getset_extended`, so that was the reason for removing the former. But I may be overlooking something?

src/core/ddsc/src/dds_sertype_builtintopic.c
dpotman: Removed in [19ca697](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1397/commits/19ca697cdd23995b85f569d55ce3e6ac6b249cc4)

src/core/ddsi/include/dds/ddsi/ddsi_cdrstream.h
dpotman: Yes, I agree, I found it hard to select the exact set of functions from the cdrstream serializer that would be useful to expose, so I minimized the set to what is currently used, and some related functions (LE/BE variants).

dpotman:
Checking the examples, that a good point! I had already checked the sources, so the functions used there are exported. Now I also confirmed that they all build (only found a minor regression, that is fixed in [24d29bf](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1397/commits/24d29bf89d59196d7759813388a174b1625435b6)). I didn't check OSS-Fuzz yet.

dpotman:
Thanks for catching this! I tested building RMW with an earlier version of the PR and did a sanity check later, with vcs import and pull to fast forward to the new commit, but apparently this went wrong (and I didn't check that it was at the right commit). 
I agree that exporting these functions used in RMW is the best option for now, in commit 16442bc I've added your patch (with one additional `#ifdef` in the test).

PR #1403, opened 2022-09-13
Fixes for symbol export test

dpotman:
Some fixes for the symbol export test: link sockets library to ddsrt on QNX, and make function calls in the test conditional based on `DDSRT_HAVE_...` macros.


eboasson (approved):
Thanks!

PR #1404, opened 2022-09-13
Initialise interface_names in network partition

eboasson:
Without this, a well-constructed bad config triggers a bad free.  With thanks to OSS-Fuzz.


dpotman (approved):
LGTM

Issue #1406, opened 2022-09-14
Some useless code in dds_topic.c


binbowang1987:
I find some useless code in dds_topic.c

In dds_create_topic, ddsi_plist_t is initialized. 
https://github.com/eclipse-cyclonedds/cyclonedds/blob/91c61f2e0c0efc19f9f0722b6b811d9d9e1c70c7/src/core/ddsc/src/dds_topic.c#L645-L664

Following the invocation path, find that the `ddsi_plist_t *sedp_plist` is not used.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/91c61f2e0c0efc19f9f0722b6b811d9d9e1c70c7/src/core/ddsc/src/dds_topic.c#L333-L348

What is the ddsi_plist used for in dds_topic.c formerly ? Can it be deleted now ?

eboasson:
It was used for publishing metadata prior to adding XTypes support, but now that all happens with different encodings and with the relevant information coming from a different place. So, based on your research and my little bit of checking the code just now, this can indeed be thrown away.

That actually holds not just for this code, but I think also for all the `..._ADLINK_TYPE_DESCRIPTION`, `plist.type_description`, as well as `ADLINK_SUBSCRIPTION_KEYS` and `qos.subscription_keys`.

As the discoverer of this bit of dead code, you are entitled the honour of removing it (but if you'd rather let us take care of it, that's ok too) 🙂

binbowang1987:
I just found this code by accident, I am not so familiar with other codes, please take care of it, thank you.:smile:

PR #1414, opened 2022-09-21
Remove some unused parameter lists

eboasson:
This removes some dead code. Thanks @binbowang1987 for pointing out the existence of some of this, cleaning up that bit of dead code then made it clear that I could remove a bit more 🙂

Closes https://github.com/eclipse-cyclonedds/cyclonedds/issues/1406

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,7 @@ if("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR
              "-Wconditional-uninitialized"
              "-Wshadow")
   add_compile_options(${wflags})
+  add_link_options(-undefined error)
   if(${WERROR})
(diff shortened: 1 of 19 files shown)


dpotman (approved):
LGTM. One minor thing: we could consider to return `UNSUPPORTED` in case `dds_qset_presentation` QoS is called with `coherent_access` scope kind?

eboasson:
> One minor thing: we could consider to return UNSUPPORTED in case dds_qset_presentation QoS is called with coherent_access scope kind?

Arguably we should do that in a couple of more cases (e.g., for transient data, too). I think that ought to be a separate PR.

I also just remembered another pair of unused, inherited settings. I'll push an update for that.

dpotman (approved):

PR #1420, opened 2022-09-28
Generate RST for config options and include it in the docs build

thijsmie:


docs/manual/config/cyclonedds_specifics.rst
```diff
@@ -70,7 +72,7 @@ or proxy writer.  In the discovery process, writers are matched with proxy reade
 readers are matched with proxy writers, based on the topic and type names and the QoS
 settings.
 
-Proxies have the same natural hierarchy that ‘normal’ DDSI entities have: each proxy
+Proxies have the same natural hierarchy that "normal" DDSI entities have: each proxy
```
eboasson, line 75: This looks like a global replace gone wrong

eboasson (approved):
Nice! I have only comment: some `'`s have changed into `"`s, but that's only nitpicking I have managed to do. It's not going to affect legibility for humans, so I'm tempted to say that it is ok to merge it, but it would be even nicer if you could fix those mishaps first.

docs/manual/config/cyclonedds_specifics.rst
thijsmie: Ah yup, my editor was highlighting unicode apostrophes it didn't like so I global replaced them, but apparently the editor tried to be smart and also replaced the correct ones. I think I've fixed all of them back to what they were.

Issue #1393, opened 2022-08-30
DDS Topic name with '-' leads to BAD PARAMTER error


AkhilTThomas:
The issue #657 is marked as closed but there is still no resolution.

https://www.omg.org/spec/DDS/1.4/PDF#page=179 

The DDS specifications mentions that '-' is a valid character but cyclonedds throws the below error:
`cyclonedds.core.DDSException: [DDS_RETCODE_BAD_PARAMETER] Bad parameter value. Occurred upon initialisation of a cyclonedds.topic.Topic`

eboasson:
The argument in #657 — many thanks for making the effort of looking it up! — holds just as much today as it did then, so it was to be expected.

How big is the problem you are facing? If you are dealing with a system originally built using one of those sloppy (😁) implementations, and are now trying to gradually replace that implementation with Cyclone (or if you have integrate with such a system), I can see how this could be a real problem and we might have to relent a bit.

AkhilTThomas:
I'm trying to define DDS topic names based on the [VSS specification](https://covesa.github.io/vehicle_signal_specification/introduction/overview/) this would require ideally a '.' separator but since that is not a valid character at least '-' would work for me.

eboasson:
A very brief look at that website suggests there is no "official" mapping to (OMG) IDL, so presumably you'd be free to translate the names? If so, why would an '_' or '/' not work?

AkhilTThomas:
'_' would be fine for now but it will not be conforming to the OMG guidelines.This can cause interoperability issues with other DDS implementations. Would it be a major change to support '-'?

eboasson:
I don't think that it would be a major change, the code base doesn't really do anything with the actual topic name (other than check it for validity) but it might well affect tooling, whether that be a scary perl script to help make sense of the trace files Cyclone can write, or things like the `cyclonedds` tool. If there are some limitations, I'm sure they can be resolved easily.

The biggest problem is probably having to step off my soap-box despite my annoyance at the quality of the OMG specs and the general sloppiness of most implementations, and all the nonsense that causes ...

[...]

PR #1426, opened 2022-10-01
Improve topic name support

clunietp:
The DDS Spec does not explicitly specify what constitutes a valid name.
Per https://github.com/eclipse-cyclonedds/cyclonedds/issues/1393#issuecomment-1248936537
"we require isprint is true and not , " or ' for the time being, then work our way to supporting UTF-8"

Resolves https://github.com/eclipse-cyclonedds/cyclonedds/issues/1393

### DIFF
--- a/src/core/ddsc/src/dds_topic.c
+++ b/src/core/ddsc/src/dds_topic.c
@@ -55,9 +55,11 @@ static bool is_valid_name (const char *name) ddsrt_nonnull_all;
 static bool is_valid_name (const char *name)
 {
   /* DDS Spec does not explicitly specify what constitutes a valid name.
-   * Per https://github.com/eclipse-cyclonedds/cyclonedds/issues/1393#issuecomment-1248936537
-   *  "we require isprint is true and not <space>, " or ' for the time being, then work our way to supporting UTF-8"
+   * Per https://github.com/eclipse-cyclonedds/cyclonedds/pull/1426
+   *  Require isprint is true and not <space>*?[]"' for the time being, then work our way to supporting UTF-8
    */
+  const char* invalid = "*?[]\"'#$";
+
   if (name[0] == '\0')
     return false;
 
@@ -65,8 +67,7 @@ static bool is_valid_name (const char *name)
     if (
         (!(isprint((unsigned char) name[i])))
         || (isspace((unsigned char) name[i]))
-        || (name[i] == '"')
-        || (name[i] == '\'')
+        || (strchr(invalid, name[i]) != NULL)
       )
       return false;
     
--- a/src/core/ddsc/src/dds_topic.c
+++ b/src/core/ddsc/src/dds_topic.c
@@ -54,20 +54,22 @@ static bool is_valid_name (const char *name) ddsrt_nonnull_all;
 
 static bool is_valid_name (const char *name)
 {
-  /* DDS Spec:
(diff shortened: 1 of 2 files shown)


thijsmie (changes requested):
I like the changes, but perhaps it is wise to forbid the characters of `fnmatch` with an eye on future topic filtering expressions. This would be `*?[]`. I guess the topic is not a match expression so having these characters can't actually hurt, I simply think that topic names as `room.temperature.*` look extremely confusing. Maybe also with an eye on our sister project Zenoh it is also wise to forbid the Zenoh key expression special characters `#$?*`. What do you thing @eboasson?

eboasson:
> I like the changes, but perhaps it is wise to forbid the characters of `fnmatch` with an eye on future topic filtering expressions. This would be `*?[]`. I guess the topic is not a match expression so having these characters can't actually hurt, I simply think that topic names as `room.temperature.*` look extremely confusing. Maybe also with an eye on our sister project Zenoh it is also wise to forbid the Zenoh key expression special characters `#$?*`. What do you thing @eboasson?

I think you're making an very good point. Should've thought about it myself ... 🤦‍♀️

clunietp:
I'm happy to make the requested changes to the topic name filter.  My only input here is that this restriction may cause issues in the future with interoperability.  For example, both Fast-DDS 2.5 and Connext 6.1 both gladly support the creation of the topic named `Hello *?[]'"\tWorld`  (just tested)

I'm not saying that creating topic names in this format is anywhere close to a good idea (at least in my mind), but there are likely use cases that I'm unaware of where some subset of these restricted characters actually makes sense and would be considered valid.  This would necessitate a future workaround/PR, similar to the issues I had when I originally created this PR.

As an alternative to restricting the topic names any more than necessary at this point, perhaps future work with `fnmatch` (which I admittedly know nothing about) would support the use of escape characters (or some other mechanism) for topic matching with special characters and/or UTF8 chars within the topic name.  

Thoughts?

eboasson:
I'd like Cyclone to put up no barriers other than those for which there exist good engineering reasons. Requiring topic names to be ASCII really is a barrier to people living in other parts of the world, and I think we need to support them. (I think that's a pretty non-controversial position 🙂.)

In practice that means we will have to accept nearly any byte sequence as a name, while strongly suggesting that these byte strings be valid UTF-8 strings. Actually validating it is way out of scope for an embedded library like Cyclone, it pulls in too much code, makes it dependent on Unicode revisions, ... This is of course a pretty standard.

What I would *also* like is log files to be fairly easy to search through and take apart with regular expressions. (Granted, my definition of "easy" is of [doubtful quality](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/tools/decode-trace) in this matter.) The way I see it, that means: an easy way of knowing where a name begins and ends, and that either means forbidding some characters or escaping some in the trace output. Getting the latter right means some thinking.

There are also some other parts in the code/configuration where a few things will have to be changed. For example, the network partition configuration & matching.

Hence the proposal to do it in two steps, first we go from whitelisting some characters to allowing all printable ASCII with a few disallowed characters, and then @thijsmie is right that the characters that play a role in wildcard matching are also best excluded. That should alleviate some issues today.

Then we do a follow-up where we lift all restrictions, escape names in the output and address any other things that need changing.

thijsmie (approved):
Thanks @clunietp for making the requested changes :smile:, LGTM!

PR #1429, opened 2022-10-03
Updated windows installer to use the new logo and show a proper license agreement

ThijsSassen:
added a copy of the LICENSE file as rtf because WiX needs an rtf as input.


thijsmie:
Hi @ThijsSassen,

I'll gladly merge this, but don't you think it is better to move the LICENCE.rtf file to the WiX folder to avoid having more things laying around in the top level directory (which is arguably already a bit messy)?

ThijsSassen:
> Hi @ThijsSassen,
> 
> I'll gladly merge this, but don't you think it is better to move the LICENCE.rtf file to the WiX folder to avoid having more things laying around in the top level directory (which is arguably already a bit messy)?

yes i agree the file is moved now

PR #1438, opened 2022-10-07
Add missing functionality to idl(c)

thijsmie:
This adds a couple small things to the IDL tooling:
 * Add `idl_fclose` and `idl_free` to make sure file opening and memory freeing across DLL bounds on Windows can function.
 * Make the `generator_annotations` actually function: you could define them, you could load them but the parser would not actually use them.

### DIFF
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -33,6 +33,7 @@ if(NOT ${_bison_hash_correct})
 
   if(BISON_FOUND)
     message(STATUS "Running Bison, parser.y was updated.")
+    message(STATUS "${source_dir}")
 
     file(REMOVE src/parser.c)
     file(REMOVE src/parser.h)
@@ -46,7 +47,7 @@ if(NOT ${_bison_hash_correct})
         ${binary_dir}/src/parser.c
         ${binary_dir}/src/parser.h
       COMMAND ${CMAKE_COMMAND} ARGS
-        "-Dsource_dir=${CMAKE_CURRENT_LIST_DIR}"
+        "-Dsource_dir=${source_dir}"
         "-Dbinary_dir=${binary_dir}"
         "-DMAIN_PROJECT_DIR=${CMAKE_SOURCE_DIR}"
         -P ${CMAKE_CURRENT_LIST_DIR}/postprocess-parser.cmake
@@ -74,7 +75,7 @@ set(headers
   include/idl/symbol.h
   include/idl/tree.h
   include/idl/visit.h
-  include/idl/stdlib.h
+  include/idl/heap.h
   ${binary_dir}/include/idl/version.h)
 
 set(sources
@@ -94,7 +95,7 @@ set(sources
   src/keylist.c
   src/hashid.c
   src/parser.c
-  src/stdlib.c
+  src/heap.c
   src/annotation.h
   src/directive.h
   src/expression.h
--- a/src/idl/CMakeLists.txt
+++ b/src/idl/CMakeLists.txt
@@ -74,6 +74,7 @@ set(headers
   include/idl/symbol.h
   include/idl/tree.h
   include/idl/visit.h
+  include/idl/stdlib.h
   ${binary_dir}/include/idl/version.h)
 
 set(sources
@@ -93,6 +94,7 @@ set(sources
   src/keylist.c
   src/hashid.c
   src/parser.c
+  src/stdlib.c
   src/annotation.h
   src/directive.h
   src/expression.h
--- a/src/idl/include/idl/print.h
+++ b/src/idl/include/idl/print.h
@@ -19,7 +19,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #if defined(_WIN32)
-# include <idl_malloc.h>
+# include <malloc.h>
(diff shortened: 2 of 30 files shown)


dpotman:
Looks good. I think it may also be a good idea to introduce `idl_malloc` and `idl_calloc` functions and use this for all allocations in idlc, and also replace `free` with `idl_free` so that idlc (including the language backends) consistently use these alloc/free functions.

src/idl/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ set(headers
   include/idl/symbol.h
   include/idl/tree.h
   include/idl/visit.h
+  include/idl/stdlib.h
```
dpotman, line 77: I'd prefer using `heap.h` (and `heap.c`), similar to ddsrt

src/idl/include/idl/print.h
```diff
@@ -19,7 +19,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #if defined(_WIN32)
-# include <malloc.h>
+# include <idl_malloc.h>
```
dpotman, line 22: Line can be removed

src/idl/include/idl/stdlib.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef IDL_STDLIB_H
```
dpotman, line 12: See above, I think `IDL_HEAP_H` would be better

src/idl/src/parser.c
```diff
@@ -1,8 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.8.2.  */
+/* A Bison parser, made by GNU Bison 3.5.1.  */
```
dpotman, line 1: Using an older bison version probably doesn't cause any issues, but to be sure we should find out what the minimum version is that is required. As this version causes quite some changes in the resulting c code, I think it is safer to use a more recent version.

src/idl/src/parser.h
```diff
@@ -53,7 +52,7 @@
 extern int idl_yydebug;
 #endif
 /* "%code requires" blocks.  */
-#line 85 "src/parser.y"
+#line 86 "/home/tmiedema/Projects/workspace-cyclonedds/cyclonedds/src/idl/src/parser.y"
```
dpotman, line 55: Should be a relative path, might be fixed by using a more recent bison version?

src/idl/src/string.c
```diff
@@ -382,7 +388,7 @@ void WINAPI idl_cdtor(PVOID handle, DWORD reason, PVOID reserved)
         return;
       _free_locale(loc);
 err_locale:
-      TlsFree(locale);
+      Tlsidl_free(locale);
```
dpotman, line 391: Shouldn't be replaced

src/idl/src/string.c
```diff
@@ -407,7 +413,7 @@ void WINAPI idl_cdtor(PVOID handle, DWORD reason, PVOID reserved)
       if (loc)
         _free_locale(loc);
       TlsSetValue(locale, NULL);
-      TlsFree(locale);
+      Tlsidl_free(locale);
```
dpotman, line 416: Same here

src/idl/src/parser.c
eboasson: A good point in principle, but we do know 3.5.1 is also fine. Maybe we should agree on a preferred version of bison and try to use that, not because of compatibility concerns, but rather to reduce the amount of noise in the history.

That probably means everyone needs to have a custom version of bison hanging around though, because there are so many Linux distributions (and other platforms) in use, and the system-provided one can be anything.

For now, I wouldn't worry about this.

thijsmie:
I addressed the review points in a separate commit for easy review, I'll squash that down when merging 😄

dpotman:
The changes look fine, but there are 2 failures in CI: the static analyzer complaining about a null dereference is most likely not related to this PR and may be a false positive. 

For the `type_builder` test failure: I tried to reproduce it locally but that was unsuccessful. I can't explain why it happens only in this build, not in e.g. the clang release build.. Maybe we can create an issue for this and investigate it later, because it is probably also unrelated to the changes in this PR?

thijsmie:
So I am unsure here what to do about the static analyzer failure. In my opinion, if `const_expr==NULL` at `cyclonedds/src/tools/idlc/src/descriptor.c:412` then you can never jump to a case in the switch statement below, `idl_type(NULL)` will _always_ return `NULL`.

dpotman (approved):

thijsmie:
I should have merged this earlier. If no more comments arrive today I will do so :smile:

eboasson:
Why wait until the end of the day? Looks to me like it has received a positive review from a respected contributor 🙂 But it is of course up to you!

(Come to think of it, the day arguably ends rather soon where I am 😁)

PR #1449, opened 2022-10-18
Add extra condition to pacify the sanitizer for descriptor.c

thijsmie:
We have been having a sanitizer failure in CI for a bit, which I think is a false positive. In any case I've  "stated the obvious" in the condition which pacifies the santizer.


dpotman (approved):

thijsmie:
Superseeded by #1465

PR #1451, opened 2022-10-19
Ubuntu 18 being phased out on Azure

thijsmie:
See title mostly. I switched the old 18 to a 22 build with GCC11 instead of just removing it, but maybe instead we want everything on 22 now and keep one 20 job?


eboasson:
> but maybe instead we want everything on 22 now and keep one 20 job?

I think just moving to the latest supported versions for all platforms (with a few older ones to make sure we don't leave users dependent on older versions stranded) makes the most sense.

PR #1453, opened 2022-10-19
Fix assumptions that `CMAKE_INSTALL_*DIR` variables are relative

lopsided98:
The `CMAKE_INSTALL_*DIR` variables are allowed to be absolute paths, which allows libraries and headers to be installed into separate prefixes. This means they cannot be naively concatenated with `CMAKE_INSTALL_PREFIX`. If an absolute path is needed, `CMAKE_INSTALL_FULL_*DIR` should be used, while more complicated handling is required to keep `${prefix}` interpolation in pkg-config files in the case where `CMAKE_INSTALL_*DIR` is relative.

More information about this problem can be found here: https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path. Note that a more concise way to construct paths for the pkg-config file is available in CMake 3.20, using `cmake_path(APPEND ...)`.


thijsmie (approved):
Hi @lopsided98,

Hook, line and sinker! Thanks for the contribution that states a problem and immediately provides the solution, including some nice CMake reading.

PR #1462, opened 2022-10-28
Missed a couple references

thijsmie:


ThijsSassen:
@thijsmie the helloworld rst also has a ref which is not changed.

eboasson (approved):

Issue #1445, opened 2022-10-14
[Bug] node deadlocks and roudi segfaults in ros galactic.


elfenpiff:
In iceoryx the following bug was reported: https://github.com/eclipse-iceoryx/iceoryx/issues/1740

In short we have a high frequency ROS publisher which sends data and a listener application which receives the first couple of thousand samples and then seems to end up in a dead lock.
This deadlock seems to cause also a seg-fault in roudi when the listener is terminated with a kill signal.

Furthermore, when the listener is in this deadlock roudi cannot be terminated any longer.

This repository provides examplary code and instructions on how to produce the issue: https://github.com/ceccocats/iceoryx_bug

I suspect that we have here a couple of issues. A deadlock in the listener should never cause roudi to be no longer able to terminate, this we have to investigate further on iceoryx side. Also the segfault which is caused when the deadlocked listener is killed.

But I think that the problem maybe originates somewhere in dds therefore I created the ticket.

sergmister:
I believe this issue might have been resolved here https://github.com/eclipse-iceoryx/iceoryx/issues/1740 after #1463.

PR #1463, opened 2022-10-28
Fixed iceoryx crash (#1445)

robotechvision:
The crash #1445 was caused by accessing non-thread-safe resources of iceoryx subscriber at the same time from different threads (one thread was [reading](https://github.com/eclipse-cyclonedds/cyclonedds/blob/cd3c6ed83899035b98f76266d4e73223a9b911da/src/core/ddsc/src/shm_monitor.c#L102) chunks, another thread was [releasing](https://github.com/eclipse-cyclonedds/cyclonedds/blob/cd3c6ed83899035b98f76266d4e73223a9b911da/src/core/ddsc/src/dds_data_allocator.c#L155) processed chunks).

Note: This fixes the particular crash scenario but there may be other places where iox locking is needed...

### DIFF
--- a/src/core/ddsc/src/dds_data_allocator.c
+++ b/src/core/ddsc/src/dds_data_allocator.c
@@ -152,7 +152,9 @@ dds_return_t dds_data_allocator_free (dds_data_allocator_t *data_allocator, void
       case DDS_IOX_ALLOCATOR_KIND_SUBSCRIBER:
         if (ptr != NULL) {
           ddsrt_mutex_lock(&d->mutex);
+          shm_lock_iox_sub(d->ref.sub);
           iox_sub_release_chunk(d->ref.sub, ptr);
+          shm_unlock_iox_sub(d->ref.sub);
           ddsrt_mutex_unlock(&d->mutex);
         }
         break;


robotechvision:
The license agreement has just been signed, you can re-run CI to verify it.

eboasson:
@robotechvision friendly ping: I suspect this is at least related to https://github.com/eclipse-cyclonedds/cyclonedds/issues/1481, but the ECA check still fails and so I can't merge it. (I just checked again, in case some database got updated with me getting a notification.)

robotechvision:
Oh we are sorry, for some reason we did not get notifications when you messaged here. Seems like our employee who needs to sign the agreement too

robotechvision:
now it should work, @eboasson could you verify it?

eboasson:
> now it should work, @eboasson could you verify it?

Yes, it works! Thank you so much ☺️

YeahhhhLi:
Can this fix pr be added to the v0.10 version?  I also encountered the same crash situation:

image

eboasson:
> an this fix pr be added to the v0.10 version

Yes, definitely. The easiest for me would be if you were to do a PR for it, is that an option? 😁

YeahhhhLi:
> Yes, definitely. The easiest for me would be if you were to do a PR for it, is that an option?

ok, let me try

eboasson:
@YeahhhhLi 

> it is more troublesome to apply for a contributor. :(

I know. I'm not sure it applies in this case because it should be a clean cherry-pick operation, but finding out is probably even more effort.

Don't worry about it. I was being lazy and tried to avoid having to go through the process of creating a branch, cherry-picking, pushing, opening a PR, setting the correct target branch ... In reality, that's 2 minutes of work. Let me sort this one out for you.

YeahhhhLi:
> @YeahhhhLi
> 
> > it is more troublesome to apply for a contributor. :(
> 
> I know. I'm not sure it applies in this case because it should be a clean cherry-pick operation, but finding out is probably even more effort.
> 
> Don't worry about it. I was being lazy and tried to avoid having to go through the process of creating a branch, cherry-picking, pushing, opening a PR, setting the correct target branch ... In reality, that's 2 minutes of work. Let me sort this one out for you.

Thanks eboasson!  waiting for your good news  :+1:

YeahhhhLi:
> @YeahhhhLi
> 
> > it is more troublesome to apply for a contributor. :(
> 
> I know. I'm not sure it applies in this case because it should be a clean cherry-pick operation, but finding out is probably even more effort.
> 
> Don't worry about it. I was being lazy and tried to avoid having to go through the process of creating a branch, cherry-picking, pushing, opening a PR, setting the correct target branch ... In reality, that's 2 minutes of work. Let me sort this one out for you.

Hey eboasson,  I see that this fix has been cp to v.10.x:
https://github.com/eclipse-cyclonedds/cyclonedds/pull/1606

Which is very efficient!

In addition, I want to know when we can make a release package of v0.10.3, so that we can use it quickly

eboasson:
@YeahhhhLi I just created PRs for backporting a few things and bumping versions. I'll probably need to massage the CI a bit to get the C++ PR to build against the C PR to get a full CI run before merging. Did I miss anything you know you need?

YeahhhhLi:
> @YeahhhhLi I just created PRs for backporting a few things and bumping versions. I'll probably need to massage the CI a bit to get the C++ PR to build against the C PR to get a full CI run before merging. Did I miss anything you know you need?

That will be enough. Appreciate it.Thanks!

PR #1468, opened 2022-11-03
Move sertype/serdata default and cdrstream serializer out of DDSI

dpotman:
This moves the default implementation for sertype and serdata to DDSC and the cdrstream serializer to a separate component within Cyclone core. See commit message for more details. 

These change also have minor impact on the Python and C++ bindings, so I'll create PRs for these bindings that need to be merged with this PR.


src/core/cdr/include/dds/cdr/dds_cdrstream.h
```diff
@@ -44,9 +71,58 @@ typedef struct dds_ostreamLE {
...
+#else
+#define MK_ALIGN(n) (n)
+#endif
+  if (size > 4)
+    return xcdr_version == DDS_CDR_ENC_VERSION_2 ? MK_ALIGN(4) : MK_ALIGN(8);
+  return MK_ALIGN(size);
+#undef MK_ALIGN
+}
+
+uint32_t dds_cdr_alignto_clear_and_resize (dds_ostream_t * __restrict s, align_t a, uint32_t extra);
```
eboasson, line 124: I wonder if it doesn't make more sense to split this function into an internal one taking `align_t` and an externally visible wrapper taking a `uint32_t` and calling the internal one. That way `align_t` and `dds_cdr_get_align` don't have to be exported.

For that matter, the only two external call sites (in this repo, haven't checked python and C++ yet) only call it with `dds_cdr_get_align (s->m_xcdr_version, 4)` to pad it out. Arguably, that means the externally visible function could simply be `dds_cdr_alignto4_clear_and_resize`.

src/core/ddsi/src/ddsi_serdata_cdr.c
```diff
@@ -0,0 +1,361 @@
...
+      const unsigned char *payload = NN_RMSG_PAYLOADOFF (fragchain->rmsg, NN_RDATA_PAYLOAD_OFF (fragchain));
+      serdata_cdr_append_blob (&d, fragchain->maxp1 - off, payload + off - fragchain->min);
+      off = fragchain->maxp1;
+    }
+    fragchain = fragchain->nextfrag;
+  }
+
+  const bool needs_bswap = !CDR_ENC_IS_NATIVE (d->hdr.identifier);
+  d->hdr.identifier = CDR_ENC_TO_NATIVE (d->hdr.identifier);
+  const uint32_t pad = ddsrt_fromBE2u (d->hdr.options) & 2;
```
eboasson, line 171: This `& 2` is weird. It was there before in `serdata_default`, but either that was in error and it should be `& 3` (and we are missing a test case 🤓) or it needs a comment why it is correct!

src/core/ddsi/src/ddsi_serdata_cdr.c
```diff
@@ -0,0 +1,361 @@
...
+
+static uint32_t serdata_cdr_get_size (const struct ddsi_serdata *dcmn)
+{
+  const struct ddsi_serdata_cdr *d = (const struct ddsi_serdata_cdr *) dcmn;
+  return d->pos + (uint32_t)sizeof (struct dds_cdr_header);
+}
+
+static bool serdata_cdr_eqkey (const struct ddsi_serdata *acmn, const struct ddsi_serdata *bcmn)
+{
+  (void) acmn;
```
eboasson, line 82: I think it would make sense to add a comment somewhere that this is specialised for types that have no key fields.

eboasson (commented):
It is always nice when things get cleaned up a bit. I remember looking at this (quite) some time ago and deciding it was too much effort at the time, but I am glad you went ahead and did it!

I have a few minor comments, I'd appreciate it if you had a look at them and see what you think.

src/core/cdr/include/dds/cdr/dds_cdrstream.h
dpotman: That's a good point, I've refactored this in [09fee43](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1468/commits/09fee435c23d47d12ad16a73bcc97688cf9dc15d)

src/core/ddsi/src/ddsi_serdata_cdr.c
dpotman: Good catch! Fixed in [616f7c4](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1468/commits/616f7c4b2a12f4a03301239d49397d68d6868dff)

src/core/ddsi/src/ddsi_serdata_cdr.c
dpotman: I've added a comment on the struct definition for `sertype_cdr` and a comment in this function ([fab4340](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1468/commits/fab4340750acdb33031ae156160d16b3cb67c7b5))

dpotman:
Thanks for reviewing this so quickly! See my comments above.

dpotman:
Added a commit to fix an issue in `dds_cdrstream_desc_from_topic_desc`, see commit message above for details.

eboasson (approved):

PR #1480, opened 2022-11-15
Unification of lease durations in plist/qos

eboasson:
This reworks the code so that participants rely on the liveliness QoS setting for their lease durations in all contexts, rather than just in the API. (The spec, for some reason, doesn't use liveliness for the participant.)

The DDSI spec requires the use of a separate "participant lease duration" parameter in the participant discovery protocol, in this PR, that is handled entirely by the (de)serialisation code. That in turn requires it to be aware of whether it is dealing with SPDP data (where "participant lease duration" is a thing, and "liveliness" is not), and all other (i.e., SEDP) data (where "liveliness" is a thing and "participant lease duration" is not).

It adds some tests to ensure that the correct parameters is used in the correct context, but it doesn't yet have unit tests that show that the parameter list (de)serialisation uses the correct context in all places. Presumably the easiest way to do that is to create some entities in a two-node setting and check the built-in data. That *may* already be covered by built-in topic tests and liveliness tests, but it needs to be looked into. Hence making this a draft PR for now.

### DIFF
--- a/src/core/ddsi/tests/plist_leasedur.c
+++ b/src/core/ddsi/tests/plist_leasedur.c
@@ -61,38 +61,40 @@ struct plist_valid {
   dds_duration_t lease_duration;
 };
 struct plist_cdr {
-  struct plist_valid exp[2]; // SPDP, SEDP
+  struct plist_valid spdp, others;
   unsigned char cdr[44];
 };
 static const struct plist_cdr plists[] = {
-  { { { true, true, DDS_LIVELINESS_AUTOMATIC, 3071111111 },
-      { true, false, (dds_liveliness_kind_t)0, 0 } },
-    { LD(3,0x12345679), SENTINEL } },
-  { { { true, false, (dds_liveliness_kind_t)0, 0 },
-      { true, true, DDS_LIVELINESS_MANUAL_BY_PARTICIPANT, 2041944443 } },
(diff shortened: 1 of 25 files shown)


src/core/ddsi/tests/plist_leasedur.c
```diff
@@ -0,0 +1,480 @@
...
+  unsigned char pkt_header[] = {
+    'R', 'T', 'P', 'S', RTPS_MAJOR, RTPS_MINOR,
+    // vendor id: major 1 is a given
+    1, NN_VENDORID_MINOR_ECLIPSE,
+    // GUID prefix: first two bytes ordinarily have vendor id, so 7,7 is
+    // guaranteed to not be used locally
+    TEST_GUIDPREFIX_BYTES,
+    // DATA (0x15); flags (4 = dataflag + big-endian); octets-to-next-header = 0
+    // means it continues until the end
+    0x15, 4, 0,0,
```
dpotman, line 262: You could use the `SMID_DATA` macro here (and other macros for similar cases below).

dpotman (commented):
Looks good, I like the test approach with constructed RTPS messages, this approach will definitely be useful in other cases. Below a few minor remarks.

dpotman (approved):
LGTM

Issue #1454, opened 2022-10-20
DDS Fails Discovery with containers in different hosts


sinsalabim:
I'm trying to make communicate containers in different hosts, however, if the containers are running in the same machine, everything runs smoothly, but different hosts do have some problem to get along.
I'll resume the two situations:
# Successful case
Host Machine OS: Windows 11
Publisher Container OS: Linux
Subscriber Container OS: Linux

Both publisher and subscriber cyclonedds.xml configuration are the same:

```xml
<CycloneDDS>
	<Domain id="any">
		<General>
			<Interfaces>
			  <NetworkInterface name="eth0" autodetermine="false" multicast="false"/>
			</Interfaces>
		</General>
		<Discovery>
			<Peers>
				<Peer address="host.docker.internal"/>
			</Peers>
		</Discovery>
	</Domain>
</CycloneDDS>
```

And everything runs smoothly.
In facts, XTypes are recognized, and communication happens.
`cyclonedds ls` works out of the box, as well as `cyclonedds ps`, these are the output of the respective commands:

## cyclonedds ls output

```
# cyclonedds ls

1666230282.019213 [0] cyclonedds: selected interface "eth0" is not multicast-capable: disabling multicast

   0:00:01 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Entities discovered: 4
╭────────────────────────────────── Participant 011051a3-bc0b-0b30-ebe0-7c2a000001c1 ───────────────────────────────────╮
│                                                                                                                       │
│ ╭───────────────────────────────  QoS ───────────────────────────────╮                                                │
│ │ Liveliness.Automatic(lease_duration='10 seconds')                  │                                                │
│ │ Property(key='__Hostname', value='3d051886342f')                   │                                                │
│ │ Property(key='__NetworkAddresses', value='udp/172.17.0.2:7410@22') │                                                │
│ │ Property(key='__Pid', value='8933')                                │                                                │
│ │ Property(key='__ProcessName', value='python')                      │                                                │
│ ╰────────────────────────────────────────────────────────────────────╯                                                │
│                                                                                                                       │
│ ╭─────────────────────────────────────────────────── Boot_Topic ────────────────────────────────────────────────────╮ │
│ │ ╭───────────────────────────────────────────────── Common QoS ──────────────────────────────────────────────────╮ │ │
[...]

PR #1488, opened 2022-11-22
Default values as code fragment in config docs

dpotman:
Add the default value in the configuration guide as a code fragment, so that it is clear that the value does not include quotes. 

Related to https://github.com/eclipse-cyclonedds/cyclonedds/issues/1454#issuecomment-1290228361

### DIFF
--- a/src/tools/_confgen/generate_rst.c
+++ b/src/tools/_confgen/generate_rst.c
@@ -26,7 +26,7 @@ static const char *xlatmd(const char *str, const char **end)
     { "<li>", " * " }, { "</li>", "\n" },
     { "<ul>", "" }, { "</ul>", "\n" },
     { "<sup>", "^" }, { "</sup>", "" },
-    { "<code>", "`" }, { "</code>", "`" },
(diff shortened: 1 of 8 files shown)


src/tools/_confgen/generate_rst.c
```diff
@@ -26,6 +26,8 @@ static const char *xlatmd(const char *str, const char **end)
     { "<li>", " * " }, { "</li>", "\n" },
     { "<ul>", "" }, { "</ul>", "\n" },
     { "<sup>", "^" }, { "</sup>", "" },
+    { "<code>", "`" }, { "</code>", "`" },
```
thijsmie, line 29: Rst code formatting should be double backticks, like this: `\`\`code\`\``

thijsmie (commented):
I just have one small thing to be fixed, for the rest it looks good 😊

src/tools/_confgen/generate_rst.c
dpotman: Fixed in 9fb41da

PR #1491, opened 2022-11-23

eboasson:

Signed-off-by: Scott K Logan <logans@cottsay.net>


eboasson:
Based on the nature of the change, the green CI for master and all non-Windows builds, and the Windows builds here failing because the world changed from underneath our feet, I think we should merge this despite some build failures. @thijsmie do you agree?

thijsmie:
Yeah seems reasonable to me. The fix for CI failures of this kind was merged to master and are not terribly relevant to backport. We can go ahead.

PR #1492, opened 2022-11-23

eboasson:

Signed-off-by: Scott K Logan <logans@cottsay.net>


eboasson:
Based on the nature of the change, the green CI for master and all non-Windows builds, and the Windows builds here failing because the world changed from underneath our feet, I think we should merge this despite some build failures. @thijsmie do you agree?

thijsmie:
Your comment was a little early, the windows builds here work just fine, the CI fix was applied between 0.9 and 0.10 :innocent:

eboasson:
> Your comment was a little early, the windows builds here work just fine, the CI fix was applied between 0.9 and 0.10 😇

The combination of impatience and copy-paste has proven dangerous again!

PR #1494, opened 2022-11-23
Fix tkmap_instance leak in WHC

### LINKED ISSUE #1487: Memory leak when ddsi_serdata.statusinfo contains NN_STATUSINFO_UNREGISTER
I found `whc_idxnode` refed `ddsi_tkmap_instance`, but the refed `ddsi_tkmap_instance` was not unrefed when deleting `whc_idxnode`. As `ddsi_tkmap_instance` is not unrefed, it will never be freed. So, this may lead to a memory leak of `ddsi_tkmap_instance`.

`ddsi_tkmap_instance`refed here when creating `whc_idxnode`.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/9995905bce6c4cf9f740d6438bbf7fcfd1c83dfd/src/core/ddsc/src/dds_whc.c#L1346-L1357

`whc_idxnode`deleted here, but `ddsi_tkmap_instance` is not unrefed.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/9995905bce6c4cf9f740d6438bbf7fcfd1c83dfd/src/core/ddsc/src/dds_whc.c#L1279-L1295

eboasson:
Fixes #1487 

A few tests keel over when asserting that the instance map is empty on shutdown, and it appears that these can be traced back to some "oddities" in registering and unregistering instances. That's some really old code that needs some TLC because it really is not quite correct (even if it mostly works ok).

So this PR simply fixes the leak. I'll create an issue for the other thing.

### DIFF
--- a/docs/manual/config/cyclonedds_specifics.rst
+++ b/docs/manual/config/cyclonedds_specifics.rst
@@ -145,42 +145,6 @@ linger duration has elapsed, whichever comes first.
   is requested to terminate.
 
 
-.. _`Start-up mode`:
-
-=============
-Start-up Mode
-=============
-
-A similar issue exists when starting |var-project-short|: DDSI discovery takes time, and when
-data is written immediately after the first participant was created, it is likely that
-the discovery process hasn't been completed yet, and some remote Readers have not yet been
-discovered. This would cause the Writers to throw away samples for lack of interest,
-even though matching Readers already existed at the starting time. For best-effort
-Writers, this is perhaps surprising but still acceptable; for reliable Writers, however,
-it would be very counter-intuitive.
-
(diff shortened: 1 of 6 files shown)


dpotman (commented):
Looks good. One thing (unrelated to this PR) I noticed during reviewing this is that `whc_default_downgrade_to_volatile` uses `free_one_instance_from_idx` and not `delete_one_instance_from_idx` (and therefore does not remove the node from `whc->idx_hash`). But it looks like `downgrade_to_volatile` is not used at all, and can be removed from the `ddsi_whc` interface?

dpotman (approved):

PR #1496, opened 2022-11-24
Fixes for build with SSM disabled

dpotman:
In case the cmake configuration option `ENABLE_SOURCE_SPECIFIC_MULTICAST` is disabled:
- In `ddsi_xqos.h` the header `ddsrt/sockets.h` is not included (via `ddsi_protocol.h`, `dds_feature_check.h`), and therefore has no definition for `dds_duration_t` and `dds_retcode_t`. This PR adds the missing includes in `ddsi_xqos.h`. 
- One of the xtypes tests was not taking the SSM build option into account when calling `ddsi_new_proxy_reader`, that is fixed in the other commit in this PR. 

It's clear that we need a better check for building with combinations of build options when merging PRs, an issue is created for this: https://github.com/eclipse-cyclonedds/cyclonedds/issues/1498


eboasson (approved):
Thanks @dpotman ! (Also for creating the issue on checking more combination options in the CI ☺️)

PR #1500, opened 2022-11-25
Align the upstream change of renaming Iceoryx_LogLevel_Verbose to Iceoryx_LogLevel_Trace

YuanYuYuan:
As title. To prevent the compiling error due to [the latest change 28. on the master branch of iceoryx](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/release-notes/iceoryx-unreleased.md).

### DIFF
--- a/src/core/ddsi/src/ddsi_shm_transport.c
+++ b/src/core/ddsi/src/ddsi_shm_transport.c
@@ -56,7 +56,7 @@ static enum iox_LogLevel to_iox_loglevel(enum ddsi_shm_loglevel level) {
         case DDSI_SHM_WARN : return Iceoryx_LogLevel_Warn;
         case DDSI_SHM_INFO : return Iceoryx_LogLevel_Info;
         case DDSI_SHM_DEBUG : return Iceoryx_LogLevel_Debug;
-        case DDSI_SHM_VERBOSE : return Iceoryx_LogLevel_Verbose;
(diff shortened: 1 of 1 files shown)


thijsmie:
Hi @YuanYuYuan,

This raises an interesting question for us maintainers. I think it is best we actually do not try to align with the `master` branch of iceoryx, but with the releases branch. If they are making API-breaking changes this will probably be a upcoming 3.0 release?

If this is the case we should also update our [shared memory guide](https://cyclonedds.io/docs/cyclonedds/latest/shared_memory.html) to tell people to check out the release branch. This is in fact [exactly what the CI does](https://github.com/eclipse-cyclonedds/cyclonedds/blob/a559b4376f58ef9dca0dd01e97bae4687d0c09fb/.azure/templates/build-test.yml#L93), hence the failing iceoryx test!

evshary:
I've just tested iceoryx v2.0.0 and it can work. Maybe we can fix at that version.
I guess the shared memory documentation might need some update since the branch should be master now.

YuanYuYuan:
Hi, @thijsmie, I agree with your point. So I think a doc update  PR #1502 can resolve the issue.
Also thanks for @evshary's help!

thijsmie:
Closed in favour of #1502, thanks again @YuanYuYuan and thanks for testing @evshary.

PR #1503, opened 2022-11-25
Export default participant QoS

reicheratwork:
Currently the default participant QoS is hardcoded inside ddsi_default_plist_participant, and not exported in the libraries To make this follow the same pattern as the default reader/writer/etc QoSes, the ddsi_default_qos_participant was created As a side effect, the ddsi_default_plist_participant was no longer necessary, as only its QoS component was used
The code was changed to use the default participant QoS in the places where the ddsi_default_plist_participant was used previously


eboasson (approved):
A nice clean-up! 🙂

Issue #682, opened 2021-01-31
Questions about Qos in cycloneDDS 


Jingzhao123:
Hi, 
I am a freshman in cyclonedds. I have a few questions about Qos in DDS. Could you please help to check it?
I notice there are 22 Qos-options in [DDS Spec](https://www.omg.org/spec/DDSI-RTPS/2.3/PDF) (8.7.2). And some of them have been implemented in dds_qset_xxx functions. My questions are:

- Have all Qos features been implemented in the latest cyclonedds version?

- Is there any document describing those Qos features in cycloneDDS?

Thanks very much!

eboasson:
> Have all Qos features been implemented in the latest cyclonedds version?

The `dds_qos_t` object can handle all QoS settings, but the functionality behind them is not yet implemented for all of them. The incomplete ones are:
* presentation (publisher/subscriber)
* time based filter
* durability kinds transient and persistent
* transport priority and latency budget (they are only hints, but still)
* reader data lifecycle

The rest is fully supported.

Do note that the "durability service" QoS setting is a bit of a odd one. Given that Cyclone does not yet implement the "transient" or "persistent" settings for the durability setting, most of the "durability service" setting is meaningless. Its "history" component is used by writers with "transient-local durability".

> Is there any document describing those Qos features in cycloneDDS?

There ought to be one, but there isn't currently. So it is good of you to ask!

Jingzhao123:
Thanks for your comments. It is a very good answer!

PR #1506, opened 2022-11-30
Implement time-based filter functionality

reicheratwork:
The QoSPolicy minimum_separation now does filtering of samples on the reader side.
Samples received within the minimum_separation window will not be put into the reader history cache, but instead be treated as being lost.
Added unittests to test this new functionality


src/core/ddsc/src/dds_rhc_default.c
```diff
@@ -1016,22 +1019,30 @@ static int inst_accepts_sample_by_writer_guid (const struct rhc_instance *inst,
 
 static int inst_accepts_sample (const struct dds_rhc_default *rhc, const struct rhc_instance *inst, const struct ddsi_writer_info *wrinfo, const struct ddsi_serdata *sample, const bool has_data)
 {
-  if (rhc->by_source_ordering)
```
eboasson, line 1019: I think the "time based filter" QoS should also apply when "by reception timestamp" order is selected - that is at least what I would expect if I imagine myself to be a regular user. This way it only does it for "by source timestamp". That is, in principle, trivially dealt with.

Another thing is that C says signed integer over/underflow is undefined behaviour, and that needs to be accounted for when checking whether the sample arrived in the "minimum separation" windor or not.

IINM, then, on input, the source time stamp of the sample and source timestamp stored in the instance can be anything (unavoidably so: data received from other implementations must be assumed to can be anything, and if the instance is created based on an sample with an "invalid" timestamp, then the only possible value for the instance is also "invalid").

If either (or both) of the timestamps is "invalid", then there is no meaningful time difference. I am tempted to say it means they are unordered with respect to each other. There is no way of building an eventually consistent system based on unordered samples, so we probably should also require that a "by source timestamp" writer *must* provide an non-"invalid" timestamp < ∞ for each eample, and discard everything else.

Then, we can observe that in the "by source timestamp" case, they are always ordered (even if the arithmetic is a bit tricky). For "by reception timestamp" we recognise they may be unordered, in which case we simply consider the new one to be sufficiently more recent. Alternatively, we could use another bit of information - e.g., the local clock - but there are dragons there too. Smart people just avoid these problems, use Cyclone DDS and are always guaranteed to have valid source timestamps 🙂

So I am wondering if it doesn't make the most sense to write it like:
```
if (by_source_ordering) { // (essentially) unchanged
  if (sample->timestamp.v > inst->tstamp.v)
    ;//possibly accept
  else if (sample->timestamp.v < inst->tstamp.v)
    return 0;// reject
  else if (!inst_accepts_sample_by_writer_guid (...))
    return 0;//reject
}
if (minimum_separation > 0 && sample->timestamp.v != INVALID && inst->tstamp.v != INVALID) {
  // new code; we can actually compute the separation because we made sure we have a
  // non-"invalid" timestamps
  if (sample->timestamp.v < INT64_MIN + minimum_separation || sample->timestamp.v - minimum_separation < inst->tstamp)
    return 0;//reject
  }
}
if (rhc->exclusive_ownership && inst->wr_iid_islive && inst->wr_iid != wrinfo->iid)
...
```

Which really only adds a bit for `minimum_separation` and the arithmetic is done in a way that avoids trouble. What do you think?

eboasson (commented):
Thanks @reicheratwork, and sorry it took so long even considering the holiday period. There are a few details that I think need another look, I'd appreciate it if you could look at my remarks and tell me what you make of them.

eboasson:
What I should have remarked is that this doesn't implement everything it says in the spec, in particular:

> 2.2.3.12 TIME_BASED_FILTER
> In the case where the reliability QoS kind is RELIABLE then in steady-state, defined as the situation where the DataWriter does not write new samples for a period “long” compared to the minimum_separation, the system should guarantee delivery the last sample to the DataReader.

Since more than a decade of dealing with DDS-based system has not resulted in getting a single solid use-case for this feature, I consider it a mis-feature and am of the opinion that the additional overhead of implementing this particular guarantee is not worth it.

I just thought of another thing, though that is more of an issue with the specification: I wonder how "exclusive ownership" and "time-based filter" are supposed to interact. I suppose we could define something here, but I am tempted to say: don't use either, and certainly never both 😁

So even though this PR does not implement everything in the spec and even though there some grey areas, I'd rather deal with those in the documentation than in the implementation. We should make a better effort at documenting what the QoS settings mean, perhaps noting these caveats would a good place to start?

src/core/ddsc/tests/time_based_filter.c
```diff
@@ -0,0 +1,109 @@
...
+{
+    pp = dds_create_participant(DDS_DOMAIN_DEFAULT, NULL, NULL);
+    CU_ASSERT_FATAL(pp > 0);
+
+    dds_entity_t tp = dds_create_topic(pp, &Space_Type1_desc, topic_name, NULL, NULL);
+    CU_ASSERT_FATAL(tp > 0);
+
+    //qos
+    dds_qos_t qos;
+    ddsi_xqos_init_empty (&qos);
```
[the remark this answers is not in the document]
reicheratwork: Agreed, I will change this to no longer put the dds_qos on the stack.

src/core/ddsc/tests/time_based_filter.c
```diff
@@ -0,0 +1,109 @@
...
+    {
+        rc = dds_get_status_changes (wr, &status);
+        CU_ASSERT_EQUAL_FATAL(rc, DDS_RETCODE_OK);
+
+        /* Polling sleep. */
+        dds_sleepfor (DDS_MSECS (20));
+    }
+}
+
+static void teardown()
```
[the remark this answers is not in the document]
reicheratwork: Okay, no problem :)

src/core/ddsc/tests/time_based_filter.c
```diff
@@ -0,0 +1,109 @@
...
+{
+    CU_ASSERT_EQUAL_FATAL (dds_delete(pp), DDS_RETCODE_OK);
```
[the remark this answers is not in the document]
reicheratwork: You are correct, this has occurred in other spots as well.
I will look into splitting up the call function and the check for its value.

src/core/ddsc/src/dds_rhc_default.c
reicheratwork: Having a `by_reception_timestamp` kind of filter will require some larger modifications, so we maybe do that in another PR and now just keep unsupported?

Maybe have some warnings or something when these "unsupported" or incompatible filter settings are attempted?

hansvanthag: The fact that a time-based filter is basically defined (in the spec) as a means to limit the number of 'seen updates' of arrived samples of instances, its (at least to me) something that's about 'local reader-side' behavior and when that's combined with the fact that most users use _by_reception_timestamp_ ordering (as source-timestamp-ordering requires time-alignment throughout the system), I'm very much in-favor of allowing time-based filtering for that (common) use-case... just wondering how that implies a 'large/larger modification'  ..

reicheratwork: The larger modification is defining/recording the reception timestamp, because at the moment it is not stored in either the `rhc_instance` or the `ddsi_serdata` entities.
And this information is needed to keep track of whether to store/drop samples in the reader.

eboasson: I think there's another possible view: that is that "destination order" switch really is (roughly) about whether something describes events you want to queue (`by_reception_timestamp`) or state you want to update (`by_source_timestamp`); and that it has nothing whatsoever to do with how you interpret "time-based filter".

That view allows you to use the source timestamp for the "time-based filter" in both cases. Of course it falls apart when people start using the source timestamp in creative ways, but the answer is simple: then don't do that.

In Cyclone (and in OpenSplice) there is *always* a source timestamp, it is never elided because someone chose "by_reception_timestamp" for the "destination_order" setting in the writer. That's one of the silliest things I have seen done ...

So in my view, the only problem is the existence of some *other* implementations where a "by reception timestamp" writer may not provide a valid timestamp. I'd just accept everything and suggest to use a proper implementation for the writing side as well to anyone who runs into that detail and considers it a problem.

src/core/ddsc/tests/time_based_filter.c
```diff
@@ -0,0 +1,109 @@
...
+
+    setup("ddsc_time_based_filter_sep", sep, dok);
+
+    test_f(DDS_MSECS(1), 0, 0);
+    test_f(DDS_MSECS(0), 1, 1);
+    test_f(DDS_MSECS(1), 2, 1);
+    test_f(DDS_MSECS(101), 2, 0);
+
+    teardown();
+}
```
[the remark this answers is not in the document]
reicheratwork: I will look in to this.

src/core/ddsc/src/dds_rhc_default.c
reicheratwork: If we are going for the option to not implement time based filters based on reception timestamps, which is the status quo at this moment in time, I do think that giving a warning when you attempt to use time filtering when the ordering is by_reception_timestamp is giving at least the correct feedback to the user.
We can then later choose to implement a reception timestamp functionality.

src/core/ddsc/tests/time_based_filter.c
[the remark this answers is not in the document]
reicheratwork: Added some extra unittests

AkhilTThomas:
Does this PR mean that time based filtering is not supported as of now even in cyclonedds-cxx? I was testing this feature out and i see that the new filter value is visible when i do ``cyclonedds ls`` but i see all the data packets, i.e. the filter has no effect.

eboasson:
> Does this PR mean that time based filtering is not supported as of now even in cyclonedds-cxx? I was testing this feature out and i see that the new filter value is visible when i do `cyclonedds ls` but i see all the data packets, i.e. the filter has no effect.

Yes, implementing the actual filtering was overlooked. Perhaps it would be fairer to say that I couldn't be bothered to implement it way back when because I have always considered it a rather useless feature that can easily lead one down a rabbit hole — and one with extremely little *real* benefits at that.

When we realised it was missing we decided to do something about it. That's this PR.

> I could not find a supported feature list for cyclonedds, apologies if i missed it. Could you point me to it? Like which QoS parameters are fully supported.

I'm pretty sure there is one somewhere ... https://github.com/eclipse-cyclonedds/cyclonedds/issues/682#issuecomment-770406960 contains one, but apparently I forgot that it needed to become part of the documentation. Or at the very least, that it should be placed somewhere that is easy to find.

It could be that I thought the list in the README was good enough:
> Cyclone DDS aims at full coverage of the specs and today already covers most of this. With references to the individual OMG specifications, the following is available:
> 
> [DCPS](https://www.omg.org/spec/DDS/1.4/PDF) the base specification
> * zero configuration discovery (if multicast works)
> * publish/subscribe messaging
> * configurable storage of data in subscribers
> * many QoS settings - liveliness monitoring, deadlines, historical data, ...
> * coverage includes the Minimum, Ownership and (partially) Content profiles
> 
> [DDS Security](https://www.omg.org/spec/DDS-SECURITY/1.1/PDF) - providing authentication, access control and encryption
> [DDS C++ API](https://www.omg.org/spec/DDS-PSM-Cxx/1.0/PDF)
> [DDS XTypes](https://www.omg.org/spec/DDS-XTypes/1.3/PDF) - the structural type system (some [caveats](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/xtypes_relnotes.md) here)
> [DDSI-RTPS](https://www.omg.org/spec/DDSI-RTPS/2.5/PDF) - the interoperable network protocol

But looking at it, no, it is not very clear for someone who doesn't know the specs inside out ...

AkhilTThomas:
@eboasson Thank you for your answer. [#682 (comment)](https://github.com/eclipse-cyclonedds/cyclonedds/issues/682#issuecomment-770406960) was helpful.

> Yes, implementing the actual filtering was overlooked. Perhaps it would be fairer to say that I couldn't be bothered to implement it way back when because I have always considered it a rather useless feature that can easily lead one down a rabbit hole — and one with extremely little _real_ benefits at that.

I wouldn't say its useless. I'm collecting use-cases for QoS and an immediate benefit of this is to have a sort of rate limiting on the network based on the slowest Reader. Idea is that all reader publish their readability rate via time based filter and the writer can adjust its output rate based on this.

reicheratwork:
Had to rebase on master + force-push since the "old" version still had the issue with the Conan version not working correctly

eboasson:
> @eboasson Thank you for your answer. [#682 (comment)](https://github.com/eclipse-cyclonedds/cyclonedds/issues/682#issuecomment-770406960) was helpful.
> 
> > Yes, implementing the actual filtering was overlooked. Perhaps it would be fairer to say that I couldn't be bothered to implement it way back when because I have always considered it a rather useless feature that can easily lead one down a rabbit hole — and one with extremely little _real_ benefits at that.
> 
> I wouldn't say its useless. I'm collecting use-cases for QoS and an immediate benefit of this is to have a sort of rate limiting on the network based on the slowest Reader. Idea is that all reader publish their readability rate via time based filter and the writer can adjust its output rate based on this.

If it is used to rate-limit the network traffic, then yes, there is/can be a benefit. How large that benefit is in practice is an interesting question, but there are certainly usage scenarios where it would be useful. The problem is with the implementation.

If I understand you correctly, you propose that the application adjust its publication rate, and of course that works out well because at the application level, the meaning of the data, the next time data will be generated, the variation in "minimum separation" values that can occur, &c., are all under the control of the designer.

What I have considered several times is what it would take to make it work predictably if DDS were to do it automatically, but then all those variations work against you. Then you do have to look at all possible combinations that can occur, and you have to do that without knowing when the next write will occur or how bad it is if something gets missed or delayed a bit. For example, you have to decide what to do when there is one reader with a "minimum separation" of 0 and one with 1s, when the writer is generating data at 1kHz. The data needs to go out, but not always to all ... (similar problems occur in writer-side content filtering, but at least there the temporal aspect is absent).

And then if you try to be smart about it, at what point do you give up and start sending everything to everyone? And does that then mean a small change in the system can cause a dramatic increase in network load? Is that actually desirable?

The net result is that until those questions have satisfactory answers, in all likelihood, Cyclone will send all samples to all readers regardless of the "minimum separation" setting. That in turn means you only avoid (roughly) a pointer swap in the reader history cache, and that's not so exciting ...

eboasson (approved):
Thanks @reicheratwork !

PR #1522, opened 2022-12-19
Add metadata to interface and some cleanup

dpotman:
This PR adds a meta-data property `Component` to the doxygen comments for local and globally exposed functions in DDSI, DDSC and CDR. For function that already had a comment or a parameter with a comment, a complete doxygen comment block is added. For functions that did not have any comment, a single-line comment with only the component name is used. 

A custom doxygen command `@component` is included, which takes the component handle (see below) as a single parameter. For user documentation generation this command is an alias for a no-op, as the component is not included in the user docs. 

The defined components are included as a csv file, but there is not yet an automated check that (a) all function have a component set and (b) the component handle used in the doxygen comment is valid. 

Additionally, this PR includes some minor cleaning up for some of the interfaces and sources.


dpotman:
I've force-pushed the branch to resolve the conflict with master (the new vendor code), and I've added some minor fixes in an additional commit. I think this PR is now in a state that it could be reviewed and merged; the sanity checks for completeness and correctness of component names can be added later in a separate PR.

PR #1524, opened 2022-12-20
Connext Micro interop + other low-level stuff

eboasson:
They say making branches in Git is easy, but "they" say so much ... 🤓

Everything in here is low-level stuff:
* The "oneliner" tests all ran with shared memory disabled, that's not practical; everyone but is smart enough to avoid it, so this is really just for me 😁
* Since https://github.com/eclipse-cyclonedds/cyclonedds/pull/1396 it is straightforward to map specific traffic onto a specific interface, but now it turns out that the port number in the locators was 0 if the participant index was left at "none". Slipped through the cracks because the unit tests for this operate on fake network configurations and so have to avoid some code paths. Now there's a test that verifies it does get set
* Someone ran into a case where some addresses would match two interfaces based on netmask. Now it will choose the one with the most selective netmask. "Find first set" got added to implement that, but then it also makes sense to use it in the hopscotch implementation
* RTI Connext Micro turns out to generate malformed ACKNACK messages when sending pre-emptive ACKNACKs. RTI Connext Pro and eProsima FastDDS do this too. Micro uses a different vendor code, so the workaround didn't trigger. I'd prefer not to do a work around, but RTI is not a company known for fixing their interop bugs, so what can one do? 🤷
* In the process of adding the network partition port number check, I found out that an illegal combination of settings wasn't properly flagged, so fixed that as well
* The DDSI spec requires that the "octetsToNextHeader" is a multiple of 4, except when it doesn't. Which is when it is the last submessage in the RTPS message, and according to the spec "one can use the length of the RTPS message" to find out which case it is. It is not straightforward. The silly thing is that it *also* allows 0 to be used in that case, so it isn't clear to me what the benefit of this complication is. It is legal, so we should handle it, and it so happens that Connext Micro generates it. I don't quite know whether the rather strict interpretation Cyclone takes with this commit will be ok in all cases, we'll see.


src/core/ddsc/tests/test_oneliner.h
```diff
@@ -303,6 +309,7 @@ struct oneliner_ctx {
   ddsrt_cond_t g_cond;
   struct oneliner_cb cb[3];
   
+  bool disable_sharedmemory;
```
dpotman, line 312: This variable seems to be unused (or am I overlooking something?)

dpotman (commented):
LGTM, except for a detail in the oneliner test for shm (see below)

src/core/ddsc/tests/test_oneliner.h
eboasson: That is indeed a remnant of an experiment. I'll delete it and force push the updated version. Thanks!

Issue #1508, opened 2022-11-30
IOS Support


jc211:
Hi, does cyclonedds compile for ios?

eboasson:
I once did that, but I had difficulty getting CMake to generate what I needed and manually constructed an XCode project instead, and that obviously didn't make it into the repository. (No doubt this is because *my* brain is incompatible with CMake.)

We're perhaps using a few more functions in the OS, but I bet they're all optional, so I see no reason why it wouldn't work today if you're willing to put some effort into getting the build process to do something at all.

I could probably salvage the test application (which was somewhat compatible with the "shapes" demo that various DDS vendors use — I think some kind souls even made one for Cyclone, but it is not part of the Cyclone repositories), but I am sure that you're not asking for information on building an ios application, only on building Cyclone for ios 😂

jc211:
Would be great to see the test application! 

Re: Compiling for iOS
This header could not be found https://github.com/eclipse-cyclonedds/cyclonedds/blob/46eda2d7786b55ca9704470b58c15cff3b93621f/src/ddsrt/src/netstat/darwin/netstat.c#L13.

eboasson:
> Re: Compiling for iOS
> This header could not be found [..] `if_types.h` in `netstat.c`

This is really exists only so `ddsperf` can gather some information on network loads, and it is a no-op on some other platforms. So a bit of `#ifdef`ing so that this "netstat" thing gets built on macOS but stubbed on iOS is a perfectly sensible trick. I'd be happy with a PR that does that, too.

> Would be great to see the test application!

I'm sure it won't build or work like this, it is 5 years old! But here it is:

[iosshapes.zip](https://github.com/eclipse-cyclonedds/cyclonedds/files/10139444/iosshapes.zip)

On a side note: as I am a mac user and do all my Cyclone development work on it (except for the rare occasion where I have no choice to use Windows as well as when Linux happens to be too different), and have an iPhone, I'm quite interested in your project ☺️

jc211:
Thanks for your help and for sharing the project! 

I got the library compiling for iOS but haven't integrated it into a project yet. 
FYI: I had to disable this section of the code: 
https://github.com/eclipse-cyclonedds/cyclonedds/blob/7fb79b81456a203d02878aa4f9ef53f60fb202fa/src/ddsrt/src/ifaddrs/posix/ifaddrs.c#L82
since if_media.h does not exist for iOS. 

[...]

PR #1531, opened 2023-01-02
Fix compiling for iOS

jc211:
#1508 Small changes to allow cmake to build for ios


eboasson (commented):
LGTM, Thanks @jc211 🙂

The rules for contributing to an Eclipse project require that an [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) is on file with the Eclipse Foundation. Would you be willing to sign one?

eboasson (approved):

PR #1536, opened 2023-01-09
Fix connext micro compat 0.10.x

eboasson:
Backport from #1524 — whether it actually makes sense to update 0.10.x or it makes more sense to do 11.0.0 is still to be decided.


eboasson (commented):
The identical changes have been in master for a couple of months without any adverse effect. So I will merge this.

Issue #1490, opened 2022-11-23
CycloneDDS+Iceoryx on Windows


EttoreVidotto:
Dear CycloneDDS Team,
I have a question for you, since the support of Iceoryx by CycloneDDS is not 100% clear to me.
On the official GitHub page of CycloneDDS I could not find any limitations regarding the usage of Iceoryx, since there only the CMake variable ENABLE_SHM is described. However, trying to build CycloneDDS with Iceoryx support on my laptop with Windows 10 is apparently not supported, since I get the CMake error:
```cmake
-- This project seems to be plain C, using 'MSVC' compiler
-- Library y found D:/dev/.conan/data/bison/3.7.1/_/_/package/58dfdc10302cb712787b358dceba3d18737fc982/lib/y.lib
-- Library iceoryx_posh_gateway found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_posh_gateway.lib
-- Library iceoryx_posh_config found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_posh_config.lib
-- Library iceoryx_posh_roudi found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_posh_roudi.lib
-- Library iceoryx_binding_c found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_binding_c.lib
-- Library iceoryx_posh found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_posh.lib
-- Library iceoryx_hoofs found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_hoofs.lib
-- Library iceoryx_platform found D:/dev/.conan/data/iceoryx/2.0.2/_/_/package/a1163d1672fcc9f437f21accbe5c1b6c02993e3a/lib/iceoryx_platform.lib
-- Looking for getopt.h
-- Looking for getopt.h - not found
[...]

PR #1539, opened 2023-01-12
Iceoryx on Windows

EttoreVidotto:
* Allow activation of shared memory support on Windows
* Throw an error during CMake configuration, if CycloneDDS is built as a shared library
* Link security_core as BUILD_INTERFACE
* Define `ddsrt_eth_get_mac_addr` for Windows


EttoreVidotto:
Issue https://github.com/eclipse-cyclonedds/cyclonedds/issues/1490 requires the definition of the function `ddsrt_eth_get_mac_addr` for Windows. 
Does anyone have time/interest to take a look at my implementation?

EttoreVidotto:
@eboasson What do you think about this implementation?

eboasson (commented):
LGTM, although I admit I have never figured out the intricacies of wide character strings in Windows-land, so I have no idea if it should be `strcmp` is correct ...

Perhaps we should add a test to ddsrt to check that it either returns `UNSUPPORTED` or something plausible if the result buffer is all-0 on input. It is still somewhat tricky to do that, not in the least because it would require picking an interface with a MAC address, and so running it over the loopback interface would then fail.

I'm probably in favour of not worrying too much about that testing detail: it would be a shame to leave something valuable out for want of a good testing strategy.

EttoreVidotto:
@eboasson Thank you very much for the answer. I'm not very familiar with such details as well, but with this implementation I was able to retrieve the MAC address and carry on the inclusion of iceoryx in my project. The code now hangs somewhere else, but I describe this information in my issue report https://github.com/eclipse-cyclonedds/cyclonedds/issues/1490.
Can maybe someone else give a quick look at it?

eboasson (approved):
Thanks @EttoreVidotto !

PR #1541, opened 2023-01-18
Fix memory leak on writing invalid data

### LINKED ISSUE #1540: Memory leak when cdr failed
When serialization failed (for example a boolean object has a value of 2 other than 0/1), `dds_write` returns error code `-3`. If I ignore this error and go on calling `dds_write`, memory comsumption grows.

After debug with valgrind, it shows `serdata_default_from_sample_cdr_commo` has memory leak. When `dds_stream_write_sample` returns NULL, the ostream buffer or the serdata object is not freed.

eboasson:
The memory allocated for the CDR representation wasn't freed on detecting an invalid input sample while serialising. Fixes #1540.


src/core/ddsc/src/dds_serdata_default.c
```diff
@@ -603,8 +603,10 @@ static struct dds_serdata_default *serdata_default_from_sample_cdr_common (const
       }
       break;
     case SDK_DATA:
-      if (!dds_stream_write_sample (&os, sample, &tp->type))
+      if (!dds_stream_write_sample (&os, sample, &tp->type)) {
+        dds_ostream_fini (&os);
```
dpotman, line 607: Does this also work when a serdata pointer is used from the freelist (line 570 set `os` to `d`, which can be from the freelist). And does `d` also need to be freed on lines 602 and 612?

eboasson: Thank you @dpotman! Clearly I am no longer as clever as I used to be 😂

It is such a small change that I decided to try again and I simply force-pushed it. I'd appreciate it if you would have another look.

dpotman (commented):
Looks good, the only thing I noticed (not related to these code changes) is that the call to `dds_stream_write_key` on line 258 could also fail (e.g. enum value out of range), although the error is not propagated so `gen_serdata_key_from_sample` doesn't return the error. And in case of `serdata_default_from_sample_cdr_common` it will not get to `dds_stream_write_key` because `dds_stream_write_sample` will also fail in that case.

eboasson:
@dpotman thanks, and yes, I also noticed this detail. I suspect addressing that properly requires reworking quite a bit, including updating all the return types (`bool` to `dds_retcode_t`) so we can also distinguish properly between invalid data and running out of memory.

Doing that as part of this PR doesn't seem so sensible. Probably I should create a new issue for it so it doesn't get forgotten, but I'll have a look at the code first.

PR #1549, opened 2023-01-25
Require Windows static build only if SHM enabled

eboasson:
This addresses

  CMake Error at src/CMakeLists.txt:50 (message):
    CycloneDDS with enabled shared memory support can only be built as a static
    library on Windows

on Windows when ENABLE_SHM=AUTO is in effect.


reicheratwork:
Looks good to me.
Approved.

Issue #1546, opened 2023-01-23
Chain of trust issues with a single CA certificate


ruffsl:
I'll quote the same succinctly put issue filled over for OpenDDS here:

https://github.com/OpenDDS/OpenDDS/pull/3992#issue-1548228907

> ## Problem
> Assume a usage of DDS Security where the same CA is used for both permissions and identity. The certificates issued to particpants allow them to sign documents. Assume the participant generates a permissions file and then signs it. Chain verification causes verification attempts to succeed since the signing certificate, i.e., the participant's certificate, can be chained back to the permission CA's certificate (which is also the identity CA).
> 
> This problem was identified in [ros2/sros2#282](https://github.com/ros2/sros2/issues/282).
> 
> ## Solution
> Implement the suggestion in [ros2/sros2#282](https://github.com/ros2/sros2/issues/282). Specifically, use PKCS7_NOINTERN to not accept any signatures in the signed document. This, in turn, requires the use of the `certs` parameter to `PKCS7_verify`. PKCS7_NOVERIFY is used since the permissions CA certificate will not be chain verified.

Example call site for reference:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/65981fb935e028c3fbb3b0fbafcbebf3013aa59a/src/security/builtin_plugins/access_control/src/access_control_utils.c#L251

PR #1551, opened 2023-01-27
Add fix suggested in sros2 bug report.

jpace121:
### Description

Fixes https://github.com/eclipse-cyclonedds/cyclonedds/issues/1546 by taking the solution from https://github.com/OpenDDS/OpenDDS/pull/3992#issue-1548228907. As mentioned in the OpenDDS PR:

"Specifically, use PKCS7_NOINTERN to not accept any signatures in the signed document. This, in turn, requires the use of the certs parameter to PKCS7_verify. PKCS7_NOVERIFY is used since the permissions CA certificate will not be chain verified."

Related to: https://github.com/ros2/sros2/issues/282

### Verification
I built ros2 rolling from scratch with version of 0.10.2 cyclonedds and used the script in the sros2 linked above to verify the error was replicate-able. I then cherry-picked the commit in this PR to the 0.10.2 version of cyclonedds and repeated the test and confirmed the issue did not replicate.

### DIFF
--- a/src/security/builtin_plugins/access_control/src/access_control_utils.c
+++ b/src/security/builtin_plugins/access_control/src/access_control_utils.c
@@ -248,7 +248,7 @@ static bool PKCS7_document_verify(PKCS7 *p7, X509 *cert, BIO *inbio, BIO **outbi
   else
   {
     X509_STORE_add_cert(store, cert);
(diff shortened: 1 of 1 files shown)


jpace121:
Converted to draft. While this fixes the observed issue, I would like to take a second look in the morning and make sure the API call makes sense again. If someone sees something I missed initially I would appreciate the heads up.

jpace121:
Nope.... Way jumped the gun here...

PR #1552, opened 2023-01-27
Add fix suggested in sros2 bug report.

### LINKED ISSUE #1546: Chain of trust issues with a single CA certificate

jpace121:
### Description
Fixes #1546 using the same method as https://github.com/OpenDDS/OpenDDS/pull/3992#issue-1548228907.

As stated in: https://github.com/OpenDDS/OpenDDS/pull/3992#issue-1548228907

"Implement the suggestion in ros2/sros2#282.
Specifically, use PKCS7_NOINTERN to not accept any signatures in the signed document. This, in turn, requires the use of the certs parameter to PKCS7_verify. PKCS7_NOVERIFY is used since the permissions CA certificate will not be chain verified."

### Verification
I built ros2 rolling from source with cyclonedds tag  0.10.2 checked out and used the scripts [here](https://gist.github.com/jpace121/b14b614089741480cd5894fdfc6851ea) copied and slightly modified from the sros2 issue to verify the reported issue. Specifically, the should-fail script should not allow the subscriber to receive messages at the end.

I then cherry-picked this issue a branch forking from the 0.10.2 tag and verified the should-fail script now failed. I used the should-succeed script to test a configuration where the permissions did allow the two nodes to talk, and verified that worked as well.

I referenced the openssl docs [here](https://www.openssl.org/docs/man1.0.2/man3/PKCS7_verify.html) to confirm the API and heavily based this PR on the OpenDDS PR linked above.

Fixes: https://github.com/eclipse-cyclonedds/cyclonedds/issues/1546 
Related to: https://github.com/ros2/sros2/issues/282

Signed-off-by: James Pace <jpace121@gmail.com>

(Apologies for the last PR. I was a little quick to the draw there...)

### DIFF
--- a/src/security/builtin_plugins/access_control/src/access_control_utils.c
+++ b/src/security/builtin_plugins/access_control/src/access_control_utils.c
@@ -234,7 +234,7 @@ static bool PKCS7_document_from_data(const char *data, size_t len, PKCS7 **p7, B
 static bool PKCS7_document_verify(PKCS7 *p7, X509 *cert, BIO *inbio, BIO **outbio, DDS_Security_SecurityException *ex)
 {
   bool result = false;
-  X509_STORE *store = NULL;
+  STACK_OF(X509) *certStack = NULL;
 
   assert(p7);
   assert(cert);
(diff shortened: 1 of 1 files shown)


eboasson (approved):
That took some reading and double-checking to even feel I am allowed to write "LGTM" 😅 I concur with the analysis and proposal on sros and then this is the logical consequence. Thank you very much @jpace121 !

Issue #1554, opened 2023-02-01
Seg fault in cyclonedds publisher when using uint8 in IDL and OpenDDS subscriber


smnrgrs:
Most easily demonstrated using i11eperf. 
Add a uint8 field to the test topic.
Run the osub subscriber. 
The cyclone publishers (asub, csub, python) seg fault.
Tested using cyclone 0.11.0, from October (so may be a little out of date)

```
#0  xt_is_fully_descriptive (t=t@entry=0x0) at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:2046
#1  0x00007ffff7f36cf5 in xt_is_non_hash (xt=xt@entry=0x0) at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:2625
#2  0x00007ffff7f37640 in xt_validate_impl (gv=0x55555555c7f8, t=0x0, validate_hash_type=<optimised out>)
    at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:846
#3  0x00007ffff7f37a80 in xt_validate_impl (gv=0x55555555c7f8, t=0x7fffe80061e0, validate_hash_type=<optimised out>)
    at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:860
#4  0x00007ffff7f3a699 in ddsi_xt_validate (t=0x7fffe80061e0, gv=0x55555555c7f8)
    at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:959
#5  ddsi_xt_type_add_typeobj (gv=gv@entry=0x55555555c7f8, xt=xt@entry=0x7fffe80061e0, to=to@entry=0x7ffff70f9370)
    at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typewrap.c:1235
#6  0x00007ffff7efcd86 in ddsi_type_add_typeobj (gv=0x55555555c7f8, type=0x7fffe80061e0, type_obj=0x7ffff70f9370)
    at /home/simon/Development/cyclonedds/src/core/ddsi/src/ddsi_typelib.c:414
#7  0x00007ffff7efd7b5 in ddsi_type_add_typeobj (gv=gv@entry=0x55555555c7f8, type=type@entry=0x7fffe80061e0, 
[...]

PR #1557, opened 2023-02-07
Add error handling for deps in type wrapper initialization

dpotman:
The XTypes type wrapper failed to verify the return value of the register dependencies function calls, which resulted in ignoring validation errors in dependent types (e.g. a structs member types) and could result in other problems, e.g. the issue reported in #1554. This fixes the issue by checking the return codes and propagating errors.

### DIFF
--- a/src/core/ddsi/src/ddsi_typewrap.c
+++ b/src/core/ddsi/src/ddsi_typewrap.c
@@ -38,6 +38,7 @@
 static ddsi_typeid_kind_t ddsi_typeid_kind_impl (const struct DDS_XTypes_TypeIdentifier *type_id);
 static void ddsi_xt_get_typeid_impl (const struct xt_type *xt, struct DDS_XTypes_TypeIdentifier *ti, ddsi_typeid_kind_t kind);
 static bool xt_is_non_hash (const struct xt_type *xt);
+static void xt_applied_member_annotations_fini (struct xt_applied_member_annotations *ann);
 
 void ddsi_typeid_copy_impl (struct DDS_XTypes_TypeIdentifier *dst, const struct DDS_XTypes_TypeIdentifier *src)
 {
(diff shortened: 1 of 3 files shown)


src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -1141,7 +1188,17 @@ static dds_return_t add_complete_typeobj (struct ddsi_domaingv *gv, struct xt_ty
       {
         xt->_u.union_type.members.seq[n].id = cto->_u.union_type.member_seq._buffer[n].common.member_id;
         xt->_u.union_type.members.seq[n].flags = cto->_u.union_type.member_seq._buffer[n].common.member_flags;
-        ddsi_type_register_dep (gv, &xt->id, &xt->_u.union_type.members.seq[n].type, &cto->_u.union_type.member_seq._buffer[n].common.type_id);
+        if ((ret = ddsi_type_register_dep (gv, &xt->id, &xt->_u.union_type.members.seq[n].type, &cto->_u.union_type.member_seq._buffer[n].common.type_id)) != DDS_RETCODE_OK)
+        {
+          for (uint32_t m = 0; m < n - 1; m++)
+          {
+            ddsi_type_unref_locked (gv, xt->_u.union_type.members.seq[m].type);
+            ddsrt_free (xt->_u.union_type.members.seq[m].label_seq._buffer);
```
eboasson, line 1196: Shouldn't there be a `xt_applied_member_annotations_fini` here, too, like in the structure case?

dpotman: Yes, good catch, I've fixed this in [f65235c](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1557/commits/f65235ca3c8c75b06a7790cadab765302ca6400a)

eboasson (approved):
Thanks @dpotman!

PR #1558, opened 2023-02-07
Add OSS fuzzer test for type object validation

dpotman:
This adds an OSS fuzzer test for type object validation. To get to the point where a type object can be validated, the input for this test is a type-map that contains a set of type objects, which are added to the type library. So that means that this test also (or mainly, as it does not get to the validation in many cases) tests the deserialization of a typemap (which includes deserialization of its embedded type objects).

@eboasson I'm not sure if this is the fuzzer test you had in mind for #1554, but at least this test has discovered a few bugs in the cdrstream serializer (for which I'll create separate PRs with the fixes)


eboasson (approved):
Thanks @dpotman ☺️ Yes, at some point we should make an effort to fuzz a greater variety of types, but this looks like an excellent first step to me.

PR #1560, opened 2023-02-07
Fix in cdrstream skip_default function for delimited and PL-CDR

dpotman:
The skip-default functions in the cdrstream serializer were not taking delimited and PL-CDR into account when skipping members and setting defaults. This fixes the issue, which was discovered by running the OSS fuzzer tests from #1558.

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -1768,6 +1768,13 @@ static const uint32_t *dds_stream_skip_adr_default (uint32_t insn, char * __rest
   void *addr = data + ops[1];
   /* FIXME: currently only implicit default values are used, this code should be
      using default values that are specified in the type definition */
+
+  if (op_type_external (insn))
+  {
+    *(void **) addr = NULL;
+    return dds_stream_skip_adr (insn, ops);
+  }
+
   switch (DDS_OP_TYPE (insn))
   {
     case DDS_OP_VAL_BLN: case DDS_OP_VAL_1BY: *(uint8_t *) addr = 0; return ops + 2;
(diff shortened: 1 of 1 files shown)


src/core/cdr/src/dds_cdrstream.c
```diff
@@ -1810,9 +1817,76 @@ static const uint32_t *dds_stream_skip_adr_default (uint32_t insn, char * __rest
...
 
+static const uint32_t *dds_stream_skip_delimited_default (char * __restrict data, const uint32_t * __restrict ops)
+{
+  return dds_stream_skip_default (data, ++ops);
+}
+
+static bool dds_stream_skip_pl_member_default (char * __restrict data, const uint32_t * __restrict ops)
+{
+  uint32_t insn;
+  while ((insn = *ops) != DDS_OP_RTS)
```
eboasson, line 1832: Methinks this function can be rewritten in a slightly clearer manner: the only successful way out for `dds_stream_skip_pl_member_default` is hitting `DDS_OP_JSR`, and so having a loop that has as its declared termination condition reaching a `DDS_OP_RTS`, that doesn't update the loop variable and which unconditionally `abort`s on reaching the termination condition seems ever so slightly obfuscated.

Maybe I'm missing something, but isn't this function really just:
```
insn = *ops;
assert (DDS_OP (insn) == DDS_OP_JSR);
return dds_stream_skip_pl_member_default (data, ops + DDS_OP_JUMP (insn));
```

dpotman: Yeah, that implementation was definitely incorrect, I made a stripped-down version of `dds_stream_skip_default` and forgot to including updating the `ops` pointer (and the final `abort` shouldn't be there). Because the `ADR` for a PL-member is actually a type on itself delimited by an `RTS`, it can just use `dds_stream_skip_default` for skip-with-default of that member. It slipped through because the type object deserialization only required skipping appendable types, not mutable. In [8409126](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1560/commits/840912631c92ad9229968815364af338d6b96c15) I've added an improved version, but there should be some tests for this, so I'll add these in another commit.

eboasson (approved):
Thanks @dpotman !

PR #1564, opened 2023-02-09
Add BUILD_IDLC_ONLY option to only build the IDL preprocessor

PetervdPerk-NXP:
This PR adds a CMake option to only build the idlc tool.
Which is useful for projects using cyclonedds as a depedency but are not interested in using the dds part.


eboasson (commented):
I see the value of being able to build only IDLC, but doing it by having an option that then builds a special subset seems likely to suffer from bit-rot.

The more elegant way would be to reorganise things a little bit so that IDLC is essentially a separate subdirectory (instead of being spread out of over `tools/idl*` and `idl`). It seems it would then not be too far from allowing the IDLC CMake to also work as a stand-alone project.

There are some complicating factors, one is `ddsrt`, and the other is the "dependency" on the core library. That dependency is really limited to the code for supporting the constructing of XTypes TypeObjects, and those are very intimately tied to DDS. The IDLC code can run without generating those, so we could leave that bit out of a stand-alone IDLC build.

So I suppose the question is: would that direction be of interest to you?

@dpotman do feel free to chime in if you have some good ideas 🙂

PetervdPerk-NXP:
Makes sense to refactor, a long term goal would also to make the CDR serializer/deserializer standalone option.

The idea would be to only build the IDLC compiler and the CDR serializer as a library so that other modules i.e. Zenoh can use it serialize IDL to communicate with CycloneDDS/ROS2 using [zenoh-plugin-dds](https://github.com/eclipse-zenoh/zenoh-plugin-dds).

The question is if we can nicely decuple both IDLC and `dds_cdrstream.c` from `ddsrt` and if there's interest from the CycloneDDS maintainers to do so.

dpotman:
@PetervdPerk-NXP, regarding building only the IDL compiler, you could also consider to build only the specific cmake target: `cmake --build . --target idlc` (although in that case you can't use the `install` target to install only the IDLC binary).

PR #1569, opened 2023-02-15
Fix race condition with GC in get_matched_..._data

eboasson:
Race was introduced in bb9fdf481ac759078aea67497ef96ce7dabf6287


dpotman (approved):
LGTM!

PR #1573, opened 2023-02-16
Add tests for cdrstream skip-default and fix some issues

dpotman:
This PR adds and improves tests for the skip-default functionality in the cdrstream serializer for appendable and mutable types. These tests revealed a few new issues, that are fixed in c66e943eaf52c224fbe2537d586651b5792facbb:
- Memory for members of an existing sample that is passed when deserializing data is not freed in all cases:
   - in the skip-default functionality of the cdrstream serializer
   - when reading a sample of a mutable type in case the reader type has additional members that are not in the data
- Skip-default for an array with enum or bitmask element type was not updating the ops pointer correctly

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -1041,7 +1041,7 @@ static const uint32_t *skip_array_default (uint32_t insn, char * __restrict data
     case DDS_OP_VAL_ENU: case DDS_OP_VAL_BMK: {
       const uint32_t elem_size = DDS_OP_TYPE_SZ (insn);
       memset (data, 0, num * elem_size);
-      return ops + 3 + (subtype == DDS_OP_VAL_BMK ? 1 : 0);
+      return ops + 4 + (subtype == DDS_OP_VAL_BMK ? 1 : 0);
     }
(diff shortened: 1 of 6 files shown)


src/core/cdr/src/dds_cdrstream.c
```diff
@@ -1769,10 +1769,14 @@ static const uint32_t *dds_stream_skip_adr_default (uint32_t insn, char * __rest
...
-  if (op_type_external (insn))
+  /* Free memory in sample and set pointer to null in case of optional or external member.
+     test for optional (which also gets the external flag) is added because string type
+     is the exception for this rule, that does not get the external flag */
+  if (op_type_external (insn) || op_type_optional (insn))
   {
-    *(void **) addr = NULL;
-    return dds_stream_skip_adr (insn, ops);
+    ops = stream_free_sample_adr(insn, data, ops);
+    *(char **) addr = NULL;
```
eboasson, line 1778: For "external", this assignment is unnecessary because `stream_free_sample_adr` already does it. However, it does not store a null pointer after freeing the memory and this makes me wonder: if it is needed in the one case, is it then not needed in the other case?

eboasson (approved):
Thanks @dpotman!

Issue #1572, opened 2023-02-16
can not deserialize TypeInformation, which serialized in XCDR


binbowang1987:
In SEDP message, reader anounced that it supports XCDR and XCDR2.
![图片](https://user-images.githubusercontent.com/8586061/219262762-591cb8f0-331a-485d-b492-8fcf2cad7fb1.png)

But, when look into the `TypeInformation` deserialization code. CycloneDDS only support desialize `TypeInformation` in CDR2.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/011f80a68e3b44fdcc07b0065473f4cbfa5470a5/src/core/ddsi/src/ddsi_plist.c#L781-L808

eboasson:
That's because they concern different things. `DATA_REPRESENTATION` deals with the encodings the reader accepts for *application* data, but `TypeInformation` is not application data but internal meta-data.

The `TypeInformation` is simply spec'd as little-endian XCDR2:

> 7.3.4.5 TypeObject serialization
>
> The serialization of a TypeObject shall happen in accordance to its IDL declaration and the general serialization rules defined in this specification (see Clause 7.4) for XCDR encoding version 2. Additional restrictions are placed such that the serialized result is bitwise identical independently of the vendor or platform where the serialization occurs. Specifically:
> * The serialization shall use Little Endian encoding.
> * ...

Interesting screenshot by the way: I hadn't expected to see screenshots of Wireshark that don't recognise parameter id 0x75 as `TypeInformation`, nor can I recall having seen a "Compression Id Mask" before.

And last but not least, would you be willing to tell me which DDS implementation this is coming from? I like to keep track of implementations that I know Cyclone interoperates with, and I can't see how this could've come out of Cyclone: lifespan wouldn't be part of a reader endpoint description. 😁

(And besides, ∞ would be the default anyway, and so Cyclone won't publish it unless it is configured to publish all QoS settings, but then it doesn't fit the order I would expect.)

binbowang1987:
The DDS Implementation in wireshark is CycloneDDS(0.10.2).
Recently, I am comparing the implementation of XTypes(1.2) between CycloneDDS and FastDDS. Find some diffrences, FastDDS use `PL_CDR_LE` to encode TypeLookupService related requests, and CycloneDDS uses `CDR2_LE`. FastDDS also use CDR1 to encode TypeInformation. And CycloneDDS can not interoerate with FastDDS through XTypes(1.2)
[...]

PR #1574, opened 2023-02-17
Do not inherit irrelevant topic QoS in rd/wr

eboasson:
The set of topic QoS is an extended subset of the reader QoS (e.g., the "lifespan" QoS). Allowing those that are irrelevant to the reader to end up in the reader's QoS object means it shows up as a difference with the default QoS settings and causes it to be published in the endpoint discovery data.

As the topic QoS object is deliberately left sparse, this only happens when the application explicitly sets a QoS on the topic that is not relevant to the reader. For example, setting the lifespan in the topic QoS -- and even explicitly setting it to the default value -- causes this.

The impact is very low because the recipients of the endpoint discovery data ignore irrelevant (but otherwise valid) QoS values.

See #1572 Thanks @binbowang1987!


dpotman (approved):
LGTM

PR #1577, opened 2023-02-23
Add fix for Conan update

reicheratwork:
The default Conan version was updated to 2.0, so now we force fix the version to 1.59.0 as the current CUnit version (2.3.1) is currently not compatible with Conan v2.0

@NoxPardalis @dpotman , could you give this a look?

### DIFF
--- a/.azure/templates/install-conan.yml
+++ b/.azure/templates/install-conan.yml
@@ -29,7 +29,7 @@ steps:
     name: cache_conan
   - bash: |
       set -e
-      pip install conan --user --upgrade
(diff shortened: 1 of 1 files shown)


NoxPardalis (approved):
Looks like a sensible workaround to me. I guess we should come back to this at some point when `cunit` supports installation via `conan` v2.0 as well?

eboasson:
> Looks like a sensible workaround to me. I guess we should come back to this at some point when `cunit` supports installation via `conan` v2.0 as well?

I think we should do so, but for now I, too, think this is the sensible thing to do.

PR #1556, opened 2023-02-07
Add support for (u)int8 type kind in IDLC and type library

dpotman:
This adds support for the int8 and uint8 type kinds in the IDL compiler and the type library and XTypes wrapper. The IDL file describing the type object that is provided by the OMG spec was not updated for these missing type kinds, which was fixed in the spec text (including the IDL that is included in the spec as annex). See also https://issues.omg.org/issues/DDSXTY14-52

The Python binding, that also has its own implementation for building type objects, also needs an update for these kinds. I'll create a separate PR for that. 

Fixes #1554

PR #1578, opened 2023-02-23
Dynamic Type API

dpotman:
This PR introduces the Dynamic Type API, part of the Dynamic Language Binding as defined in section 7.5.2 of the DDS XTypes 1.3 specification. 

Note that the first commits in this PR is cherry-picked from #1556. That PR needs to be merged before this PR.


dpotman:
The CI issues are fixed and more tests are included, so that makes this PR ready for review. The API is not fully complete, some features are not supported yet (e.g. extensibility on enumerated types, union discriminant as key), but I think the current set of features is sufficient for most use-cases. 

Note that this PR depends on (and includes a commit from) #1556.

eboasson (approved):
Thanks for this extension @dpotman! It should make several people happy, and it certainly will make possible some interesting improvements in the ROS 2 RMW layer. 🥳

PR #1591, opened 2023-03-02
Iceoryx partition cross-talk

eboasson:
The mapping of readers/writers to Iceoryx failed to take the partition QoS into consideration, and so allow data from a writer using partition A to arrive at a reader using partition B.

This changes the logic to:

* not use Iceoryx for any complicated cases, that is: where there is more than one partition or a wildcard is involved; and

* extends the "topic" name in Iceoryx to include the (now known-to-be-simple) partition name

This eliminates the crosstalk and hopefully is flexible enough to not avoid using Iceoryx in important cases.

### DIFF
--- a/src/core/ddsc/CMakeLists.txt
+++ b/src/core/ddsc/CMakeLists.txt
@@ -47,7 +47,9 @@ prepend(srcs_ddsc "${CMAKE_CURRENT_LIST_DIR}/src/"
   dds_loan.c)
 
 if(ENABLE_SHM)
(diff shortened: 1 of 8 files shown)


eboasson:
@dpotman would you be willing to review this?

eboasson (commented):
Merging after timing out on the review process ... because it really fixes a serious issue and there is confirmation that it indeed addresses the problem.

dpotman:
The notification for the review request ended up in my spam folder, but the changes look good to me!

PR #1594, opened 2023-03-07
Fix detection of annotation identifier clash and add -Werror to IDLC

dpotman:
This fixes the issue that a name clash with a (built-in) annotation was not correctly detected by the IDL compiler. This could result in overwriting built-in annotations, making these annotations a no-op when applied later in the same IDL fragment. 

IDLC shows a warning in case an unknown (overwritten) annotation is used. This PR adds an `-Werror` option to treat this warning as an error (from cmake using the `WERROR` option for the `idlc_generate` function). 

Fixes https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/375

### DIFF
--- a/src/idl/src/scope.c
+++ b/src/idl/src/scope.c
@@ -50,9 +50,8 @@ create_declaration(
   declaration->kind = kind;
   if (!(identifier = idl_strdup(name->identifier)))
     goto err_identifier;
-  if (idl_create_name(pstate, idl_location(name), identifier, &declaration->name))
+  if (idl_create_name(pstate, idl_location(name), identifier, name->is_annotation, &declaration->name))
     goto err_name;
-  declaration->name->is_annotation = name->is_annotation;
   *declarationp = declaration;
   return IDL_RETCODE_OK;
 err_name:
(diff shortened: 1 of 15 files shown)


src/idl/tests/annotation.c
```diff
@@ -942,6 +942,39 @@ CU_Test(idl_annotation, parameter_scope)
 #undef A
 #undef TA
 
+CU_Test(idl_annotation, identifier_clash)
+{
+  static const struct {
+    const char *str;
+    idl_retcode_t ret;
+  } tests[] = {
+    { "struct key { @key long f1; };", IDL_RETCODE_SEMANTIC_ERROR },
```
eboasson, line 951: Is this really correct?

In XTypes 1.3 it says:

> 7.3.1.2.4 Defining Annotations
> Annotation types shall be represented as described in this clause. An annotation type is defined using the new token @annotation, as in the following example:
> ```
> @annotation MyAnnotation { // ...
> };
> ```
> Annotation identifiers are orthogonal to any other kind of type and therefore do not conflict with other types that may use the same identifier name even when defined in the same module. This is because the application of an annotation prefixes the annotation identifier with the “@” character, see Sub Clause 7.3.1.2.5.

That suggests that defining a `struct key` is perfectly acceptable.

dpotman: You're right, I missed that section. In the updated commit I've rewritten the fix, which I think now correctly handles the cases with an annotation that has the same name as a type in the same scope. What still does not work is using annotations with a fully scoped name, e.g. `@module1::my_annotation` (and also: `@module1::key`, which I think is allowed). I'll create an issue for this.

src/idl/src/scope.c
```diff
@@ -38,6 +52,7 @@ create_declaration(
     goto err_identifier;
   if (idl_create_name(pstate, idl_location(name), identifier, &declaration->name))
     goto err_name;
+  declaration->name->is_annotation = name->is_annotation;
```
eboasson, line 55: I have only one comment: would it not be better to let `idl_create_name` set the `is_annotation` field based on a parameter?

dpotman: Yes, I also considered that, but the first try didn't work out well (because the call from generic `identifier` in the parser), so I decided not to do that. But looking at this again, I realized that based on the parser state, I can set the value for that flag at that point. I've changed this in e402ec2

eboasson (approved):
Thanks @dpotman!

PR #1600, opened 2023-03-09
Parameterize allocator functions in CDR stream serializer

### LINKED ISSUE #1586: idlc: Wrong memory alloc/free between dll and exe
I am currently updating a conan recipe for cyclonedds, and as part of this, I'm building on Windows with compiler runtime MTd.

Used Version: Current master branch (https://github.com/eclipse-cyclonedds/cyclonedds/commit/7e7bcd2b8b9d8a4022283867b1e018d2727fe065)

I'm hitting the following error during compilation:

```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Assertion Failed!

Program: C:\.conan\b8fe73\1\build\bin\Debug\ddsc.dll
File: minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp
Line: 996

Expression: __acrt_first_block == header

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts.

(Press Retry to debug the application)

---------------------------
Abort   Retry   Ignore   
---------------------------
```

image

As you can see in the screenshot this happens when calling idlc.exe.

The error message typically means that the memory allocations do not match, i.e., malloc is done in the .exe, while free is done in the .dll (or the other way around).

The master branch should already include the fixes from https://github.com/eclipse-cyclonedds/cyclonedds/pull/1438

Seems there is still at least one place where malloc/free are not properly called via `idl_malloc` and `idl_free`

dpotman:
This parameterizes the memory allocator functions in the CDR stream serializer, so that an alloc/free is not crossing the library boundary (when used from IDLC). 

Fixes #1586

### DIFF
--- a/src/core/cdr/include/dds/cdr/dds_cdrstream.h
+++ b/src/core/cdr/include/dds/cdr/dds_cdrstream.h
@@ -233,9 +233,6 @@ uint32_t dds_stream_type_nesting_depth (const uint32_t * __restrict ops);
 bool dds_stream_extensibility (const uint32_t * __restrict ops, enum dds_cdr_type_extensibility *ext);
 
 /** @component cdr_serializer */
-DDS_EXPORT void dds_cdrstream_desc_from_topic_desc (struct dds_cdrstream_desc *desc, const dds_topic_descriptor_t *topic_desc);
-
-/** @component cdr_serializer */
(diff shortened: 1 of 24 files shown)


dpotman:
I've also included a commit that moves `dds_cdrstream_desc_from_topic_desc` to DDSC, so that CDR stream does not depend on DDSC. And created https://github.com/eclipse-cyclonedds/cyclonedds-python/pull/185, that has the changes required in the Python binding for the updated CDR stream interface.

PR #1602, opened 2023-03-10
Allow CDR to be compiled as standalone project/library

PetervdPerk-NXP:
Allows CDR to compiled as its own project.
Which creates libcdr.so library target so that external applications can utilize the cdr serializer/deserializer.

### DIFF
--- a/src/core/cdr/CMakeLists.txt
+++ b/src/core/cdr/CMakeLists.txt
@@ -10,19 +10,64 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 
+if(NOT ${CMAKE_PROJECT_NAME} STREQUAL "CycloneDDS")
+  cmake_minimum_required(VERSION 3.16)
+  project(cdr VERSION 0.11.0 LANGUAGES C)
+endif()
+
 set(srcs_cdr
-  dds_cdrstream.c)
+  "${CMAKE_CURRENT_LIST_DIR}/src/dds_cdrstream.c")
 
 set(hdrs_private_cdr
-  dds_cdrstream.h)
+  "${CMAKE_CURRENT_LIST_DIR}/include/dds/cdr/dds_cdrstream.h")
 
-prepend(hdrs_private_cdr "${CMAKE_CURRENT_LIST_DIR}/include/dds/cdr/" ${hdrs_private_cdr})
-prepend(srcs_cdr "${CMAKE_CURRENT_LIST_DIR}/src/" ${srcs_cdr})
+if(${CMAKE_PROJECT_NAME} STREQUAL "CycloneDDS")
+  target_sources(ddsc PRIVATE ${srcs_cdr} ${hdrs_private_cdr})
+  target_include_directories(ddsc PRIVATE "${CMAKE_CURRENT_LIST_DIR}/include")
 
-target_sources(ddsc PRIVATE ${srcs_cdr} ${hdrs_private_cdr})
-target_include_directories(ddsc PRIVATE "${CMAKE_CURRENT_LIST_DIR}/include")
+  install(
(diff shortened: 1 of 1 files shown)


bperseghetti:
@PetervdPerk-NXP awesome thanks! @cguimaraes This is really important for those of us using zenoh-pico with CDR for the DDS plugin for Cyclone. @eboasson what's the feasibility of getting this or something exactly like it merged?

eboasson:
Awesome! Thank you so much @PetervdPerk-NXP for working on this

> what's the feasibility of getting this or something exactly like it upstream?

I have always tried to keep all of Cyclone inside a single shared object, and so having the CDR serialiser in a separate library kind of makes me unhappy. That means I'd be tempted to see if I can get it inside `libcyclonedds.so` and only generate a `libcdr.so` (or `libcycloneddscdr.so` or ...) on request.

It may be that I shouldn't be so attached to having everything in a single library anyway. Comments or thoughts on this are welcome.

We also just merged #1600, solving a bug in a way specifically meant to help with using the serialiser outside Cyclone. It really is just a matter of finding the best form.

PetervdPerk-NXP:
> I have always tried to keep all of Cyclone inside a single shared object, and so having the CDR serialiser in a separate library kind of makes me unhappy. That means I'd be tempted to see if I can get it inside libcyclonedds.so and only generate a libcdr.so (or libcycloneddscdr.so or ...) on request.'

That's perfectly fine for me, the point is that CycloneDDS could act as a dependency for another project and there could be an option to make `libcdr.so` on request for this specific use case. We don't have any intention for change the default behavior but are also looking for middle ground to please both use cases.

PetervdPerk-NXP:
I've changed it into an optional mechanism, which adds the option `BUILD_CDRLIB` so that external projects can include CycloneDDS and only build the cdr library.

For example the Zenoh-pico CDR example (https://github.com/NXPHoverGames/zenoh-demos/blob/CYCLONE_CDR_SO/ROS2/zenoh-pico-cyclonedds-cdr-message-log/CMakeLists.txt) 

I've updated it to incorporate the changes from #1600 and the changes in this proposed draft so that the application only links to libzenohpico.so and libcdr.so

```
ldd z_pub_ros2
        linux-vdso.so.1 (0x00007fdc881de000)
        libzenohpico.so => /home/peter/zenoh/zenoh-demos/ROS2/zenoh-pico-cyclonedds-cdr-message-log/build/lib/libzenohpico.so (0x00007fdc88199000)
        libcdr.so => /home/peter/zenoh/zenoh-demos/ROS2/zenoh-pico-cyclonedds-cdr-message-log/build/lib/libcdr.so (0x00007fdc88184000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdc87f3e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fdc881df000)
```

It's not a full clean approach i.e. we need to have `idlc` pre-compiled to generate the msgs and all headers from CycloneDDS are required to compile the messages and CDR. But I think we can improve iteratively on this.

PetervdPerk-NXP:
I'm no CMake expert either, so I wasn't aware of the `project()` trick you did.

I've update my PR to reflect on this. It's pretty much what I'm looking for now. Any external project can add CycloneDDS as a git submodule or use the `FetchContent_Declare` macro. And then just include the `cyclonedds/src/core/cdr/CMakeLists.txt` to add CDR serialization to their application.

PetervdPerk-NXP:
@eboasson 
I'm using this solution for a couple of weeks now.
And it seems to be working as intended.

Would you be able to take a look at these changes?

eboasson (approved):
Hi @PetervdPerk-NXP, so sorry I overlooked your comment from a month ago. Fortunately the update from a few days ago bumped it in my inbox. Yes, I am totally happy with this, thanks!

PR #1604, opened 2023-03-13
Retry relwr_unrelrd_network test on failure

eboasson:
The test is flakey by design, relying on a very specific sequence of received and lost messages with unreliable communication, and it turns out to fail more often than hoped for.

This changes the test to retry it its entirety a limited number of times. The alternative would be to write it in C. Let's see if we can avoid that.


dpotman (approved):
LGTM

PR #1613, opened 2023-03-16
Enable idlc_generate() to support "dot-separated" idl files	(#1612)

### LINKED ISSUE #1612: [IDLC] idlc_generate() failed generate proper files for idl containing "." e.g. "org.eclipse.HellloWorld.idl"
Is there any special consideration for this?

The current idlc_generate() in .cmake do not support idl files like “module1.module2.module3.MyStuff.idl”, which cause some trouble in a large project where all the idl files are name like that.

I do have a solution, just wanna to confirm if I can/should create a pull request to resolve this issue.

tez1szh:
Fix #1612

### DIFF
--- a/cmake/Modules/Generate.cmake
+++ b/cmake/Modules/Generate.cmake
@@ -156,7 +156,7 @@ function(IDLC_GENERATE_GENERIC)
   endif()
   set(_outputs "")
   foreach(_file ${_files})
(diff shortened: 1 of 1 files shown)


eboasson (approved):
Nice fix! Much simpler than I feared when I saw #1612 😀

PR #1622, opened 2023-03-21
Refactor xevent

eboasson:
This finally (almost 12 years after the introduction of a generic callback 😱) removes the various special event types from the timed event handling mechanism (xevents) and uses callbacks for everything. Doing so significantly reduces the set of dependencies of xevent on other components of the system and generally allows shuffling stuff around a bit. So that, too, is in this PR.

There are some other changes that kind-of happened, but that really need a good look. Those are in ec5ed364057b5c5ce7f7864143a4809a15f6923a


dpotman (commented):
Looks good, a few comments on some minor issues. I cannot really assess what the impact is of the new approach of rounding (or actually not rounding) the schedule time and the change in reading the clock while processing timed events. Shouldn't we make the latter configurable, to avoid the issue (in certain specific scenarios) that was the reason for adding this extra update of `tnow`?

dpotman (approved):
LGTM!

PR #1626, opened 2023-03-23
Const serialized meta-data in topic descriptor

dpotman:
This makes the serialized type information and type mapping `const` in the topic descriptor


eboasson (approved):
Thanks @dpotman!

PR #1639, opened 2023-04-06
Fix typebuilder max key size for unbounded strings

### LINKED ISSUE #1637: Command (dynsub "DDSPerfCPUStats")  is running incorrectly
Hi ,
I want to try using example/dynsub to subscribe to the topic "DDSPerfCPUStats" in ddsperf. Then dynsub will crash.
image
I type the command (ddsperf pub) on one terminal, then I type (dynsub DDSPerfCPUStats) on the other terminal.

Debugging in master branch :(ddsperf.c:2282 CPUStats_desc.m_flagset) != (dynsub.c:397 descriptor->m_flagset) .  

Is it caused dynsub crash? How to solve it ? 

thanks

eboasson:
An unbounded string key field was treated as sizeof(char*) in computing the maximum serialized key size (in dynamically constructed serializer operations for the C binding), rather than as infinity.  If this computed maximum is no more than some fixed size, then (for the C binding):

* Samples rely on a pre-allocated buffer for storing the serialized key, and handling a key for which the serialised form is actually longer will result in trying to reallocate a non-malloc'd buffer;

* The decisions whether the DDSI keyhash is the serialised key or the MD5 of it will be taken incorrectly.  (Cyclone doesn't really use them, but needs to support them regardless.)

Fixes #1637

### DIFF
--- a/src/core/ddsc/tests/TypeBuilderTypes.idl
+++ b/src/core/ddsc/tests/TypeBuilderTypes.idl
@@ -401,4 +401,8 @@ module TypeBuilderTypes {
     t27_1 f1;
   };
 
+  /* unbounded string key */
+  struct t28 {
+    @key string f1;
+  };
 };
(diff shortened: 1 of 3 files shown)


dpotman (approved):
Looks good! The IDL compiler handles this correctly (src/tools/idlc/src/descriptor.c:2053) and doesn't need this fix.

PR #1641, opened 2023-04-06
Avoid passing a null pointer to ddsrt_memdup

eboasson:
It is currently handled correctly, but the interface specification states the argument must be non-null (like all the standard C library functions of similar character), and the undefined behaviour sanitizer flags this.


dpotman (approved):
LGTM

PR #1642, opened 2023-04-06
Doc updates

ThijsSassen:
and cleanups


docs/manual/ddsc.rst
```diff
@@ -9,7 +9,11 @@
 
    SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 
-C API Reference
+.. index::
+   single: C API reference
+   single: API; C reference
```
poetinger, line 14: Should we also say `C API reference` like above?

poetinger (approved):

PR #1643, opened 2023-04-06
add doxygen comments to hopscotch.h

mvandenhoek:
Aside from the comments, I also have a bit of feedback regarding the user friendliness of the API:

- I don't like the fact that you need to make a dummy object to do the lookup. Even if you made a wrapper around it to improve the interface, you still unnecessarily create the whole object, which could be large. Therefore I also included a workaround that lets you use just the key.
- When you remove an element, you still need to free the memory of the element yourself, so you need the pointer. Thus it would be convenient if the remove function returned the pointer to the removed element. It also wouldn't cost anything to do so since the function internally already does a lookup.

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/hopscotch.h
+++ b/src/ddsrt/include/dds/ddsrt/hopscotch.h
@@ -251,8 +251,11 @@ void ddsrt_chh_free (struct ddsrt_chh * __restrict hh);
 /**
  * @brief Concurrent version of @ref ddsrt_hh_lookup
  * 
- * The lookup is lock-free and wait-free, but can fail (and return NULL) despite the element existing,
- * if another thread was busy with it for too long. It can work in parallel to @ref ddsrt_chh_remove,
+ * The lookup is lock-free and wait-free, and an element that exists in the hash table is guaranteed to be found.
+ * Note that wait-free doesn't mean the operation is unimpeded by actions of other threads,
+ * but rather it means that the operation will complete within a bounded number of steps.
+ * 
+ * It can work in parallel to @ref ddsrt_chh_remove,
  * but a successful lookup doesn't guarantee that the element hasn't been removed in the mean time.
  * 
  * @param[in] rt the hash table
@@ -273,6 +276,9 @@ int ddsrt_chh_add (struct ddsrt_chh * __restrict rt, void * __restrict data);
 
 /**
  * @brief Concurrent version of @ref ddsrt_hh_remove
+ * 
+ * It can work in parallel to @ref ddsrt_chh_lookup, but in that case the looked up element may not be
+ * part of the hash table anymore by the time the caller of @ref ddsrt_chh_lookup accesses the element.
  * 
  * @param[in,out] rt the hash table
  * @param[in] keyobject user data to remove
--- a/src/ddsrt/include/dds/ddsrt/hopscotch.h
+++ b/src/ddsrt/include/dds/ddsrt/hopscotch.h
@@ -19,7 +19,22 @@
 extern "C" {
 #endif
 
-/*
+/** \file hopscotch.h
+ * The hopscotch hash table is implemented in three versions:
+ * - Standard version
+ * - Concurrent version (for shared use by multiple threads)
+ * - Embedded data version (stores a copy of the data)
+ * 
+ * With exception of the embedded data version, the hash table stores pointers to user data,
+ * but doesn't manage the memory of the data itself.
+ * 
+ * Note that elements in a hash table are unordered. If an ordered set is desired,
+ * then a different data structure such as a tree should be used.
+ */
+
+/**
+ * @brief User defined hash function.
+ * 
  * The hopscotch hash table is dependent on a proper functioning hash.
  * If the hash function generates a lot of hash collisions, then it will
  * not be able to handle that by design.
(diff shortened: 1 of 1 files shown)


src/ddsrt/include/dds/ddsrt/hopscotch.h
```diff
@@ -32,42 +47,172 @@ extern "C" {
...
+ * @param[in] rt the hash table
+ * @param[in] keyobject is the object with which to do the lookup
+ * @return pointer to the matching element, NULL if failed
+ */
 DDS_EXPORT void *ddsrt_hh_lookup (const struct ddsrt_hh * __restrict rt, const void * __restrict keyobject) ddsrt_nonnull_all;
+
+/**
+ * @brief Add an element to the hash table.
+ * 
+ * It is not allowed to add duplicates to the hash table. Trying to do so results in failure.
```
eboasson, line 125: I would phrase this differently, because this phrasing leaves it unclear whether you're allowed to try (you are).

Something like: attempts to store an element with a key that is already present will be rejected.

And then the return value becomes something like: 1 is success, element stored; 0 is failure, element key already present.

src/ddsrt/include/dds/ddsrt/hopscotch.h
```diff
@@ -32,42 +47,172 @@ extern "C" {
...
+ * 
+ * The walk function 'f' receives in parameter 'a' the pointer to user data, and has 'f_arg' as an optional extra argument.
+ * It is allowed to modify the user data. The @ref ddsrt_hh_enum is useful to free user data prior to calling @ref ddsrt_hh_free.
+ * 
+ * @param[in,out] rt the hash table
+ * @param[in] f user defined walk function to apply to each element
+ * @param[in] f_arg extra argument for walk function
+ * @see @ref ddsrt_hh_iter_next
+ */
 DDS_EXPORT void ddsrt_hh_enum (struct ddsrt_hh * __restrict rt, void (*f) (void *a, void *f_arg), void *f_arg) ddsrt_nonnull ((1, 2)); /* may delete a */
```
eboasson, line 181: The "dangling" comment "may delete a" should be moved into the doxygen comment, I'd say. It should also be made clear that it is `f` that is allowed to delete a. `enum` by itself doesn't modify anything.

src/ddsrt/include/dds/ddsrt/hopscotch.h
```diff
@@ -32,42 +47,172 @@ extern "C" {
...
+
+/**
+ * @brief Initialize the iterator and get the first element.
+ * 
+ * @param[in] rt the hash table
+ * @param[out] iter iterator object
+ * @return pointer to first element
+ * @see @ref ddsrt_hh_iter_next
+ */
 DDS_EXPORT void *ddsrt_hh_iter_first (struct ddsrt_hh * __restrict rt, struct ddsrt_hh_iter * __restrict iter) ddsrt_nonnull_all; /* may delete nodes */
```
eboasson, line 191: "May delete nodes" should be deleted, `iter_first` initializes an iterator object, it doesn't modify the table.

src/ddsrt/include/dds/ddsrt/hopscotch.h
```diff
@@ -81,29 +226,194 @@ struct ddsrt_chh_iter {
...
+/**
+ * @brief Concurrent version of @ref ddsrt_hh_free
+ * 
+ * @param[in,out] hh the hash table to destroy
+ * @see @ref ddsrt_chh_new
+ */
 void ddsrt_chh_free (struct ddsrt_chh * __restrict hh);
+
+/**
+ * @brief Concurrent version of @ref ddsrt_hh_lookup
```
eboasson, line 251: I think it would be worth pointing out that a lookup is lock- and wait-free and can occur in parallel to a remove operation, and that a successful lookup doesn't mean the returned element is still part of the hash table by the time control has been returned to the caller.

src/ddsrt/include/dds/ddsrt/hopscotch.h
```diff
@@ -81,29 +226,194 @@ struct ddsrt_chh_iter {
...
+ * @param[in] keyobject user data to remove
+ * @return int meant to be interpreted as bool (1 is success, 0 is failure)
+ * @see @ref ddsrt_chh_add
+ */
 int ddsrt_chh_remove (struct ddsrt_chh * __restrict rt, const void * __restrict keyobject);
+
+/**
+ * @brief Concurrent version of @ref ddsrt_hh_enum
+ * 
+ * Called unsafe because it doesn't use a mutex. The user must ensure no other thread is accessing the hash table in the mean time.
```
eboasson, line 282: It is not that another thread may not access the hash table. Concurrent lookups are self-evidently safe (because they don't alter the table), but adds and removes may also be performed while enumerating.

What is "unsafe" about it is that:
* if another thread is removing an element, then there is no guarantee that the element for which `f` is invoked is still in the table at the time of invocation (similar to a concurrent `remove` and `lookup`)
* if another thread is adding an element, there is no guarantee whether that new element will, or will not, be visited
* if another thread is adding an element, there is no guarantee that existing elements will be visited
* if another thread is adding an element, there is no guarantee that an element won't be visited multiple times

The first two are quite ordinary uncertainties. The latter two are nasty enough to make it deserve to be called "unsafe". I think it is better to be precise about this in the documentation.

mvandenhoek: > It is not that another thread may not access the hash table. Concurrent lookups are self-evidently safe (because they don't alter the table), 

Wouldn't it still be problematic if say:
- Thr A calls ddsrt_chh_enum_unsafe with a free funtion.
- Whilst A's call is ongoing, Thr B does a lookup of element x.
- Thr A has freed element x.
- Thr B tries to read and modify data value of element x.

src/ddsrt/include/dds/ddsrt/hopscotch.h
mvandenhoek: As a further feedback I'd say you might want to provide a return value indicating why it failed, since there are two options: 

1. It failed because the element doesn't exist
2. It failed because another thread was busy with it for too long.

I think it would be wrong for the program to assume 1 upon failure when it actually failed due to 2, since in the latter case it might decide to retry at some later time and still get it.

src/ddsrt/include/dds/ddsrt/hopscotch.h
eboasson: Well, yes, but if the caller frees memory that it still uses, it's the callers fault 😀 It is true that this really does place a burden on the callers, see, e.g., https://github.com/eclipse-cyclonedds/cyclonedds/blob/1aeea0faea21ea14d587d46c02736c74d520440f/src/core/ddsi/src/ddsi_tkmap.c#L157 and https://github.com/eclipse-cyclonedds/cyclonedds/blob/1aeea0faea21ea14d587d46c02736c74d520440f/src/core/ddsi/src/ddsi_tkmap.c#L213

eboasson:
I left some responses, hopefully those will allow you to update the documentation comments. When that's done, I think I'll be happy to merge it.

One request for the future: please don't just force-push when updating a PR based on a review. Sometimes it makes the most sense, but often it makes re-reviewing harder.

eboasson (commented):
LGTM, but I think I'll let #1662 go first. That means this PR will probably need some minor changes to avoid breaking MISRA 3.1 again.

mvandenhoek:
The branch was rebased on master

eboasson (approved):
Thanks @mvandenhoek !

PR #1651, opened 2023-04-13
Fix re-using existing types when registering a dynamic type

dpotman:
In case a dynamic type is registered and it exists in the type library, the existing type is used. But in the scenario that the existing type was (partially) unresolved, the meta-data from the dynamic type should be added to the existing type. This fixes this problem and adds a test-case.


eboasson (approved):
LGTM!

PR #1653, opened 2023-04-13
Fix a warning when building in release mode.

clalancette:
When building the 0.10.x branch in release mode, we are seeing the following warning:

In function ‘dds_serdata_builtin_from_topic_definition’,
    inlined from ‘dds__builtin_make_sample_topic_impl’ at ros2_rolling/src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_builtin.c:302:34:
ros2_rolling/src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_serdata_builtintopic.c:468:3: warning: writing 16 bytes into a region of size 0 [-Wstringop-overflow=]
  468 |   memcpy (&d->common.key.raw, key, sizeof (d->common.key.raw));

I believe what is happening here is that the compiler is finding a path through the code where serdata_builtin_new() can allocate and return a struct of 0 size.  That can theoretically happen if the tp->entity_kind is not one of the known types when serdata_builtin_new() is called. Given that this is an exceptional situation, detect this and abort().  Note that we use abort() and not assert() here as the latter will be compiled out in release builds.

While we are in here, change the syntax of the memcpy command as well.  While the previous version wasn't wrong, this one makes it much more clear that we are copying the key bytes from one structure to another.  If this part is objectionable, it can be dropped.

To be totally transparent, I do *not* see this issue with the `master` branch, even though the code here is largely the same between the `0.10.x` branch and the `master` branch.  I don't really understand how that can be, but I'd suggest porting this to the `master` branch anyway.

### DIFF
--- a/src/core/ddsc/src/dds_serdata_builtintopic.c
+++ b/src/core/ddsc/src/dds_serdata_builtintopic.c
@@ -102,6 +102,8 @@ static struct ddsi_serdata_builtintopic *serdata_builtin_new(const struct ddsi_s
     case DSBT_WRITER:
       size = sizeof (struct ddsi_serdata_builtintopic_endpoint);
       break;
+    default:
+      abort();
   }
   struct ddsi_serdata_builtintopic *d = ddsrt_malloc(size);
   ddsi_serdata_init (&d->c, &tp->c, serdata_kind);
@@ -465,7 +467,7 @@ struct ddsi_serdata *dds_serdata_builtin_from_topic_definition (const struct dds
   const struct ddsi_sertype_builtintopic *tp = (const struct ddsi_sertype_builtintopic *) tpcmn;
   assert (tp->entity_kind == DSBT_TOPIC);
   struct ddsi_serdata_builtintopic_topic *d = (struct ddsi_serdata_builtintopic_topic *) serdata_builtin_new (tp, kind);
-  memcpy (&d->common.key.raw, key, sizeof (d->common.key.raw));
+  memcpy (d->common.key.raw, key->d, sizeof (d->common.key.raw));
   if (tpd != NULL && kind == SDK_DATA)
     from_qos (&d->common, tpd->xqos);
   return fix_serdata_builtin (&d->common, DSBT_TOPIC, tp->c.serdata_basehash);
--- a/src/core/ddsc/src/dds_serdata_builtintopic.c
+++ b/src/core/ddsc/src/dds_serdata_builtintopic.c
@@ -83,29 +83,29 @@ static void serdata_builtin_free(struct ddsi_serdata *dcmn)
   ddsrt_free (d);
 }
 
-static struct ddsi_serdata_builtintopic *serdata_builtin_new(const struct ddsi_sertype_builtintopic *tp, enum ddsi_serdata_kind serdata_kind)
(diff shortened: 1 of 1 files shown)


eboasson:
It's definitely a false positive.

In `serdata_builtin_new` the `&d->c` in the call to `ddsi_serdata_init` means offsetting a pointer, and offsetting a null pointer is undefined behaviour, therefore `d` can't be null, therefore `size` cannot have been 0, etc. Of course there is the trouble of being able to do arithmetic on `enum`s , but then you check all the paths and it can't be any other value than the ones accounted for in the switch ...

It is also quite incorrect of it to claim `&d->common.key.raw` points to a region of size 0 where the sizeof of the type is simply 16 bytes.

That said, I have a simple policy of not accepting any compiler warnings, even if sometimes it means changing the code a bit to accommodate a compiler's idiosyncrasies, with "disable warning for this line" as an option of last resort only.

I fully agree with the `memcpy` change, but not with the rewrite of the `switch`. Aborting is fine with me, but adding a `default` clause is not because that suppresses a compiler warning that not all cases are covered if I ever add a symbol to the enum and forget this switch statement.

Would you be ok with something like:
```C
static size_t serdata_builtin_sizeof(enum ddsi_serdata_kind serdata_kind)
{
  switch (tp->entity_kind)
  {
    case DSBT_PARTICIPANT:
      return sizeof (struct ddsi_serdata_builtintopic_participant);
    case DSBT_TOPIC:
#ifdef DDS_HAS_TOPIC_DISCOVERY
      return sizeof (struct ddsi_serdata_builtintopic_topic);
#else
      break;
#endif
    case DSBT_READER:
    case DSBT_WRITER:
      return sizeof (struct ddsi_serdata_builtintopic_endpoint);
  }
  abort ();
  return 0;
}
```
and calling it?

(Or any of many other variants on the theme. I do always struggle with the abort/return bit, and it seems to me that `__builtin_unreachable` is the proper solution but I am not sure when that was introduced, nor how well supported that is in other compilers.)

I have to admit that your proposal violates fewer Misra rules then mine, but both are much better than any Misra compliant one 😄

eboasson:
Something else: the "eclipsefdn/eca" check fails but I'm pretty sure you used to have the Eclipse Contributor Agreement sorted. Maybe the email address changed and you forgot to update the ECA?

clalancette:
> Aborting is fine with me, but adding a `default` clause is not because that suppresses a compiler warning that not all cases are covered if I ever add a symbol to the enum and forget this switch statement.

That's reasonable.

> Would you be ok with something like:

> `static size_t serdata_builtin_sizeof(enum ddsi_serdata_kind serdata_kind)`

Yeah, that works for me.  I'll give it a shot here and update this PR assuming it works.  Thanks!

> Something else: the "eclipsefdn/eca" check fails but I'm pretty sure you used to have the Eclipse Contributor Agreement sorted. Maybe the email address changed and you forgot to update the ECA?

I've been using my personal address, but my CLA is with my clalancette@openrobotics.org one.  I'll rebase this one so that the author is my clalancette@openrobotics.org one.

clalancette:
And I've now implemented it in 3aca321 .  Please take a look when you have time @eboasson .

eboasson (approved):
LGTM, thanks @clalancette !

(The CI test failure is unrelated.)

clalancette:
> LGTM, thanks @clalancette !

Thank you!

PR #1654, opened 2023-04-14
atomics.h: add doxygen comments

mvandenhoek:


src/ddsrt/include/dds/ddsrt/atomics.h
```diff
@@ -19,6 +19,20 @@
 #include "dds/ddsrt/endian.h"
 #include "dds/ddsrt/types.h"
 
+/**
+ * @file atomics.h
+ * @brief Contains functions and types for atomic operations on various types
+ *
+ * This header file declares functions and types for atomic operations on uint32_t,
```
eboasson, line 26: This is true (I would add `void *` as well) but then the comments only deal with the 64-bit integer types. The problem of course is that the others are all defined in platform-specific files, e.g., https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/ddsrt/include/dds/ddsrt/atomics/gcc.h

Perhaps you were just making your way there. Marking the PR as a draft in such a case would be useful.

mvandenhoek:
Because it's very similar between platforms, this turned out to be a lot of copy paste work, however I am not sure what would be a proper alternative. I know there is a copydoc command, but there isn't really one master version from which to copy. It may be a bit odd for example if the msvc copied from sun.

PR #1657, opened 2023-04-17
Fix crash on getting key from Iceoryx chunk

eboasson:
dds_serdata_default (so the default sample implementation used only by the C binding) always stores the serialized key in native XCDR2 format, but this was broken for Iceoryx chunks.

For the case of a local write, the code assumed that the iox chunk contains a "raw sample" rather than a serialized sample.  This was correct until https://github.com/eclipse-cyclonedds/cyclonedds/commit/8845a81ce77f78df2816a74b01cf89686f96581d (pr https://github.com/eclipse-cyclonedds/cyclonedds/pull/1031) changed that.

The typical failure mode (but not the only one) is a crash if the key contains a string and if the data is volatile and all subscribers are addressed via Iceoryx.  In practice this turns out to have been rare enough that no-one noticed it until now.

For the case of data coming in over Iceoryx, there was a confusion between the DDSI keyhash and the key representation in serdata_default (the former is the big-endian serialisation of the key or the MD5 hash thereof; the latter is always native-endianness XCDR2 serialisation of the key and so resistant to collision attacks).

The typical failure mode here is one of instance confusion: either mapping to different key values to the same instance, or mapping what really is the same key value to different instances.

With this PR, it now always extracts the key from the actual data in the Iceoryx chunk.  In the case of a local write, we know we have the sample in memory somwhere, even if the chunk contains the serialized representation.  Nonetheless, here it extracts the key from the serialized form because at the point where we need to do it, we do not have a pointer to the original sample and changing the interface for this rare case is not worth it at this point in time.

Another consequence of the problem is that the instance handle can be wrong. That is addressed here, too.

The PR also extends the test cases in separate commits.

This should address https://github.com/ros2/rmw_cyclonedds/issues/449 . Needs backporting to 0.9 and 0.10.

### DIFF
--- a/src/core/ddsc/src/dds_serdata_default.c
+++ b/src/core/ddsc/src/dds_serdata_default.c
@@ -467,54 +467,38 @@ static struct ddsi_serdata *serdata_default_from_keyhash_cdr_nokey (const struct
 }
 
 #ifdef DDS_HAS_SHM
-static struct ddsi_serdata* serdata_default_from_received_iox_buffer(const struct ddsi_sertype* tpcmn, enum ddsi_serdata_kind kind, void* sub, void* iox_buffer)
+static struct dds_serdata_default *serdata_default_from_iox_common (const struct ddsi_sertype *tpcmn, enum ddsi_serdata_kind kind, void *sub, void *iox_buffer)
 {
-  const iceoryx_header_t* ice_hdr = iceoryx_header_from_chunk(iox_buffer);
-
-  const struct dds_sertype_default* tp = (const struct dds_sertype_default*)tpcmn;
-
-  struct dds_serdata_default *d = serdata_default_new_size (tp, kind, ice_hdr->data_size, tp->write_encoding_version);
-
+  struct dds_sertype_default const * const tp = (const struct dds_sertype_default *) tpcmn;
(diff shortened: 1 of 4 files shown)


dpotman (approved):
Looks good! Nit-picking: you could avoid some code duplication in `serdata_default_from_iox` and `serdata_default_from_iox_nokey`.

eboasson:
> Looks good! Nit-picking: you could avoid some code duplication in `serdata_default_from_iox` and `serdata_default_from_iox_nokey`.

I do agree, I've reworked it a bit. If you're happy with the amended version, I'll squash the `serdata_default` changes and merge the PR.

dpotman (approved):

bperseghetti:
@eboasson what do you expect the timeline is for back-porting into 0.10 and 0.9?

eboasson:
The PRs are ready (#1658 and #1659), I am really just waiting for someone to review them.

eboasson:
@bperseghetti I just merged those two PRs, so the fix is now available in 0.9, 0.10 and master 🙂

bperseghetti:
> @bperseghetti I just merged those two PRs, so the fix is now available in 0.9, 0.10 and master 🙂

That's great, thanks!

PR #1660, opened 2023-04-19
Add a "writer batching" QoS

eboasson:
This adds a QoS setting to control whether or not a writer does batching without any (possible) dependency on asynchronous publications. Batching means small samples get combined in a single RTPS message, improving performance.

This message only gets sent automatically when it is full, and if it is important that the preceding samples are actually sent, then the application code must explicitly flush the batch by call `dds_flush`.


dpotman (approved):
Looks good. Maybe we could move the documentation for the deprecated `WriteBatch` option to the doxygen comments of the qset/qget functions in dds.h?

eboasson:
> Looks good. Maybe we could move the documentation for the deprecated `WriteBatch` option to the doxygen comments of the qset/qget functions in dds.h?

Good point. How about this?

dpotman (approved):

PR #1662, opened 2023-04-20
Address MISRA rule 3.1 violations across Cyclone

NoxPardalis:
This resolves the issue of `//` within `/*...*/` comments. These typically feature within the URLs present in the copyright blurbs and are allowed within comments starting with a `//` but not the `/*...*/` style comments. This is applied across all `.c`, `.h`, and `.idl` files and also results in the following files being regenerated:

- docs/manual/config/config_file_reference.rst
- docs/manual/options.md
- etc/cyclonedds.rnc
- etc/cyclonedds.xsd
- src/core/ddsi/defconfig.c


NoxPardalis:
I finally got an efficient workflow for checking the MISRA violations locally. Pushed some additional commits addressing a couple more violations of Rule 3.1. This has now addressed everything according to my local testing and so I've changed the title of the PR accordingly.

src/core/ddsi/src/ddsi_discovery_spdp.c
```diff
@@ -182,7 +181,7 @@ void ddsi_get_participant_builtin_topic_data (const struct ddsi_participant *pp,
     if (ddsrt_gethostname(node, sizeof(node)-1) < 0)
 #endif
       (void) ddsrt_strlcpy (node, "unknown", sizeof (node));
-    size = strlen(node) + strlen(DDS_VERSION) + strlen(DDS_HOST_NAME) + strlen(DDS_TARGET_NAME) + 4; /* + ///'\0' */
+    size = strlen(node) + strlen(DDS_VERSION) + strlen(DDS_HOST_NAME) + strlen(DDS_TARGET_NAME) + 4; // + '\0';
```
eboasson, line 184: That should now read `// 3x'/' + '\0'` or something similar

src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -2624,7 +2622,8 @@ bool ddsi_xt_is_assignable_from (struct ddsi_domaingv *gv, const struct xt_type
     {
       case DDS_XTypes_TK_BITMASK:
         return bb == t_other->_u.bitmask.bit_bound;
-      // case TK_UINT8:   /* FIXME: TK_UINT8 not defined in idl */
+      // FIXME: TK_UINT8 not defined in idl
+      // case TK_UINT8:   
       //   return bb >= 1 && bb <= 8;
```
eboasson, line 2627: @dpotman shouldn't this one be

```suggestion
      case DDS_XTypes_TK_UINT16:   
         return bb >= 1 && bb <= 8;
```

now?

eboasson (commented):
Thanks @NoxPardalis, we indeed have to do try to adhere to MISRA (at least within reason and taking into account their section 5 ("don't completely rewrite code with a proven track record"), and this is a sensible step.

This particular rule is probably going to crop up a few more times unless we integrate a checker in the CI, at least for the rules where it the code should no longer contain any violations. Based on your comment I'm thinking you may know of a good way to do that. I'd be interested to learn a bit more.

Other than that, I did find two things, one really unimportant and the other a real bug that I spotted through sheer luck. (Or perhaps because I made a perl script to de-noise the diff 🤣)

src/core/ddsi/src/ddsi_typewrap.c
dpotman: Yes, you're right (almost right, UINT8 instead of UINT16 :wink:). Apparently I missed that one when adding the type kind for UINT8, thanks for catching that!

eboasson: I'm not really good at copy-paste it seems ...

It should be separate commit (ideally a separate PR) anyway.

src/core/ddsi/src/ddsi_discovery_spdp.c
NoxPardalis: Ah, now it makes sense. I wasn't exactly sure how to grok the original comment. Resolved in 92fe0c90a8b2571f8ad95aa7b4d2332dc4b8daa7.

NoxPardalis:
> This particular rule is probably going to crop up a few more times unless we integrate a checker in the CI, at least for the rules where it the code should no longer contain any violations. Based on your comment I'm thinking you may know of a good way to do that. I'd be interested to learn a bit more.

@eboasson sounds sensible. I did some preliminary tests hooking up the MISRA checks I'm using via GitHub actions. I'll coordinate with you to see if we can integrate the checks in the mainline.

eboasson (approved):

PR #1667, opened 2023-04-21
circlist.h: add doxygen comments

mvandenhoek:
Aside from the comments, I noticed a missing closing bracket for `#if defined(__cplusplus)`, so I added that as well.

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/circlist.h
+++ b/src/ddsrt/include/dds/ddsrt/circlist.h
@@ -36,23 +36,21 @@ struct ddsrt_circlist {
   struct ddsrt_circlist_elem *latest; /**< pointer to latest inserted element */
 };
 
-/**
- * @brief The circlist element is the basic element of circular doubly linked list structure.
- * 
- * To store user data in the list, the element must be embedded in a user element, which is a struct containing the user data.
- * All API calls with an element as input or output, deal with the circlist element. Example:
- * @code{.c}
- * typedef struct num_s{ // user element
- *   uint64_t val; // user data
- *   ddsrt_circlist_elem elem;
- * }num_t;
- * @endcode
- * 
- * For functions like @ref ddsrt_circlist_latest, this means that to get a pointer to the user element, you can use the macro @ref DDSRT_FROM_CIRCLIST :
- * @code{.c}
- * num_t* num = DDSRT_FROM_CIRCLIST(num_t, elem, ddsrt_circlist_latest(listptr));
- * @endcode
- */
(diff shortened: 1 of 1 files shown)


src/ddsrt/include/dds/ddsrt/circlist.h
```diff
@@ -22,22 +26,102 @@
 extern "C" {
 #endif
 
+/**
+ * @brief Convenience macro to get the pointer to the user node, see @ref ddsrt_circlist_elem for an example.
```
eboasson, line 30:
```suggestion
 * @brief Macro to get the pointer to the user node, see @ref ddsrt_circlist_elem for an example.
```

eboasson (commented):
It is much better to require using `DDSRT_FROM_CIRCLIST` then to call it a "convenience": writing out pointer arithmetic like this without a good reason in application code is generally a bad idea and it moreover reduces the implementation freedom for the library.

src/ddsrt/include/dds/ddsrt/circlist.h
```diff
@@ -22,22 +26,95 @@
...
+  struct ddsrt_circlist_elem *latest; /**< pointer to latest inserted element */
 };
 
+/**
+ * @brief The circlist element is the basic element of circular doubly linked list structure.
+ * 
+ * To store user data in the list, the element must be embedded in a user element, which is a struct containing the user data.
+ * All API calls with an element as input or output, deal with the circlist element. Example:
+ * @code{.c}
+ * typedef struct num_s{ // user element
```
eboasson, line 45: This a violation of MISRA rule 3.1 that @NoxPardalis just fixed elsewhere. For the comparable cases, he changed the documentation comment to use `///` instead of `/**` ... `*/`. I think that would be a practical solution here, too.

src/ddsrt/include/dds/ddsrt/circlist.h
```diff
@@ -22,22 +26,95 @@
...
+ * @return true if empty, false if not empty
+ */
 bool ddsrt_circlist_isempty (const struct ddsrt_circlist *list);
+
+/**
+ * @brief append an element to the list
+ * 
+ * @param[in,out] list the circular list
+ * @param[in] elem the element to append
+ * @see @ref ddsrt_circlist_remove
```
eboasson, line 81: That should be `See @ref` instead of `@see @ref` (Clang taught me this)

eboasson (commented):
Almost there 🙂

eboasson (approved):
Thanks @mvandenhoek !

PR #1672, opened 2023-04-25
Allow dds_write_flush on publisher, participant

eboasson:
This PR adds support for invoking `dds_write_flush` on a publisher or participant with the obvious meaning of flushing the buffers for all writers contained in that entity. The technique used is the same as that for a variety of other operations that need to act on all descendants (like `set_qos`, `set_listener`, &c.)

The test is a bit fragile in the sense that it is quite tightly bound to the behaviour of the stack, but I don't see a good way around that. After all, the batching doesn't guarantee that it never goes out, it merely stops guaranteeing that it will go out before one calls `dds_write_flush`.

### DIFF
--- a/src/core/ddsc/src/dds_entity.c
+++ b/src/core/ddsc/src/dds_entity.c
@@ -1700,3 +1700,71 @@ dds_return_t dds_get_entity_sertype (dds_entity_t entity, const struct ddsi_sert
   dds_entity_unpin (e);
   return ret;
 }
+
+static dds_return_t pushdown_write_flush (dds_entity *e)
+{
+  /* Note: e is claimed, no mutexes held */
+  dds_return_t rc = DDS_RETCODE_OK;
+  struct dds_entity *c;
+  dds_instance_handle_t last_iid = 0;
+  ddsrt_mutex_lock (&e->m_mutex);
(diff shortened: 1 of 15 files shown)


src/core/ddsc/src/dds_entity.c
```diff
@@ -1700,3 +1700,71 @@ dds_return_t dds_get_entity_sertype (dds_entity_t entity, const struct ddsi_sert
...
+        case DDS_KIND_WRITER: {
+          dds_write_flush_impl ((dds_writer *) c);
+          break;
+        }
+        case DDS_KIND_PUBLISHER:
+        case DDS_KIND_PARTICIPANT:
+        case DDS_KIND_DOMAIN: {
+          dds_return_t rc1;
+          if ((rc1 = pushdown_write_flush (c)) < 0 && rc1 == 0)
+            rc = rc1;
```
dpotman, line 1730: Looks like `pushdown_write_flush` can only return `DDS_RETCODE_OK`, so the return code checking (that should be `rc == 0` I think?) can be removed and return type can be `void`?

dpotman (commented):
Looks good, just a few details in the comments below.

eboasson:
Thanks @dpotman! I updated the PR accordingly.

I did force-push it, but the changes I made are in the last 3 commits, so that should not pose a problem. I hope you don't mind.

dpotman (approved):

PR #1680, opened 2023-05-02
fibheap.h: add doxygen comments

mvandenhoek:
Aside from the comments, I also have an idea for potential improvement: Since extract_min has a bad worst case performance due to rebalancing, it might be useful to extract take the rebalancing part of the algorithm in extract_min and make that a separate function, which is then also exposed in the API. The user can then "pay off the debt" by rebalancing when they can afford the slower execution, as opposed to a forced expensive rebalancing later when it may be inconvenient.

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/fibheap.h
+++ b/src/ddsrt/include/dds/ddsrt/fibheap.h
@@ -11,6 +11,21 @@
 #ifndef DDSRT_FIBHEAP_H
 #define DDSRT_FIBHEAP_H
 
+/** @file fibheap.h
+  The main purpose of a fibonacci heap is for use as a priority queue i.e. repeatedly extracting the minimum of the set.
+  Certain operations i.e. insert, merge, decrease_key are very efficient in that they have O(1) time complexity,
+  but they they degrade the shape of the heap, which requires rebalancing to fix.
+  The necessary rebalancing is delayed until the next call to extract_min.
+
+  For example, after lots of calls to insert and/or decrease_key, the next call to extract_min is very expensive
+  since it has to rebalance, and this could take as much as O(N) in the worst case. This is where you pay off the debt
+  for procrastinating on rebalancing. Once balanced, following calls to extract_min will be O(log(N)).
+
+  It is worth mentioning that the avl tree is a viable alternative as it can support the same use cases with its
+  insert, find_min, and delete operations. Whilst the avl tree is not necessarily more efficient as a priority queue overall,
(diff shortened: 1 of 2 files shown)


src/ddsrt/include/dds/ddsrt/fibheap.h
```diff
@@ -11,6 +11,21 @@
...
+  The main purpose of a fibonacci heap is for use as a priority queue i.e. repeatedly extracting the minimum of the set.
+  Certain operations i.e. insert, merge, decrease_key are very efficient in that they have O(1) time complexity,
+  but they they degrade the shape of the heap, which requires rebalancing to fix.
+  The necessary rebalancing is delayed until the next call to extract_min.
+
+  For example, after lots of calls to insert and/or decrease_key, the next call to extract_min is very expensive
+  since it has to rebalance, and this could take as much as O(N) in the worst case. This is where you pay off the debt
+  for procrastinating on rebalancing. Once balanced, following calls to extract_min will be O(log(N)).
+
+  It is worth mentioning that the avl tree is a viable alternative as it can support the same use cases with its
```
eboasson, line 24: Up to here I completely agree with the comment — so including the criticism of the choice of a fibonacci heap in this particular context. For the way Cyclone uses the priority queues, a fibonacci heap is indeed not the best choice. It has simply been able to stick around because it has never really been an issue in practice, but ideally it should be replaced with one that has better performance characteristics (for the usage patterns).

That solution is not an AVL tree. (It actually was used for that once upon a time: https://github.com/ADLINK-IST/opensplice/blob/20789cdc061cd7097b7f7df8af759c39a8f6cedc/src/services/ddsi2/code/rtps.c#L754 and *that* did show up as a bottleneck.)

There are other data structures, probably a classic binary heap would be a sensible choice that fits Cyclone's needs better than a fibonacci heap, but many other tricks exist. Replacing it with a binary heap would probably be good, re-inventing some (timer) wheel would probably be a waste of time ...

eboasson (commented):
Just a little bit of nitpicking 🙂

Regarding your suggestion, I think there's no point in doing that. IINM the fibonacci heap was invented to improve the asymptotic running time of algorithms that depend on very cheap "merge" operations like Dijkstra's single-source shortest path. Modifying it here to try to mitigate the downside of abusing it doesn't seem quite right to me.

eboasson (approved):

eboasson:
I missed the update ... apologies for the delay @mvandenhoek !

PR #1681, opened 2023-05-03
[CP-277] Misra check github workflow: fix filter_sarif error due to m…

mvandenhoek:
The comment was interpreted as a pattern, and that caused an error (https://github.com/eclipse-cyclonedds/cyclonedds/actions/runs/4863665172/jobs/8671671156?pr=1680):

`File "/home/runner/work/_actions/advanced-security/filter-sarif/v1/filter_sarif.py", line 44, in parse_pattern
    raise Exception('Invalid pattern: "' + line + '" Contains more than one separator!')
Exception: Invalid pattern: "# Justification: Based on the rationale presented by MISRA:" Contains more than one separator!`

The error complains about the separator ':' within the comment, because it thinks it's parsing a pattern rather than a comment. The solution is to place the comment outside of the "pattern:" command.


NoxPardalis (approved):
Thanks for catching and fixing this! I just noticed it this morning as well but I'm confused as it seemed to work before :shrug:

eboasson (approved):
Thanks @mvandenhoek !

PR #1694, opened 2023-05-17
Fixes for creating type-info from dynamic type

dpotman:
This fixes a number of issues in the dynamic type API implementation, related to type-info creation. Additionally a commit with some fixes for setting flags and hash-names in the dynamic type API is included. 

- Handle plain-collection types and nested hash types correctly
- Fix locking for operations on the type library
- Refcounting minimal DDSI types is fixed

See the commit messages for more details.


src/core/ddsi/src/ddsi__typelib.h
```diff
@@ -66,10 +63,27 @@ dds_return_t ddsi_type_ref_proxy (struct ddsi_domaingv *gv, struct ddsi_type **t
...
+ * with a refcount of 1. This applies to the top-level minimal type and
+ * all underlying dependencies.
+ *
+ * @param gv  domain globals
+ * @param type_c  the complete DDSI type
+ * @param type_info  the type_info structure to create
+ * @param type_m  the minimal DDSI type corresponding to the provided complete type is stored in this variable
+ * @return dds_return_t
+ */
+dds_return_t ddsi_type_get_typeinfo_locked (struct ddsi_domaingv *gv, const struct ddsi_type *type_c, struct ddsi_typeinfo *type_info, struct ddsi_type **type_m);
```
eboasson, line 86: It appears from the rest of the code that in all (successful) cases the `type_c` and `*type_m` point to the complete and corresponding minimal type and have this pointer reflected in their reference counts. (I don't know how to put it otherwise ...)

For the minimal type this is suggested by the description, although it only mentions the case where the type is added. For the complete type, there is no hint whatsoever that this is the case, and it is contradictory to the `const` bit. I am always a bit ambivalent about `const` vs. reference counts and mutexes, whether the object changes from an application perspective or from a memory contents perspective ...

`const` in C really only means the second, and so only says it is ROMmable. In reality, that's often the less useful thing ... All of these `ddsi_type`s are constructed dynamically, so cannot be in ROM, so I am generally not against a refcount changing during the operation. In this case however, it is not merely that the function has a net-zero effect on the refcount.

So I do think it would be better to mark it non-`const`.

eboasson:
One more thing: please just update the commits based on the review and force-push it. I'll manage 🙂

eboasson (approved):
Thanks @dpotman !

PR #1696, opened 2023-05-22
Use full version in shared objects

eboasson:
One mishap too many occurred where a symbol was added to the core library, the C++ binding depended on it and the library path pointed to an older core library.

In traditional C binary interface compatibility terms, the old binaries can run with the new library and so it is compatible, but the practice turns out to be different.


paulsohn:
@eboasson For me it builds fine, it reduced exceptionally setting `LD_LIBRARY_PATH=""` only for building cyclonedds.

PR #1698, opened 2023-05-22
Include type-info in sertype_equal for default sertype implementation

dpotman:
When comparing sertypes that have xtypes type meta-data, the type information should be included. This PR adds the check to the default sertype implementation and adds a test for this.

### DIFF
--- a/src/core/ddsc/src/dds_sertype_default.c
+++ b/src/core/ddsc/src/dds_sertype_default.c
@@ -57,6 +57,29 @@ static bool sertype_default_equal (const struct ddsi_sertype *acmn, const struct
     return false;
   assert (a->type.opt_size_xcdr1 == b->type.opt_size_xcdr1);
   assert (a->type.opt_size_xcdr2 == b->type.opt_size_xcdr2);
+
+#ifdef DDS_HAS_TYPE_DISCOVERY
+  if (a->type.flagset & DDS_TOPIC_XTYPES_METADATA)
+  {
+    ddsi_typeinfo_t *ti_a, *ti_b;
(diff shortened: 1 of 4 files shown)


src/core/ddsc/src/dds_sertype_default.c
```diff
@@ -57,6 +57,29 @@ static bool sertype_default_equal (const struct ddsi_sertype *acmn, const struct
     return false;
   assert (a->type.opt_size_xcdr1 == b->type.opt_size_xcdr1);
   assert (a->type.opt_size_xcdr2 == b->type.opt_size_xcdr2);
+
+#ifdef DDS_HAS_TYPE_DISCOVERY
+  if (a->type.flagset & DDS_TOPIC_XTYPES_METADATA)
+  {
+    ddsi_typeinfo_t *ti_a, *ti_b;
+    ti_a = ddsi_typeinfo_deser (a->typeinfo_ser.data, a->typeinfo_ser.sz);
```
eboasson, line 65: Do we really need to deserialize here? Or can we get away with comparing the serialised representation?

For that, we would have to be certain about the endianness and padding, and perhaps something else. The first two are reasonably easy to deal with, the third is of course tricky 😄

dpotman: The problem is the set of hash type identifiers the type depends on, that is included in a type-info object. The spec (7.6.3.2.1) does not specify any ordering for this list (`TypeInformation.[minimal/complete].dependent_typeids`), and it is also allowed to leave out one or more type identifiers, as long as the dependent type count is correct (`TypeInformation.[minimal/complete].dependent_typeid_count`).

PR #1707, opened 2023-05-27
Move project() statements below cmake_minimum_required()

roehling:
The `cmake_minimum_required()` statement configures compatibility settings and CMake policies which also affect `project()`. Therefore, it should be the first statement executed. Starting with CMake 3.26, a warning is issued otherwise. This PR fixes the examples and silences the warning.


eboasson (approved):
Thanks @roehling !

PR #1711, opened 2023-05-30
Fix error handling in type wrapper and type object fuzzing test

dpotman:
This fixes a bug in the error handling for the functions that add type details to a type in the type library. A fix in the error handling in the OSS type object fuzzer is also included.


eboasson (approved):
Thanks @dpotman !

PR #1718, opened 2023-06-05
add doxygen comments to mh3.h and md5.h

mvandenhoek:
One question I have about the md5.h is why explicitly define typedef `ddsrt_md5_byte_t` and `ddsrt_md5_word_t` here, rather than just using the standard `uint8_t` and `uint32_t` ?

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/md5.h
+++ b/src/ddsrt/include/dds/ddsrt/md5.h
@@ -61,7 +61,8 @@
 #include "dds/export.h"
 
 /** @file md5.h
- * The message digest algorithm computes a 128-bit hash from a message.
+ * The MD5 hash (a 128-bit hash, still suitable for detecting unintentional corruption,
+ * but no longer to be used where a cryptographic hash is needed)
  * 
  * This package supports both compile-time and run-time determination of CPU
  * byte order.  If ARCH_IS_BIG_ENDIAN is defined as 0, the code will be
--- a/src/ddsrt/include/dds/ddsrt/md5.h
+++ b/src/ddsrt/include/dds/ddsrt/md5.h
@@ -60,7 +60,9 @@
 #include <stddef.h>
 #include "dds/export.h"
 
-/*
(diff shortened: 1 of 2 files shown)


eboasson:
> One question I have about the md5.h is why explicitly define typedef `ddsrt_md5_byte_t` and `ddsrt_md5_word_t` here, rather than just using the standard `uint8_t` and `uint32_t` ?

That's simply to keep the changes to the imported sources to a minimum.

src/ddsrt/include/dds/ddsrt/md5.h
```diff
@@ -60,7 +60,9 @@
 #include <stddef.h>
 #include "dds/export.h"
 
-/*
+/** @file md5.h
+ * The message digest algorithm computes a 128-bit hash from a message.
```
eboasson, line 64: I would say "the MD5 hash (a 128-bit hash, still suitable for detecting unintentional corruption, but no longer to be used where a cryptographic hash is needed)". The parenthesised bit is for those who are a bit behind.

eboasson (commented):
Perhaps a bit nitpicky, but I think either one should simply say "MD5" and assume that everyone knows what that means, or give a little bit additional information including pointing out what it may be used to.

src/ddsrt/include/dds/ddsrt/md5.h
mvandenhoek: I know some people might have a different opinion, but my rule of thumb is that if something is an abbreviation or acronym, there should be at least in one place a short explanation (in a report that would be where it is first mentioned, in code I think the header file of that component is a suitable place).

eboasson: It is a good rule of thumb, but "the message digest algorithm" is actually leaving undefined *which* digest, and everyone refers to it simply as "MD5" so expanding it doesn't really help anyone.

And I dare say it was quite obvious from the function name what it would do anyway ...

PR #1719, opened 2023-06-06
remove leading and trailing white spaces from distinguished name when…

SeanYu81:
DDS Security v1.1 standard mentions distinguished name (DN) is used as subject name, and also DN should follow RFC4514 (String Representation of Distinguished Name). RFC2253 (obsoleted by RFC4514) gives some explanations about how to process whitespace characters in DN.  (refer to https://www.rfc-editor.org/rfc/rfc2253, chapter 4).

So following the standard, below two DNs should be the same.
CN=KeyBoard,OU=AGV,L=Graz,C=AT
CN=Keyboard, OU=AGV, L=Graz, C=AT

### DIFF
--- a/src/ddsrt/src/string.c
+++ b/src/ddsrt/src/string.c
@@ -219,3 +219,41 @@ ddsrt_strndup(const char *str, size_t len)
 
   return s;
 }
+
+void
+ddsrt_str_trim(char *str)
+{
+  int index, i;
+
+  if (str == NULL) {
+    return;
+  }
+  
+  /* trim leading white spaces */
+  index = 0;
+  while(str[index] == ' ' || str[index] == '\t' || str[index] == '\n') {
+    index++;
+  }
+
+  /* shift all trailing characters to its left */
+  i = 0;
+  while(str[i + index] != '\0') {
+    str[i] = str[i + index];
+    i++;
+  }
+  str[i] = '\0'; /* terminate string with NULL */
+
+  /* trim trailing white spaces */
+  i = 0;
+  index = -1;
+  while(str[i] != '\0') {
+    if(str[i] != ' ' && str[i] != '\t' && str[i] != '\n') {
+      index = i;
+    }
+    i++;
+  }
+
+  /* mark the next character to last non white space character as NULL */
+  str[index + 1] = '\0';
+}
+
--- a/src/security/builtin_plugins/access_control/src/access_control_utils.c
+++ b/src/security/builtin_plugins/access_control/src/access_control_utils.c
@@ -302,6 +302,8 @@ static bool string_to_properties(const char *str, DDS_Security_PropertySeq *prop
     if (strlen(tok) == 0)
       continue;
     char *name = ddsrt_strsep (&tok, "=");
+    ddsrt_str_trim(name);
+    ddsrt_str_trim(tok);
     if (name == NULL || tok == NULL || properties->_length >= properties->_maximum)
     {
       ddsrt_free (copy);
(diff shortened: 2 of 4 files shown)


src/ddsrt/src/string.c
```diff
@@ -219,3 +219,41 @@ ddsrt_strndup(const char *str, size_t len)
...
+    index++;
+  }
+
+  /* shift all trailing characters to its left */
+  i = 0;
+  while(str[i + index] != '\0') {
+    str[i] = str[i + index];
+    i++;
+  }
+  str[i] = '\0'; /* terminate string with NULL */
```
eboasson, line 244: For the way it is used, it could also be changed to only trim the right-hand white space and return a pointer to the first non-white space character, eliminating the shifting. That's probably what I would have done, but this interface is probably less likely to be used incorrectly 🙂

What I don't really like is doing the shifting even over a distance of 0, which I suspect is the common case. That's easily remedied, of course.

```suggestion
  if (index > 0) {
    i = 0;
    while(str[i + index] != '\0') {
      str[i] = str[i + index];
      i++;
    }
    str[i] = '\0'; /* terminate string with NULL */
  }
```

src/ddsrt/src/string.c
```diff
@@ -219,3 +219,41 @@ ddsrt_strndup(const char *str, size_t len)
 
   return s;
 }
+
+void
+ddsrt_str_trim(char *str)
+{
+  int index, i;
```
eboasson, line 226: The other string functions consistently use `size_t` to index the string. Given how they are input, I doubt the distinguished names can ever be > 2GB, but using `int` here means there is a limit that one would not expect based on the other string functions.

src/ddsrt/src/string.c
```diff
@@ -219,3 +219,41 @@ ddsrt_strndup(const char *str, size_t len)
...
+  i = 0;
+  while(str[i + index] != '\0') {
+    str[i] = str[i + index];
+    i++;
+  }
+  str[i] = '\0'; /* terminate string with NULL */
+
+  /* trim trailing white spaces */
+  i = 0;
+  index = -1;
```
eboasson, line 248: Making `index` of type `size_t` would of course mean this here becomes a minor problem. Initializing to 0, setting it to `i+1` and writing the zero at the end at `str[index]` instead would fix that.

src/security/builtin_plugins/access_control/src/access_control_utils.c
```diff
@@ -302,6 +302,8 @@ static bool string_to_properties(const char *str, DDS_Security_PropertySeq *prop
     if (strlen(tok) == 0)
       continue;
     char *name = ddsrt_strsep (&tok, "=");
+    ddsrt_str_trim(name);
```
eboasson, line 305: Based on the RFC (if I am reading it correctly), it should really only be trimming space (ASCII 32), not other white space characters. I'm rather doubtful that the `string_to_properties` is correct, given that it doesn't account for escaping, and I am sure this makes it less incorrect.

Am I right in assuming you actually ran into this problem? Because then I think it would make sense to at least do this improvement.

src/ddsrt/src/string.c
```diff
@@ -219,3 +219,41 @@ ddsrt_strndup(const char *str, size_t len)
...
+{
+  int index, i;
+
+  if (str == NULL) {
+    return;
+  }
+  
+  /* trim leading white spaces */
+  index = 0;
+  while(str[index] == ' ' || str[index] == '\t' || str[index] == '\n') {
```
eboasson, line 234: I would expect a generic "trim whitespace" to rely on `isspace`, so also handle `'\r'`, `\v` and `\f`. Or (see comment below) strictly limit it to the ordinary space character (`' '`), but then I think the name/description should be updated to reflect that.

eboasson (commented):
Thanks @SeanYu81, even if it meant trying to wrap my head around the RFCs ... I do have a couple of questions concerning some details.

Also, maybe, while you're at it and since we're trying to up the test coverage: would you be willing to add a few test cases for `ddsrt_str_trim` to `src/ddsrt/tests/string.c`?

PR #1720, opened 2023-06-08
Fix for idl_generate_generic with files directly on BASE_DIR

reicheratwork:
An error occurs in CMake when you supply the idlc_generate_generic function a path which contains a file directly located on a supplied BASE_DIR, in this case the _mid_dir_path component does not exist, so the REGEX operation does not get enough arguments Fixed this by converting the _mid_dir_path to a string explicitly, so even when this parameter does not exist, you get an empty string


eboasson (approved):
Nice, looks good to me, thanks @reicheratwork . I could merge it already, but I think we can wait a little while to see if it is confirmed fixed on https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/417 too.

PR #1723, opened 2023-06-13
Reject arrays with empty bounds or 0 bound

eboasson:
As per spec; affects both the type object validation and IDLC.


dpotman (approved):
LGTM

PR #1724, opened 2023-06-14
Vlan support

MarcelJordense:
Added vlan-tagging support for raweth transport


src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -115,22 +158,48 @@ static ssize_t ddsi_raweth_conn_write (struct ddsi_tran_conn * conn, const ddsi_
...
+    vhdr.e.proto = htons ((uint16_t) ETH_P_8021Q);
+    vhdr.vtag = htons (vtag);
+    vhdr.proto = htons ((uint16_t) uc->m_base.m_base.m_port);
+    hdrlen = sizeof(vhdr);
+  } else {
+    vhdr.e.proto = htons ((uint16_t) uc->m_base.m_base.m_port);
+    hdrlen = 14;
+  }
+
+  iov_base = (ddsrt_iovec_t *)(iov - 1);
```
eboasson, line 189: Just a bit of nitpicking, I'd simply use `iov[-1]` instead of `iov_base`.

I believe all the shifting is correct, but I do think it would make sense to give it a name (like `#define DDSI_TRAN_RESERVED_IOV_SLOTS 1` and then document for `ddsi_con_write` that it has this expectation, and do a static assert here just in case).

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -210,6 +302,34 @@ static dds_return_t ddsi_raweth_create_conn (struct ddsi_tran_conn **conn_out, s
     return DDS_RETCODE_ERROR;
   }
 
+  memset(&mreq, 0, sizeof (mreq));
+  mreq.mr_ifindex = (int)intf->if_index;
+  mreq.mr_type = PACKET_MR_PROMISC;
```
eboasson, line 307: It is somewhat surprising to me that promiscuous mode is needed, or is that because of the `ETH_P_ALL`?

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -397,7 +526,7 @@ int ddsi_raweth_init (struct ddsi_domaingv *gv)
   fact->gv = gv;
   fact->m_free_fn = ddsi_raweth_deinit;
   fact->m_typename = "raweth";
-  fact->m_default_spdp_address = "raweth/ff:ff:ff:ff:ff:ff";
+  fact->m_default_spdp_address = "raweth/01:00:5e:7f:00:01";
```
eboasson, line 529: (I presume this comes from the spec? Then of course I am happy with the change. I used a broadcast originally because, well, why not for a quick hack.)

eboasson (commented):
Thanks @MarcelJordense it basically looks good to me, but there are a few things I think could be done more elegantly. Some of it really feels like dusting off some old code, some others I am not 100% sure about.

I'll see if I can try out a few of these things, see if what I have in my head works out. Then you can decide whether you agree with it or not 😄 .

eboasson:
@MarcelJordense I did manage to try out something: https://github.com/eboasson/cyclonedds/tree/vlan_support. It is just a rebase of this PR on current master, plus some fiddling on my side. It is not polished, mostly because I don't have the equipment at hand to really try it out ...

What it effectively does is change the ethernet locators to a mac_address.vlanid.priority form, and the rest of it is mostly trying to find a compromise between squeezing the vlan information into the port field while still being able to manipulate it from the generic code. That is a bit messy and things would probably work out better if the port number manipulation is moved into the UDP transport code. I didn't want to do that right now, especially not until I have confirmation supporting VLAN in this particular manner actually does work.

I'd be grateful if you'd take a look at it. Then we can decide whether to discard it, polish it, or even move the port number calculations to the UDP transport altogether.

src/core/ddsi/src/ddsi_xmsg.c
```diff
@@ -146,7 +146,8 @@ struct ddsi_xpack
   ddsrt_atomic_uint32_t calls;
   uint32_t call_flags;
   size_t niov;
-  ddsrt_iovec_t *iov;
+  ddsrt_iovec_t *iov_base; /* the list of iov's where the first one is reserved to be used by the transport layer */
+  ddsrt_iovec_t *iov; /* points to iov_base[1] */
```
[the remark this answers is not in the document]
MarcelJordense: Ok, will remove iov_base

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -172,12 +241,35 @@ static int ddsi_raweth_conn_locator (struct ddsi_tran_factory * fact, struct dds
   return ret;
 }
 
+static dds_return_t ddsi_raweth_set_filter (struct ddsi_tran_factory * fact, ddsrt_socket_t sock, uint32_t port)
+{
+  struct sock_filter code[] = {
+      { 0x28, 0, 0, 0x0000000c },
```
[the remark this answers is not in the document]
MarcelJordense: It seems that the linux kernel removes the vlan tag, when it is present, from the ethernet header before passing the frame to the filter function.  Thus the check on only the ethernet type field seems to be sufficient. I will add the alternative filter expression in a comment.

src/core/ddsi/src/ddsi_raweth.c
MarcelJordense: I do not exactly understand what you mean here with a static assert?

src/core/ddsi/src/ddsi_raweth.c
MarcelJordense: It appears that promiscuous mode is necessary because the linux kernel will drop frames with a vlan tag for which no vlan is configured.

src/core/ddsi/include/dds/ddsi/ddsi_config.h
```diff
@@ -90,6 +90,8 @@ struct ddsi_config_networkpartition_listelem {
 #ifdef DDS_HAS_SSM
   struct ddsi_networkpartition_address *ssm_addresses;
 #endif
+  uint16_t vlan_id;
```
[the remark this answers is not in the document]
MarcelJordense: Ok, as agreed we will for now use your proposal to extend the MAC address optional with the vlan information. Thanks for implementing this. I have tested it and it appears to work fine.

src/core/ddsi/src/ddsi_raweth.c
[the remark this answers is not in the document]
eboasson: Good to know that that's how it (seems) to work! Thanks for checking

src/core/ddsi/src/ddsi_raweth.c
eboasson: Just like a normal assert, except one that works at compile time. C11 has _Static_assert for this, but we emulate it by abusing array type definitions in `DDSRT_STATIC_ASSERT`.

So here, basically `DDSRT_STATIC_ASSERT(DDSI_TRAN_RESERVED_IOV_SLOTS >= 1)`. It makes the assumption explicit and helps catch trouble if someone decides to change `DDSI_TRAN_RESERVED_IOV_SLOTS`. That's all 🙂

eboasson (approved):

PR #1726, opened 2023-06-15
Fix MISRA rule 21-19

Splinter1984:
This addresses rule 21.19 .

```
returned by the Standard Library functions 'localeconv', 'getenv', 'setlocale' or, 'strerror' shall only be used as if they have pointer to const-qualified type
```


eboasson (approved):
Long live MISRA?? 🧐 More importantly: thanks @Splinter1984 !

PR #1727, opened 2023-06-16
cdtors.h: add doxygen comments

mvandenhoek:


src/ddsrt/include/dds/ddsrt/cdtors.h
```diff
@@ -14,16 +14,40 @@
...
+ * Initializes and finalizes mutexes of ddsrt/random and ddsrt/atomics, and for Windows also the ddsrt/socket
+ * (not necessary for other platforms).
+ */
+
 #if defined (__cplusplus)
 extern "C" {
 #endif
 
+/**
+ * @brief Initialize ddsrt units
```
eboasson, line 30: I think it needs two extra bit of information:
* I counts the number of invocations of `ddsrt_init`, and only the "first" (scare quotes because of `ddsrt_fini`) actually initializes it.
* Any concurrent attempts at initializing it will wait until it has been initialized.

src/ddsrt/include/dds/ddsrt/cdtors.h
```diff
@@ -14,16 +14,40 @@
...
+ */
 void ddsrt_init(void);
 
+/**
+ * @brief Finalize ddsrt units
```
eboasson, line 35: This then also needs a mention of the reference counting and note that only the last one actually de-initializes it.

Moreover: it does not allow for calling `init` in parallel to the `fini` call that de-initializes everything. Arguably, that's a bug, but not a urgent one.

eboasson (approved):

PR #1728, opened 2023-06-16
Fix MISRA rule 21-17

Splinter1984:
This addresses rule 21.17 .

```
Avoid overflow due to reading a not zero terminated string
```


src/core/ddsc/src/dds_serdata_builtintopic.c
```diff
@@ -118,7 +118,7 @@ struct format_address_arg {
 static void format_address (const ddsi_xlocator_t *n, void *varg)
 {
   struct format_address_arg *arg = varg;
-  char buf[DDSI_LOCSTRLEN];
+  char buf[DDSI_LOCSTRLEN] = {'\0'};
```
eboasson, line 121: If this is truly necessary, then `ddsi_xlocator_to_string` is broken: it is supposed to always return a properly terminated string in `buf` (assuming the size argument is at least 1).

If that guarantee doesn't hold, I am quite certain there are many more places where something would have to change. A quick check shows that is mostly built on `snprintf` (which gives this guarantee), but that there is a potential risk with `inet_ntop`: at least on macOS the manpage isn't clear about this detail.

PR #1730, opened 2023-06-16
Fix MISRA rule 21-18

Splinter1984:
This addresses rule 21.18.
```
The size_t argument passed to any function in string.h shall have an appropriate value 
```

### DIFF
--- a/src/ddsrt/src/expand_vars.c
+++ b/src/ddsrt/src/expand_vars.c
@@ -95,7 +95,9 @@ static char *expand_varbrace (const char **src, expand_fn expand, expand_lookup_
         goto err;
     }
     name = ddsrt_malloc ((size_t) (*src - start) + 1);
-    memcpy (name, start, (size_t) (*src - start));
+    if (*src > start) {
+      memcpy (name, start, (size_t) (*src - start));
+    }
     name[*src - start] = 0;
     if (**src == '}') {
         (*src)++;
--- a/src/ddsrt/src/string.c
+++ b/src/ddsrt/src/string.c
@@ -99,7 +99,9 @@ ddsrt_strlcpy(
     if (size <= srclen) {
       len = size - 1;
     }
-    memcpy(dest, src, len);
+    if (len > 0) {
+      memcpy(dest, src, len);
+    }
     dest[len] = '\0';
   }
 
--- a/src/ddsrt/src/xmlparser.c
+++ b/src/ddsrt/src/xmlparser.c
@@ -192,7 +192,9 @@ static int make_chars_available (struct ddsrt_xmlp_state *st, size_t nmin)
     /* ensure buffer space is available */
     if (st->fp != NULL) {
         if (pos + nmin > st->cbufmax) {
-            memmove (st->cbuf, st->cbuf + pos, st->cbufn - pos);
+            if (st->cbufn > pos) {
+              memmove (st->cbuf, st->cbuf + pos, st->cbufn - pos);
+            }
             st->cbufn -= pos;
             st->cbufp -= pos;
             if (st->cbufmark != NOMARKER) {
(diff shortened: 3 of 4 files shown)


src/ddsrt/src/expand_vars.c
```diff
@@ -95,7 +95,9 @@ static char *expand_varbrace (const char **src, expand_fn expand, expand_lookup_
         goto err;
     }
     name = ddsrt_malloc ((size_t) (*src - start) + 1);
-    memcpy (name, start, (size_t) (*src - start));
```
eboasson, line 98: Here, `src > start` necessarily holds under the assumption that the input is a null-terminated string starting with `{`, that this is the case is easily proven (short of arbitrary memory corruption). That makes the conditional always true and so it shouldn't be present. (And besides, there's another MISRA rule (rule 14.3) that forbids such a conditional!)

I don't know (yet) how to suppress the warning from codeql, but that's what we need to do here.

src/ddsrt/src/string.c
```diff
@@ -99,7 +99,9 @@ ddsrt_strlcpy(
     if (size <= srclen) {
       len = size - 1;
     }
-    memcpy(dest, src, len);
+    if (len > 0) {
+      memcpy(dest, src, len);
```
eboasson, line 103: Here I simply have yet another problem with MISRA. (Nothing new here.)

Per the C spec (I use the draft, just like everyone else), 7.21.1 paragraph 2, concerning string function conventions, of which `memcpy` is one: "Where an argument declared as **size_t n** specifies the length of the array for a function, **n** can have the value zero on a call to that function." And further: "Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values, as described in 7.1.4"

In this case, assuming `src`, `dest` and `size` are valid (and they must be, or else one may not call `ddsrt_strlcpy`), then `0 <= len && len <= srclen && len < size`. That makes all arguments valid according to the C standard.

MISRA completely ridiculously claims that 0 is not an "appropriate" value, apparently because it "may result in in read or write access beyond the bounds of an object passed as a parameter". But the the C spec says:

> On such a call, a function that locates a character finds no occurrence, a function that compares two character sequences returns zero, and a function that copies characters copies zero characters.

I've read the MISRA document in its entirety. I truly fear software written by its authors. I cannot agree to a change based on this rule where the number of bytes is 0. The most generous interpretation is that the MISRA authors don't know the difference between "positive" and "non-negative". (As if `size_t` could *ever* by negative ... but that's beside the point.)

src/ddsrt/src/xmlparser.c
```diff
@@ -192,7 +192,9 @@ static int make_chars_available (struct ddsrt_xmlp_state *st, size_t nmin)
     /* ensure buffer space is available */
     if (st->fp != NULL) {
         if (pos + nmin > st->cbufmax) {
-            memmove (st->cbuf, st->cbuf + pos, st->cbufn - pos);
```
eboasson, line 195: I have difficulty accepting this change, but I can live with it on the grounds that the code doesn't come with a *proof* that `0 <= pos <= st->cbufn`.

Still, I would prefer at least a comment explaining why this test is here (so a reference to MISRA 21.18).

However, that is not the last of it. If `pos == cbufn`, then fine. If `pos > cbufn`, which must be assumed possible if one requires this test because there is no proof that `pos <= cbufn`, then the values of `cbufn` and `cbufp` will become huge (on a 2 complement machine), and certainly *much* larger than the underlying buffer, and so skipping the `memcpy` to avoid undefined behaviour means we then have a problem on *all* memory accesses of `st->cbuf`. Checking those is infeasible in C. If only this were Ada and not C ...

So the only sensible action on encountering `st->cbufn < pos` is to call `abort()`. But, in all there wisdom, the MISRA authors decided that aborting execution is not allowed, not even when you know there is no recovery possible ...

eboasson (commented):
I can't really accept these changes. The problem lies with MISRA ...

src/ddsrt/src/expand_vars.c
Splinter1984: codeql doesn't detect violations here. but another tool detects it as an violation!🤷

PR #1732, opened 2023-06-16
Fix MISRA rule 9-1

Splinter1984:
This addresses rule 9.1.
```
Avoid use before initialization 
```


eboasson (commented):
LGTM, except for the typo in the commit message. This one typo I think ought to be corrected.

eboasson (approved):
Thanks!

PR #1734, opened 2023-06-19
ifaddrs.h: add doxygen comments

mvandenhoek:


src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -13,16 +13,23 @@
...
+  @brief This header deals with interface addresses
+*/
+
 #if defined (__cplusplus)
 extern "C" {
 #endif
 
+/** @brief The interface type */
 enum ddsrt_iftype {
   DDSRT_IFTYPE_UNKNOWN,
```
eboasson, line 27: These are pretty self-explanatory, but I think the documentation "standard" includes documenting the values. Same holds for documenting the fields of `struct ddsrt_ifaddrs`.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -36,15 +43,38 @@ struct ddsrt_ifaddrs {
 
 typedef struct ddsrt_ifaddrs ddsrt_ifaddrs_t;
 
+/**
+ * @brief Get the interface addresses (@ref ddsrt_ifaddrs) for specific address families
+ * 
+ * Note that array 'afs' must be terminated with an element 'DDSRT_AF_TERM'.
+ * 
+ * @param[out] ifap interface address pointer
```
eboasson, line 51: This is a bit too vague.
* on success, `*ifap` will be point to a newly allocated list of interfaces — that it is allocated matters because it means you can don't have to free it before calling it again.
* on failure, `*ifap` is left untouched

it is also not simply "interface addresses", you get interface information, including addresses.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -36,15 +43,38 @@ struct ddsrt_ifaddrs {
...
+ * @param[in] afs an array of address families
```
eboasson, line 52: And this should state that the returned interfaces are restricted to the address families in `afs`.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -36,15 +43,38 @@ struct ddsrt_ifaddrs {
...
+ */
 dds_return_t
 ddsrt_getifaddrs(
   ddsrt_ifaddrs_t **ifap,
   const int *afs);
 
+/**
+ * @brief Free the interface addresses
+ * 
+ * @param[in] ifa the interface addresses to free
```
eboasson, line 63: It is not just an arbitrary `ddsrt_ifaddrs_t *`, rather it must be the pointer that was returned by `ddsrt_getifaddrs`.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -36,15 +43,38 @@ struct ddsrt_ifaddrs {
...
 typedef struct ddsrt_ifaddrs ddsrt_ifaddrs_t;
 
+/**
+ * @brief Get the interface addresses (@ref ddsrt_ifaddrs) for specific address families
+ * 
+ * Note that array 'afs' must be terminated with an element 'DDSRT_AF_TERM'.
+ * 
+ * @param[out] ifap interface address pointer
+ * @param[in] afs an array of address families
+ * @return a DDS_RETCODE
```
eboasson, line 53: And the possible return codes are quite limited: `OK`, `NOT_ALLOWED`, `OUT_OF_RESOURCES` or the catch-all `ERROR`.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -36,15 +43,38 @@ struct ddsrt_ifaddrs {
...
   ddsrt_ifaddrs_t *ifa);
 
+/**
+ * @brief Get the mac address for a given interface name
+ * 
+ * Copies the first 6 bytes of the mac address into the buffer 'mac_addr'.
+ * 
+ * @param[in] interface_name the name of the interface
+ * @param[out] mac_addr a buffer to copy the mac address into
+ * @return a DDS_RETCODE
```
eboasson, line 76: It would be good to note that this is one of `OK`, `NOT_ALLOWED`, `OUT_OF_RESOURCES` or simply `ERROR` and that specifying an unknown interface simply results in `ERROR`.

If the interface exists and doesn't actually have a MAC address, like a loopback interface, the behaviour is platform-dependent. (On Linux, you get all 0, but that may not be universally true.)

eboasson (commented):
Thanks @mvandenhoek for getting started on documenting this, but it needs a bit more detail before it documents anything other than the obvious.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
mvandenhoek: Very well then, it's just that I had a gut feeling that these would be over the top in the sense of pointing out the obvious (I know you have a different standard for what's obvious, but that would result in even less comments).

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
-  uint32_t index;
-  uint32_t flags;
-  enum ddsrt_iftype type;
-  struct sockaddr *addr;
-  struct sockaddr *netmask;
-  struct sockaddr *broadaddr;
+  struct ddsrt_ifaddrs *next; ///< next pointer of the linked list
+  char *name; ///< name of the interface
+  uint32_t index; ///< the index of the interface
+  uint32_t flags; ///< flags from 'struct ifaddrs'
```
eboasson, line 37: You do realize that `struct ifaddrs` doesn't exist on Windows and that this reference is quite opaque to anyone using that ... operating system? If you search for, e.g., `IFF_MULTICAST` you'll find it, but the problem with that is that it presupposes someone knows what to look for.

I think it would be better to write something like
```suggestion
  uint32_t flags; ///< interface flags, e.g., IFF_MULTICAST, IFF_LOOPBACK, IFF_BROADCAST, ...
```
that at least gives someone not familiar with these interfaces a good starting point

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
-  char *name;
-  uint32_t index;
-  uint32_t flags;
-  enum ddsrt_iftype type;
-  struct sockaddr *addr;
-  struct sockaddr *netmask;
-  struct sockaddr *broadaddr;
+  struct ddsrt_ifaddrs *next; ///< next pointer of the linked list
+  char *name; ///< name of the interface
+  uint32_t index; ///< the index of the interface
```
eboasson, line 36: The "index of the interface" is a really just a unique number associated with the interface by the operating system, it just happens to be called "index" historically, from what I can tell. The socket abstractions are much less well-documented than you might be expecting ...

So
```suggestion
  uint32_t index; ///< a unique integer associated with the interface
```
might be more useful to the average user

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
 struct ddsrt_ifaddrs {
-  struct ddsrt_ifaddrs *next;
-  char *name;
-  uint32_t index;
-  uint32_t flags;
-  enum ddsrt_iftype type;
-  struct sockaddr *addr;
-  struct sockaddr *netmask;
-  struct sockaddr *broadaddr;
+  struct ddsrt_ifaddrs *next; ///< next pointer of the linked list
```
eboasson, line 34: Wouldn't that be
```suggestion
  struct ddsrt_ifaddrs *next; ///< pointer to the next interface in the returned list of interfaces
```
I am aware it is pretty much the same, but it doesn't hurt to make an effort to make it easier for users ...

(And yes, I am aware that I often fall short of that in writing documentation, but let's try to raise the bar.)

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
 
 /** @brief The interface type */
 enum ddsrt_iftype {
-  DDSRT_IFTYPE_UNKNOWN,
-  DDSRT_IFTYPE_WIRED,
-  DDSRT_IFTYPE_WIFI
+  DDSRT_IFTYPE_UNKNOWN, ///< An unknown interface
```
eboasson, line 27:
```suggestion
  DDSRT_IFTYPE_UNKNOWN, ///< An unknown interface type
```

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
+  DDSRT_IFTYPE_WIRED, ///< A wired interface (e.g. ethernet)
+  DDSRT_IFTYPE_WIFI ///< A wifi interface
```
eboasson, line 29:
```suggestion
  DDSRT_IFTYPE_WIFI ///< A wifi (IEEE 802.11) interface
```

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
-  struct sockaddr *addr;
-  struct sockaddr *netmask;
-  struct sockaddr *broadaddr;
+  struct ddsrt_ifaddrs *next; ///< next pointer of the linked list
+  char *name; ///< name of the interface
+  uint32_t index; ///< the index of the interface
+  uint32_t flags; ///< flags from 'struct ifaddrs'
+  enum ddsrt_iftype type; ///< @see ddsrt_iftype
+  struct sockaddr *addr; ///< network address of this interface
+  struct sockaddr *netmask; ///< netmask of this interface
```
eboasson, line 40:
```suggestion
  struct sockaddr *netmask; ///< netmask of this interface or a null pointer if it doesn't have one
```

src/ddsrt/include/dds/ddsrt/ifaddrs.h
```diff
@@ -24,41 +24,47 @@ extern "C" {
...
+  struct sockaddr *broadaddr; ///< broadcast address of this interface
```
eboasson, line 41:
```suggestion
  struct sockaddr *broadaddr; ///< broadcast address of this interface or a null pointer if it doesn't have one
```

eboasson (commented):
Thanks @mvandenhoek for updating the comments and adding the missing bits of documentation. I've done some digging and I am afraid I am of the opinion that a bit more information is needed. Some of it is just nitpicking, but not all of it.

I've tried to make it easy for you by providing suggestions — and please note that are just that, suggestions.

src/ddsrt/include/dds/ddsrt/ifaddrs.h
mvandenhoek: Okay, I really just wasn't sure what to do with this. Essentially most of this unit isn't really our code, they are just very thin wrappers (with arguably a bit better error handling) around OS calls. Therefore I'm tempted to just say some form of "go look at the OS documentation".

src/ddsrt/include/dds/ddsrt/ifaddrs.h
mvandenhoek: I'm not sure what this adds. Surely I can assume the reader knows what a linked list is? Then the next pointer needs no further explanation. It is just a pointer to another struct of the same type.

eboasson (approved):
Thanks @mvandenhoek !

PR #1737, opened 2023-06-26
fix MISRA rule 8.2

Splinter1984:
This addresses rule 8.2
```
Function types shall have named parameters.
```


eboasson (approved):
Thanks @Splinter1984 !

PR #1738, opened 2023-06-27
Fix misra rule 17.4 (missing `dds_cdstream.c` file violations)

Splinter1984:
This PR complements the violation fixes for 17.4 MISRA-RULE.

PR #1740, opened 2023-06-27
Fix MISRA 17.4 (missing violation)

Splinter1984:
Following the suggestion from next [https://github.com/eclipse-cyclonedds/cyclonedds/pull/1738#issuecomment-1609083552](comment)

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -449,6 +449,7 @@ static uint32_t get_collection_elem_size (uint32_t insn, const uint32_t * __rest
       break;
   }
   abort ();
+  return 0u;
 }
 
 static uint32_t get_adr_type_size (uint32_t insn, const uint32_t * __restrict ops)
@@ -954,6 +955,7 @@ static uint32_t read_union_discriminant (dds_istream_t * __restrict is, uint32_t
       break;
     default: return 0;
   }
+  return 0;
 }
 
 static const uint32_t *find_union_case (const uint32_t * __restrict union_ops, uint32_t disc)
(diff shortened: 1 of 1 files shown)


src/core/cdr/src/dds_cdrstream.c
```diff
@@ -954,6 +955,7 @@ static uint32_t read_union_discriminant (dds_istream_t * __restrict is, uint32_t
       break;
     default: return 0;
   }
+  return 0;
```
eboasson, line 958: I may be nitpicking here: but I don't like changing *this* function in *this* way. The reason is the `default: return 0` just above it, and the fact that there is only one (dead-end) path through the function that runs into triggers this MISRA violation.

There are a number of ways it can be done that would make me happy (insofar as anything MISRA can make me happy):
* there is the obvious `default: abort (); return 0;` solution (line 953 in new numbers)
* or `default: break;` (line 956)
* or, recognise that the default case on line 956 is already an impossible one, and so instead make both default cases a simple `break`, and here do
```suggestion
  abort ();
  return 0;
```

@dpotman I *am* correct in that the default cases are both impossible, am I not?

eboasson (commented):
Thanks for rebasing @Splinter1984. I had one final look before merging and noticed a detail that I don't quite like. It is not wrong, it rather is a case of a function having been written with two different ideas in mind and your proposed change making me notice it.

src/core/cdr/src/dds_cdrstream.c
```diff
@@ -954,6 +955,7 @@ static uint32_t read_union_discriminant (dds_istream_t * __restrict is, uint32_t
       break;
     default: return 0;
```
Splinter1984, line 956: I would suggest the following which won't actually change anything in how the function works now.
```suggestion
    default: break;
   }
   return 0;
```

src/core/cdr/src/dds_cdrstream.c
dpotman: @eboasson yes, that's correct (and also also ensured in `normalize_uni_disc`), the IDL compiler or dynamic type builder is expected to only use one of the listed types as union discriminant.

eboasson (approved):

PR #1741, opened 2023-06-27
Use `bool` instead of `int` in hash tables, RHC, XML parser

eboasson:
This PR replaces some of the `int`s in the code by `bool`s where they actually are booleans. The better the types match the meaning, the better, in my view. Curiously, in this particular case, I am in agreement with MISRA 😱

There is a bit of collateral damage: in the XML parser I ended up changing some documentation (and indentation, just look at a diff ignoring whitespace). If that's a complication in reviewing I'll split it 🙂

There are plenty other places where the same change can be made but there's no need to do all of them in one go.


noxpardalis (approved):
Looks good to me!

As an aside, I didn't realize that `/**< ...*/` was the syntax for documenting a field with doxygen. I wonder why the `<` was needed as part of their syntax 🤔.

eboasson:
> As an aside, I didn't realize that `/**< ...*/` was the syntax for documenting a field with doxygen. I wonder why the `<` was needed as part of their syntax 🤔.

Haha! Yes, I too can't escape the feeling that doxygen and cmake are related somehow ...

PR #1743, opened 2023-06-28
example/throughput/readme.rst fixes

JEnoch:
The names of Throughput example binaries are wrong in the `readme.rst`:

 - `./publisher` vs. `./ThroughputPublisher`
 - `./subscriber` vs. `./ThroughputSubscriber`


eboasson (approved):
Thanks @JEnoch !

PR #1744, opened 2023-06-28
Add read/take API calling an application-provided function for each sample

eboasson:
This PR proposes a new pair of API functions:

* dds_read_with_collector
* dds_take_with_collector

The key difference with all the other read/take functions is that this pair takes a callback function as an argument, to be called for each sample matched by the conditions given in the other arguments.

The RHC interface has been updated to make this the only model for reading/taking data.  Consequently, the regular read/take has been reworked and readcdr/takecdr are now simply wrappers around the new API functions.  Some of the common code in setting up read/take variants is now factored out into separate functions.

This also eliminates the abomination that was the "lock samples" operation and the magic value for the maximum number of samples to read to prevent locking.  This therefore also fixes some odd issues caused by the magic number and elimates the conditional lock/unconditional unlock in the RHC that was one of the internal consequences of the abomination.

The C++ binding is (hopefully) the only piece of code in existence that relied on the abomination. Updating the read/take functions in the C++ binding to use these new functions should improve performance a bit as well. That needs to be done before this can be merged.

### DIFF
--- a/src/core/ddsc/src/dds__reader.h
+++ b/src/core/ddsc/src/dds__reader.h
@@ -31,25 +31,6 @@ void dds_reader_invoke_cbs_for_pending_events(struct dds_entity *e, uint32_t sta
 /** @component reader */
 dds_return_t dds_return_reader_loan (dds_reader *rd, void **buf, int32_t bufsz);
 
-
-#define DDS_READ_WITHOUT_LOCK (0xFFFFFFED)
-
-/**
- * @component reader
- *
- * Returns number of samples in read cache and locks the reader cache to make
- * sure that the samples content doesn't change. Because the cache is locked,
- * you should be able to read/take without having to lock first. This is done
- * by passing the DDS_READ_WITHOUT_LOCK value to the read/take call as maxs.
- * Then the read/take will not lock but still unlock.
- *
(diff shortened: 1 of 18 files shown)


src/core/ddsc/src/dds__reader.h
```diff
@@ -25,25 +25,6 @@ void dds_reader_status_cb (void *entity, const struct ddsi_status_cb_data * data
...
- * you should be able to read/take without having to lock first. This is done
- * by passing the DDS_READ_WITHOUT_LOCK value to the read/take call as maxs.
- * Then the read/take will not lock but still unlock.
- *
- * Used to support LENGTH_UNLIMITED in C++.
- *
- * @param entity reader entity
- * @return the number of samples
- */
-DDS_EXPORT uint32_t dds_reader_lock_samples (dds_entity_t entity);
```
dpotman, line 45: This function is also declared in `dds_internal_api.h:52`, so that one also needs to be removed (same for `DDS_READ_WITHOUT_LOCK`).

eboasson:
The new (second) commit eliminates the need for the caller of the new API function to patch the generation and sample ranks. That should be a bit nicer.

PR #1745, opened 2023-06-29
Fix misra rule 5 3

Splinter1984:
This PR addresses rule 5.3.
```
Identifier declared in a local or function prototype scope shall not hide an identifier declared in a global or namespace scope.
```

### DIFF
--- a/src/tools/ddsperf/ddsperf.c
+++ b/src/tools/ddsperf/ddsperf.c
@@ -2434,9 +2434,11 @@ int main (int argc, char *argv[])
 
   /* Just before starting the threads but after setting everything up, wait for
      the required number of peers, if requested to do so */
+  dds_time_t tnow;
+  bool ok;
   if (initmaxwait > 0)
   {
-    dds_time_t tnow = dds_time ();
+    tnow = dds_time ();
     const dds_time_t tendwait = tnow + (dds_duration_t) (initmaxwait * 1e9);
     ddsrt_mutex_lock (&disc_lock);
     while (matchcount < minmatch && tnow < tendwait)
@@ -2446,15 +2448,15 @@ int main (int argc, char *argv[])
       ddsrt_mutex_lock (&disc_lock);
       tnow = dds_time ();
     }
-    const bool ok = (matchcount >= minmatch);
-    if (!ok)
+    const bool status_ok = (matchcount >= minmatch);
+    if (!status_ok)
     {
       /* set minmatch to an impossible value to avoid a match occurring between now and
          the determining of the exit status from causing a successful return */
       minmatch = UINT32_MAX;
     }
     ddsrt_mutex_unlock (&disc_lock);
-    if (!ok)
+    if (!status_ok)
       goto err_minmatch_wait;
     dds_sleepfor (DDS_MSECS (100));
   }
@@ -2540,7 +2542,7 @@ int main (int argc, char *argv[])
 
   /* Run until time limit reached or a signal received.  (The time calculations
      ignore the possibility of overflow around the year 2260.) */
-  dds_time_t tnow = dds_time ();
+  tnow = dds_time ();
   const dds_time_t tstart = tnow;
   if (tref == DDS_INFINITY)
     tref = tstart;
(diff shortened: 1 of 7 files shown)


src/tools/ddsperf/ddsperf.c
```diff
@@ -2434,9 +2434,11 @@ int main (int argc, char *argv[])
 
   /* Just before starting the threads but after setting everything up, wait for
      the required number of peers, if requested to do so */
+  dds_time_t tnow;
+  bool ok;
   if (initmaxwait > 0)
```
eboasson, line 2439: As far as I am concerned this a false positive from the analyzer: there is no `tnow` nor an `ok` in scope ... It all hinges ...

MISRA rule 5.3 (emphasis mine):
> An identifier declared in an inner scope shall be distinct from any identifier declared in an outer scope.
> [...]
> Rationale
> If an identifier is declared in an inner scope but is not distinct from an identifier *that already exists* in an outer scope, then the inner-most declaration will "hide" the outer one. This may lead to developer confusion.

Great, so obviously the second `tnow` is not in the scope yet, and is not hiding the inner. Easy-peasy. The MISRA checker has a false positive. Except ...

C99 6.2.1 
> For each different entity that an identifier designates, the identifier is visible (i.e., can be used) only within a region of program text called its scope.

C99 6.5.1
> An identifier is a primary expression, provided it has been declared as designating an object (in which case it is an lvalue) or a function (in which case it is a function designator). (footnote 79)
> (footnote 79) Thus, an undeclared identifier is a violation of the syntax.

The crazy thing is that without the footnote I would say the second `tnow` is not in scope (yet), and with the footnote I read that the C specification leaves it undefined whether that the second `tnow` is already in the outer scope but that there is no valid *syntax* for referring to it because it has not been declared yet!

Why would it say "no valid *syntax*"? There are cases where you need to know the declared type in order to parse correctly, so that does make sense, weird though it is. (When in doubt, the C standard is more likely to be written by people that know what they are doing than the MISRA spec.)

I am convinced that the MISRA people forgot to actually reconsider what they wrote when they slapped the "C99" label on it, because in C89 this particular problem doesn't exist. A slightly more careful wording on their side would have stated it in terms of declaring an identifier in an inner scope when said identifier has already been declared in an outer scope: then the confusion is gone, because it avoids that ambiguity in the C spec about when exactly an identifier exists in a scope. With that it would have been clear that this code is perfectly acceptable because nothing is being hidden. But hey, no need to be precise when dealing with software ...

But fear not! I can't help noticing that this bit can trivially moved into a separate function, which I like better than leaving it as was, and that should also satisfy the MISRA checker.

```suggestion
  if (initmaxwait > 0 && !wait_for_initial_matches())
    goto err_minmatch_wait;
```
in combination with:
```C
static bool wait_for_initial_matches (void)
{
  dds_time_t tnow = dds_time ();
  const dds_time_t tendwait = tnow + (dds_duration_t) (initmaxwait * 1e9);
  ddsrt_mutex_lock (&disc_lock);
  while (matchcount < minmatch && tnow < tendwait)
  {
    ddsrt_mutex_unlock (&disc_lock);
    dds_sleepfor (DDS_MSECS (100));
    ddsrt_mutex_lock (&disc_lock);
    tnow = dds_time ();
  }
  const bool ok = (matchcount >= minmatch);
  if (!ok)
  {
    /* set minmatch to an impossible value to avoid a match occurring between now and
       the determining of the exit status from causing a successful return */
    minmatch = UINT32_MAX;
  }
  ddsrt_mutex_unlock (&disc_lock);
  if (!ok)
    return false;
  dds_sleepfor (DDS_MSECS (100));
  return true;
}
```
should do the trick as well, but in a way I can approve of 🙂

eboasson (commented):
@Splinter1984 thanks! To me all these changes are to deal with a non-issue as far as I am concerned, but for the vast majority of them, I couldn't care less and have no problem merging them. *One* of them, however ... I think you'll be ok with my counter proposal, and if so, we have an elegant way out of the rule 5.3 morass.

src/tools/ddsperf/ddsperf.c
Splinter1984: Thanks @eboasson for such a detailed answer! As you can see, I accepted your changes and indeed MISRA let me go 😅.

eboasson (approved):

PR #1747, opened 2023-07-03
Fix MISRA rule 22-8.

Splinter1984:
This PR addresses rule 22.8.
```
Properly use errno value 
```


src/ddsrt/src/strtod.c
```diff
@@ -102,12 +102,11 @@ ddsrt_strtod(const char *nptr, char **endptr, double *dblptr)
...
 
   assert(nptr != NULL);
-  assert(dblptr != NULL);
-
-  orig_errno = errno;
-  errno = 0;
+  assert(dblptr != NULL); 
 
   if (os_lcNumericGet() == '.') {
+    orig_errno = errno;
```
eboasson, line 108: I don't understand why this got moved: the only difference is that whatever `os_lcNumericGet` did to `errno` is now restored on a successful call to `ddsrt_strtod`, but why is that desirable?

Splinter1984:
```diff
diff --git a/src/ddsrt/src/strtod.c b/src/ddsrt/src/strtod.c
index 559d39ef..63b957b4 100644
--- a/src/ddsrt/src/strtod.c
+++ b/src/ddsrt/src/strtod.c
@@ -104,8 +104,8 @@ ddsrt_strtod(const char *nptr, char **endptr, double *dblptr)
   assert(nptr != NULL);
   assert(dblptr != NULL); 
 
+  orig_errno = errno;
   if (os_lcNumericGet() == '.') {
-    orig_errno = errno;
     errno = 0;
     /* The current locale uses '.', so strtod can be used as is. */
     dbl = strtod(nptr, endptr);
@@ -139,7 +139,6 @@ ddsrt_strtod(const char *nptr, char **endptr, double *dblptr)
 
     /* Now that we have a copy with the proper locale LC_NUMERIC, we can use
        strtod() for the conversion. */
-    orig_errno = errno;
     errno = 0;
     dbl = strtod(nptrCopy, &nptrCopyEnd);
```
In this case we can move `origin_errno` back and keep it store prev `errno`.

eboasson (approved):

PR #1754, opened 2023-07-06
Delete .travis.yml

mosfet80:
some configuration using ubuntu1804   and gcc7, both unused


eboasson (approved):
We have left Travis CI behind, so yes, better to remove it.

Here, too, the (Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) needs to be on file. If you would be so kind as to fill it out, that'd be great.

PR #1756, opened 2023-07-06
Do not assert on encoding in serdata_pserop_new

eboasson:
This function is called with data from a network packet for the participant message data, and the generic receiver code currently only filters on any of the defined encodings, not on the encodings supported by a specific sertype.


dpotman (approved):
LGTM

PR #1757, opened 2023-07-06
Fix MISRA-RULE 11.1

Splinter1984:
This addresses rule 11.1.

> Conversions shall not be performed between non compatible pointer to a function types 

A bit of a controversial rule in relation to us. But something needs to be done about it.
There are also triggers on several controversial points.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/fd1d979e4a8ff45d63a2c505b7ec28810bd1d573/src/core/ddsc/src/dds_listener.c#L93-L117
https://github.com/eclipse-cyclonedds/cyclonedds/blob/fd1d979e4a8ff45d63a2c505b7ec28810bd1d573/src/tools/ddsperf/ddsperf.c#L2659-L2660
With exclude of the lines above, we have solved this rule. 👾


src/ddsrt/include/dds/ddsrt/avl.h
```diff
@@ -200,7 +200,7 @@ typedef void (*ddsrt_avl_augment_t) (void *vnode, const void *left, const void *
 /**
  * @brief User defined walk function to perform an action on each node in the tree.
  */
-typedef void (*ddsrt_avl_walk_t) (void *vnode, void *arg);
+typedef void (*ddsrt_avl_walk_t) (const void *vnode, void *arg);
```
eboasson, line 203: This one is incorrect: the point of `ddsrt_avl_cwalk_t` defined immediately below is that there are two versions, one which is `const` and one which is not (and thereby allows the callback function to modify the contents within reason).

src/core/ddsi/tests/sysdeps.c
```diff
@@ -162,7 +162,7 @@ static ddsrt_avl_tree_t helper_tree = {
   &helper_node.avlnode
 };
 
-static void log_stacktrace_helper (void *vnode, void *varg)
+static void log_stacktrace_helper (const void *vnode, void *varg)
```
eboasson, line 165: If you make this one `const` (I'm in favour of that change) then `log_stacktrace_thread` should use `ddsrt_avl_const_walk`

eboasson (commented):
Thanks @Splinter1984 !

I suspect the problematic ones you mentioned are caused by `void (*dst)(void)` and similar, right? I can see how MISRA would be upset by someone making use of:

> A pointer to a function of one type may be converted to a pointer to a function of another type and back again; the result shall compare equal to the original pointer. If a converted pointer is used to call a function whose type is not compatible with the pointed-to type, the behavior is undefined.

(C99, 6.3.2.3, para 8) They are, after all, already upset by void pointers.

Splinter1984:
Yeah this was the reason why i called this violations controversial. Looks like MISRA just cut (C99, 6.3.2.3, para 8)  spec without deeply understand of all sentences. 💁

eboasson (approved):
LGTM, with apologies for the delay ☺️

PR #1761, opened 2023-07-11
Fix type object validation for types with no members/labels

dpotman:
This fixes an issues in the type object validation, for the case that a type object is received for an aggregated or enumerated type with no members/elements.


eboasson (approved):
Thanks @dpotman !

PR #1763, opened 2023-07-12
Work-around for CI failure on CUnit

dpotman:
This is a workaround for the CI builds failing on installation of CUnit (via Conan). I've tried several other solutions, e.g. adding the `with_curses` option as suggested in the error message, but that results in build failures of ncurses on some of the builds. Switching cunit to use it as a static library is also not working, that results in linker errors and conflicting symbols in certain builds. So switching back to a previous version of the conan recipe for cmake seems to be the best option for now.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -106,7 +106,8 @@ steps:
       set -e -x
       mkdir build
       cd build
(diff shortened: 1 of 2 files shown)


dpotman:
This worked fine on my fork, but apparently it doesn't work here, so I'll close this PR, we'll have to find a different solution.

PR #1765, opened 2023-07-12
Fix MISRA rule 22-1.

Splinter1984:
This addresses rule 22-1.
> Ensure resources are freed.


eboasson (approved):
Thanks @Splinter1984!

PR #1767, opened 2023-07-15
Eliminate Conan, CUnit

### LINKED ISSUE #1678: Minor Grammar Mistakes in README
# Description
Hello! Just suggesting a couple minor grammar changes to the [README.md](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/README.md) file. 

Initially, I cloned the repository, made a branch and tried to open a pull request with the proposed changes.... then realised there were more steps when looking through the [CONTRIBUTING](https://github.com/eclipse-cyclonedds/cyclonedds/blob/53a92b1e6643d9818b741898b95e698aba5016d9/CONTRIBUTING.md) file. 

Instead of going through the contributor process, perhaps someone can throw these changes into an open pull request? If I have larger suggestions, perhaps I'll take the plunge! Thanks.

## Change 1
Screenshot 2023-05-01 at 18 08 18

## Change 2
Screenshot 2023-05-01 at 18 08 25

eboasson:
We have been using Conan to get the prerequisites in the CI, but that's become problematic: Conan 2.0 dropped the model for integrating with CMake builds that used to be the way in Conan 1.x, but the new models assume a tighter integration with Conan than I think is wise for Cyclone. So Conan has to go at some point.

The only prerequisites for which we used Conan were CUnit and OpenSSL. Azure has OpenSSL pre-installed, so we might as well use that. That leaves only CUnit, which is easy to get on macOS and Ubuntu (via homebrew and apt), but turns out to be hard on Windows without using Conan, and *that* recipe got broken recently. We have simply been lucky because of some caching on Azure ...

Building CUnit from the Azure build scripting is feasible and takes only a few seconds, but the problem with that is finding a good set of sources. The old CUnit sources depend on GNU autoconf/automake, which Cyclone should not start depending on, and a more recently updated fork that uses CMake only supports static builds. And as it happens, a statically linked CUnit breaks the security tests in Windows because some of the assertions are in a DLL ...

While we could probably rewrite the problematic tests, it turns out we use such a tiny bit of CUnit that it takes a mere 600 lines of C code+headers to replace it altogether. That way, we don't have to rewrite the tests *and* we drop the dependency.

Incidentally fixes #1678.


src/ddsrt/tests/log.c
```diff
@@ -315,6 +316,7 @@ CU_Test(dds_log, newline_terminates, .fini=reset)
   CU_ASSERT_PTR_NULL_FATAL(msg);
   DDS_ERROR("baz\n");
   CU_ASSERT_PTR_NOT_NULL_FATAL(msg);
+  assert(msg);
```
poetinger, line 319: Why did you add these asserts, but remove them elsewhere (e.g.  string.c)

poetinger (approved):

PR #1742, opened 2023-06-28
Fix MISRA rule 8-3(1/2) 'declarations of a function same name and type'

Splinter1984:
This PR addresses rule 8.3.
```
The identifiers used in the declaration and definition of a function shall be identical.
```

PR #1770, opened 2023-07-17
Update Zephyr port to support Zephyr V3.4

PatrickM-ZS:
Updates related to changes introduced in Zephyr v3.4:
- Change API for maddr join/leave functions (zephyrproject-rtos/zephyr@ff0fd2d7a7)
- Change main function return type from `void` to `int` (zephyrproject-rtos/zephyr#54628)
- Allow the use of SCHED_OTHER class and verify thread priorities are inside min/max range (zephyrproject-rtos/zephyr#57161).
- Fix issue caused by the refactoring in #1742
- Add a device-tree overlay for s32z270dc2_rtu0_r52 board to increase SRAM size without needing to change any Zephyr defaults

Besides these, added some clarifications in the README.md based on user feedback.


ports/zephyr/README.md
```diff
@@ -4,16 +4,20 @@
 This directory contains some proof-of-concept applications that show how to build and use CycloneDDS on [Zephyr RTOS](https://www.zephyrproject.org) for a NXP X-S32Z27X-DC board.
 
 
-Getting Started with Zephyr information can be found [here](https://docs.zephyrproject.org/3.3.0/develop/getting_started/index.html)
+Getting Started with Zephyr information can be found [here](https://docs.zephyrproject.org/latest/develop/getting_started/index.html)
```
ambroise-arm, line 7: Shouldn't we point to the version of the Zephyr documentation that is currently supported instead of `latest`? It will need to be updated on further version updates, but it ensures the link doesn't break and that the external information doesn't get out of date compared to this document. (same for the other links in the PR)

ports/zephyr/README.md
```diff
@@ -4,16 +4,20 @@
...
+Documentation for the NXP X-S32Z27X-DC board can be found [here](https://docs.zephyrproject.org/latest/boards/arm/s32z270dc2_r52/doc/index.html)
 
+## Usage:
```
ambroise-arm, line 10: There is usually no punctuation in titles

ports/zephyr/README.md
```diff
@@ -4,16 +4,20 @@
...
+The applications can be treated similar to other Zephyr sample applications.
```
ambroise-arm, line 11: > can be treated similar to

"similarly to"?

ports/zephyr/README.md
```diff
@@ -4,16 +4,20 @@
...
+The `CMakeLists.txt` is currently set up to build some of the CycloneDDS examples or ddsperf but can be easily modified or extended for other application code.
 
-Documentation for the NXP X-S32Z27X-DC board can be found [here](https://docs.zephyrproject.org/3.3.0/boards/arm/s32z270dc2_r52/doc/index.html)
+:warning: While CycloneDDS can be built in-tree, it does not support multiple different builds. Therefore the CycloneDDS source directory needs to be clean in order to build for Zephyr and must not contain an in-tree build for eg. the host system. Alternatively, this directory with Zephyr examples can be copied outside the CycloneDDS source directory but that requires updating the `ExternalProject_Add` directive in `CMakeLists.txt` to point to the CycloneDDS source directory.
```
ambroise-arm, line 14: > needs to be clean

"cleaned"?

ambroise-arm (changes requested):

ports/zephyr/README.md
PatrickM-ZS: I see your point but linking to an 'old' version also implies newer versions are not supported, while they may just work as-is. We intent for latest CycloneDDS version to always work with latest Zephyr version.

ports/zephyr/README.md
PatrickM-ZS: Fixed!

ports/zephyr/README.md
PatrickM-ZS: Fixed!

ports/zephyr/README.md
PatrickM-ZS: Fixed!

ports/zephyr/README.md
eboasson: If I understand correctly, details in Zephyr change often enough that even a minor version change can already affect the port. So I have to agree with @ambroise-arm here, it makes sense to link to the latest version of Zephyr for which the port has been validated. I don't see a problem with (also) noting that it is likely to work on later versions, or something of the sort.

Having a versioned (and stable) documentation link is fine, but I think the version number should be noted at the top-level of the readme, so more like:

> This directory contains some proof-of-concept applications that show how to build and use CycloneDDS on [Zephyr RTOS](https://www.zephyrproject.org) for a NXP X-S32Z27X-DC board. We aim for compatibility with the latest release of Zephyr, at the time of writing this is version 3.3.

(Or something along those lines)

src/ddsrt/src/threads/posix/threads.c
```diff
@@ -348,6 +348,17 @@ ddsrt_thread_create (
...
+    if ((result = pthread_attr_getschedparam(&pattr, &sched_param)) != 0)
+    {
+      DDS_ERROR("ddsrt_thread_create(%s): pthread_attr_getschedparam() failed with error %d\n", name, result);
+      goto err;
+    }
+    if ((result = pthread_attr_getschedpolicy(&pattr, &policy)) != 0)
+    {
+      DDS_ERROR("ddsrt_thread_create(%s): pthread_attr_getschedpolicy() failed with error %d\n", name, result);
+      goto err;
+    }
```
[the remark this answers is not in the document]
eboasson: > Would it make sense to remove the bit that makes use of pthread_self() and use the new implementation in all cases as part of this PR?

I don't think so: if I am reading it correctly, that would be a significant change in default behaviour for other POSIX platforms and hence would better be done (and debated) in a separate PR.

eboasson (approved):
LGTM, thanks @PatrickM-ZS

PR #1775, opened 2023-07-19
Fix type wrapper memory leak for custom annotations

dpotman:
Fix leaking memory in the xtypes type wrapper for custom member annotations, which Cyclone doesn't fully support, but could be present in a type object received from another implementation.


eboasson (approved):
LGTM, thanks @dpotman !

PR #1776, opened 2023-07-19
Support for additional types in dynsub example

dpotman:
This adds support for additional data types in the `dynsub` example (thanks @eboasson for most of the code). Additionally, the example is split into multiple files to improve readability a bit.


eboasson (approved):
Thanks @dpotman for taking that quick hack of mine and making it somewhat presentable. Both support for some more types and printing of the type object will come in handy in the future, I suspect.

PR #1778, opened 2023-07-21
Some backports to 0.10.x

eboasson:
This backports
* #1773
* #1717 
* #1660

As well as #1767 to get the CI back to life (though that bit is not quite done yet, but at least no longer Conan-related)

It also bumps the version number to 0.10.4, I expect that doing a further patch release would appreciated.


trittsv:
Please see my comment in https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/433
which applies for this PR too 😃 
In short, our unit tests run successfully, pending events seems to be fixed and compiles for a loooot of platforms.

PR #1779, opened 2023-07-21
Unittest fixes

reicheratwork:
Fixes for a number of unittests that were either not doing what they said they were or were checking for the wrong things.

### DIFF
--- a/src/core/ddsc/tests/qosmatch.c
+++ b/src/core/ddsc/tests/qosmatch.c
@@ -269,12 +269,12 @@ static uint32_t pub_thread (void *varg)
     {
       for (size_t j = 0; j < NWR_PUB; j++)
       {
-        if (chk[i][j])
-          continue;
         dds_instance_handle_t ih;
         dds_builtintopic_endpoint_t *ep;
         rc = dds_get_matched_subscriptions (wr[i][j], &ih, 1);
         CU_ASSERT_FATAL (rc == 0 || rc == 1);
+        if (chk[i][j])
+          continue;
         if (rc == 1)
         {
           ep = dds_get_matched_subscription_data (wr[i][j], ih);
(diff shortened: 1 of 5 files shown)


src/core/ddsc/tests/qosmatch.c
```diff
@@ -269,12 +269,12 @@ static uint32_t pub_thread (void *varg)
     {
       for (size_t j = 0; j < NWR_PUB; j++)
       {
-        if (chk[i][j])
-          continue;
         dds_instance_handle_t ih;
         dds_builtintopic_endpoint_t *ep;
         rc = dds_get_matched_subscriptions (wr[i][j], &ih, 1);
         CU_ASSERT_FATAL (rc == 0 || rc == 1);
+        if (chk[i][j])
```
eboasson, line 276: I don't understand what this fixes, the way I read it, the only difference it makes is that after moving this test, `dds_get_matched_subscription` gets called a few more times. That's not doing any harm, but it is also not changing anything.

src/core/ddsc/tests/qosmatch.c
```diff
@@ -411,12 +411,12 @@ static uint32_t sub_thread (void *varg)
     {
       for (size_t j = 0; j < NWR_PUB; j++)
       {
-        if (chk[i][j])
-          continue;
         dds_instance_handle_t ih;
         dds_builtintopic_endpoint_t *ep;
         rc = dds_get_matched_publications (rd[i][j], &ih, 1);
         CU_ASSERT_FATAL (rc == 0 || rc == 1);
+        if (chk[i][j])
```
eboasson, line 418: Same here

src/core/ddsc/tests/qosmatch.c
reicheratwork: This was more of an optimization rather than a fix for something explictly being done wrong.

eboasson: I don't get it, you're trading a call to `dds_get_matched_subscriptions` with loading a boolean from a small 2-dimensional array. Surely one would expect the latter to be cheaper than the former?

At the same time, as there's serious underlying issue, I really don't think it is a very interesting subject. Merging it is by far the easiest alternative 🙂

Issue #1772, opened 2023-07-17
Denial of Service due to an invalid locator in DATA submessage


squizz617:
Hi,

After the initial PDP handshake, the **second** PDP message that contains an invalid `METATRAFFIC_UNICAST_LOCATOR`, i.e., `10.0.0.15:0`, seems to cause a denial of service.

* DATA submessage:
```
0000   15 05 00 00 00 00 10 00 00 01 00 c7 00 01 00 c2
0010   00 00 00 00 01 00 00 00 00 03 00 00 15 00 04 00
0020   02 02 00 00 16 00 04 00 01 0f 00 00 50 00 10 00
0030   01 0f 11 3e f6 42 cd 90 00 00 00 00 00 00 01 c1
0040   32 00 18 00 01 00 00 00 00 00 00 00 00 00 00 00
0050   00 00 00 00 00 00 00 00 0a 00 00 0f 31 00 18 00
0060   01 00 00 00 f3 1c 00 00 00 00 00 00 00 00 00 00
0070   00 00 00 00 0a 00 00 0f 02 00 08 00 14 00 00 00
0080   00 00 00 00 58 00 04 00 3f 0c 00 00 62 00 10 00
0090   0a 00 00 00 70 75 62 6c 69 73 68 65 72 00 00 00
00a0   59 00 28 00 01 00 00 00 11 00 00 00 50 41 52 54
00b0   49 43 49 50 41 4e 54 5f 54 59 50 45 00 00 00 00
00c0   07 00 00 00 53 49 4d 50 4c 45 00 00 01 00 00 00
```

Invalid locator param (bytes 0x40 - 0x5b):
```
0000   32 00 18 00 01 00 00 00 00 00 00 00 00 00 00 00
0010   00 00 00 00 00 00 00 00 0a 00 00 0f
```
Wireshark dissection of the above param:
```
PID_METATRAFFIC_UNICAST_LOCATOR (LOCATOR_KIND_UDPV4, 10.0.0.15:0)
    parameterId: PID_METATRAFFIC_UNICAST_LOCATOR (0x0032)
    parameterLength: 24
    locator
        Kind: LOCATOR_KIND_UDPV4 (0x00000001)
        Port: 0
            [Expert Info (Warning/Protocol): Invalid Port]
        Address: 10.0.0.15
```

* Upon the receipt of the DATA submessage above, CycloneDDS client keeps printing the following error messages to stderr:
```
[...]

PR #1780, opened 2023-07-25
Reject locators with port = 0

eboasson:
This PR prevents Cyclone from accepting locators with port = 0 and so should take care of the error messages reported in #1772. The other problems caused by accepting incorrect locators can't be fixed so easily.

### DIFF
--- a/src/core/ddsi/src/ddsi__tran.h
+++ b/src/core/ddsi/src/ddsi__tran.h
@@ -252,7 +252,7 @@ inline dds_return_t ddsi_factory_create_conn (struct ddsi_tran_conn **conn, stru
   *conn = NULL;
   if ((qos->m_interface != NULL) != (qos->m_purpose == DDSI_TRAN_QOS_XMIT_UC || qos->m_purpose == DDSI_TRAN_QOS_XMIT_MC))
     return DDS_RETCODE_BAD_PARAMETER;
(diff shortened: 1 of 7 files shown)


src/core/ddsi/src/ddsi__tran.h
```diff
@@ -252,15 +252,15 @@ inline dds_return_t ddsi_factory_create_conn (struct ddsi_tran_conn **conn, stru
   *conn = NULL;
   if ((qos->m_interface != NULL) != (qos->m_purpose == DDSI_TRAN_QOS_XMIT_UC || qos->m_purpose == DDSI_TRAN_QOS_XMIT_MC))
     return DDS_RETCODE_BAD_PARAMETER;
-  if (!ddsi_is_valid_port (factory, port))
+  if (port != 0 && !ddsi_is_valid_port (factory, port))
```
dpotman, line 255: This condition (and the ones below) confused me a bit, because port number `0` was rejected earlier in the check in `do_locator`. Because at this point the `0` has a special meaning for the underlying implementation, I think it would be more clear to replace it with a #define with a descriptive name.

dpotman (commented):
Looks good, only a detail on the meaning of port number `0`, see comment below.

dpotman (approved):

PR #1783, opened 2023-07-27
Add ENABLE_TYPELIB build option

dpotman:
This adds the `ENABLE_TYPELIB` build option (default is `ON`) that controls including the type library (which stores the type meta-data for types used in local entities and in discovered proxy entities). The `ENABLE_TYPELIB` option also controls the availability of the type-builder (to construct the type meta-data used by the serialized, e.g. used from the Python API) and the dynamic type API.

Most of the functionality that can be disabled/enabled with this option was previously controlled by the `ENABLE_TYPE_DISCOVERY` build option. That option is also still available, but now only controls the inclusion of the type lookup functionality (that allows resolving type meta-data for types from discovery endpoints or topics).

The commit in this PR also include splitting the XTypes tests into multiple files, because that fits better with the new build options.


eboasson (commented):
LGTM 😀 It does seem that CodeQL is not quite happy yet.

dpotman:
These CodeQL issues were fixed with the force-push to [ecb6992](https://github.com/eclipse-cyclonedds/cyclonedds/commit/ecb6992793150554f73302cb948f1916e370a0ce), but for some reason it seems that the new CodeQL run has used outdated header files for doing the checks.

eboasson:
With my eyes I can't see what CodeQL sees. If CodeQL still complains after merging, we'll have another look.

PR #1787, opened 2023-08-01
Export typebuilder and cdrstream functions

dpotman:
Export some additional typebuilder and cdrstream functions that are used in python binding


eboasson (approved):
Thanks @dpotman ☺️

PR #1788, opened 2023-08-01
Assorted minor fixes

noxpardalis:
These are a set of assorted minor fixes primarily to reduce some checks in other functions guarding against impossible states.

I'm not certain about my (`strtof`, `strtod`) fixes because I'm not sure why the original `ddsrt_strtod` guarded against `HUGE_VALF` and `HUGE_VALL` and not `HUGE_VAL`?

### DIFF
--- a/src/ddsrt/src/hopscotch.c
+++ b/src/ddsrt/src/hopscotch.c
@@ -313,7 +313,7 @@ static bool ddsrt_chh_data_valid_p (void *data)
     return data != NULL && data != CHH_BUSY;
 }
 
-static int ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
+static dds_return_t ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
 {
     uint32_t size;
     uint32_t i;
@@ -328,7 +328,10 @@ static int ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_ha
     rt->gc_buckets = gc_buckets;
     rt->gc_buckets_arg = gc_buckets_arg;
 
-    buckets = ddsrt_malloc (offsetof (struct ddsrt_chh_bucket_array, bs) + size * sizeof (*buckets->bs));
+    buckets = ddsrt_malloc_s (offsetof (struct ddsrt_chh_bucket_array, bs) + size * sizeof (*buckets->bs));
+    if (buckets == NULL) {
+      return DDS_RETCODE_OUT_OF_RESOURCES;
+    }
     ddsrt_atomic_stvoidp (&rt->buckets, buckets);
     buckets->size = size;
     for (i = 0; i < size; i++) {
@@ -338,7 +341,7 @@ static int ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_ha
         ddsrt_atomic_stvoidp (&b->data, NULL);
     }
     ddsrt_mutex_init (&rt->change_lock);
-    return 0;
+    return DDS_RETCODE_OK;
 }
 
 static void ddsrt_chh_fini (struct ddsrt_chh *rt)
@@ -349,8 +352,8 @@ static void ddsrt_chh_fini (struct ddsrt_chh *rt)
 
 struct ddsrt_chh *ddsrt_chh_new (uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
 {
-    struct ddsrt_chh *hh = ddsrt_malloc (sizeof (*hh));
-    if (ddsrt_chh_init (hh, init_size, hash, equals, gc_buckets, gc_buckets_arg) < 0) {
+    struct ddsrt_chh *hh = ddsrt_malloc_s (sizeof (*hh));
+    if (hh == NULL || ddsrt_chh_init (hh, init_size, hash, equals, gc_buckets, gc_buckets_arg) < 0) {
         ddsrt_free (hh);
         return NULL;
     } else {
--- a/src/ddsrt/src/hopscotch.c
+++ b/src/ddsrt/src/hopscotch.c
@@ -313,7 +313,7 @@ static bool ddsrt_chh_data_valid_p (void *data)
     return data != NULL && data != CHH_BUSY;
 }
 
-static int ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
+static void ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
 {
     uint32_t size;
     uint32_t i;
@@ -338,7 +338,6 @@ static int ddsrt_chh_init (struct ddsrt_chh *rt, uint32_t init_size, ddsrt_hh_ha
         ddsrt_atomic_stvoidp (&b->data, NULL);
     }
     ddsrt_mutex_init (&rt->change_lock);
-    return 0;
 }
 
 static void ddsrt_chh_fini (struct ddsrt_chh *rt)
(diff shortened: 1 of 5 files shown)


src/ddsrt/src/hopscotch.c
```diff
@@ -350,12 +349,8 @@ static void ddsrt_chh_fini (struct ddsrt_chh *rt)
 struct ddsrt_chh *ddsrt_chh_new (uint32_t init_size, ddsrt_hh_hash_fn hash, ddsrt_hh_equals_fn equals, ddsrt_hh_buckets_gc_fn gc_buckets, void *gc_buckets_arg)
 {
     struct ddsrt_chh *hh = ddsrt_malloc (sizeof (*hh));
-    if (ddsrt_chh_init (hh, init_size, hash, equals, gc_buckets, gc_buckets_arg) < 0) {
```
eboasson, line 353: I'm always a little sad when correct out-of-memory handling code gets dumped. Then I realise it's probably never going to get executed and so is just unhelpful ballast in the code ... At which point I consider that it really can happen in practice, especially on embedded machines and so that we should handle it correctly.

But then, doing it correctly for the entirety of DDS is a rather tricky thing because of the consistency that should be maintained between the different data readers (what if reader 1 gets the data and the application reads it, but it runs out of memory before it becomes visible in reader 2 in the same process?) and without wanting to incur overhead for this case that really never happens in practice ...

And so I keep switching sides. I know what I'd like: do it properly all throughout the code base, guaranteeing that it correctly handles any out-of-memory condition by aborting the operation without any observable state change and while maintaining consistency (save for some logging and statistics).

And failing that, giving up on it altogether, keeping the code free of error paths and just crashing on out-of-memory ...

This code happened to be some code that did handle it correctly, it is just that `ddsrt_malloc` aborts on out-of-memory instead of returning a null pointer (because, well, see above). I think I'd like to kick the can down the road a little longer and use `ddsrt_malloc_s` (a terribly named function if ever there was any) instead of `ddsrt_malloc`: then the error handling code is no longer dead code.

src/ddsrt/src/strtod.c
```diff
@@ -148,7 +148,7 @@ ddsrt_strtod(const char *nptr, char **endptr, double *dblptr)
     }
   }
 
-  if ((dbl == HUGE_VALF || dbl == HUGE_VALL || dbl == 0) && errno == ERANGE) {
+  if ((dbl == HUGE_VAL || dbl == 0) && errno == ERANGE) {
```
eboasson, line 163: I agree that `HUGE_VALF` and `HUGE_VALL` must have been put in because of a misreading of the spec, so removing those is definitely an imrpovement. There are other things amiss, though ...

Firstly, another misreading is the sign, because it states:
> 7.20.1.3:
> (¶10) If the correct value is outside the range of representable values, plus or minus `HUGE_VAL`, `HUGE_VALF`, or `HUGE_VALL` is returned (according to the return type and sign of the value),

Secondly, the way I read the spec, in particular the continuation of the sentence above:
> and the value of the macro ERANGE is stored in errno

the return of ±`HUGE_VAL` is simply the most reasonable return value, and one that the application code may be able to work with if it doesn't check for an overflow, but for detecting an overflow it only `errno` matters:
```C
errno=0;
dbl=strtod(nptr,&endptr);
if(errno==ERANGE) {
  // overflow occurred
  assert(dbl == HUGE_VAL || dbl == -HUGE_VAL);
}
```

From the spec there's another error case that seems to be handled incorrectly by this code:
> 7.20.1.3:
> (¶7) If the subject sequence is empty or does not have the expected form, no conversion is performed; the value of nptr is stored in the object pointed to by endptr, provided that endptr is not a null pointer.
> (¶10 If no conversion could be performed, zero is returned.

Obviously, zero is a completely expected return value, so I the way I read it, `endptr==nptr || errno==ERANGE` is the sufficient and necessary condition for detecting an invalid conversion.

The cases of underflow I think we can ignore, and for us it probably makes sense to reject NaNs. I checked the current uses and they are few: `ddsrt_strtof` (itself unused) and something in my crazy little [oneliner](https://github.com/eclipse-cyclonedds/cyclonedds/blob/8638e1fa1e8ae9f964c91ad4763653702b8f91e0/src/core/ddsc/tests/test_oneliner.c#L398). Maybe we should use it instead of `sscanf` more often, but I don't think we'll ever have to worry about underflows, and NaNs would probably only cause trouble, so rejecting them (by default, if need be) makes sense.

If we make the effort of fixing some part of it, let's really fix it and make sure the documentation is correct 😀

src/ddsrt/src/xmlparser.c
```diff
@@ -377,22 +377,27 @@ static void discard_payload (struct ddsrt_xmlp_state *st)
   st->tpescp = 0;
 }
 
-static int append_to_payload (struct ddsrt_xmlp_state *st, int c, int islit)
+static void append_literal_to_payload (struct ddsrt_xmlp_state *st, int c)
```
eboasson, line 380: I suppose, for the changes in this file the only issue I have is the same as for the proposed change in `hopscotch.c`.

src/ddsrt/src/xmlparser.c
```diff
@@ -377,22 +377,27 @@ static void discard_payload (struct ddsrt_xmlp_state *st)
...
+    size_t n = st->tpp - st->tpescp;
+    if (unescape_insitu (st->tp + st->tpescp, &n) < 0) {
+      discard_payload (st);
+      return -1;
     }
-    st->tp[st->tpp++] = (char) c;
-    st->tpescp = st->tpp;
+    st->tpp = st->tpescp + n;
   }
+  st->tp[st->tpp++] = (char) c;
```
eboasson, line 399: For this I would rely on (the proposed new) `append_literal_to_payload`. The only thing is to remember to retain the update of `st->tpescp` 🙂

src/ddsrt/src/strtod.c
```diff
@@ -174,18 +174,44 @@ ddsrt_strtod(const char *nptr, char **endptr, double *dblptr)
 dds_return_t
 ddsrt_strtof(const char *nptr, char **endptr, float *fltptr)
 {
-  /* Just use os_strtod(). */
```
eboasson, line 165: Even more convenient might be to delete it altogether!

eboasson (commented):
Thanks @noxpardalis! You have to look past the amount of commentary to see how close you are to a LGTM review 😀 I'm also curious to know what your view is on the matters addressed in my comments.

src/ddsrt/src/hopscotch.c
noxpardalis: I agree on this point. My preference is also be to maintain the out-of-memory code. It makes sense to me to operate assuming that allocation of resources may fail, while de-allocation of resources must not.

I've reverted my change and then swapped to using `ddsrt_malloc_s` in 979f8d0.

The question that remains is whether the use of `ddsrt_malloc_s` should be promoted wherever possible in the rest of the code-base.

src/ddsrt/src/strtod.c
noxpardalis: I wanted to clarify if you're suggesting that I just remove `ddsrt_strtof` altogether. My concern is that it would be a breaking change to remove it from the public API of `ddsrt`, but I suppose it makes some sense as Cyclone isn't actually using it for anything.

src/ddsrt/src/xmlparser.c
noxpardalis: Yup, I've swapped to using `append_literal_to_payload` in `append_to_payload`. I thought about doing this before but it slipped my mind (or perhaps I was just lazy :stuck_out_tongue_closed_eyes:). Either way thanks for catching this.

src/ddsrt/src/xmlparser.c
noxpardalis: I'll swap this out to using a `dds_retcode_t` and `ddsrt_realloc_s`. Before I make that change would you like me to revert the changes so that `append_to_payload` and `append_literal_to_payload` are unified again in `append_to_payload`?

src/ddsrt/src/strtod.c
eboasson: I don't consider `ddsrt` a public interface, so yes, I was indeed suggesting removing `ddsrt_strtof` altogether.

src/ddsrt/src/xmlparser.c
eboasson: No, there is something to be said for having two different functions there, especially with `append_to_payload` defined in terms of `append_literal_to_payload`

src/ddsrt/src/strtod.c
noxpardalis: My reading of `strtod`'s spec was the same as yours and so I've changed it to using `errno==ERANGE` and essentially `endptr==nptr` to check that the parse was good. I also added an `isnan` and an `isinf` check as a condition for resulting in a `DDS_RETCODE_OUT_OF_RANGE` + some documentation to that effect.

src/ddsrt/src/strtod.c
noxpardalis: Okay, I removed `ddsrt_strtof` completely.

src/ddsrt/src/xmlparser.c
noxpardalis: Okay, I kept the two functions separate and made `append_to_payload` return an error in case the reallocation fails (+reintroduced the previous error checking).

noxpardalis:
@eboasson I think I fixed up everything mentioned in the comments. Would you mind doing a re-review of the changes? Thanks!

eboasson (commented):
There is this one off-by-one error ... (a statement which itself probably has at least one off-by-at-least-one error, but that'd be a single-sided error 😂)

It was rather tricky and somewhat confusingly named code to begin with, so it is entirely understandable.

Fixing the error can be done in a trivial way, but I think it would be worth considering the background information I added and possibly renaming something and/or adding some comments, perhaps derived from my review comment.

eboasson (approved):
Thank you @noxpardalis ☺️

PR #1792, opened 2023-08-03
Fix duplicate condition in validate_remote_identity

dpotman:
Fix duplicate condition in `validate_remote_identity` test and add some missing copyright headers in security plugin tests


eboasson (approved):
Thanks @dpotman !

PR #1793, opened 2023-08-03
Cdrstream fix for extracting key from delimited type

dpotman:
This fixes skipping additional data in a delimited type when extracting a key from data using the cdrstream serializer.


eboasson (approved):
That's a nasty little one, thanks for finding it and fixing it!

I should say something about there not being a test case but I'm not so worried about a regression. What does have me worried is what other edge case we may have overlooked. I suspect you found it thanks to https://github.com/eclipse-cyclonedds/cyclonedds-python/blob/master/tests/test_fuzzy.py and it would probably be wiser to spend the effort on increasing the likelihood of finding things quickly through fuzzing than on trying to do it by hand.

PR #1797, opened 2023-08-08
Add CI build for Python and C++ bindings

dpotman:
This PR adds CI builds for the Python and C++ bindings. This enables detection of incompatibilities with language bindings (e.g. because of changes in the (internal) API used by these bindings). Both CI jobs include running the binding's tests, and the Python fuzzer tests are also included.


eboasson (approved):
Thanks @dpotman!

PR #1800, opened 2023-08-10
Fix cdrstream bug for union types

dpotman:
When using a sample for deserialization of CDR for a type that has a union, the sample data for the union was cleared before
deserialization (indicated by the `DDS_TOPIC_CONTAINS_UNION` flag). In case the union type is used for an external or mutable member, this could lead to reading uninitialized sample data when re-using an allocated sample for multiple reads. E.g. in this data type:

```IDL
@final @nested
union u switch(octet) {
    case 1:
        sequence<string<3> >  u1;
    default:
        string d;
};

@mutable
struct s {
    u f1;
};
```

This commit fixes the issue by introducing a sample-data state parameter in the read functions in cdrstream, and ensures that sample data is only read when the data state is 'initialized'. In case of switching cases for a union, the state for the sample data in that context is set to 'uninitialized', so that the deserializer doesn't read the data (e.g. to check if a pointer needs to be released) for nested types; in that case it will just allocate the memory and assume releasing it is not required.

The `DDS_TOPIC_CONTAINS_UNION` flag is not used anymore and removed from the IDLC generator. It is only kept in `dds_opcodes.h` for backward compatibility.

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -1414,6 +1414,30 @@ const uint32_t * dds_stream_write_with_byte_order (dds_ostream_t * __restrict os
     return dds_stream_write (os, allocator, data, ops);
 }
 
+static void malloc_sequence_buffer (dds_sequence_t * __restrict seq, const struct dds_cdrstream_allocator * __restrict allocator, uint32_t num, uint32_t elem_size)
+{
+  const uint32_t size = num * elem_size;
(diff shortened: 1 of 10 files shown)


eboasson (approved):
LGTM! There are a few non-functional details that I wanted done slightly differently, which I have addressed in the bunch of small commits I just pushed, that was the easiest way to deal with them 🙂

dennispotman:
Thanks, the changes look good to me

dpotman:
Thanks, the changes look good to me!

PR #1801, opened 2023-08-11
Avoid qsort_r because of older MUSL versions

eboasson:
There was only place where it is used, that is in the IDL compiler, to sort a vector of strings. By using qsort_r, the compare function supplied to the "sort vector" operation could operate on elements, with the dereferencing step done inside the vector implementation, which is much nicer than requiring the supplied compare function to do that dereference.

Unfortunately, qsort_r is not sufficiently widely supported.


MarcelJordense:
LGTM

PR #1806, opened 2023-08-18
Issue #1312: Generated include guards are insufficient

### LINKED ISSUE #1312: Generated include guards are insufficient
Currently `idlc` puts an include guard like `DDSC_BASENAME_H` (similar for C++). However if you have 2 files  with the same basename (in a different folder, for example) it will clash and you won't be able to include both :)

RTI solves this by appending a "random" number to the guard, not sure what it is, but it should be trivial to build md5/sha sum of the input IDL file and stick it in the guard. md5 is already a dependency from what I can tell (used in `idl_hashid`).

wadehunk:
This commit computes the MD5 hash for IDL source files and utilizes the result to generate a unique header guard macro when combined with the header file path.

Fixes #1312

### DIFF
--- a/src/idl/include/idl/processor.h
+++ b/src/idl/include/idl/processor.h
@@ -25,6 +25,7 @@
 #include "idl/scope.h"
 #include "idl/visit.h"
 #include "idl/attributes.h"
+#include "idl/md5.h"
 
 /* enable "#pragma keylist" for backwards compatibility */
 #define IDL_FLAG_KEYLIST (1u<<0)
@@ -88,6 +89,7 @@ struct idl_pstate {
   void *directive;
   idl_node_t *builtin_root, *root;
   idl_buffer_t buffer; /**< dynamically sized input buffer */
+  idl_md5_byte_t digest[16]; /**< md5 digest of idl source */
   struct {
     enum {
       IDL_SCAN,
--- a/src/tools/idlc/src/idlc/idlc.c
+++ b/src/tools/idlc/src/idlc/idlc.c
@@ -22,6 +22,7 @@
 #include "dds/features.h"
 #include "dds/cdr/dds_cdrstream.h"
 #include "idl/heap.h"
+#include "idl/md5.h"
 #include "idl/misc.h"
 #include "idl/tree.h"
 #include "idl/string.h"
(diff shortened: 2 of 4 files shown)


eboasson (approved):
Thanks @wadehunk ! 😀

This only covers the C generator (I'm sure you know) but with the digest stored in `pstate` it is totally trivial to also update the C++ generator. In fact, I decided to give it a go and it looks like it can be as simple as this:
```diff
diff --git a/src/idlcxx/src/generator.c b/src/idlcxx/src/generator.c
index efe7bb9e..bbd97b5b 100644
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -613,11 +613,14 @@ idl_type_t unalias_bitmask(const idl_node_t *node)
 }
 
 static char *
-figure_guard(const char *file)
+figure_guard(const char *file, const idl_md5_byte_t digest[16])
 {
   char *inc = NULL;
 
-  if (idl_asprintf(&inc, "DDSCXX_%s", file) == -1)
+  if (idl_asprintf(&inc, "DDSCXX_%s_%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x",
+                   file, digest[0], digest[1], digest[2], digest[3], digest[4], digest[5],
+                   digest[6], digest[7], digest[8], digest[9], digest[10], digest[11],
+                   digest[12], digest[13], digest[14], digest[15]) == -1)
     return NULL;
 
   /* replace any non-alphanumeric characters */
@@ -905,7 +908,7 @@ idl_retcode_t generate_nosetup(const idl_pstate_t *pstate, struct generator *gen
   idl_retcode_t ret;
   char *guard;
 
-  if (!(guard = figure_guard(gen->header.path)))
+  if (!(guard = figure_guard(gen->header.path, pstate->digest)))
     return IDL_RETCODE_NO_MEMORY;
   if ((ret = print_header(gen->header.handle, gen->path, gen->header.path)))
     goto err_print;
```

Because the issue relates both to C and C++, I'd like to merge updates to both repositories at the same time so that it is fixed for once and for all. Would you be willing to also do a PR for the C++? Feel free to use the diff above if you agree with it.

wadehunk:
@eboasson Will do. That was my plan to reuse that digest. Glad it works out nicely.

PR #1807, opened 2023-08-19
Reject NaN in uf_int64_unit

eboasson:
sscanf will happily parse a "nan" as a floating-point input, resulting in an attempt at converting a NaN into an integer instead of rejecting it outright.

That NaN then was manipulated and ended up in a range check:

  if (v_dbl_abs_scaled >= (double) INT64_MAX)

but because all comparisons involving a NaN evaluate as false, it passes this check and, at least on some platforms, causes undefined behaviour by trying to evaluate -INT64_MIN.

This commit inverts the sense of the check.


MarcelJordense:
LGTM

PR #1812, opened 2023-08-24
dynsub: fix bounded strings

eboasson:


examples/dynsub/print_sample.c
```diff
@@ -81,12 +81,15 @@ static void print_sample1_ti (const unsigned char *sample, const DDS_XTypes_Type
...
+      const char *p = align (sample, c, _Alignof (char *), sizeof (char *));
       if (c->key || c->valid_data)
       {
         printf ("%s", sep);
         if (label) printf ("\"%s\":", label);
-        printf ("\"%s\"", *p);
+        if ((typeid->_d == DDS_XTypes_TI_PLAIN_SEQUENCE_SMALL) ? typeid->_u.string_sdefn.bound : typeid->_u.string_ldefn.bound)
+          printf ("\"%s\"", p);
+        else
+          printf ("\"%s\"", *((const char **) p));
```
noxpardalis, line 92: I wanted to clarify my comprehension that this is involving some sort of small string/sequence optimization where the characters are immediately available at `p` for small strings but requires pointer indirection otherwise?

noxpardalis (approved):
LGTM! I had one point of clarification I asked in the comments as I'm a little unfamiliar with the code.

PR #1817, opened 2023-08-30
Resolve `snprintf` buffer overflow under GCC 12.3.0

noxpardalis:
The following tests were crashing for me when compiled with GCC 12.3.0 (and surprisingly not under either 12.2.0 nor 13.2.0 :confounded:).

```
275 - CUnit_ddsc_config_simple_udp (Subprocess aborted)
  - 275: *** buffer overflow detected ***: terminated
545 - CUnit_ddsc_participant_create_with_conf_no_env (Subprocess aborted)
  - 545: *** buffer overflow detected ***: terminated
```

Both of these arise from `do_print_uint32_bitset`:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/7df82fbdf36ea68c03bf29b96891ea91b86692ea/src/core/ddsi/src/ddsi_config.c#L812
Here the out-of-bounds check from `__snprintf_chk` correctly fails when `resp` is incremented ahead of `res` while the `bufsz` stays at 256. Note, during actual execution of the code with older versions of GCC the code never actually reaches out-of-bounds but it's certainly better to fix this.

The fix involves bounding the max `bufsz` expected by `snprintf` to the actual length that's remaining. I added some assertions for the results of `snprintf` too.


eboasson (approved):
🤦‍♀️

Thank you @noxpardalis ! ☺️

PR #1821, opened 2023-09-05
Introduce ddsrt_allocation_ops and ddsrt_set_allocators

poetinger:
This commit introduces the `ddsrt_allocation_ops_t` struct, which encapsulates memory allocation function pointers (malloc, calloc, realloc, free). Additionally the user may change these from the default functions of `stdlib` with the new function `ddsrt_set_allocators`.

The new struct makes it easier to change allocation strategies.

Changes:
- Added `ddsrt_allocation_ops_t` struct
- Added `ddsrt_set_allocators(ddsrt_allocation_ops)`
- Added `static ddsrt_allocator`, with default members from `stdlib`
- Changed `stdlib` allocator calls to use `ddsrt_allocator.{function}` etc.

CP-221

### DIFF
--- a/src/ddsrt/include/dds/ddsrt/heap.h
+++ b/src/ddsrt/include/dds/ddsrt/heap.h
@@ -26,6 +26,21 @@
 extern "C" {
 #endif
 
+typedef struct ddsrt_allocation_ops {
+  void* (*malloc)  (size_t size);
+  void* (*calloc)  (size_t count, size_t size);
+  void* (*realloc) (void *memblk, size_t size);
+  void  (*free)    (void *ptr);
+} ddsrt_allocation_ops_t;
+
+/// @brief Set the memory allocators
+///
+/// Set the functions to malloc, calloc, realloc, and free memory. By default
+/// the functions of `stdlib` will be used.
+DDS_EXPORT void
+ddsrt_set_allocator(
+  ddsrt_allocation_ops_t custom_ops);
+
 /**
  * @brief Allocate memory from heap.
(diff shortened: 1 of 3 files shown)


src/ddsrt/include/dds/ddsrt/heap.h
```diff
@@ -26,6 +26,21 @@
 extern "C" {
 #endif
 
+typedef struct ddsrt_allocation_ops {
+  void* (*malloc)  (size_t);
```
eboasson, line 30: It is a tantamount to nitpicking, but I'd prefer to see names for the arguments here. They're meaningless to the compiler and any C programmer may reasonably be assumed to know the what the arguments of these functions are, but it is cheap documentation and sometimes comes in very handy.

poetinger: No problem, I will add these!

poetinger:
> Thanks @poetinger, it is indeed about time applications can override the malloc implementation. However, I have found a few (very minor) things that in my view would benefit from some changes.

Fixed the "very minor" issues 🙂. I have one worry about the "Symbol Export Sanity Test" that seems to fail when enabled though.

eboasson (commented):
Thanks @poetinger!

The `symbol_export` test indeed needs to be updated: it picked up the introduction of a new exported symbol. In this case it is intentional, and so it needs to be listed in `symbol_export`.

eboasson (approved):
Thanks @poetinger !

PR #1826, opened 2023-09-12
remove deprecate warnings when using openssl v3

MarcelJordense:
Some of the Openssl v1.1.1 function used by the authentication security plugin have become deprecated in Openssl v3. These functions have been replaced by the corresponding function provided by Openssl v3.


src/security/openssl/include/dds/security/openssl_support.h
```diff
@@ -38,7 +38,7 @@
 #include <WinSock2.h>
 #endif
 
-#define OPENSSL_API_COMPAT 10101
+#define OPENSSL_API_COMPAT 30000
```
eboasson, line 41: I guess this should be defined to 10101 when using OpenSSL < 3.0.0, but I checked and OpenSSL 1.1.1 doesn't care if it is set too high. Making it conditional on the OpenSSL version number seems overkill, but a comment to this effect might be useful.

src/security/builtin_plugins/authentication/src/auth_utils.c
```diff
@@ -74,7 +74,7 @@ char *get_certificate_subject_name(X509 *cert, DDS_Security_SecurityException *e
 dds_time_t get_certificate_expiry(const X509 *cert)
 {
   assert(cert);
-  ASN1_TIME *asn1 = X509_get_notAfter(cert);
+  const ASN1_TIME *asn1 = X509_get0_notAfter(cert);
```
eboasson, line 77: Wouldn't it be better to change all uses of  `X509_get_notAfter` and `X509_get_notBefore` to the `get0` variant (or none of them)?

eboasson (commented):
I think I should've merged this a long time ago already ... or at least reviewed it properly because I do have two questions/remarks. They're totally trivial details, to be honest.

Now that the CI is working again on Windows on `master` — the problems were caused by a combination of CMake, OpenSSL 3.2.1 and Chocolatey — it would be also be nice to see those builds green with this PR.

Perhaps you would be willing to consider my two comments, possibly update the code and rebase it once more to `master`?

eboasson (commented):
LGTM, but now there is an issue with the email address used in the commits: the short one is registered with the Eclipse Foundation, the long one isn't. Could you please edit the commits to use the correct address? Thank you!

eboasson (approved):
Thanks!

PR #1829, opened 2023-09-18
Always respect SPDPInterval setting

eboasson:
The code used to silently limit the SPDP interval to approximately 80% of the lease duration for addressing some interoperability issues at some point in the past.  That meant that even in a Cyclone-only system it would be impossible to raise the interval beyond this limit.

This extends the valid values with the special value "default" in which case it does exactly the same as it used to do if the setting was left at the default (30s).  With this change, if it is set to a numerical value, that specified interval is respected without taking the lease duration into consideration.

That means that a configuration in which the SPDP interval was set to some value greater than 80% of the lease duration, the rate now goes down.  In all other cases, the behaviour is unchanged.

### DIFF
--- a/src/core/ddsi/src/ddsi_discovery_spdp.c
+++ b/src/core/ddsi/src/ddsi_discovery_spdp.c
@@ -355,8 +355,12 @@ static void ddsi_spdp_directed_xevent_cb (struct ddsi_domaingv *gv, struct ddsi_
              PGUID (pp->e.guid), PGUIDPREFIX (arg->dest_proxypp_guid_prefix), DDSI_ENTITYID_PARTICIPANT);
 
   // Directed events are used to send SPDP packets to newly discovered peers, and used just once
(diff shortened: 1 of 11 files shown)


src/core/ddsi/src/ddsi_discovery_spdp.c
```diff
@@ -371,22 +375,28 @@ static void ddsi_spdp_directed_xevent_cb (struct ddsi_domaingv *gv, struct ddsi_
 static void resched_spdp_broadcast (struct ddsi_xevent *ev, struct ddsi_participant *pp, ddsrt_mtime_t tnow)
 {
   /* schedule next when 80% of the interval has elapsed, or 2s
-   before the lease ends, whichever comes first (similar to PMD),
-   but never wait longer than spdp_interval */
+     before the lease ends, whichever comes first (similar to PMD),
+     but never wait longer than spdp_interval */
```
dpotman, line 379: With these code changes this code comment now applies to the `else` branch below, and 'interval' concerns the participant's lease duration. So may be better to either move it down or include the SPDP interval in the text.

dpotman (commented):
LGTM! Only a minor remark below.

dpotman (approved):

PR #1831, opened 2023-09-19
Build and doxygen fixes

dpotman:
- Add -Werror to linker options 
- Doxygen comment fixes in security header
- Use chocolatey on Windows to install OpenSSL


eboasson (approved):
Thanks @dpotman !

PR #1841, opened 2023-09-27
Configurable #fragments to send on NACK of sample

eboasson:
This changes the hard-coded policy of responding to a NACK of a large sample with only the first fragment of that sample into one where the number is configurable.  The default is 1 and so there is by default no change in behavior.

Because of the change, the code handling HeartbeatFrag messages in the case of a retransmit is now no longer dead code.


dpotman (approved):
LGTM

PR #1842, opened 2023-09-28
Add dds peek

eboasson:
Every now and then the problem of wanting to peek at the contents of a reader history cache without marking samples as read pops up. It is a perfectly sensible operation to support and it is mystery to me why the DDS specification did not include it.

Also in this PR:

* a `(void)x` to suppress a warning in release builds in some test code for a variable only read in an `assert`
* an update to clang 13 in the CI configuration because clang 12 is no longer included in the Azure Pipelines images

If desired I could do a separate PR for these, but I don't think there's a real need for that.


src/core/ddsc/include/dds/dds.h
```diff
@@ -3988,6 +4265,115 @@ dds_take_with_collector (
...
+ *
+ * If the sample/view/instance state component in the mask is 0 and there is no read or query condition,
+ * to combine it with, it is treated as equivalent to any sample/view/instance state.
+ *
+ * Return value provides information about the number of samples read, which will
+ * be <= maxs. Based on the count, the buffer will contain serialized data to be
+ * read only when valid_data bit in sample info structure is set.
+ * The buffer required for data values, could be allocated explicitly or can
+ * use the memory from data reader to prevent copy. In the latter case, buffer and
+ * sample_info should be returned back, once it is no longer using the data.
```
dpotman, line 4290: Not a problem initiated by this PR, and the paragraph has been used in `dds.h` for a while for several read functions, but I think it's not completely clear under what conditions the caller should return the buffer (and not sample_info..). And shouldn't the `buf` parameter be `in,out`?

dpotman (approved):
Looks good! No problem to have the unrelated commits in this PR. Only a detail in the comments for the read (and now also peek) functions in `dds.h` (which is something not introduced by this PR, so I'm fine with improving these comments later). 

The build failure on CI seems unrelated, however something we should really look into...

src/core/ddsc/include/dds/dds.h
eboasson: Good point, I can't even make sense of that description myself! There'll be a separate PR for the documentation of those functions shortly.

eboasson:
I agree on the CI issue. I suspect it will turn out to be an issue with the test rather than the code-under-test, but it needs to be sorted out.

Issue #1834, opened 2023-09-22
Problem using VLAN tagging


MarcelJordense:
It appears that the configured VLAN tag is not applied. It appears that the VLAN tag is applied to the user data multicast locator but not to the meta data multicast locator of the corresponding reader.

poetinger:
This issue should be resolved by #1848

PR #1848, opened 2023-10-09
set the same vlan id on all the locators

MarcelJordense:
To use a specific vlan for a particular NetworkPartition then the vlan id is configured as part of the address (multicast) set for the NetworkPartition. This results in that the vlan id is set on the corresponding multicast locator. However it is not set on the unicast locator associated with the chosen ethernet interface.


MarcelJordense:
fixes #1834

PR #1857, opened 2023-10-19
CP-492 Add documentation for setting allocation strategy

poetinger:
This will make it obvious to users that they can tailor the memory (de)allocation strategy to their specific needs.


eboasson (approved):
Thanks @poetinger !

PR #1859, opened 2023-10-23
Check for required serdata ops before enabling PSMX on an endpoint

gmartin82:
This PR adds an additional check to dds_ensure_valid_psmx_instances() which validates that the required serdata operations are present for the sertype. For a writer, it checks for the from_loaned_sample() operation; for a reader, it checks for the from_psmx() operation. If the required operation is not present no PSMX instances will be added to the supported_psmx list.

Without this (or a similar) check, Cyclone may call these operations on a custom sertype implementation that doesn't provide implementations potentially resulting in a segmentation fault. As an example of this, the [Zenoh DDS Bridge](https://github.com/eclipse-zenoh/zenoh-plugin-dds) uses such a custom sertype which is used when type information isn't available.


eboasson (approved):
Thank you @gmartin82 for eliminating that nice little gotcha.

PR #1862, opened 2023-10-24
fix ddsrt_getprocessname() for FreeRTOS

CenTr1X:
Now `ddsrt_getprocessname(void)` returns the copy of string of process name, so it can be free correctly.


eboasson (approved):
Thank you very much @CenTr1X !

PR #1864, opened 2023-10-26
Static build support

eboasson:
Master currently doesn't support a static build including Iceoryx support.

This PR aims to address that. It works locally for me on macOS but there are some lingering issues. The nice bit is that it now also supports statically linking in the security plugins, so a statically linked build including DDS Security is now possible.

I have been rather careless (reckless?) in the CMake bits and it is horribly ugly now. I don't actually know the detailed behaviour of `OBJECT` libraries, how much pain is caused by `install` entry on some internal things just to be allowed to reference it elsewhere, how to set properties for include directories so that I don't have to write them out by hand, etc.

Maybe someone with a good idea on how to clean this up sees it and is willing to help. That's actually the primary reason for making it a draft PR now.


CMakeLists.txt
```diff
@@ -29,6 +30,28 @@ else()
...
+#  CMAKE_POSITION_INDEPENDENT_CODE=1
+#
+# resulting in a PIC static library, but that is in turn incompatible with the
+# system-provided static libraries for OpenSSL.  So much easier to avoid it for the rare
+# case where it is needed.
+#
+# It appears that on macOS, all code is position independent and it'll work regardless.
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+if(NOT BUILD_SHARED_LIBS AND NOT CMAKE_CROSSCOMPILING)
+  message(WARNING "It is probably best to build a static library as-if cross compiling (e.g., use -DCMAKE_CROSSCOMPILING=1 -DCMAKE_SYSTEM_NAME=${CMAKE_SYSTEM_NAME})")
```
dpotman, line 52: Maybe also add a note that in this case `CMAKE_PREFIX_PATH` needs to include the install directory of a shared-lib build, in order to find the `cycloneddsidlc`?

src/security/core/tests/CMakeLists.txt
```diff
@@ -11,14 +11,44 @@
...
+        init_test_${libname}_local_writer_not_allowed
+        init_test_${libname}_local_reader_not_allowed
+        init_test_${libname}_local_permissions_not_allowed
+        init_test_${libname}_remote_participant_not_allowed
+        init_test_${libname}_remote_topic_not_allowed
+        init_test_${libname}_remote_writer_not_allowed
+        init_test_${libname}_remote_reader_not_allowed
+        init_test_${libname}_remote_reader_relay_only
+        init_test_${libname}_remote_permissions_not_allowed
+        finalize_test_${libname}_not_allowed)
```
dpotman, line 45: Updating these symbol lists can easily be missed when adding functions, but I guess that leads to an error in the static build on CI?

dpotman (commented):
Looks good and works as expected on my dev machine (nice macro hack for the dynamic lib loader 😄). Below a few minor remarks/questions, but nothing blocking.

CMakeLists.txt
eboasson: Good point, I added that.

src/security/core/tests/CMakeLists.txt
eboasson: Yes, leaving them out means that `dlsym` will fail to find the symbols and that causes test failures. It would be nicer to automatically generate this list, but then I need to figure out a way to build the library first, then extract the symbols, add them to the plug-in properties, then continue building the library. I'd have done it if it were normal `make` ... Maybe one day I can find the courage to try it with CMake!

PR #1867, opened 2023-10-30
Introduce ddsrt abort, exit, assert

poetinger:
It would be nice to have a single location where `abort` from the `stdlib` is actually called.

This PR will:
- Introduce `ddsrt_abort` in `ddsrt::process`
- Replace `abort` with `ddsrt_abort`, and include the `ddsrt/process.h` if it was not already included, in:
  - cdr
  - ddsc
  - ddsi
  - ddsrt
- Add the symbol export test


poetinger:
The large number of failing CI is a surprise to me. I checked on my Mac and Linux machines and don't seem to get the same issues.

PR #1877, opened 2023-11-13
Introduce `ddsrt_abort` and `ddsrt_exit`

poetinger:
This pull request:

- Introduces `ddsrt_abort` and `ddsrt_exit`
- Replaces `abort` with `ddsrt_abort` in `ddsc`, `ddsrt`, `cdr`, and `ddsi`
- Addes a symbol export test for each

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -15,6 +15,7 @@
 #include "dds/ddsrt/endian.h"
 #include "dds/ddsrt/md5.h"
 #include "dds/ddsrt/heap.h"
+#include "dds/ddsrt/process.h"
 #include "dds/ddsrt/string.h"
 #include "dds/ddsrt/static_assert.h"
 #include "dds/cdr/dds_cdrstream.h"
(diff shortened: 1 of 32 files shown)


eboasson (changes requested):
Thanks @poetinger, I do see the practical value of not calling `abort()` directly but putting something in between if one is using an embedded platform — at least I have found something of the sort useful in the past [🙂](https://github.com/atolab/zhe/blob/bef745b00e75b3436424f2e79d9428395f1c4189/example/zbot/zbot.ino#L120).

So that's the good bit.

The bad bit is that not all the `abort`s in the core code have been replaced, for example `src/core/cdr/src/dds_cdrstream_keys.part.c` have a bunch left.

Not replacing `abort` in *all* locations (like tests) is a defensible choice, but I think it only makes life more complicated to sometimes use `abort` and sometimes use `ddsrt_abort`: now one has to decide which one to use. So I think *all* should be replaced and only a single call to `abort` should remain, in `ddsrt_abort`.

The subject of `exit` is another story. Cyclone DDS comes in multiple parts: the core bit is a library—this should never terminate and so should never call `exit` to begin with, and it doesn't—and then there are some tools/examples/tests that use it. I don't quite see why something like `ddsperf` can't just call `exit`.

So unless you have a good argument why `exit` must be avoided at all cost, I'd take out the 11 changes related to it.

poetinger:
Sounds reasonable to me. I'll remove `ddsrt_exit` and do another pass for missing aborts, including the ones in tests 👍.

PR #1879, opened 2023-11-15
Introduce the concept of a domain lifecycle state.

poetinger:
This PR will:

- Introduce the concept of a lifecycle state for domains
- Introduce the ability for a user to transition the lifecycle state

Limitations:

- The only two lifecycle states that will exist, for now, are `initialisation` and `operational`
- The only supported transition is `initialisation`->`operational`
- For now, there is no meaning to the two lifecycle states, but this can/will change in the future
- For now, the transition and state can only be applied to _all_ domains


poetinger:
Converted to draft, still needs tests.

poetinger:
Converted back to draft. Will include the ability to set a function to be called when the state transition is triggered.

Issue #1873, opened 2023-11-08
Formatting broken


Divelix:
Configuration File Reference page formatting in docs seems broken.
Link: https://cyclonedds.io/docs/cyclonedds/latest/config/config_file_reference.html

Issue #1883, opened 2023-11-20
Is shared memory no longer supported?


ex-purple:
Hi there,
I have a problem with shared memory usage on master:
`config: //CycloneDDS/Domain/SharedMemory: 'Enable': deprecated element`
`config: //CycloneDDS/Domain/SharedMemory: 'LogLevel': deprecated element`

Is shared memory no longer supported? Or maybe [this documentation](https://cyclonedds.io/docs/cyclonedds/latest/shared_memory/shared_mem_config.html) is deprecated?

eboasson:
Oops. The documentation is outdated.

The code for integrating Iceoryx used to be "hard-coded", with references to Iceoryx in various place in the Cyclone sources. We refactored it out into a plugin-based mechanism (see #1676) and added a plugin for Iceoryx in the sources. The "correct" way of configuring now looks like:

```
<General>
  <Interfaces>
    <PubSubMessageExchange name="iox" library="psmx_iox" config="LOG_LEVEL=INFO;"/>
  </Interfaces>
</General>
```

The old settings continue to work with a warning. They just translate the old (unchanging) configuration to the new format so they'll probably be around for quite some time.

ciandonovan:
Are there plans to use this new plugin interface to provide other shared-memory providers beside Iceoryx? Maybe a daemon-less one, similar to what FastDDS does? Use the loopback just for discovery, handshakes, etc. before coordinating on the shared-memory segment to use? Have had issues with Iceoryx crashing with a large amount of nodes, would be nice if there wasn't a single point of failure as currently everything must be restarted if Iceoryx crashes AFAIK.

Gummum:
The shared memory example has been removed. How can I test it

dibsonmuad:
On ROS2 Humble, 
```
<General>
  <Interfaces>
    <PubSubMessageExchange name="iox" library="psmx_iox" config="LOG_LEVEL=INFO;"/>
  </Interfaces>
</General>
```
did not work. I 
What worked for me was 
```
<?xml version="1.0" encoding="UTF-8" ?>
<CycloneDDS xmlns="https://cdds.io/config"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="https://cdds.io/config https://raw.githubusercontent.com/eclipse-cyclonedds/cyclonedds/iceoryx/etc/cyclonedds.xsd">
    <Domain id="any">
        <SharedMemory>
[...]

PR #1888, opened 2023-11-29
Documentation updates

ThijsSassen:
Fixed issue #1873 and #1883
Added missing psmx, dynamic types, log and loan documentation.
Fix missing refs and group refs from several header files.


docs/manual/config/config_file_reference.rst
```diff
@@ -21,7 +21,7 @@ CycloneDDS configuration
 //CycloneDDS/Domain
 *******************
 
-Attributes: [Id](`//CycloneDDS/Domain[@Id]`_)
```
eboasson, line 24: This is definitely much better, but I wonder if it would be possible to show `Id` the reader but have it be a link target for the definition of the attribute (as in `<a href="...">Id</a>`): that would make it read like:

Attributes: Id
Children: Compatibility, Discovery, General, ...

which in my opinion would look much nicer.

(The same goes for the child elements, of course.)

ThijsSassen: @eboasson i agree and fixed it

eboasson (approved):
Thank you @ThijsSassen !

PR #1889, opened 2023-11-29
Fix confusion between raw samples and CDR in PSMX

eboasson:
There were various issues causing confusion in the contents and size of
loans:

* dds_write_impl performed a look-ahead and could decide to allocate
  memory for a PSMX loan, only taking into account the type to decide
  whether to allocate based on the raw sample size or the serialized size,
  but failing to account for the possibility of a serialized key value.

* What used to be the "force serialization" flag was interpreted by
  serdata_default as (also) requiring that the data in a loan would be in
  serialized form, even though that wasn't the case.  It will handle
  serialized data just fine even if it is not necessary, but there is no
  reason for serializing it into shared memory if the data can just be
  memcpy'd around.

* This resulted in serdata_default using slightly different rules for
  deciding whether or not to serialize and so could (attempt to) put
  serialized data in a loan allocated for a raw sample.

  The metadata would be set correctly, and the amount of memory copied
  was that of the destination buffer, so as long as the CDR was no
  longer than the raw sample (unlikely) and the serdata buffer happened
  to be large enough, it was ok.  But those conditions are not always
  satisfied.

Taken together, this meant: out-of-bounds reads and inefficiencies.

This commit refactors dds_write_impl, taking the burden of doing things
with the contents of the loans from the serdata implementation (it
leaves the interface unchanged), instead doing it in dds_write_impl
where all the relevant information is available.

It also moves the publishing via PSMX to *before* the network part.
This ensures that the latency via shared memory is not affected by the
speed with which the network can absorb the data.

In this refactored version, supporting zero-copy within the process (and
without PSMX) was missing.  That has also been added.

This PR is dependent on #1890 for compiling warning-free (hence its inclusion of cc47314)

### DIFF
--- a/src/core/ddsc/src/dds__write.h
+++ b/src/core/ddsc/src/dds__write.h
@@ -32,13 +32,16 @@ typedef enum {
 
 /** @component write_data */
 DDS_EXPORT_INTERNAL_FUNCTION
-dds_return_t dds_write_impl (dds_writer *wr, const void *data, dds_time_t tstamp, dds_write_action action);
+dds_return_t dds_write_impl (dds_writer *wr, const void *data, dds_time_t timestamp, dds_write_action action)
+  ddsrt_attribute_warn_unused_result ddsrt_nonnull_all;
 
 /** @component write_data */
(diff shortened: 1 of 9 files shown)


dpotman (approved):
LGTM!

PR #1890, opened 2023-11-30
Write-like operations: consistent check of data/timestamp

eboasson:
Some functions, notable dds_write_ts rejected negative time stamps, a choice that I think is reasonable.

Unregister and dispose did not always check, but those occur necessarily after a write, and therefore allowing negative time stamps if they are rejected by a write operation makes no sense.  So adding a check here is a bug fix.

All write-like functions taking a data argument check for a null pointer before anything else.  dds_writedispose_ts failed to do this, which then results in a different error code.  That bug is also fixed here.

### DIFF
--- a/src/core/ddsc/tests/dispose.c
+++ b/src/core/ddsc/tests/dispose.c
@@ -177,9 +177,11 @@ CU_TheoryDataPoints(ddsc_writedispose, non_writers) = {
 };
 CU_Theory((dds_entity_t *writer), ddsc_writedispose, non_writers, .init=disposing_init, .fini=disposing_fini)
 {
+    Space_Type1 data = { 0, 22, 22 };
     dds_return_t ret;
+
     DDSRT_WARNING_MSVC_OFF(6387); /* Disable SAL warning on intentional misuse of the API */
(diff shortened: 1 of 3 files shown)


src/core/ddsc/tests/dispose.c
```diff
@@ -177,9 +177,11 @@ CU_TheoryDataPoints(ddsc_writedispose, non_writers) = {
 };
 CU_Theory((dds_entity_t *writer), ddsc_writedispose, non_writers, .init=disposing_init, .fini=disposing_fini)
 {
+    Space_Type1 data = { 0, 22, 22 };
     dds_return_t ret;
+
     DDSRT_WARNING_MSVC_OFF(6387); /* Disable SAL warning on intentional misuse of the API */
```
dpotman, line 183: Just a nitpick: I think these (this one and below) disable-warnings can be removed.

dpotman (approved):
LGTM, just a nitpick below.

src/core/ddsc/tests/dispose.c
eboasson: Indeed. I think there are several more of those, I'll do that in a separate PR.

PR #1907, opened 2023-12-18
Add support for QNX SDP 7.1

PatrickM-ZS:
This PR adds support for building CycloneDDS for [QNX SDP 7.1](https://www.qnx.com/developers/docs/7.1/)

- Added toolchain files for `aarch64` and `x86_64` targets. Note I've only tested on Linux host but it should work also on Windows and MacOS.
- Added math lib as a link dependency for ddsrt to resolve `isinf()`, which somehow doesn't get picked up automatically by the QNX toolchain.
- FIxed IDLC Generate CMake module to find external `idlc` and `libcycloneidlc.so` when building IDLC is disabled.


eboasson (approved):
A very belated LGTM! Thank you!

PR #1931, opened 2024-02-08
initial qnx vlan support

MarcelJordense:
Include the option to use raw ethernet transport and on QNX

### DIFF
--- a/src/core/ddsi/src/ddsi_raweth.c
+++ b/src/core/ddsi/src/ddsi_raweth.c
@@ -114,7 +114,7 @@ static char *ddsi_raweth_to_string (char *dst, size_t sizeof_dst, const ddsi_loc
 static void set_locator(ddsi_locator_t *srcloc, const uint8_t * addr, uint16_t port, uint16_t vtag)
 {
   srcloc->kind = DDSI_LOCATOR_KIND_RAWETH;
-  srcloc->port = (uint32_t)(port +  ((vtag & 0xfff) << 20) + ((vtag & 0xf000) << 4));
+  srcloc->port = (uint32_t)port +  (((uint32_t)vtag & 0xfff) << 20) + (((uint32_t)vtag & 0xf000) << 4);
   memset(srcloc->address, 0, 10);
   memcpy(srcloc->address + 10, addr, 6);
 }
@@ -151,7 +151,6 @@ static ssize_t ddsi_raweth_conn_read (struct ddsi_tran_conn * conn, unsigned cha
   struct tpacket_auxdata *pauxd;
   struct cmsghdr *cptr;
   uint16_t vtag = 0;
-//  uint32_t port;
   struct iovec msg_iov[2];
   socklen_t srclen = (socklen_t) sizeof (src);
   (void) allow_spurious;
@@ -223,7 +222,6 @@ static ssize_t ddsi_raweth_conn_write (struct ddsi_tran_conn * conn, const ddsi_
   struct msghdr msg;
   struct sockaddr_ll dstaddr;
   struct ddsi_vlan_header vhdr;
-//  uint16_t vtag;
   size_t hdrlen;
 
   assert(msgfrags->niov <= INT_MAX - 1); // we'll be adding one later on
@@ -331,7 +329,6 @@ static dds_return_t ddsi_raweth_create_conn (struct ddsi_tran_conn **conn_out, s
   memset(&addr, 0, sizeof(addr));
   addr.sll_family = AF_PACKET;
   addr.sll_protocol = htons(ETH_P_ALL);
-
   addr.sll_ifindex = (int)intf->if_index;
   addr.sll_pkttype = PACKET_HOST | PACKET_BROADCAST | PACKET_MULTICAST;
   rc = ddsrt_bind(sock, (struct sockaddr *)&addr, sizeof(addr));
@@ -477,7 +474,7 @@ static ssize_t ddsi_raweth_conn_read (struct ddsi_tran_conn * conn, unsigned cha
       {
         memcpy(buf, ptr, (size_t)ret);
         if (srcloc)
-          set_locator(srcloc, eth_hdr->smac, ntohs (eth_hdr->proto), (vtag ? vtag->tag : 0));
+          set_locator(srcloc, eth_hdr->smac, ntohs (eth_hdr->proto), (vtag ? ntohs(vtag->tag) : 0));
       }
       else
       {
--- a/src/core/ddsi/src/ddsi_raweth.c
+++ b/src/core/ddsi/src/ddsi_raweth.c
@@ -20,30 +20,68 @@
 #include "ddsi__mcgroup.h"
 #include "ddsi__pcap.h"
 
-#if defined(__linux) && !LWIP_SOCKET
+#if (defined(__linux) || defined(__FreeBSD__) || defined(__QNXNTO__) || defined(__APPLE__)) && !LWIP_SOCKET
+#include <sys/types.h>
+#include <string.h>
+
+#if defined(__linux)
 #include <linux/if_packet.h>
 #include <linux/if_ether.h>
 #include <linux/filter.h>
 #include <sys/types.h>
 #include <ifaddrs.h>
-#include <string.h>
 #include <assert.h>
 #include <stdlib.h>
 #include <errno.h>
+#elif defined(__FreeBSD__) || defined(__QNXNTO__) || defined(__APPLE__)
+#define DDSI_USE_BSD (1)
+#include <sys/uio.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#if defined(__FreeBSD__) || defined(__APPLE__)
+#include <net/ethernet.h>
+#elif defined(__QNXNTO__)
+#define DDSI_BPF_IS_CONING_DEV (1)
+#include <net/ethertypes.h>
+#include <net/if_ether.h>
+#endif    
+#include <net/bpf.h>
+#include <net/if_dl.h>
+#endif
+
+#if defined(__linux) 
+#define DDSI_ETH_ADDR_LEN ETH_ALEN
+#define DDSI_LINK_FAMILY AF_PACKET
+
+union ddsi_cmessage {
+  struct cmsghdr chdr;
+  char buf[CMSG_SPACE(sizeof(struct tpacket_auxdata))];
+};
+
+#elif DDSI_USE_BSD
+#define DDSI_ETH_ADDR_LEN ETHER_ADDR_LEN
+#define DDSI_LINK_FAMILY AF_LINK
+#endif
 
 typedef struct ddsi_raweth_conn {
   struct ddsi_tran_conn m_base;
   ddsrt_socket_t m_sock;
   int m_ifindex;
+#if DDSI_USE_BSD
+  ddsrt_mutex_t lock;
+  char *buffer;
+  uint32_t buflen;
+  ssize_t avail;
+  char *bptr;
+#endif
 } *ddsi_raweth_conn_t;
 
-
 struct ddsi_ethernet_header {
-  unsigned char dmac[ETH_ALEN];
-  unsigned char smac[ETH_ALEN];
+  unsigned char dmac[DDSI_ETH_ADDR_LEN];
+  unsigned char smac[DDSI_ETH_ADDR_LEN];
   unsigned short proto;
 } __attribute__((packed));
-
 
 struct ddsi_vlan_header {
   struct ddsi_ethernet_header e;
@@ -51,10 +89,9 @@ struct ddsi_vlan_header {
   unsigned short proto;
 } __attribute__((packed));
 
-union ddsi_cmessage {
-  struct cmsghdr chdr;
-  char buf[CMSG_SPACE(sizeof(struct tpacket_auxdata))];
-};
+static ddsrt_socket_t ddsi_raweth_conn_handle (struct ddsi_tran_base * base);
+static int ddsi_raweth_conn_locator (struct ddsi_tran_factory * fact, struct ddsi_tran_base * base, ddsi_locator_t *loc);
+
 
 static char *ddsi_raweth_to_string (char *dst, size_t sizeof_dst, const ddsi_locator_t *loc, struct ddsi_tran_conn * conn, int with_port)
 {
@@ -72,6 +109,7 @@ static char *ddsi_raweth_to_string (char *dst, size_t sizeof_dst, const ddsi_loc
   return dst;
 }
 
+#if defined(__linux)
 static ssize_t ddsi_raweth_conn_read (struct ddsi_tran_conn * conn, unsigned char * buf, size_t len, bool allow_spurious, ddsi_locator_t *srcloc)
 {
   dds_return_t rc;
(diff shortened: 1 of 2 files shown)


src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -20,41 +20,78 @@
...
+#elif defined(__FreeBSD__) || defined(__QNXNTO__) || defined(__APPLE__)
+#define DDSI_USE_BSD (1)
+#include <sys/uio.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#if defined(__FreeBSD__) || defined(__APPLE__)
+#include <net/ethernet.h>
+#elif defined(__QNXNTO__)
+#define DDSI_BPF_IS_CONING_DEV (1)
```
eboasson, line 45: I suspect a typo here, that is meant to be `DDSI_BPF_IS_CLONING_DEV`

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+  struct ddsi_network_interface const * const intf = qos->m_interface ? qos->m_interface : &gv->interfaces[0];
+  uint32_t buflen;
+
+  if (port == 0 || port > 65535)
+  {
+    DDS_CERROR (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u - using port number as ethernet type, %u won't do\n", mcast ? "multicast" : "unicast", port, port);
+    return DDS_RETCODE_ERROR;
+  }
+
+#if defined(DDSI_BPF_IS_CONING_DEV)
```
eboasson, line 577: If my suspicion on this above is correct, then this should also be `DDSI_BPF_IS_CLONING_DEV`.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+    return DDS_RETCODE_ERROR;
+  }
+
+#if defined(__FreeBSD__)
+  uint32_t direction = BPF_D_IN;
+  if ((r = ioctl (sock, BIOCGDIRECTION, &direction)) == -1 ) {
+      ddsrt_close (sock);
+      DDS_CWARNING (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u could not set directiion ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
+  }
+#elif defined(__QNXNTO__) || defined(__APPLLE__)
```
eboasson, line 633:
```suggestion
#elif defined(__QNXNTO__) || defined(__APPLE__)
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+        ptr += sizeof(*vtag);
+        ret -= (ssize_t)sizeof(*vtag);
+      }
+      if ((size_t)ret <= len)
+      {
+        memcpy(buf, ptr, (size_t)ret);
+        port = (uint32_t)(ntohs (eth_hdr->proto));
+        if (vtag)
+        {
+          port += ((uint32_t)(vtag->tag & 0xfff) << 20) + ((uint32_t)(vtag->tag & 0xf000) << 16);
```
eboasson, line 457: I think it makes sense not to duplicate this. I have no problem with the magic numbers and the bit fiddling, but it is a fiddly bit of code and it is nice to have only one copy to check.

Setting the source locator is also the same, I'd also de-dup that.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+  return (rc == DDS_RETCODE_OK ? ret : -1);
+}
+
+static dds_return_t ddsi_raweth_set_filter (struct ddsi_tran_factory * fact, ddsrt_socket_t sock, uint32_t port)
+{
+  int r;
+  ushort etype = (ushort)(port & 0xFFFF);
+  struct bpf_insn insns[] = {
+    BPF_STMT(BPF_LD+BPF_H+BPF_ABS, 12),
+	  BPF_JUMP(BPF_JMP+BPF_JEQ+BPF_K, etype, 3, 0),
```
eboasson, line 540:
```suggestion
    BPF_JUMP(BPF_JMP+BPF_JEQ+BPF_K, etype, 3, 0),
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+
+  return (rc == DDS_RETCODE_OK ? ret : -1);
+}
+
+static dds_return_t ddsi_raweth_set_filter (struct ddsi_tran_factory * fact, ddsrt_socket_t sock, uint32_t port)
+{
+  int r;
+  ushort etype = (ushort)(port & 0xFFFF);
+  struct bpf_insn insns[] = {
+    BPF_STMT(BPF_LD+BPF_H+BPF_ABS, 12),
```
eboasson, line 539: Nice to have a somewhat readable representation compared to Linux. Does this by any chance work on Linux as well?

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+    {
+      DDS_CERROR (&conn->m_base.gv->logconfig, "ddsi_raweth_create_conn read failed ... retcode = %"PRIdSIZE"\n", ret);
+      rc = DDS_RETCODE_ERROR;
+      goto error;
+    }
+    uc->avail = ret;
+    uc->bptr = uc->buffer;
+  }
+
+  if (uc->bptr < uc->buffer + uc->avail)
```
eboasson, line 433: For all the `bptr` manipulations/checks: these structures come straight from the OS kernel, so I suppose we can trust them. I do think it would be good to point out in a comment that this is the reason there is so little checking, to help any future reader.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+  (void) flags;
+
+  assert(msgfrags->niov <= INT_MAX - 1); // we'll be adding one later on
+ 
+  vtag = (uint16_t)(((dst->port >> 20) & 0xfff) + (((dst->port >> 16) & 0xe) << 12));
+  
+  memcpy(vhdr.e.dmac, dst->address + 10, 6);
+  memcpy(vhdr.e.smac, uc->m_base.m_base.gv->interfaces[0].loc.address + 10, 6);
+  
+  if (vtag) {
```
eboasson, line 509: Constructing the ethernet header also looks like a good candidate for de-duping.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+  
+  // activate immediate mode (therefore, buf_len is initially set to "1")
+  int mode = 1;
+  if ((r = ioctl (sock, BIOCIMMEDIATE, &mode)) == -1 ) {
+    ddsrt_close (sock);
+    DDS_CERROR (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u failed ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
+    return DDS_RETCODE_ERROR;
+  }
+
+   buflen = gv->config.socket_rcvbuf_size.max.value;
```
eboasson, line 603:
```suggestion
  buflen = gv->config.socket_rcvbuf_size.max.value;
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+    ddsrt_close (sock);
+    DDS_CERROR (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u failed ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
+    return DDS_RETCODE_ERROR;
+  }
+
+#if defined(__FreeBSD__)
+  uint32_t direction = BPF_D_IN;
+  if ((r = ioctl (sock, BIOCGDIRECTION, &direction)) == -1 ) {
+      ddsrt_close (sock);
+      DDS_CWARNING (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u could not set directiion ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
```
eboasson, line 631:
```suggestion
    ddsrt_close (sock);
    DDS_CWARNING (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u could not set direction ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +395,334 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+  }
+#elif defined(__QNXNTO__) || defined(__APPLLE__)
+  uint32_t direction = 0;
+  if ((r = ioctl (sock, BIOCSSEESENT, &direction)) == -1 ) {
+      ddsrt_close (sock);
+      DDS_CWARNING (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u could not set directiion ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
```
eboasson, line 637:
```suggestion
    ddsrt_close (sock);
    DDS_CWARNING (&fact->gv->logconfig, "ddsi_raweth_create_conn %s port %u could not set direction ... retcode = %d\n", mcast ? "multicast" : "unicast", port, r);
```

eboasson (commented):
Thanks @MarcelJordense, I have only done a quick review, but it generally looks fine. I have a few minor comments, perhaps you could take a look?

I am curious to try it on macOS, but I haven't had a chance yet.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -121,7 +151,7 @@ static ssize_t ddsi_raweth_conn_read (struct ddsi_tran_conn * conn, unsigned cha
   struct tpacket_auxdata *pauxd;
   struct cmsghdr *cptr;
   uint16_t vtag = 0;
-  uint32_t port;
+//  uint32_t port;
```
eboasson, line 154:
```suggestion
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -200,7 +223,7 @@ static ssize_t ddsi_raweth_conn_write (struct ddsi_tran_conn * conn, const ddsi_
   struct msghdr msg;
   struct sockaddr_ll dstaddr;
   struct ddsi_vlan_header vhdr;
-  uint16_t vtag;
+//  uint16_t vtag;
```
eboasson, line 226:
```suggestion
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -109,6 +111,34 @@ static char *ddsi_raweth_to_string (char *dst, size_t sizeof_dst, const ddsi_loc
   return dst;
 }
 
+static void set_locator(ddsi_locator_t *srcloc, const uint8_t * addr, uint16_t port, uint16_t vtag)
+{
+  srcloc->kind = DDSI_LOCATOR_KIND_RAWETH;
+  srcloc->port = (uint32_t)(port +  ((vtag & 0xfff) << 20) + ((vtag & 0xf000) << 4));
```
eboasson, line 117: I'd prefer it if `port` and `vtag` were cast to a `uint32_t` before doing the arithmetic, so that all of it only involves unsigned 32-bit integers. I think the C99 section 6.5.7 "Bitwise shift operators" even says that if we are not careful we will get ourselves into trouble:

> The integer promotions are performed on each of the operands. The type of the result is that of the promoted left operand.
> The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with zeros.
> If E1 has a signed type and nonnegative value, and E1 × 2^E2 is representable in the result type, then that is the resulting value; otherwise, the behavior is undefined.

So if `vtag & 0xfff` ≥ 0x800, so say 0x800 = 2^11, then `(vtag & 0xfff) << 20` = 2^11 × 2^20 = 2^31, and 2^31 isn't representable in a signed 32 bit int. That would make it undefined behaviour.

It is quite possible that this hasn't been introduced by the refactoring and that I just happened to spot it now. Even if that's the case, I think it makes sense to fix it now.

```suggestion
  srcloc->port = (uint32_t)port + (((uint32_t)vtag & 0xfff) << 20) + (((uint32_t)vtag & 0xf000) << 4));
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -307,6 +331,7 @@ static dds_return_t ddsi_raweth_create_conn (struct ddsi_tran_conn **conn_out, s
   memset(&addr, 0, sizeof(addr));
   addr.sll_family = AF_PACKET;
   addr.sll_protocol = htons(ETH_P_ALL);
+
```
eboasson, line 334:
```suggestion
```

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -119,15 +187,8 @@ static ssize_t ddsi_raweth_conn_read (struct ddsi_tran_conn * conn, unsigned cha
...
-    // FIXME: ((vtag & 0xf000) << 16)) looks decidedly odd, << 4 would make more sense
-    port = (uint32_t)(ntohs (src.sll_protocol) + ((vtag & 0xfff) << 20) + ((vtag & 0xf000) << 16));
     if (srcloc)
-    {
-      srcloc->kind = DDSI_LOCATOR_KIND_RAWETH;
-      srcloc->port = port;
-      memset(srcloc->address, 0, 10);
-      memcpy(srcloc->address + 10, src.sll_addr, 6);
-    }
+      set_locator(srcloc, src.sll_addr, ntohs (src.sll_protocol), vtag);
```
eboasson, line 191: Is it really the case that `vtag` a.k.a. `pauxd->tp_vlan_tci` is already in native byte order? I'm asking only because it seems somewhat against the pattern, but it is not impossible.

src/core/ddsi/src/ddsi_raweth.c
```diff
@@ -391,6 +407,323 @@ static int joinleave_asm_mcgroup (ddsrt_socket_t socket, int join, const ddsi_lo
...
+      {
+        vtag = (struct ddsi_vlan_tag *)ptr;
+        ptr += sizeof(*vtag);
+        ret -= (ssize_t)sizeof(*vtag);
+      }
+      if ((size_t)ret <= len)
+      {
+        memcpy(buf, ptr, (size_t)ret);
+        if (srcloc)
+          set_locator(srcloc, eth_hdr->smac, ntohs (eth_hdr->proto), (vtag ? vtag->tag : 0));
```
eboasson, line 480: Like in the Linux version, on consideration I am somewhat surprised that the tag would be in native byte order already. Perhaps you can double-check?

eboasson (commented):
Thanks @MarcelJordense ☺️ Just a few tiny remarks left

src/core/ddsi/src/ddsi_raweth.c
MarcelJordense: I this case the vlan tag is already in host order

src/core/ddsi/src/ddsi_raweth.c
MarcelJordense: In this case the vtag is of course in network order. I will correct this.

eboasson (approved):
Thanks @MarcelJordense !

While I don't like merging when there are CI failures, these are definitely unrelated and so it makes no sense to not merge this PR now.

PR #1933, opened 2024-02-21
Bump docker ubuntu jammy

e-tmplr:
Use Ubuntu Jammy (LTS) for the docker images because that provides a version of CMake that is required to build the project

- Updated README.md accordingly
- Enable building the examples in build.sh
- Remove the docker containers on exit

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -58,6 +58,6 @@ fi
 
 mkdir -p build
 cd build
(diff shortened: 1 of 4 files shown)


eboasson (approved):
Thank you very much @e-tmplr. I apologise for the long delay, I have been away for a bit and it has been a bit hectic these past weeks ...

The CI failures have totally unrelated (known) causes.

PR #1939, opened 2024-02-29
add psmx test to check whether shared memory can be enabled and the l…

mvandenhoek:
…ocator can be set

### DIFF
--- a/src/core/ddsc/tests/psmx.c
+++ b/src/core/ddsc/tests/psmx.c
@@ -31,6 +31,7 @@
 
 #include "config_env.h"
 #include "test_common.h"
+#include "psmx_dummy_public.h"
 #include "Array100.h"
 #include "DynamicData.h"
 #include "PsmxDataModels.h"
@@ -1032,6 +1033,80 @@ static bool eq_DynamicData_KMsg (const void *vsent, const void *vrecvd, bool val
   }
 }
 
+/** @brief Compare two sets of stats.
+ * 
+ * @param[in] mockstats1 first stats to compare
+ * @param[in] mockstats2 second stats to compare
+ * 
+ * @return A bitmask to indicate which elements are different, zero if all are equal.
+*/
+static uint32_t dummy_mockstats_cmp(const dummy_mockstats_t* mockstats1, const dummy_mockstats_t* mockstats2)
+{
+  uint32_t cmp = 0;
+  uint32_t shift = 0;
+
+  cmp |= ((uint32_t)(mockstats1->cnt_type_qos_supported != mockstats2->cnt_type_qos_supported) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_create_topic != mockstats2->cnt_create_topic) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_delete_topic != mockstats2->cnt_delete_topic) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_deinit != mockstats2->cnt_deinit) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_get_node_id != mockstats2->cnt_get_node_id) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_supported_features != mockstats2->cnt_supported_features) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_create_endpoint != mockstats2->cnt_create_endpoint) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_delete_endpoint != mockstats2->cnt_delete_endpoint) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_request_loan != mockstats2->cnt_request_loan) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_write != mockstats2->cnt_write) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_take != mockstats2->cnt_take) << shift++);
+  cmp |= ((uint32_t)(mockstats1->cnt_on_data_available != mockstats2->cnt_on_data_available) << shift++);
+  return cmp;
+}
+
+/** @brief Convert a set of stats to a string.
+ * 
+ * Truncates the output string if the string buffer capacity is too small.
+ * 
+ * @param[in] dmock stats to convert
+ * @param[out] str_out string buffer to write the string into
+ * @param[in] str_capacity number of bytes the string buffer can hold
+ * 
+ * @return Upon successful return, it returns the number of characters printed
+ *         (excluding the null byte used to end output to strings).
+ *         If an output error is encountered, a negative value is returned.
+*/
+static int dummy_mockstats_tostring(const dummy_mockstats_t* dmock, char* str_out, size_t str_capacity)
+{
+  return snprintf(
+    str_out,
+    str_capacity,
+    "\
+  type_qos_supported: %i\n\
+  create_topic: %i\n\
+  delete_topic: %i\n\
+  deinit: %i\n\
+  get_node_id: %i\n\
+  supported_features: %i\n\
+  \n\
+  create_endpoint: %i\n\
+  delete_endpoint: %i\n\
+  \n\
+  request_loan: %i\n\
+  write: %i\n\
+  take: %i\n\
+  on_data_available: %i\n",
+    dmock->cnt_type_qos_supported,
+    dmock->cnt_create_topic,
+    dmock->cnt_delete_topic,
+    dmock->cnt_deinit,
+    dmock->cnt_get_node_id,
+    dmock->cnt_supported_features,
+    dmock->cnt_create_endpoint,
+    dmock->cnt_delete_endpoint,
+    dmock->cnt_request_loan,
+    dmock->cnt_write,
+    dmock->cnt_take,
+    dmock->cnt_on_data_available
+  );
+}
+
 CU_Test(ddsc_psmx, one_writer, .timeout = 240)
 {
   failed = false;
@@ -1332,6 +1407,9 @@ CU_Test (ddsc_psmx, basic)
 /// - Create some entities
 /// - Check if shared memory is enabled.
 /// - Expectation: Shared memory is enabled iff the psmx interface supports it (use queried value).
+/// - Delete the domain
+/// - Check the function call counts of the dummy psmx.
+/// - Expectation: The counts match expectations. In particular, create counts must match their delete counterpart.
 /// 
 /// - Create a configuration with a psmx interface capable of shared memory and don't specify a locator.
 /// - Create a domain using this configuration.
@@ -1341,12 +1419,19 @@ CU_Test (ddsc_psmx, basic)
 /// - Create some entities
 /// - Check if shared memory is enabled.
 /// - Expectation: Shared memory is enabled iff the psmx interface supports it (use queried value).
+/// - Delete the domain
+/// - Check the function call counts of the dummy psmx.
+/// - Expectation: The counts match expectations. In particular, create counts must match their delete counterpart.
 /// 
-/// - The above two cases (with and without locator) are repeated for a
-///   psmx interface that supports shared memory, and a psmx interface that doesn't.
-/// 
-CU_Test(ddsc_psmx, shared_memory)
+CU_Test(ddsc_psmxif, shared_memory)
 {
+  const char* CDDS_PSMX_NAME = NULL;
+  ddsrt_getenv("CDDS_PSMX_NAME", &CDDS_PSMX_NAME);
+  ddsrt_setenv("CDDS_PSMX_NAME", "dummy"); // Make `create_participant()` use the dummy psmx.
+  dummy_mockstats_t dmock, dmock_expected;
+  char strbuf1[512];
+  char strbuf2[512];
+  size_t strbuf_size = sizeof(strbuf1);
   {
     // Check that the data types I'm planning to use are actually suitable for use with shared memory.
     dds_data_type_properties_t props;
@@ -1384,10 +1469,12 @@ CU_Test(ddsc_psmx, shared_memory)
       dds_domain* dom = NULL;
       CU_ASSERT_FATAL(domain_pin(domain, &dom));
       CU_ASSERT_FATAL(dom->psmx_instances.length == 1); // There shall be exactly one psmx instance.
-      dds_psmx_t* psmx_ptr = dom->psmx_instances.instances[0];
+      struct dummy_psmx* dpsmx = (struct dummy_psmx*)dom->psmx_instances.instances[0];
+      dpsmx->mockstats_get_ownership(&dmock);
+
       // The psmx must have an instance_name that is not an empty string.
-      CU_ASSERT_FATAL(psmx_ptr->instance_name != NULL && strcmp(psmx_ptr->instance_name, "") != 0);
-      supports_shared_memory = ((psmx_ptr->ops.supported_features(psmx_ptr) & DDS_PSMX_FEATURE_SHARED_MEMORY) == DDS_PSMX_FEATURE_SHARED_MEMORY);
+      CU_ASSERT_FATAL(dpsmx->c.instance_name != NULL && strcmp(dpsmx->c.instance_name, "") != 0);
+      supports_shared_memory = ((dpsmx->c.ops.supported_features(&dpsmx->c) & DDS_PSMX_FEATURE_SHARED_MEMORY) == DDS_PSMX_FEATURE_SHARED_MEMORY);
       domain_unpin(dom);
     }
     dds_entity_t writer1, reader1, writer2, reader2;
@@ -1416,6 +1503,7 @@ CU_Test(ddsc_psmx, shared_memory)
       CU_ASSERT_FATAL(reader2 > 0);
       dds_delete_qos(qos);
     }
+
     {
       // Check that shared memory is enabled when it should, and not enabled when it shouldn't.
       bool psmx_enabled;
@@ -1436,6 +1524,29 @@ CU_Test(ddsc_psmx, shared_memory)
       CU_ASSERT_FATAL(dds_is_shared_memory_available(reader2) == supports_shared_memory);
     }
     dds_delete(domain);
+
+    // Check number of calls against expected counts.
+    memset(&dmock_expected, 0, sizeof(dummy_mockstats_t));
+    dmock_expected.cnt_type_qos_supported = 10;
+    dmock_expected.cnt_create_topic = 2;
+    dmock_expected.cnt_delete_topic = 2;
+    dmock_expected.cnt_deinit = 1;
+    dmock_expected.cnt_get_node_id = 1;
+    dmock_expected.cnt_supported_features = 5;
+    dmock_expected.cnt_create_endpoint = 4;
+    dmock_expected.cnt_delete_endpoint = 4;
+    dmock_expected.cnt_on_data_available = 2; // This is perhaps unexpected, see below.
+    /*
+    NOTE: The `on_data_available` is triggered in
+    `dds_create_reader_int()`, even if no samples have been received at all.
+    */
+    uint32_t cmp = dummy_mockstats_cmp(&dmock, &dmock_expected);
+    if( cmp != 0 ){
+      dummy_mockstats_tostring(&dmock, strbuf1, strbuf_size);
+      dummy_mockstats_tostring(&dmock_expected, strbuf2, strbuf_size);
+      printf("actual calls:\n%s\nexpected calls:\n%s\n", strbuf1, strbuf2);
+      CU_ASSERT_FATAL(cmp == 0);
+    }
   }
 }
 
--- a/src/core/ddsc/tests/psmx.c
+++ b/src/core/ddsc/tests/psmx.c
@@ -64,17 +64,35 @@ static void fail_too_much_data (void) { fail (); }
 #define TRACE_CATEGORY "discovery"
 #endif
 
-static dds_entity_t create_participant (dds_domainid_t int_dom)
+/**
+ * @brief Create a participant object
+ * 
+ * @param[in] int_dom the domain id
+ * @param[in] locator an array of 16 bytes, NULL if not used
+ * @return dds_entity_t the participant
+ */
+static dds_entity_t create_participant (dds_domainid_t int_dom, const uint8_t* locator)
 {
   assert (int_dom < MAX_DOMAINS);
-  const unsigned char *l = psmx_locators[int_dom].a;
   char *configstr;
+  char locator_str[74];
+  if ( locator == NULL ) {
+    locator_str[0] = 0;
+  } else {
+    const uint8_t *l = locator;
+    snprintf(
+      locator_str,
+      sizeof(locator_str),
+      "LOCATOR=%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x;",
+      l[0], l[1], l[2], l[3], l[4], l[5], l[6], l[7], l[8], l[9], l[10], l[11], l[12], l[13], l[14], l[15]
+    );
+  }
   ddsrt_asprintf (&configstr, "\
 ${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}\
 <General>\
   <AllowMulticast>spdp</AllowMulticast>\
   <Interfaces>\
-    <PubSubMessageExchange name=\"${CDDS_PSMX_NAME:-cdds}\" library=\"psmx_${CDDS_PSMX_NAME:-cdds}\" priority=\"1000000\" config=\"LOCATOR=%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x;SERVICE_NAME=psmx%d;KEYED_TOPICS=true;\" />\
+    <PubSubMessageExchange name=\"${CDDS_PSMX_NAME:-cdds}\" library=\"psmx_${CDDS_PSMX_NAME:-cdds}\" priority=\"1000000\" config=\"%sSERVICE_NAME=psmx%d;KEYED_TOPICS=true;\" />\
   </Interfaces>\
 </General>\
 <Discovery>\
@@ -86,8 +104,8 @@ ${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}\
   <OutputFile>cdds.log.%d</OutputFile>\
 </Tracing>\
 ",
-    l[0], l[1], l[2], l[3], l[4], l[5], l[6], l[7], l[8], l[9], l[10], l[11], l[12], l[13], l[14], l[15],
-    (int) l[0],   // This prevents Iceoryx and Cyclone-based plugins from forwarding across the "network"
+    locator_str,
+    locator ? (int) locator[0] : -1,   // This prevents Iceoryx and Cyclone-based plugins from forwarding across the "network"
     (int) int_dom // log file name
   );
   char *xconfigstr = ddsrt_expand_envvars (configstr, int_dom);
@@ -100,28 +118,18 @@ ${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}\
   return pp;
 }
 
-static void config_psmx_free_content(struct ddsi_config_psmx* cfg)
-{
-  assert(cfg);
-  ddsrt_free(cfg->name);
-  ddsrt_free(cfg->library);
-  ddsrt_free(cfg->config);
-}
-
-static void config_psmx_elm_free(struct ddsi_config_psmx_listelem* elm)
-{
-  assert(elm);
-  config_psmx_free_content(&elm->cfg);
-  ddsrt_free(elm);
-}
-
 static bool domain_pin(dds_entity_t domain, dds_domain** dom_out)
 {
   dds_entity* x = NULL;
   dds_return_t rc = dds_entity_pin(domain, &x);
-  if ( rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_DOMAIN ) {
-    *dom_out = (dds_domain*)x;
-    return true;
+  if ( rc == DDS_RETCODE_OK ) {
+    if( dds_entity_kind(x) == DDS_KIND_DOMAIN ) {
+      *dom_out = (dds_domain*)x;
+      return true;
+    } else {
+      dds_entity_unpin(x);
+      return false;
+    }
   }
   return false;
 }
@@ -141,24 +149,6 @@ static bool domain_get_psmx_locator(dds_entity_t domain, ddsi_locator_t* l_out)
     domain_unpin(dom);
   }
   return ret;
-}
-
-static bool domain_has_psmx_instance_name(dds_entity_t domain, const char* inst_name)
-{
-  dds_domain* dom = NULL;
-  bool ret = domain_pin(domain, &dom);
-  bool match = false;
-  if ( ret ) {
-    uint32_t len = dom->psmx_instances.length;
-    for (uint32_t idx = 0; idx < len; ++idx) {
-      if ( strcmp(dom->psmx_instances.instances[idx]->instance_name, inst_name) == 0 ) {
-        match = true;
-        break;
-      }
-    }
-    domain_unpin(dom);
-  }
-  return ret && match;
 }
 
 struct tracebuf {
@@ -697,7 +687,7 @@ static void dotest (const dds_topic_descriptor_t *tpdesc, const void *sample, bo
   create_unique_topic_name ("test_psmx", topicname, sizeof (topicname));
   for (int i = 0; i < MAX_DOMAINS; i++)
   {
-    pp[i] = create_participant ((dds_domainid_t) i); // FIXME: vary shm_enable for i > 0
+    pp[i] = create_participant ((dds_domainid_t) i, psmx_locators[(dds_domainid_t) i].a); // FIXME: vary shm_enable for i > 0
     tp[i] = dds_create_topic (pp[i], tpdesc, topicname, NULL, NULL);
     CU_ASSERT_FATAL (tp[i] > 0);
     gvs[i] = get_domaingv (pp[i]);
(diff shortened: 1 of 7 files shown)


firuze-cagliyan (approved):
Review ok.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1326,378 @@ CU_Test (ddsc_psmx, basic)
...
+        psmx_enabled = endpoint_has_psmx_enabled(writer2);
+        CU_ASSERT_FATAL(psmx_enabled);
+        CU_ASSERT_FATAL(dds_is_shared_memory_available(writer2));
+
+        psmx_enabled = endpoint_has_psmx_enabled(reader2);
+        CU_ASSERT_FATAL(psmx_enabled);
+        CU_ASSERT_FATAL(dds_is_shared_memory_available(reader2));
+      }
+      {
+        // Write and read samples
```
dpotman, line 1473: I wouldn't include writing and reading samples in this test, because it's about testing the `is_shared_memory` flag. Maybe move this to a different test case?

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1326,378 @@ CU_Test (ddsc_psmx, basic)
...
+            cnt += result;
+          }
+          CU_ASSERT_FATAL(match);
+          CU_ASSERT_FATAL(cnt == sample_cnt);
+        }
+      }
+      dds_delete(domain);
+    }
+    {
+      // Test case using a config without specifying a locator.
```
dpotman, line 1566: Same as above: use the helper function and avoid assumptions on plugin internals.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1326,378 @@ CU_Test (ddsc_psmx, basic)
...
+        cfg.psmx_instances->next = NULL;
+        domain = dds_create_domain_with_rawconfig(domainId, &cfg);
+        config_psmx_elm_free(cfg.psmx_instances);
+      }
+      CU_ASSERT_FATAL(domain > 0);
+      CU_ASSERT_FATAL(domain_has_psmx_instance_name(domain, "CycloneDDS-IOX-PSMX"));
+      dds_entity_t participant = dds_create_participant(domainId, NULL, NULL);
+      CU_ASSERT_FATAL(participant > 0);
+      {
+        // Check that the second half of the locator is all zeros.
```
dpotman, line 1587: This assumption is based on details in the `iox` plugin implementation; the value assigned to the locator is the return value of the `get_node_id` call on the PSMX interface that returns a `dds_psmx_node_identifier_t`, which is 16 bytes. This happens to be an 64 bit value with trailing zero's, but could in theory be anything.

src/core/ddsc/tests/psmx.c
```diff
@@ -98,6 +99,67 @@ static dds_entity_t create_participant (dds_domainid_t int_dom)
...
+
+static bool domain_pin(dds_entity_t domain, dds_domain** dom_out)
+{
+  dds_entity* x = NULL;
+  dds_return_t rc = dds_entity_pin(domain, &x);
+  if ( rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_DOMAIN ) {
+    *dom_out = (dds_domain*)x;
+    return true;
+  }
+  return false;
```
dpotman, line 125: This should unpin the entity if `rc == DDS_RETCODE_OK`

dpotman (commented):
A few things that need a fix or can be improved, see review comments below.

src/core/ddsc/tests/psmx.c
firuze-cagliyan: I thought as 'setting the flag' only is not enough and we should also see that it works as it should be so that we can read/write samples when shared memory is enabled. Putting in another test case is also an option but it means that we should repeat some beginning steps of this test case in that one, so I think keeping these steps here is easier for us.

dpotman: We have other tests (in this same source file) that do test communication over psmx, including over iox (so shared memory enabled). In fact, in certified, there is only PSMX communication and we only test over IOX, so all tests that do any communication test over shared memory. So I think we don't need to test that here.

src/core/ddsc/tests/psmx.c
mvandenhoek: The unpin happens in `domain_unpin()`.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1326,378 @@ CU_Test (ddsc_psmx, basic)
...
+    props = dds_stream_data_types(PsmxType1_desc.m_ops);
+    CU_ASSERT_FATAL((props & DDS_DATA_TYPE_IS_MEMCPY_SAFE) == DDS_DATA_TYPE_IS_MEMCPY_SAFE);
+  }
+
+  const dds_domainid_t domainId = 42;
+  char* CDDS_PSMX_NAME = ddsrt_expand_envvars("${CDDS_PSMX_NAME:-cdds}", domainId);
+  bool using_psmx_iox = (strcmp(CDDS_PSMX_NAME, "iox") == 0);
+  ddsrt_free(CDDS_PSMX_NAME);
+
+  if ( using_psmx_iox ) {
```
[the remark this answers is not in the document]
mvandenhoek: I checked it, but `dds_is_shared_memory_available()` internally uses `dds_psmx_supported_features()`. In other words, if I call `dds_psmx_supported_features()` first, it can't help but give the same result for `dds_is_shared_memory_available()`. So it seems a bit pointless to me to do it that way.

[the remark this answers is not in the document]
dpotman: What I meant is that the `else` block (from line 1636) can be completely removed (and so the `if` on line 1376) if you parameterize the "is shared memory available". That can be done by either hard-coding that the `iox` implementation uses shared memory, and `cdds` not, or by requesting the features from the plugin. The latter will also work if we decide to add more implementations to run this test on, so I think that's the way to go. Calling `dds_is_shared_memory_available()` on the endpoints could be used to check if the returned value is consistent with the features the PSMX instance claims to support.

src/core/ddsc/tests/psmx.c
dpotman: But `domain_unpin` is not called in case this function returns `false` because the entity kind is incorrect.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1326,378 @@ CU_Test (ddsc_psmx, basic)
...
+        cfg.psmx_instances->cfg.library = ddsrt_strdup("psmx_iox");
+        cfg.psmx_instances->cfg.config = ddsrt_strdup(config_str); // config including a locator
+        cfg.psmx_instances->cfg.priority.isdefault = 1;
+        cfg.psmx_instances->cfg.priority.value = 1000000;
+        cfg.psmx_instances->next = NULL;
+        domain = dds_create_domain_with_rawconfig(domainId, &cfg);
+        config_psmx_elm_free(cfg.psmx_instances);
+      }
+      CU_ASSERT_FATAL(domain > 0);
+      CU_ASSERT_FATAL(domain_has_psmx_instance_name(domain, "CycloneDDS-IOX-PSMX"));
```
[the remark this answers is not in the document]
mvandenhoek: The `SERVICE_NAME` ends up in `_service_name` in the `struct iox_psmx`, which inherits from `dds_psmx_t`.
The `iox_create_psmx()` returns the `struct dds_psmx`, which doesn't contain the `_service_name` (it is in the derived class, not the base class). I couldn't find any getter functions for this. Perhaps it should be added to `dds_psmx_ops_t` ?

mvandenhoek:
I think this addresses the comments, but I bumped into a problem with the cdds_psmx. When you run the same test twice in a loop (after cleaning up all entities of course, by means of `dds_delete(domain);`), it runs into an assert when the test code calls `dds_create_topic()` on the second iteration of the test.
```psmx_cdds_impl.c:143: cdds_psmx_create_topic: Assertion `g_domain >= 0' failed.```

The value of `g_domain` is `-4`, which stands for returncode `DDS_RETCODE_PRECONDITION_NOT_MET`.

The assert happens here:
https://github.com/eclipse-cyclonedds/cyclonedds/blob/f724b4fbdaba6939c71e3412bdeec29d1fa94946/src/core/ddsc/tests/psmx_cdds_impl.c#L135-L145

The source of the returncode:
https://github.com/eclipse-cyclonedds/cyclonedds/blob/f724b4fbdaba6939c71e3412bdeec29d1fa94946/src/core/ddsc/src/dds_domain.c#L224-L241

It executes `dom = dds_domain_find_locked (id);`, then goes to `domh = DDS_RETCODE_PRECONDITION_NOT_MET;`.

**Comment by mvandenhoek** [comment]
Added a dummy psmx interface, which is now used by the shared memory test.

File: src/core/ddsc/tests/psmx.c [hunk 7]
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+      ddsi_locator_t locator_out;
+      CU_ASSERT_FATAL(domain_get_psmx_locator(domain, &locator_out));
+      CU_ASSERT_FATAL(memcmp(locator_in, locator_out.address, sizeof(locator_out.address)) == 0);
+    }
+
+    {
+      // Query whether shared memory is supported.
+      dds_domain* dom = NULL;
+      CU_ASSERT_FATAL(domain_pin(domain, &dom));
+      CU_ASSERT_FATAL(dom->psmx_instances.length == 1); // There shall be exactly one psmx instance.
```
eboasson, line 1470: That the dummy plugin has indeed been loaded and initialised can easily be verified without checking this by relying on your `dummy_mockstats_t`.

So that leaves only the question whether no other plugins have been loaded. I don't find this test very convincing that will only ever load a single one if there is only a single one in the configuration, and I can't quite think of a test that would convince me.

I would prefer code inspection that shows the set of plugins loaded corresponds to the set of configured plugins, and a configuration test that shows it rejects a configuration with two plugins.

That would also eliminate the need to look inside the `dds_domain` for this.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+      struct dummy_psmx* dpsmx = (struct dummy_psmx*)dom->psmx_instances.instances[0];
+      dpsmx->mockstats_get_ownership(&dmock);
```
eboasson, line 1472: I don't even understand this. Well, maybe I do, but it seems overly complicated for what you're trying to achieve.

This is just test code. We know there is exactly one such plugin loaded. We know the plugin. We can even *link against it from the test code* (these are just libraries after all).

So instead of carefully extracting a pointer to the `dds_psmx` struct, why not store it in a global variable inside the plugin and have the plugin export a function that returns the address of it? Then you just declare that function in `psmx_dummy_public.h` and call it.

Indeed, instead of doing that with the `dds_psmx` one, I would just let it store a global `dummy_mockstats_t` and return a pointer to that. No need to fiddle with the innards.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+
+      // The psmx must have an instance_name that is not an empty string.
+      CU_ASSERT_FATAL(dpsmx->c.instance_name != NULL && strcmp(dpsmx->c.instance_name, "") != 0);
```
eboasson, line 1475: The instance name you could also check in a (slightly convoluted way) by specifying the PSMX plugins to use when create a reader/writer: create one and read back the PSMX instances QoS. It should work anyway.

It is not really *better* but it decouples the test a bit more from the details gory implementation details.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+      supports_shared_memory = ((dpsmx->c.ops.supported_features(&dpsmx->c) & DDS_PSMX_FEATURE_SHARED_MEMORY) == DDS_PSMX_FEATURE_SHARED_MEMORY);
```
eboasson, line 1476: I would do this by calling `dds_is_shared_memory_available` on a reader/writer. What you want to know is that that function calls the plugin's `supported_features` and that its return value corresponds to what the plugin replied.

This merely checks the dummy plugin implementation, not the core.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+      CU_ASSERT_FATAL(reader2 > 0);
+      dds_delete_qos(qos);
+    }
+
+    {
+      // Check that shared memory is enabled when it should, and not enabled when it shouldn't.
+      bool psmx_enabled;
+      psmx_enabled = endpoint_has_psmx_enabled(writer1);
+      CU_ASSERT_FATAL(psmx_enabled);
+      CU_ASSERT_FATAL(dds_is_shared_memory_available(writer1) == supports_shared_memory);
```
eboasson, line 1511: So here is the aforementioned test for `supported_features` already. Of course you do have to run through this code *twice*, because otherwise `dds_is_shared_memory_available` could get away with returning a constant.

eboasson (commented):
Thank you for this @mvandenhoek but I'm afraid I have some different ideas on how to go about testing whether the functions on the PSMX interface get invoked (and the results used) correctly by the core library. I've tried to sketch these ideas in the comments.

I won't say it has to be done the way I sketched, but on the whole I think it makes more sense the approach you chose, specifically also because it makes fewer assumptions and keeps the different tests a bit cleaner.

src/core/ddsc/tests/psmx.c
```diff
@@ -62,17 +64,35 @@ static void fail_too_much_data (void) { fail (); }
 #define TRACE_CATEGORY "discovery"
 #endif
 
-static dds_entity_t create_participant (dds_domainid_t int_dom)
+/**
+ * @brief Create a participant object
```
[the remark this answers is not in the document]
mvandenhoek: This was the least intrusive way I could think of to reuse this function. That is, without changing the interface too much, and without changing the behavior in a way that affects the other testcases. But if the dummy psmx testcase goes into another file as suggested in your earlier comment, then there is no reason to use the "ductape approach" with `ddsrt_setenv()`, so the problem will solve itself.

src/core/ddsc/tests/psmx.c
mvandenhoek: The issue with using a global variable inside the plugin, is that it's gone as soon as you delete the domain, so you would miss all of the calls done during cleanup.

src/core/ddsc/tests/psmx.c
```diff
@@ -1264,6 +1391,164 @@ CU_Test (ddsc_psmx, basic)
...
+    dmock_expected.cnt_get_node_id = 1;
+    dmock_expected.cnt_supported_features = 5;
+    dmock_expected.cnt_create_endpoint = 4;
+    dmock_expected.cnt_delete_endpoint = 4;
+    dmock_expected.cnt_on_data_available = 2; // This is perhaps unexpected, see below.
+    /*
+    NOTE: The `on_data_available` is triggered in
+    `dds_create_reader_int()`, even if no samples have been received at all.
+    */
+    uint32_t cmp = dummy_mockstats_cmp(&dmock, &dmock_expected);
```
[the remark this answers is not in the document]
mvandenhoek: I considered the way you suggest, but I thought this was a nicer way to do it. Probably more a matter of taste. The conversion to string is primarily for the actual values, and only intended for inspection/debugging purposes, not for the "good path", when the tests are all passed.

src/core/ddsc/tests/psmx.c
eboasson: `dlclose` doesn't unconditionally unload the library, it only unloads it when there is no further reference to it. `dlopen` is one way to get such a reference, linking the executable with the library is another.

So you could simply add the dummy plugin to the set of libraries to link the test executable with, or you could let the test also `dlopen`/`dlcose` it. (I'd probably go for the first option, but they are equivalent.)

src/core/ddsc/tests/psmx_dummy_impl.c
```diff
@@ -0,0 +1,274 @@
...
+}
+
+static struct dds_psmx_topic* dummy_psmx_create_topic(
+  struct dds_psmx* psmx,
+  const char* topic_name,
+  const char* type_name,
+  dds_data_type_properties_t data_type_props
+) {
+  (void)data_type_props;
+  struct dds_psmx_topic* topic = dds_alloc(sizeof(struct dds_psmx_topic));
```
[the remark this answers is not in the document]
mvandenhoek: Am I interpreting this correctly that the test code should prepare some data in the global variable (which can be accessed by both the test code and the plugin), before calling `dds_create_topic()`, and then let the plugin verify from inside `dummy_psmx_create_topic()`, that it gets a request for the correct topic?

mvandenhoek:
> Thank you for this @mvandenhoek but I'm afraid I have some different ideas on how to go about testing whether the functions on the PSMX interface get invoked (and the results used) correctly by the core library. I've tried to sketch these ideas in the comments.
> 
> I won't say it has to be done the way I sketched, but on the whole I think it makes more sense the approach you chose, specifically also because it makes fewer assumptions and keeps the different tests a bit cleaner.

I believe these changes address your comments.

src/core/ddsc/tests/psmx.c
mvandenhoek: > ..., and a configuration test that shows it rejects a configuration with two plugins.

I've added this test as well, and it fails because `dds_create_domain()` doesn't reject a configuration with two plugins. Instead, it simply falls out of the `while` loop in `dds_pubsub_message_exchange_init()`, having only processed the first interface. It falls out due to condition `domain->psmx_instances.length < DDS_MAX_PSMX_INSTANCES`, with values `DDS_MAX_PSMX_INSTANCES == 1` and `domain->psmx_instances.length == 1` after processing the first plugin.

https://github.com/eclipse-cyclonedds/cyclonedds/blob/2ad37f4cc3d545149f680d6a8b0ff31e0ea84edc/src/core/ddsc/src/dds_psmx.c#L270-L299

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+ * @param[in] int_dom the domain id
+ * @param[in] cdds_psmx_name the name of the psmx interface to use
+ * @param[in] locator an array of 16 bytes, NULL if not used
+ * @return dds_entity_t the participant
+ */
+static char* create_config(dds_domainid_t int_dom, const char* cdds_psmx_name, const uint8_t* locator)
+{
+  char *configstr;
+  char locator_str[74];
+  if ( locator == NULL ) {
```
eboasson, line 60: These complications with locators are no longer used in any meaningful way, they should be removed. A single, fixed, configuration string suffices for the purpose of the test.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}\
+<General>\
+  <AllowMulticast>spdp</AllowMulticast>\
+  <Interfaces>\
+    <PubSubMessageExchange name=\"%s\" library=\"psmx_%s\" priority=\"1000000\" config=\"%sSERVICE_NAME=psmx%d;KEYED_TOPICS=true;\" />\
+  </Interfaces>\
+</General>\
+<Discovery>\
+  <Tag>${CYCLONEDDS_PID}</Tag>\
+  <ExternalDomainId>0</ExternalDomainId>\
```
eboasson, line 81: This just runs a single instance of Cyclone, no need for the tricks that allow you to pretend there's a network in between when everything is in a single process.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+</Discovery>\
+<Tracing>\
+  <OutputFile>cdds.log.%d</OutputFile>\
+</Tracing>\
+",
+    cdds_psmx_name,
+    cdds_psmx_name,
+    locator_str,
+    locator ? (int) locator[0] : -1, // This prevents plugins from forwarding across the "network".
```
eboasson, line 90: This comment doesn't apply at all.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      psmx_enabled = (rd->m_endpoint.psmx_endpoints.length > 0);
+      break;
+    }
+    case DDS_KIND_WRITER: {
+      struct dds_writer const * const wr = (struct dds_writer *) x;
+      psmx_enabled = (wr->m_endpoint.psmx_endpoints.length > 0);
+      break;
+    }
+    default: {
+      CU_ASSERT_FATAL (dds_entity_kind (x) == DDS_KIND_READER || dds_entity_kind (x) == DDS_KIND_WRITER);
```
eboasson, line 118: Writing this assert kinda defeats the purpose of using a `switch` to begin with. Then it makes more sense to do `if{READER}else{CU_ASSERT_FATAL(WRITER)`

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+/// @brief Check that creating a domain with more than one psmx interface fails.
+/// @methodology
+/// - Create a config string with two psmx interfaces.
+/// - Try to create a domain using this config string.
+/// - Expectation: Failed to create the domain.
+/// 
+CU_Test(ddsc_psmxif, config_multiple_psmx)
+{
+  dds_domainid_t domainId = 0;
+  const char* cdds_psmx_name1 = "dummy";
```
eboasson, line 185: All of this ends up being a constant string. Why not simply write a string constant and pass it to `ddsrt_expand_envvars`?

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+  }
+
+  const dds_domainid_t domainId = 0;
+  bool supports_shared_memory_expected = true;
+  dds_psmx_endpoint_t* psmx_endpt_expected = NULL;
+  dds_psmx_topic_t* psmx_topic_expected = NULL;
+  bool specify_locator[] = {true, false};
+  uint8_t locator_in[16];
+  memset(locator_in, 0x0, sizeof(locator_in)); // Avoid warning 'uninitialized value'.
+  ((uint64_t*)locator_in)[0] = (uint64_t)0x4a4d203df6996395;
```
eboasson, line 275: These are not meaningful anymore and so should go — but I also want to point out that `uint8_t a[16];*((uint64_t*)a)=...` risks misaligned access, and with that, undefined behaviour. In this case, rather than a `memset` followed by two 64-bit integer assignments, you could just have written out the value as an initializer, or.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+  ((uint64_t*)locator_in)[1] = (uint64_t)0xe1412fbecc2de4b6;
+
+  int N = sizeof(specify_locator) / sizeof(specify_locator[0]);
+  for (int i = 0; i < N; ++i)
```
eboasson, line 279: This loop then becomes superfluous. That saves me from having to comment that otherwise the `N` would at the very least best be made a `const int` and in practice better be eliminated altogether.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      // Query whether shared memory is supported.
+      dds_domain* dom = NULL;
+      {
+        dds_entity* x = NULL;
+        dds_return_t rc = dds_entity_pin(domain, &x);
+        CU_ASSERT_FATAL(rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_DOMAIN);
+        dom = (dds_domain*)x;
+      }
+      assert(dom->psmx_instances.length >= 1);
+      struct dummy_psmx* dpsmx = (struct dummy_psmx*)dom->psmx_instances.instances[0];
```
eboasson, line 297: I have commented before that I don't like this peeking deep inside the internal data if there's no good reason to. The entire purpose of suggesting linking the dummy implementation into the test executable was so you could directly use global variables or call functions inside the plugin.

Exporting `dummy_mockstats_get_ptr` as suggested replaces this entire block with a library call and an assert, much simpler and cleaner.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      dmock = dpsmx->mockstats_get_ptr();
+      CU_ASSERT_FATAL(dmock->cnt_create_psmx == 1); // Confirm the dummy psmx has been loaded.
+      dds_entity_unpin(&dom->m_entity);
+    }
+    // Check that the config string passed to `dds_create_domain()` has been correctly forwarded to the dummy psmx.
+    CU_ASSERT_FATAL(strstr(configstr_in, dmock->config) != NULL); // dmock->config is a substring of the original xml.
```
eboasson, line 303: Well, yes, it is a substring. But we know *exactly* what contents it has: those that were passed in. A exact check is just as easy and doesn't require a comment. So that is what it should be.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      create_unique_topic_name("shared_memory", topicname, sizeof(topicname));
+      psmx_topic_expected = (dds_psmx_topic_t*)dmock->topics._buffer + dmock->topics._length;
+      dds_entity_t topic1 = dds_create_topic(participant, &SC_Model_desc, topicname, NULL, NULL);
+      CU_ASSERT_FATAL(topic1 > 0);
+      {
+        dds_entity* x = NULL;
+        dds_return_t rc = dds_entity_pin(topic1, &x);
+        CU_ASSERT_FATAL(rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_TOPIC);
+        struct dds_psmx_topics_set* topics_set = &((dds_topic*)x)->m_ktopic->psmx_topics;
+        CU_ASSERT_FATAL(topics_set->length == 1 && topics_set->topics[0] == psmx_topic_expected);
```
eboasson, line 321: This is one way of verifying that the returned address is picked up correctly. Still, I prefer my original suggestion: we know the address must be passed to any `dummy_psmx_create_endpoint` call that is expected based on creating an endpoint of topic `topic1`.

And so rather I would much rather not dig so deeply in the internals and instead check in `dummy_psmx_create_endpoint` that the expectations are met. That eliminates a fair amount of ugly code here.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      }
+
+      psmx_endpt_expected = (dds_psmx_endpoint_t*)dmock->endpoints._buffer + dmock->endpoints._length;
+      writer1 = dds_create_writer(participant, topic1, NULL, NULL);
+      CU_ASSERT_FATAL(writer1 > 0);
+      {
+        dds_entity* x = NULL;
+        dds_return_t rc = dds_entity_pin(writer1, &x);
+        CU_ASSERT_FATAL(rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_WRITER);
+        struct dds_psmx_endpoints_set* endpt_set = &((dds_writer*)x)->m_endpoint.psmx_endpoints;
```
eboasson, line 332: Here we can use the same: if one calls `write` or `request_loan`, one must end up in the dummy plugin with the correct endpoint.

Indeed, simply deleting the writer will also give that information.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+        CU_ASSERT_FATAL(endpt_set->length == 1 && endpt_set->endpoints[0] == psmx_endpt_expected);
+        dds_entity_unpin(x);
+      }
+
+      psmx_endpt_expected = (dds_psmx_endpoint_t*)dmock->endpoints._buffer + dmock->endpoints._length;
+      reader1 = dds_create_reader(participant, topic1, NULL, NULL);
+      CU_ASSERT_FATAL(reader1 > 0);
+      {
+        // Check the dummy psmx instance_name.
+        dds_qos_t* qos = dds_create_qos();
```
eboasson, line 342: Why check this here, but not for the writer?

From a test structure perspective, it would also make more sense to check these in a separate test. With appropriately factored out entity creation code it takes only a few lines extra to turn it into a separate test case.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+        CU_ASSERT_FATAL(strs_len == 1 && strcmp(strs[0], "dummy_psmx") == 0);
+        free_strings(strs_len, strs);
+        dds_delete_qos(qos);
+      }
+      {
+        dds_entity* x = NULL;
+        dds_return_t rc = dds_entity_pin(reader1, &x);
+        CU_ASSERT_FATAL(rc == DDS_RETCODE_OK && dds_entity_kind(x) == DDS_KIND_READER);
+        struct dds_psmx_endpoints_set* endpt_set = &((dds_reader*)x)->m_endpoint.psmx_endpoints;
+        CU_ASSERT_FATAL(endpt_set->length == 1 && endpt_set->endpoints[0] == psmx_endpt_expected);
```
eboasson, line 356: I think the only practical way to verify this in my preferred implementation is to check in the plug-in on deleting the reader. But that needs to be done anyway.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+        dds_entity_unpin(x);
+      }
+
+      create_unique_topic_name("shared_memory", topicname, sizeof(topicname));
```
eboasson, line 360: All things `topic2`: more of the same.

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+        CU_ASSERT_FATAL(endpt_set->length == 1 && endpt_set->endpoints[0] == psmx_endpt_expected);
+        dds_entity_unpin(x);
+      }
+    }
+
+    {
+      // Check that shared memory is enabled when it should, and not enabled when it shouldn't.
+      bool psmx_enabled;
+      psmx_enabled = endpoint_has_psmx_enabled(writer1);
+      CU_ASSERT_FATAL(psmx_enabled);
```
eboasson, line 402: This is covered already: the preceding tests are stricter than this one. (I would like those preceding ones done differently, but the net effect would still be good enough to say that this test is superfluous.)

src/core/ddsc/tests/psmxif.c
```diff
@@ -0,0 +1,441 @@
...
+      CU_ASSERT_FATAL(dds_is_shared_memory_available(writer1) == supports_shared_memory_expected);
```
eboasson, line 403: `supports_shared_memory_expected` is semantically a variable, but it is assigned only once, at declaration and its address is never taken. So it is a constant. I would prefer `CU_ASSERT_FATAL(dds_is_shared_memory_available(writer1))` in such a case.

However, here, you're trying to test `dds_is_shared_memory_available` but with the current set-up, an implementation of `dds_is_shared_memory_available` that simply returns `true` will pass. It is therefore not a good test.

A much better way would be to:
* Create a reader/writer that does not involve PSMX, verify that it returns false. (This would best be done in a different test.)
* Run this test once with the plugin claiming to support shared memory and verifying it returns true. (This is the current case.)
* Run this test once with the plugin claiming to *not* support shared memory and verifying it returns false. (This case is quite obviously missing.)

With the "direct" access to the dummy plugin, it is trivial to make the plugin claim one or the other.

src/core/ddsc/tests/psmxif.c
mvandenhoek: This function was copied from `psmx.c`. I suppose then it should be modified in both places.

src/core/ddsc/tests/psmxif.c
mvandenhoek: The locator is now part of the config string.

src/core/ddsc/tests/psmxif.c
mvandenhoek: I've also added a check that the loan pointer from `request_loan()` ends up in `write()`.

eboasson (commented):
Sorry that I forgot to review this again in a reasonable time.

It is mostly in a mergeable state as far as the contents of the PR are concerned. What is still a problem is that it fails to build properly on Windows (https://dev.azure.com/eclipse-cyclonedds/fc8c2212-deca-471a-94ad-bc26de82c2b6/_apis/build/builds/6303/logs/128) because of a DLL linkage issue — that is not something caused by recent changes on `master`.

Another thing is that `CUnit_ddsc_psmxif_config_multiple_psmx` fails, presumably because the core doesn't reject the configuration as the test expects. For now, it is probably best to reject such a configuration, but I am not going to merge this PR until I have a fix in place. (If you provide a fix on this PR that would be appreciated, but otherwise I'll do it myself.)

The other good thing is that I noticed some things in the code that really need cleaning up in the core, in part because the division of labour is weird and in part because it is just written clumsily.

So it has served a purpose 🙂

mvandenhoek:
@eboasson , thanks for having a look.

> It is mostly in a mergeable state as far as the contents of the PR are concerned. What is still a problem is that it fails to build properly on Windows (https://dev.azure.com/eclipse-cyclonedds/fc8c2212-deca-471a-94ad-bc26de82c2b6/_apis/build/builds/6303/logs/128) because of a DLL linkage issue — that is not something caused by recent changes on `master`.
 
> Another thing is that `CUnit_ddsc_psmxif_config_multiple_psmx` fails, presumably because the core doesn't reject the configuration as the test expects. For now, it is probably best to reject such a configuration, but I am not going to merge this PR until I have a fix in place. (If you provide a fix on this PR that would be appreciated, but otherwise I'll do it myself.)

Yes, it fails the config multiple psmx test exactly as intended, so the test has served its purpose. I'll put these on my todo list.

mvandenhoek:
It looks like all of the builds are successful now. There is one platform with one test failure unrelated to this PR.

eboasson (approved):
Thanks @mvandenhoek! I personally would've kept the loop/sorting for loading the plugins and checked the number separately, but it is so easy to change that back when the time comes that I prefer to merging this PR now.

PR #1941, opened 2024-03-09
Fix subscriber's chunk leak in shm_monitor when unknown chunk type is retrieved from iceoryx

ksuszka:
This MR fixes a memory leak in shm_monitor in release/0.10.4 in the receive_data_wakeup_handler function.

Typical flow in this function is:
- retrieve chunk from iceoryx [link](https://github.com/eclipse-cyclonedds/cyclonedds/blob/1be07de395e4ddf969db2b90328cdf4fb73e9a64/src/core/ddsc/src/shm_monitor.c#L103)
- wrap this chunk into ddsi_serdata structure [link](https://github.com/eclipse-cyclonedds/cyclonedds/blob/1be07de395e4ddf969db2b90328cdf4fb73e9a64/src/core/ddsc/src/shm_monitor.c#L149)
- prepare it for processing
- release reference to ddsi_serdata structure so chunk will be released from iceoryx when ddsi_serdata wrapper won't be needed [link](https://github.com/eclipse-cyclonedds/cyclonedds/blob/1be07de395e4ddf969db2b90328cdf4fb73e9a64/src/core/ddsc/src/shm_monitor.c#L175)

However in some cases if chunk type [is not recognized](https://github.com/eclipse-cyclonedds/cyclonedds/blob/1be07de395e4ddf969db2b90328cdf4fb73e9a64/src/core/ddsc/src/shm_monitor.c#L141) it is not wrapped into ddsi_serdata strcture and currently it is never released. This MR fixes this execution path.

This seems to fix https://github.com/ros2/rmw_cyclonedds/issues/471 for me.

### DIFF
--- a/src/core/ddsc/src/shm_monitor.c
+++ b/src/core/ddsc/src/shm_monitor.c
@@ -142,6 +142,9 @@ static void receive_data_wakeup_handler(struct dds_reader* rd)
     {
       // Ignore that doesn't match a known writer or proxy writer
       DDS_CLOG (DDS_LC_SHM, &gv->logconfig, "unknown source entity, ignore.\n");
+      shm_lock_iox_sub(rd->m_iox_sub);
+      iox_sub_release_chunk(rd->m_iox_sub, chunk);
+      shm_unlock_iox_sub(rd->m_iox_sub);
       continue;
     }
 


ksuszka:
I've just noticed it was already fixed on master in https://github.com/eclipse-cyclonedds/cyclonedds/commit/f4ea4fcc79d25c7a21df8f827eaac6002639a610 2 years ago, however it was not back-ported to release 0.10.x which is currently in use in ROS2 Humble.

Could you make a backport of this fix to release 0.10.x so it is fixed in the next update of ROS Humble?

eboasson (approved):
LGTM, and I agree this is a fix that needs to be easily available for Humble. Thanks!

PR #1947, opened 2024-03-21
CI updates

eboasson:
This bumps the CI platforms. That in turn causes some trouble with (in particular) gcc and macOS.


dpotman (approved):
Looks good!

PR #1951, opened 2024-03-22
Drop support for OpenSSL pre-1.1.1

### LINKED ISSUE #1925: Failure to compile with boringssl
In `src/core/ddsi/src/ddsi_ssl.c` https://github.com/eclipse-cyclonedds/cyclonedds/blob/6896b29a0f1a8972bda5a4fb46551a5437c443be/src/core/ddsi/src/ddsi_ssl.c#L374
```
#if OPENSSL_VERSION_NUMBER < 0x10100000L
  {
    const int locks = CRYPTO_num_locks ();
    assert (locks >= 0);
    ddsi_ssl_locks = ddsrt_malloc (sizeof (CRYPTO_dynlock_value) * (size_t) locks);
    for (int i = 0; i < locks; i++)
      ddsrt_mutex_init (&ddsi_ssl_locks[i].m_mutex);
  }
#endif
  /* Leave these in place: OpenSSL 1.1 defines them as no-op macros that not even reference the symbol,
     therefore leaving them in means we get compile time errors if we the library expects the callbacks
     to be defined and we somehow failed to detect that previously */
  CRYPTO_set_id_callback (ddsi_ssl_id);
  CRYPTO_set_locking_callback (ddsi_ssl_lock);
  CRYPTO_set_dynlock_create_callback (ddsi_ssl_dynlock_create);
  CRYPTO_set_dynlock_lock_callback (ddsi_ssl_dynlock_lock);
  CRYPTO_set_dynlock_destroy_callback (ddsi_ssl_dynlock_destroy);
```
But in boringssl these are no-op functions not no-op macros so even if the callbacks are never called, they should be provided otherwise we would get compile errors. https://github.com/google/boringssl/blob/db7308de87ea138e7bbcbbb00dfc9b841774ba2f/include/openssl/thread.h#L144 So these statements should be put in the endif block.

eboasson:
OpenSSL versions older than 1.1.1 have all been dead for over 4 years. I don't see why an Cyclone would have to continue supporting the bad practice of not updating EOL'd security sensitive libraries full of known vulnerabilities.

Of course nobody should be using OpenSSL 1.1.1 anymore (it has been EOL'd about half a year ago), but I know there are still plenty of systems in the field that rely on it and even the CI on Azure gets the latest Linux images with it pre-installed.

Fixes #1925 — or so I expect anyway. @zhzhzoo-autra I'd appreciate it if you could give it a try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ In order to build Cyclone DDS you need a Linux, Mac or Windows 10 machine (or, w
   * C compiler (most commonly GCC on Linux, Visual Studio on Windows, Xcode on macOS);
   * Optionally GIT version control system;
   * [CMake](https://cmake.org/download/), version 3.16 or later;
-  * Optionally [OpenSSL](https://www.openssl.org/), preferably version 1.1;
+  * Optionally [OpenSSL](https://www.openssl.org/), we recommend a fully patched and supported version but 1.1.1 will still work;
   * Optionally [Eclipse Iceoryx](https://iceoryx.io) version 2.0 for shared memory and zero-copy support;
   * Optionally [Bison](https://www.gnu.org/software/bison/) parser generator. A cached source is checked into the repository.
 
--- a/docs/dev/dds_security_effort.md
+++ b/docs/dev/dds_security_effort.md
@@ -137,9 +137,7 @@ No major changes between the DDS Security plugins in OpenSplice and Cyclone
 are expected.
 
 The DDS Security plugins require OpenSSL. Cyclone DDS already uses OpenSSL.
-However, it expects (or at least it's preferred to have) version 1.1 or newer,
-while the OpenSplice Security plugins are build against 1.0.2. There are some
-API changes between the two versions. This will take some porting effort.
+We recommend a fully patched and supported version but 1.1.1 will still work.
 
 The build system should be ported from makefiles to cmake files.
 
--- a/src/core/ddsi/src/ddsi_ssl.c
+++ b/src/core/ddsi/src/ddsi_ssl.c
@@ -129,52 +129,6 @@ static ssize_t ddsi_ssl_write (SSL *ssl, const void *buf, size_t len, dds_return
   return sent;
 }
 
-/* Standard OpenSSL init and thread support routines. See O'Reilly. */
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
-static unsigned long ddsi_ssl_id (void)
-{
-  return (unsigned long) ddsrt_gettid ();
-}
-
-typedef struct CRYPTO_dynlock_value {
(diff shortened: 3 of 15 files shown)


dpotman (commented):
Looks good! In the readme there is a reference to OpenSSL 1.1, that line also needs an update: https://github.com/eclipse-cyclonedds/cyclonedds/blob/2246c005a0b5ada7f22da1e68f9c841006f29363/README.md?plain=1#L117

And here: https://github.com/eclipse-cyclonedds/cyclonedds/blob/2246c005a0b5ada7f22da1e68f9c841006f29363/docs/dev/dds_security_effort.md?plain=1#L140

eboasson:
> Looks good! In the readme there is a reference to OpenSSL 1.1, that line also needs an update:
> 
> https://github.com/eclipse-cyclonedds/cyclonedds/blob/2246c005a0b5ada7f22da1e68f9c841006f29363/README.md?plain=1#L117
> 
> And here:
> 
> https://github.com/eclipse-cyclonedds/cyclonedds/blob/2246c005a0b5ada7f22da1e68f9c841006f29363/docs/dev/dds_security_effort.md?plain=1#L140

Done. I also had to solve some conflicts because of merging #1826 and force-pushing it seemed the most sensible solution.

dpotman (approved):

zhzhzoo-autra:
Thanks for your attention! I tried to compile with boringssl with this patch, the compile errors disappeared, but there's a linker error:
```
bazel-out/k8-opt/bin/external/cyclonedds/_objs/cyclonedds/ddsi_ssl.pic.o:ddsi_ssl.c:function ddsi_ssl_accept: error: undefined reference to 'BIO_do_accept'
```
I thinks it's because boringssl removed some of openssl's api. I have no understanding of Openssl so I don't know if this if a big or small problem...

eboasson:
> Thanks for your attention! I tried to compile with boringssl with this patch, the compile errors disappeared, but there's a linker error:
> 
> ```
> bazel-out/k8-opt/bin/external/cyclonedds/_objs/cyclonedds/ddsi_ssl.pic.o:ddsi_ssl.c:function ddsi_ssl_accept: error: undefined reference to 'BIO_do_accept'
> collect2: error: ld returned 1 exit status
> ```
> 
> I thinks it's because boringssl removed some of openssl's api. I have no understanding of Openssl so I don't know if this if a big or small problem...

Thanks for trying it!

The easiest way to work around this would be to just not include the TCP / TCP+TLS support code (hardly anyone uses it anyway). There is not a separate build option to do that, by I think there should be, so: #1953. If you can give that a whirl after doing `cmake -DENABLE_TCP_TLS=0` ...

The real solution is of course to see what `BIO_do_accept` maps to in boringssl. I'm sure it is straightforward.

PR #1953, opened 2024-03-26
Add build-time option for TCP+TLS support

eboasson:
ENABLE_SSL did two things:

* Enable the built-in plugins (if ENABLE_SECURITY)
* Enable TCP+TLS support

TCP support is rarely used and has some problems, TCP+TLS support is even more rarely used
and we recommend using DDS Security instead. If TCP+TLS is not recommended, it makes sense
to support builds that leave it out. This also has the advantage of being able to
include support for DDS Security with the built-in plugins but without introducing a
dependency on OpenSSL for the core library.


dpotman (approved):
LGTM

PR #1955, opened 2024-03-28
qos_provider support

Splinter1984:
This PR add support for qos_provider
* `ENABLE_QOS_PROVIDER` flag added (`ON` by default)
* documentation include user-manual
* qos_provider tests


docs/manual/about_dds/qos_provider.rst
```diff
@@ -0,0 +1,413 @@
...
+        </datareader_qos>
+    </qos_profile>
+
+XML Example
+-----------
+
+Consider the following XML file that describes two QoS profiles:
+
+- FooQosProfile
+    - DataWriterQos - Transient
```
eboasson, line 165: I would pick some QoS other than durability because we don't yet support transient/persistent (although transient-local would be fine). It doesn't matter much which it is, as long as everything is fully supported.

docs/manual/about_dds/qos_provider.rst
```diff
@@ -0,0 +1,413 @@
...
+            (void) argv;
+
+            // provider will contains:
+            // myqoslib::foo_profile                     READER
+            // myqoslib::foo_profile                     WRITER
+            // myqoslib::foo_profile::my_topic           TOPIC
+            // myqoslib::bar_profile                     WRITER
+            // myqoslib::bar_profile::my_topic           TOPIC
+            dds_qos_provider_t *provider;
+            dds_return_t ret = dds_create_qos_provider ("/path/to/qos_definitions.xml", &provider);
```
eboasson, line 248: I think it is worth considering to make the QoS provider an entity with a handle that can be created without having anything else yet. (Much like the guardcondition, also hardly a "real" entity.) Then also `dds_delete_qos_provider` would not be necessary.

The downside is that (I think) you do want the returned QoS objects to be owned by the QoS provider so that you don't have to free them all individually. That doesn't play nicely with the implicit deleting of the QoS provider when the parent entity disappears. Fixing that would mean changing QoS objects to become entities as well ... so that's not a good path to go down.

On balance, I think the choice made here is ok.

docs/manual/about_dds/qos_provider.rst
```diff
@@ -0,0 +1,413 @@
...
+            assert (ret == DDS_RETCODE_OK);
+
+            dds_qos_t *tp_qos, *wr_qos;
```
eboasson, line 251: But then these should be `dds_qos_t const * tp_qos` etc. If I am getting my own copy, then the lifetime issue mentioned above doesn't hold.

src/core/ddsc/include/dds/ddsc/dds_public_qos_provider.h
```diff
@@ -0,0 +1,121 @@
...
+ * @param[in] provider - Pointer to the Qos Provider structure.
+ * @param[in] type - Type of entity which Qos to get.
+ * @param[in] key - Full qualify name of Qos to get in format '<library name>::<profile name>::<entity name>'.
+ * @param[in,out] qos - Pointer to the Qos structure.
+ *
+ * @return a DDS return code
+ */
+DDS_EXPORT dds_return_t
+dds_qos_provider_get_qos (const dds_qos_provider_t *provider, dds_qos_kind_t type,
+                          const char *key, dds_qos_t **qos);
```
eboasson, line 101: I think it should be `const dds_qos_t **qos`

src/core/ddsc/src/dds_qos_provider.c
```diff
@@ -0,0 +1,290 @@
...
+  }
+  *qos = item->qos;
+
+err2:
+  ddsrt_free(it.full_name);
+err:
+  return ret;
+}
+
+#define FILL_TOKEN_STR(start,end,token) \
```
eboasson, line 221: I don't see why this (and `FILL_TOKENS_STR`, `EMPTY_TOKEN_STR` and `EMPTY_TOKENS_STR`) has to be a macro, and that means it is probably better done as a function.

src/ddsrt/include/dds/ddsrt/log.h
```diff
@@ -78,6 +78,10 @@ extern "C" {
 #define DDS_LC_CONTENT (131072u)
 /** Output full dump of malformed messages as warnings */
 #define DDS_LC_MALFORMED (262144u)
+/** Debug/trace messages related to qos provider. */
+#define DDS_LC_QOSPROV (262144u)
```
eboasson, line 82: This here is a conflict with `DDS_LC_MALFORMED`

src/core/xtests/symbol_export/symbol_export.c
```diff
@@ -615,6 +620,18 @@ int main (int argc, char **argv)
   DDS_Security_get_secret_size_from_secret_handle (1);
 #endif
 
+#ifdef DDS_HAS_QOS_PROVIDER
+  dds_create_qos_provider(ptr,ptr);
+  dds_create_qos_provider_scope(ptr,ptr,ptr);
+  dds_qos_provider_get_qos(ptr,0,ptr,ptr);
+  dds_delete_qos_provider(ptr);
+  dds_sysdef_init_sysdef(ptr,ptr,0);
```
eboasson, line 628: Do we really need to export `dds_sysdef_...`?

src/core/ddsc/src/dds__sysdef_validation.h
```diff
@@ -0,0 +1,29 @@
...
+#define DDS__SYSDEF_VALIDATION_H
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+#include "dds/dds.h"
+#include "dds__sysdef_model.h"
+
+#define SYSDEF_QOS_USERDATA_MAXLEN ((size_t) 255)
```
eboasson, line 20: I don't see why we would hard-code a maximum length in the QoS provider for user data and partition names. (There is a hard limit in the discovery message format, but that is 64kB-1.) If we drop it here, we also no longer need the check elsewhere.

src/core/ddsc/src/dds__sysdef_model.h
```diff
@@ -0,0 +1,706 @@
...
+};
+
+#define QOS_POLICY_SYSDEF_STRUCT(p,t) \
+  struct dds_sysdef_ ## p { \
+    struct xml_element xmlnode; \
+    t values; \
+    uint32_t populated; \
+  };
+
+#define QOS_POLICY_DEADLINE_PARAM_PERIOD 1 << 0u
```
eboasson, line 245:
```suggestion
#define QOS_POLICY_DEADLINE_PARAM_PERIOD (1 << 0u)
```
(not only for this one, obviously)

If I am not overlooking anything, it is today only used for `QOS_POLICY_DEADLINE_PARAMS` (and analogously for all other cases) and in that context the parentheses won't make any difference, but it is a common trap and we should not set it.

eboasson (commented):
Thanks @Splinter1984! I have a few remarks, and I was going to point out the failing builds but I see you're looking into those already ☺️

Splinter1984:
@eboasson, all your remarks have been taken into account and fixed by [ea6589f](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1955/commits/ea6589f78bb34325cf87bd9ae6351e14d8e7f66b) 😉 (I hope ...)

eboasson (approved):
Thank you @Splinter1984 ☺️

PR #1956, opened 2024-03-28
Correctly initialize defaulted external sequence

eboasson:
An external mutable struct containing a sequence member that for which no value is
provided by the CDR was not initialized correctly: only the length field was set to 0, not
the other fields.  If the _release field happens to be false, this is harmless; if not it
results in freeing of unallocated memory.

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -1579,6 +1579,18 @@ static bool stream_is_member_present (uint32_t insn, dds_istream_t * __restrict 
   return !op_type_optional (insn) || is_mutable_member || dds_is_get1 (is);
 }
 
+static const uint32_t *initialize_and_skip_sequence (dds_sequence_t *seq, uint32_t insn, const uint32_t * __restrict ops, enum sample_data_state sample_state)
+{
+  if (sample_state == SAMPLE_DATA_UNINITIALIZED)
+  {
+    seq->_buffer = NULL;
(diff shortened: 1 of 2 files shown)


src/core/cdr/src/dds_cdrstream.c
```diff
@@ -1579,6 +1579,18 @@ static bool stream_is_member_present (uint32_t insn, dds_istream_t * __restrict
   return !op_type_optional (insn) || is_mutable_member || dds_is_get1 (is);
 }
 
+static const uint32_t *initialize_and_skip_sequence (dds_sequence_t *seq, uint32_t insn, const uint32_t * __restrict ops, enum sample_data_state sample_state)
+{
+  if (sample_state == SAMPLE_DATA_UNINITIALIZED)
+  {
+    seq->_buffer = NULL;
+    seq->_maximum = 0;
+    seq->_release = true;
```
dpotman, line 1588: I think this could be set to `false`. Seems that this could result in freeing a NULL pointer in `src/core/cdr/src/dds_cdrstream.c:3283`, but that won't do any harm in current allocators.

eboasson: At first I wrote initialization code at line 1958 (or thereabouts) and wondered about how to initialise `_release`. Then I realised I could refactor things and have a single implementation. The `true` is simply what it used to do. Therefore I am tempted to leave it as-is in this PR.

dpotman (approved):

PR #1957, opened 2024-03-28
Always make idlc_generate available in install

eboasson:
With cross-builds, and now therefore also static builds, one does need IDLC.  Installing Generate.cmake unconditionally means one can always write

  idlc_generate(TARGET konijn FILES wortel.idl)

and one only has to make sure that CMake can find "idlc" and the backend library.  Adding the host build to the CMAKE_PREFIX_PATH as in:

  cmake -DCMAKE_PREFIX_PATH=.../target-install\;.../host-install

results in CMake finding the CycloneDDS package for the target in the target-install directory while also looking (and finding) idlc and the backend somewhere in the host-install directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,10 @@ install(
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
   COMPONENT dev
 )
+install(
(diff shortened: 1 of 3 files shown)


dpotman (commented):
LGTM!

PR #1965, opened 2024-04-04
Delete some left-over code

eboasson:
Hasn't been necessary since 0845337, so for 4 years.


eboasson:
Fixes #1962

noxpardalis (approved):
Looks good to me and I took a look at the CI and the failing test case is the stress test which I'm guessing is Azure CI performance weirdness.

PR #1966, opened 2024-04-04
Fix padding issue in authentication plugin

eboasson:
The discovery data is to be serialized as a parameter list, serialising each parameter in turn as a 4-byte aligned, 32-bit header followed by arbitrary payload, itself serialised according to the rules of the original CDR spec (i.e., XCDR1 in XTypes' Newspeak) as-if that payload starts at offset 0.  There is never any padding other than that required by these rules.

The code for handling strings and octet sequences used by the authentication plugin got this wrong by (inconsistently) padding out strings and octet sequences to the next multiple of 4.  In the deserialization this also interferes with the bounds checking.

In principle this can lead to a failure to read/write well-formed discovery data.  In practice, the types happen to be such that that padding always ends up being present because of the next field.

### DIFF
--- a/src/security/core/src/dds_security_serialize.c
+++ b/src/security/core/src/dds_security_serialize.c
@@ -250,7 +250,6 @@ DDS_Security_Serialize_string(
 
     memcpy(&(ser->buffer[ser->offset]), str, len);
     ser->offset += len;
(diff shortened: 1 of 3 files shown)


dpotman (approved):
Looks good to me!

PR #1967, opened 2024-04-05
Add fuzzer for security deserializer

robertfemmer:
This PR adds a fuzzer targeting the security deserializer.


fuzz/fuzz_security_deser/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+  find_package(CycloneDDS REQUIRED)
+endif()
+
+add_executable(fuzz_security_deser fuzz_security_deser.c)
+target_include_directories(
+  fuzz_security_deser PRIVATE
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../src/core/ddsi/include>"
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../src/security/core/include>"
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../src/security/api/include>")
+target_link_libraries(fuzz_security_deser CycloneDDS::ddsc $ENV{LIB_FUZZING_ENGINE})
```
eboasson, line 15: I learnt the hard way that a variety of fuzzing engines is supported and used by the oss-fuzz infrastructure — at least "libfuzzer", "afl", "honggfuzz" and "centipede". "Centipede" appears to require linking in C++ code, and the easiest way to do that is to set the linker language to C++. I don't think it is necessary at this time to prevent build error reports from oss-fuzz, but it is quite a trivial change to make.

```suggestion
set_target_properties(fuzz_security_deser PROPERTIES LINKER_LANGUAGE CXX)
target_link_libraries(fuzz_security_deser CycloneDDS::ddsc $ENV{LIB_FUZZING_ENGINE})
```

fuzz/oss-fuzz-build.sh
```diff
@@ -21,7 +21,7 @@ cmake \
     -DBUILD_TESTING=ON \
     -DBUILD_SHARED_LIBS=OFF \
     -DBUILD_EXAMPLES=NO \
-    -DENABLE_SECURITY=NO \
+    -DENABLE_SECURITY=ON \
```
eboasson, line 24: I did run into some trouble, it appears it was caused by flipping this switch. I think the trouble is ultimately caused by erroneously trying to integrate wrapper libraries for the plugins where the plugins are not getting built. On consideration it is somewhat silly to waste all those CPU cyclones on building the tests when we don't even want to run those tests. The reason the tests need to be built is that we want to access symbols that are not ordinarily exported by the library, but it seems to me it would make more sense to support that directly with a build option (e.g., `EXPORT_ALL_SYMBOLS`).

#1974 should take care of that, if that gets approved, then the build should go faster and not exhibit this problem anymore.

eboasson (commented):
Thank you! This looks perfectly fine to me and is a very welcome addition to the set of fuzzers, but I did run into some small details with the build infrastructure that I think requires a trivial change and a rebase.

fuzz/oss-fuzz-build.sh
robertfemmer: I also ran into trouble building this, but only after I cleared out all build artifacts. I merged master and it does build.

fuzz/fuzz_security_deser/CMakeLists.txt
robertfemmer: I have to admit, I only tried libFuzzer. Added the suggest in a separate commit.

eboasson (approved):
Thank you!

PR #1968, opened 2024-04-05
Fuzz security handshake

robertfemmer:
This PR adds a fuzzer targeting the security handshake state machine using structure aware fuzzing.


eboasson:
I should have responded earlier, but every time I decided to try one more thing ...

Anyway, I'm struggling a bit with actually building and running this because of the interactions between static linking, plugin libraries and a symbol clash between a plugin and the fuzzer harness. None of that I can yet put on the ticket as review comments because I'm slowly working on finding a solution that I like that fits all the constraints. I suspect it works for you because you have the plugins lying around as shared libraries from a previous build. (Clearly doing two builds would then be an option, too.)

I have good hopes that it will only take a few more attempts (so let's say another couple of days) and then it'll be building and running nicely.

eboasson:
Hi @robertfemmer, it took a while but I solved my build issues and also managed to solve all the memory caused by running with the usual background threads present. It ran fine for several hours and also looks ok with the `libfuzzer`, `afl` and `honggfuzz` engines in the oss framework.

Please see: https://github.com/eboasson/cyclonedds/commit/022ba7afcd02143fe35a35612bd564a9071aeab3 for the changes I am proposing, in lieu of "normal" review comments. I'd do a PR to your fork if I knew how to do that without jumping through hoops.

eboasson (approved):
Thank you @robertfemmer !

PR #1973, opened 2024-04-10
Add TOPIC_DATA to ddsi_default_qos_topic

eboasson:
No externally visible effect for applications, but not setting is wrong and affects the C++ QoS provider implementation because it directly looks at the default QoS objects.


e-hndrks (approved):
Looks good to me. I did the same thing in my branch to make the QosProvider work correctly.

PR #1981, opened 2024-04-17
Initialize the ret variable to fix compiler warnings.

t0ny-peng:
This is a simple fix to a compiler warning we found with Clang. Please review. Thanks.


eboasson (approved):
Thanks @t0ny-peng !

PR #1982, opened 2024-04-18
Fix Iceoryx locator generation

eboasson:
For Iceoryx (or PSMX more generally) we need a unique identifier for the Iceoryx instance, but the way it is trying to get one is broken: a race condition can cause two processes on the same machine to use different locators. That causes stuttering.

This addresses that by computing a locator from the network configuration. Not quite fool-proof, but a lot better at least when the networking configuration of the machine is stable.

### DIFF
--- a/src/core/ddsc/src/dds__psmx.h
+++ b/src/core/ddsc/src/dds__psmx.h
@@ -51,6 +51,8 @@ typedef dds_return_t (*dds_psmx_create_fn) (
   const char *config // PSMX specific configuration
 );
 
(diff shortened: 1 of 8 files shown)


eboasson:
> you probably considered reading `/var/lib/dbus/machine-id` if available, with a fallback to this ip/mac addresses approach?

No ... I googled a bit but didn't find this one. Linux is a tier-2 platform for me (and Windows a tier-3 ... 🤣) so ...

Let's go with this then and keep an eye out for improvements

PR #1983, opened 2024-04-24
Fix leading underscore handling in IDLC

eboasson:
IDL identifiers start with a letter, and a (single) leading underscore is allowed as an escape character to distinguish them keywords.  That is, `Gloeiworm` and `_Gloeiworm` are the same identifier. This it implemented correctly.

However, it follows that where an identifier is expect, a single underscore is forbidden, and so is anything with a leading double underscore.  This detects those cases and raises an error.

### DIFF
--- a/src/idl/src/scanner.c
+++ b/src/idl/src/scanner.c
@@ -436,7 +436,7 @@ scan_pp_number(idl_pstate_t *pstate, const char *cur, const char **lim)
 }
 
 static int32_t
-scan_identifier(idl_pstate_t *pstate, const char *cur, const char **lim)
+scan_identifier(idl_pstate_t *pstate, const char *cur, const char **lim, bool allow_leading_underscores)
 {
   int32_t cnt = 0;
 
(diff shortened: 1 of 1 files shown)


dpotman (approved):
Would be nice to have some additional tests for the scanner (including some with line-continuation chars in identifiers). I did some manual testing, and the fix works as expected. So it's fine with me to merge this PR; we can improve the tests later.

eboasson:
That made me think of two things:

1. It is really the scanner that should check for keywords do the stripping of the leading underscore, but that's somewhat more involved than I would like to do now.
2. Returning `SYNTAX_ERROR` from the scanner seems to interfere with the way the scanner tests work. `YYUNDEF` is the magic cookie for an invalid token, so it is quite possible that that would be better the return code in this case. Changing that without fully understanding the choices made in the past seems unwise ...

So, yes, I agree. Better put this fix in and look into the tests and error handling some other time.

PR #1987, opened 2024-04-26
B64 sysdef  `_data`

Splinter1984:
According to DDS-XML spec. `group/user/topic_data` `<value>` element should be presented in `Base64` format.
With this PR:
- decode b64 format data from xml configuration to dds_octetseq value before set.
- update qos_provider test to encode tested qos `group/user/topic_data` before form xml-string.
- remove wrong validation of history_kind.

> [!WARNING]
> cxx binding qos_provider related tests will fail, and will be fixed by [#PR488](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/488).

### DIFF
--- a/src/core/ddsc/src/dds__sysdef_parser.h
+++ b/src/core/ddsc/src/dds__sysdef_parser.h
@@ -10,8 +10,8 @@
 #ifndef DDS_SYSDEF_PARSER_H
 #define DDS_SYSDEF_PARSER_H
 
-#include "dds/dds.h"
 #include "dds/ddsrt/log.h"
+#include "dds/ddsrt/retcode.h"
 
 #if defined (__cplusplus)
 extern "C" {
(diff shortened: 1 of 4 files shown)


eboasson (commented):
I was going to comment on a number of details, but it turns out that there are many more variants of base64 than I knew existed. That makes it doubtful that much would be gained by strict input validation, or that anything meaningful would be gained by some slight alterations. (I did check that it decodes correctly for what the `base64` tool that is part of macOS generates, assuming that one is correct.)

So: LGTM 🙂 

The C++ sanity check is failing on the CI because of the QoS provider, this you warned me about in the PR description already. I can't find the PR for the C++ repository yet, and I'd like to merge both together.

So I am stalling this one until I can also approve the C++ one.

Splinter1984:
@eboasson, thank you for your review. cxx binding changes now available [#PR488](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/488)

eboasson (approved):
Thanks again @Splinter1984 🙂 I did a local test run with C and C++ merged, and considering the changes in the C++ code, that's good enough to go ahead with merging it.

PR #1988, opened 2024-04-26
Fix building for Zephyr V3.6, QNX and other platforms that don't supp…

PatrickM-ZS:
This PR targets embedded builds (Zephyr, QNX) with the following changes:
- Fix some minor issues introduced recently that caused builds without single-source multicast and dynamic library loading support to fail
- Update Zephyr port to support latest version V3.6.0[https://docs.zephyrproject.org/latest/releases/release-notes-3.6.html]
- Support building unit-tests for QNX. Note to run tests you'll need to set the `CMAKE_CROSSCOMPILING_EMULATOR` variable to a script that executes the binary on eg. qemu or uses tools like ftp, ssh or telnet to upload and execute it on a remote target.


eboasson (approved):
LGTM — and in the absence of CI for these platforms, that's also all I can say 🙂

PR #1990, opened 2024-04-30
Export additional symbols that are required by the Iceoryx PSMX plugin.

gmartin82:
This PR exports the ddsrt_getifaddrs and ddsrt_freeifaddrs symbols that are required by the Iceoryx PSMX plugin.


eboasson (approved):
Thanks @gmartin82 for catching and fixing that oversight of mine. Clearly the CI matrix is missing a relevant case: a "normal" (non-testing) build with Iceoryx. I'll see if I can squeeze one in, preferably in the margins of the current builds.

PR #1999, opened 2024-05-15
Improve create_participant/domain documentation

### LINKED ISSUE #1997: Need clarification on the domain_id passed to dds_create_participant
I'm reading the doc for `dds_create_participant` and have some questions about its function doc. 

```/**
 * @brief Creates a new instance of a DDS participant in a domain
 * @ingroup domain_participant
 * @component participant
 *
 * If domain is set (not DDS_DOMAIN_DEFAULT) then it must match if the domain has also
 * been configured or an error status will be returned.
 * Currently only a single domain can be configured by providing configuration file.
 * If no configuration file exists, the default domain is configured as 0.
 *
 *
 * @param[in]  domain The domain in which to create the participant (can be DDS_DOMAIN_DEFAULT).
 *                       DDS_DOMAIN_DEFAULT is for using the domain in the configuration.
 * @param[in]  qos The QoS to set on the new participant (can be NULL).
 * @param[in]  listener Any listener functions associated with the new participant (can be NULL).

 * @returns A valid participant handle or an error code.
 *
 * @retval >0
 * ...
 DDS_EXPORT dds_entity_t
dds_create_participant(
  const dds_domainid_t domain,
  const dds_qos_t *qos,
  const dds_listener_t *listener);
 ```

I'm curious what this line means: `If domain is set (not DDS_DOMAIN_DEFAULT) then it must match if the domain has also been configured or an error status will be returned.`. E.g., if a non-DDS_DOMAIN_DEFAULT domain id 10 is passed, which has already be configured before by `dds_create_domain_with_rawconfig`, which part in the function parameter must match what?

Thanks!

### PR DESCRIPTION
**Pull request by eboasson**
Fixes #1997 — I hope 🙂

### DIFF
--- a/src/core/ddsc/include/dds/dds.h
+++ b/src/core/ddsc/include/dds/dds.h
@@ -976,13 +976,27 @@ dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
  * @ingroup domain_participant
  * @component participant
  *
- * If domain is set (not DDS_DOMAIN_DEFAULT) then it must match if the domain has also
- * been configured or an error status will be returned.
- * Currently only a single domain can be configured by providing configuration file.
- * If no configuration file exists, the default domain is configured as 0.
+ * This function creates a new participant in the specified domain id, implicitly creating
+ * a domain entity if one doesn't exist for the specified domain id.  The domain entity
+ * can exist as a consequence of the existence of another participant in it, or because it
+ * was explicitly created using @ref dds_create_domain or @ref
+ * dds_create_domain_with_rawconfig.  If the domain entity has not been created yet, a new
+ * domain entity is created using a configuration taken from the CYCLONEDDS_URI
+ * environment variable.
  *
+ * The domain id may be specified as DDS_DOMAIN_DEFAULT, in which case the behaviour
+ * depends on whether some domain entity already exists or not.  If there is at least one,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/core/ddsc/include/dds/dds.h [hunk 1]
```diff
@@ -976,13 +976,26 @@ dds_set_listener(dds_entity_t entity, const dds_listener_t * listener);
...
- * Currently only a single domain can be configured by providing configuration file.
- * If no configuration file exists, the default domain is configured as 0.
- *
- *
- * @param[in]  domain The domain in which to create the participant (can be DDS_DOMAIN_DEFAULT). DDS_DOMAIN_DEFAULT is for using the domain in the configuration.
+ * This function creates a new participant in the specified domain id, implicitly creating
+ * a domain entity if one doesn't exist for the specified domain id.  The domain entity
+ * can exist as a consequence of the existence of another participant in it, or because it
+ * was explicitly created using @ref dds_create_domain or @ref
+ * dds_create_domain_with_rawconfig.  If a domain entity is created, its configuration is
```
t0ny-peng, line 983: "If a domain entity is created" -> "If the domain entity has not been created, a new domain entity is created using configuration taken from the CYCLONEDDS_URI environment variable". 

This might be easier to understand for noobs like me.

t0ny-peng (approved):
Thanks! It is much clearer to me now how the configuration works.

PR #2004, opened 2024-05-17
Fix Zephyr ipv4 multicast join/leave support

PatrickM-ZS:
Since i neglected to remove the now deprecated `struct ip_mreq` workaround for Zephyr in my last PR #1988, it went unnoticed that Zephyr's POSIX implementation (or should i say interpretation?) demands the use of `ip_mreqn` without a fall-back to `ip_mreq` as supported by almost any other OS out there.


eboasson (approved):
I take your word for it, @PatrickM-ZS 🙂

What I don't like is the state of the CI. It is obviously not caused by this PR, more likely by an update to images used by Azure, and fortunately there are various successful builds.

PR #2007, opened 2024-05-23
Add flag to disable participant and endpoint discovery

dpotman:
The changes in this PR make it possible to disable the participant and endpoint discovery when creating a new DDS participant, by passing the `DDS_PARTICIPANT_FLAGS_NO_DISCOVERY` flag. Additionally, some other minor changes are included, including exporting some additional internal functions that are useful in specific applications.

### DIFF
--- a/docs/manual/config/config_file_reference.rst
+++ b/docs/manual/config/config_file_reference.rst
@@ -2591,7 +2591,7 @@ The default value is: ``false``
 ------------------------------------
 
 One of:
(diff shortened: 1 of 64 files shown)


eboasson (approved):
I think this is a good first step, discovery is a known cause of pain in DDS systems and being able to create the entities and proxy entities programmatically at least provides the basics for an application to run without requiring discovery.

In the future it really needs to become a bit easier to create those and also to be possible to mix participants that use dynamic discovery and ones that don't (I can easily see that work, I have a bit more trouble seeing how it could be done for individual readers/writers, though that'd be greater still).

I do not see any reason why such changes can't be built on this, so I think this is a worthwhile addition to the code. Thanks! ☺️

PR #2009, opened 2024-05-23
tsn `mac_addresses`/`vlan_tag` support for sysdef_parser

Splinter1984:
This adds support to the XML sysdef parser for tsn mac_addresses/vlan_tag according to DDS-XML spec. Also applies changes from [PR488](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/488#discussion_r1598190628) of cyclonedds-cxx repo.


src/core/ddsc/tests/qos_provider.c
```diff
@@ -391,12 +391,13 @@ static uint32_t b64_encode (const unsigned char *text, const uint32_t sz, unsign
   for (size_t i = 0, j = 0; i < buff_len && j < sz; i+=4, j+=3)
   {
     unsigned char chunk[4] = {0x00, 0x00, 0x00, 0x00};
-    chunk[0] = base64_etable[(text[j] >> 0x02U)];
```
eboasson, line 394: Ah, this one. I remember it from C++ 🙂

Can you split the commit so that this becomes a separate one?

src/core/ddsc/src/dds_sysdef_parser.c
```diff
@@ -2625,11 +2644,23 @@ static int proc_elem_open (void *varg, UNUSED_ARG (uintptr_t parentinfo), UNUSED
       else if (ddsrt_strcasecmp (name, "tsn_listener") == 0)
         CREATE_NODE_LIST (pstate, dds_sysdef_tsn_listener_configuration, ELEMENT_KIND_DEPLOYMENT_CONF_TSN_LISTENER, NO_INIT, fini_conf_tsn_listener, tsn_listener_configurations, dds_sysdef_tsn_configuration, ELEMENT_KIND_DEPLOYMENT_CONF_TSN, pstate->current);
       else if (ddsrt_strcasecmp (name, "data_frame_specification") == 0)
-        CREATE_NODE_LIST (pstate, dds_sysdef_tsn_data_frame_specification, ELEMENT_KIND_DEPLOYMENT_CONF_TSN_TALKER_DATA_FRAME_SPEC, NO_INIT, fini_conf_tsn_data_frame_specification, data_frame_specification, dds_sysdef_tsn_talker_configuration, ELEMENT_KIND_DEPLOYMENT_CONF_TSN_TALKER, pstate->current);
```
eboasson, line 2628: Can you give a little bit more background why this changes from `LIST` to `SINGLE`? I do think it is really strange to configure TSN and then send the same data to multiple flows, and to restrict it (at least until we have a good reason to not restrict it) makes sense to me. I just want to be sure I understand the reason.

Splinter1984: According to DDS-XML spec. data_frame and others changed has one or no entries in configuration of TSN(Listener|Talker).

eboasson (approved):
Thanks @Splinter1984 ! 😂

PR #2014, opened 2024-05-29
Fix bug in dds_request_loan_of_size

reicheratwork:
The return code in case of dds_entity_pin failing would be "false", which is not a valid dds_return_t, also this equates to 0, which is DDS_RETCODE_OK, which is exactly NOT what it should return here


eboasson (approved):
Ouch.

Thanks @reicheratwork !

Issue #2019, opened 2024-06-03
Build failure: `ddsi_udp.c: error: storage size of 'mreqn' isn't known` etc.


barracuda156:
Build fails on earlier macOS with the following errors:
```
[  9%] Building C object src/tools/idlpp/CMakeFiles/idlpp.dir/src/main.c.o
cd /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_cyclonedds/cyclonedds/work/build/src/tools/idlpp && /opt/local/bin/gcc-mp-13 -DHAVE_CONFIG_H=1 -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_cyclonedds/cyclonedds/work/build/src/tools/idlpp/src -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_cyclonedds/cyclonedds/work/cyclonedds-0.10.5/src/tools/idlpp/src -pipe -Os -DNDEBUG -I/opt/local/include -std=gnu99 -arch ppc -mmacosx-version-min=10.6 -fPIC -Wall -Wextra -Wconversion -Wmissing-prototypes -MD -MT src/tools/idlpp/CMakeFiles/idlpp.dir/src/main.c.o -MF CMakeFiles/idlpp.dir/src/main.c.o.d -o CMakeFiles/idlpp.dir/src/main.c.o -c /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_cyclonedds/cyclonedds/work/cyclonedds-0.10.5/src/tools/idlpp/src/main.c
[  9%] Building C object src/tools/idlpp/CMakeFiles/idlpp.dir/src/mbchar.c.o
[...]

PR #2020, opened 2024-06-04
Fix build for earlier macOS versions

barracuda156:
@eboasson These are necessary, but insufficient. At the moment 0.10.5 still fails further at linking, while the master fails earlier: https://github.com/eclipse-cyclonedds/cyclonedds/issues/2019

`MAXTHREADNAMESIZE` is defined in `Libc` internal header. Since the code uses it unconditionally, it should be defined to something. For w/e reason the value from `Libc` is inaccessible, and it also fails on CI on x86_64 up to 10.10: https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/271089/steps/install-port/logs/stdio

Then up to 10.12 there is a failure due to wrong number of arguments in `memcpy`: https://build.macports.org/builders/ports-10.12_x86_64-builder/builds/275235/steps/install-port/logs/stdio
(This is a known issue, Apple changed the declaration for `memcpy`.)

Unfortunately, I have no idea at the moment why linking fails with undefined `_dds_stream_extract_keyBE_from_key`.

### DIFF
--- a/src/security/builtin_plugins/cryptographic/src/crypto_transform.c
+++ b/src/security/builtin_plugins/cryptographic/src/crypto_transform.c
@@ -331,7 +331,7 @@ static bool read_submsg_header (tainted_input_buffer_t *input, uint8_t smid, dds
   // silly C can't deal with assignment to *submsg_view in any way because of endp
   // memcpy to the rescue!
   // coverity[store_writes_const_field]
-  memcpy (submsg_view, &(tainted_input_buffer_t){ .ptr = input->ptr, .endp = input->ptr + hdr->octetsToNextHeader }, sizeof (*submsg_view));
+  memcpy (submsg_view, (&(tainted_input_buffer_t){ .ptr = input->ptr, .endp = input->ptr + hdr->octetsToNextHeader }), sizeof (*submsg_view));
   input->ptr += hdr->octetsToNextHeader;
   return true;
 }
--- a/src/core/ddsi/src/ddsi_udp.c
+++ b/src/core/ddsi/src/ddsi_udp.c
@@ -12,6 +12,10 @@
 #define __APPLE_USE_RFC_3542
 #define _GNU_SOURCE
 
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#endif
+
 #include <assert.h>
 #include <string.h>
 #include "dds/ddsrt/atomics.h"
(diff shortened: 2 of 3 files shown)


barracuda156:
Well, looks like I cannot sign the ECA, since log in is unavailable to non-members. Feel free to cherry-pick from here and then close the PR.

barracuda156:
Given that Debian faced the same issue, apparently, it is a genuine bug: https://sources.debian.org/patches/cyclonedds/0.10.5-1/0005-Fix-unresolved-symbol-on-big-endian-architectures.patch

barracuda156:
@eboasson These 4 commits plus a patch from Debian fix 0.10.5 for me.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
```diff
@@ -331,7 +331,7 @@ static bool read_submsg_header (tainted_input_buffer_t *input, uint8_t smid, dds
   // silly C can't deal with assignment to *submsg_view in any way because of endp
   // memcpy to the rescue!
   // coverity[store_writes_const_field]
-  memcpy (submsg_view, &(tainted_input_buffer_t){ .ptr = input->ptr, .endp = input->ptr + hdr->octetsToNextHeader }, sizeof (*submsg_view));
+  memcpy (submsg_view, (&(tainted_input_buffer_t){ .ptr = input->ptr, .endp = input->ptr + hdr->octetsToNextHeader }), sizeof (*submsg_view));
```
eboasson, line 334: Interesting ... I guess `memcpy` was a macro ...

I think it needs a comment that the parentheses are not superfluous, otherwise someone (me, perhaps) might just remove them again.

eboasson (commented):
LGTM, except that the gcc-13 on macOS run has a big problem that I suspect to be triggered by including the `AvailabilityMacros.h` and to have more to do with gcc/homebrew/macOS then with your proposed changes. I'm sure that'll be easy to work around, it just needs to be sorted out before this can be merged.

eboasson:
> Well, looks like I cannot sign the ECA, since log in is unavailable to non-members. Feel free to cherry-pick from here and then close the PR.

Maybe they changed something and now require you to register? Whatever, thank you for trying ☺️ Cherry-picking is a nice pragmatic solution for this particular PR. I'll try to deal with the problem the CI ran into as well.

Thanks!

barracuda156:
> LGTM, except that the gcc-13 on macOS run has a big problem that I suspect to be triggered by including the `AvailabilityMacros.h` and to have more to do with gcc/homebrew/macOS then with your proposed changes. I'm sure that'll be easy to work around, it just needs to be sorted out before this can be merged.

Since I do not use Homebrew, I have no idea what could be the issue. `AvailabilityMacros.h` is a standard header, and it certainly works with gcc13, so perhaps the bug is on Homebrew side, if something fails there.

P. S. gcc13 did have an issue with `libc++` headers (fixed in the master after 13.2.0), but a) hardly anyone gonna even need to use this combo and b) AFAIK, there is no relation of that bug to `AvailabilityMacros.h`.

src/security/builtin_plugins/cryptographic/src/crypto_transform.c
barracuda156: Yes, see: https://github.com/Mbed-TLS/mbedtls/issues/8994

PR #2023, opened 2024-06-07
Fixed memory leak in idl preprocessor

reicheratwork:
This leak was caused by not freeing the strings generated when resolving paths
To reproduce this leak: parse an idl file which includes another idl file

### DIFF
--- a/src/tools/idlpp/src/system.c
+++ b/src/tools/idlpp/src/system.c
@@ -3293,6 +3293,7 @@ found_name:
     if (open_include( filename, (delim == '"'), next)) {
         /* 'fname' should not be free()ed, it is used as file->         */
         /*      real_fname and has been registered into fnamelist[]     */
+        free( filename);
         return  TRUE;
     }
 
@@ -3612,6 +3613,7 @@ search:
         put_depend( fullname);          /* Output dependency line   */
 
 true:
+    free( fullname);
     return  TRUE;
 false:
     free( fullname);
@@ -5086,3 +5088,8 @@ void    clear_filelist( void)
 }
 #endif
 
+void clean_system(void)
+{
+  if (sharp_filename != NULL)
+    free(sharp_filename);
+}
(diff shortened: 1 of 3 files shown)


src/tools/idlpp/src/system.c
```diff
@@ -3293,6 +3293,7 @@ int     do_include(
     if (open_include( filename, (delim == '"'), next)) {
         /* 'fname' should not be free()ed, it is used as file->         */
```
eboasson, line 3294: The comment is wrong: `set_fname` stores a copy. So the comment should be deleted.

That said, I do think `set_fname` is a bit nasty in that it returns the stored copy if it is new but the address passed in if it wasn't (you might want to double-check that statement).

If so, that means `cur_file` points to either the argument to `add_file` or to said copy. So it appears as if `free`ing here means `cur_file` may become a pointer to freed memory.

I therefore suspect that for this new `free` to be correct, `set_fname` needs to always return a pointer to the name in the list, and never to the argument that was passed in.

Would you do me a favor and double-check? 🙏

reicheratwork: > Would you do me a favor and double-check? 🙏

Will do

reicheratwork: I don't exactly know what you mean with `cur_file`, maybe you mean `cur_fname `?

Going over the lifecycle of the `filename` string:
- it is copied from `fname`, which is itself allocated on the heap through `save_string`
  - if there is no previous error ofcourse
- then it is passed to `open_include`, which through a number of function calls makes a copy of this string in the list of files
- lastly, `filename` is deallocated
So that looks completely fine in my eyes.

The memory in `cur_fname` always either points to a const string assigned during the initialization phase or something from the `set_fname` function, which always returns either a newly assigned entry in its administration, which is on the heap, or something already registered, which is also on the heap. So `cur_fname` will never point to the same address as `filename` in the `do_include` function, as far as I can see.

reicheratwork: Let me check this :)

reicheratwork: > Let me check this :)

Seems like the current version causes `double_free` and `invalid_read` errors with your example, and your correction removes this.

So let me pull this through the usual compile checks and then push this.

eboasson (approved):
Thank you @reicheratwork !

PR #2025, opened 2024-06-08
implement custom idlpy location

trittsv:
Previously the `_idlpy` could only be loaded from a location returned by `python3` program.

This PR adds a command-line option `-y <path>` to use a custom `_idlpy` lib.

This is needed when you want to use idlc python without any pip package installation.
For example `CycloneDDS-Insight` needs this because there idlc should use the binaries inside the App.

@eboasson can you have a look at this PR? 😊

### DIFF
--- a/src/tools/idlc/include/idlc/generator.h
+++ b/src/tools/idlc/include/idlc/generator.h
@@ -43,7 +43,7 @@ struct idlc_generator_plugin {
   idlc_generate_t generate;
 };
 
-int idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang);
+int idlc_load_generator(idlc_generator_plugin_t *plugin, const char *lang, const char *idlpy_overwrite);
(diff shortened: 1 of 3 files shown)


eboasson (commented):
I often have a fight with the python backend, one that I assume is caused by my rather unsophisticated use of `pip install`, locally built stuff, multiple versions, homebrew, perhaps some macOS quirks, ... I always solve it by simply finding the correct library on the disk and passing the path to the backend library with `-l`, e.g.,
```
bin/idlc -l/Users/erik/C/cdds-py/build/lib.macosx-14-arm64-cpython-311/cyclonedds/_idlpy.cpython-311-darwin.so
```
and that works fine.

It seems to me that this is exactly the same as this new option would do — or am I overlooking something?

trittsv:
oh i missed that with `-l` its possible to pass a path ... sorry

eboasson:
No worries 🙂

PR #2035, opened 2024-06-14
Do not change type in typelib making key holder

eboasson:
Type assignability for structs can require constructing a key holder type, which in turn involves marking all fields as key fields for a struct that has no key fields defined, e.g., when:

    struct Inner {
      long a, b;
    };
    struct Outer {
      @key Inner k;
    };

the key holder type will constructing a type where "a" and "b" are annotated as keys.  The code for doing this incorrectly modified the input type already in the library, rather than setting the flag on the new key holder type.

Obviously, silently changing the type in the library is bad.  It was observed because the change affects the type identifier (without changing where the type is in the index) and the mismatch got flagged while processing a type lookup reply.


dpotman (approved):
Good catch!

PR #2037, opened 2024-06-14
Dynsub extensions

eboasson:
* Also respond to `DCPSSubscription`
* Also print minimal type object
* Fix printing of `bitmask` types


dpotman (approved):

eboasson:
@dpotman I also added support for arrays, I found it too annoying that it crashed on the type involved in #2040 🙂

dpotman (approved):

PR #2040, opened 2024-06-17
Skip optionals completely when deserializing a key

eboasson:
The serialized key only contains the key fields and when deserializing a key one must skip everything else.  The deserialization (and printing) code incorrectly read the boolean indicating presence of an optional field in cases like:

    struct S {
      @optional long x;
      @key long y;
    };

This commit fixes it by first checking whether a key is being deserialized, skipping the checks for the presence of an optional field if it is.  (Note that optional fields can not be key fields.)


dpotman (approved):
Was going to type that `stream_normalize_adr` also needs this fix, but now it's fixed in the additional commit 😄 , so LGTM!

PR #2042, opened 2024-06-17
Trace reason for declaring types non-assignable

eboasson:
This records some information during assignability checking why a pair of types is considered non-assignable and adds this information to the trace.  One still needs to correlate the type ids with the actual type definitions in most cases, but in many cases even this little bit of information will probably help.  involved in the rejection), but that is not


dpotman (approved):
Very useful information for debugging type incompatibilities, looks good to me!

PR #2045, opened 2024-06-19
Turn off PACKET_DESTINATION_INFO for mingw-w64.

### LINKED ISSUE #2027: HelloWorld example broken on master with MinGW/ucrt64
The HelloWorld example (and likely, more than just the example) seems to be broken on master, when compiling with gcc (14.1) under MSYS2/MinGW/ucrt64. It does work as expected using the same code with a different compiler (VS2022).

It looks like 83b81f54d75893a004598bc35cdff5fef26fb919 is the offending commit, if that may help to locate the issue.

tonichedgehog:
The CMSG_* macros used in CycloneDDS are not provided by mingw-w64-gcc (only WSA*). Set the number of bytes received also for nonzero return from wsarecvmsg. Closes #2027

### DIFF
--- a/src/ddsrt/src/sockets/windows/socket.c
+++ b/src/ddsrt/src/sockets/windows/socket.c
@@ -570,6 +570,7 @@ ddsrt_recvmsg(
     if (sockext->wsarecvmsg (sockext->sock, &wsamsg, &n, NULL, 0) != 0)
     {
       int err = WSAGetLastError();
+      *rcvd = (ssize_t) n;
       return recv_error_to_retcode(err);
     }
     else
--- a/src/core/ddsi/src/ddsi_udp.c
+++ b/src/core/ddsi/src/ddsi_udp.c
@@ -49,6 +49,9 @@ DDSRT_STATIC_ASSERT (DDSI_LOCATOR_UDPv4MCGEN_INDEX_MASK_BITS <= 32 - UDP_MC_ADDR
 #  endif
 #endif
 #ifndef PACKET_DESTINATION_INFO
+#  if defined (__MINGW32__) && !defined (CMSG_SPACE)
+#    define PACKET_DESTINATION_INFO 0
+#  endif
 #  if defined CMSG_SPACE && (defined IP_PKTINFO || (DDSRT_HAVE_IPV6 && defined IPV6_PKTINFO))
 #    define PACKET_DESTINATION_INFO 1
 #  else


src/ddsrt/src/sockets/windows/socket.c
```diff
@@ -570,6 +570,7 @@ ddsrt_recvmsg(
     if (sockext->wsarecvmsg (sockext->sock, &wsamsg, &n, NULL, 0) != 0)
     {
       int err = WSAGetLastError();
+      *rcvd = (ssize_t) n;
```
eboasson, line 573: I really don't understand what prompted this addition.

Reading the code, the (there's only one outside the tests) call site assumes that `ddsrt_recvmsg`:
1. returns OK with `*rcvd` > 0
2. returns OK with `*rcvd` = 0
3. returns not-OK with `*rcvd` untouched or set to 0

Case 1 is the perfectly ordinary case.

Case 2 is treated as a spurious but successful read. This is somewhat weird: the socket triggers because data is available and so a read should return something. IIRC it is there because with OpenSSL 3 and TCP+TLS, I saw the received bytes all getting consumed internally by OpenSSL during a handshake. Anyway, it exists.

Case 3 is for all errors. The `*rcvd` untouched or set to 0 is important because `ddsi_udp_conn_read` first checks whether it is > 0. It initializes it to 0, so if it is untouched on error or set to 0, all is well.

The documentation of `ddsrt_recvmsg` is not quite as detailed, I am afraid. I also think it is a bad decision to first check the number of received bytes instead of whether the call succeeded. Indeed, I might just change it to: OK with `*rcvd` >= 0, or not-OK with `*rcvd` undefined. The change to that one caller is trivial and this interface is much less likely to trip someone (me, probably ...) up in the future.

Given the above, adding this assignment seems superfluous at best and possibly introduces a vulnerability at worst. That potential vulnerability is because the documentation of `WSARecvMsg` doesn't state that `n` will be 0 on error (or I overlooked it), which in turn means assigning it here may (more precisely: must be assumed to) result in setting `*rcvd` to an arbitrary value. Since it is the size of the received data, this is a buffer overflow waiting to happen.

I am sure you had a good reason, but I just don't see it just from looking at the code. Therefore, I decided to try building it on MinGW, but that is too complicated a mess for me ... Can you please help me understand what the problem is?

tonichedgehog: I _think_ that for some reason `WSARecvMsg` can return nonzero also for some (successful?) case where `*rcvd > 0`, or the return value is not interpreted properly by `recv_error_to_retcode()` .

You don't strictly need a MinGW build for testing, the same code path can be triggered with an msvc build, forcing `PACKET_DESTINATION_INFO=0`. I don't know if that's relevant for a Linux build, but when I tried, I did see some messages indicating something was not quite right (without the assignment).

eboasson: I made #2053 after some thinking and @dpotman's suggestion that the reason for `WSARecvMsg` sometimes returning non-zero while still being successful could be the `WSAEMSGSIZE` error.

eboasson:
Closing (with thanks!) because of merging #2053

PR #2046, opened 2024-06-19
Update ddsi_new_proxy_participant call in fuzzer

eboasson:
This should fix a build problem with the handshake fuzzer.


dpotman (approved):
LGTM! Nit-pick: you could replace line 254 by passing harness.proxy_pp to `ddsi_entidx_lookup_proxy_participant_guid`

PR #2047, opened 2024-06-19
Fix serdata size in serdata_default_from_psmx

dpotman:
The size of a serdata should include the padding bytes. This fix adds these padding bytes from the PSMX meta-data in the call to serdata_default_new_size when receiving data from PSMX.


eboasson (approved):
🤦 Yes. Thank you! ☺️

PR #2052, opened 2024-07-01
check return value of `X509_digest`

poetinger:
Section 4.3.1.1 of X41's security audit found a potential vulnerability where two potentially uninitialized stack values could result in incorrect behaviour.


eboasson (approved):
Thanks @poetinger !

PR #2053, opened 2024-07-01
mingw-w64 + improve ddsrt_recvmsg interface

### LINKED ISSUE #2027: HelloWorld example broken on master with MinGW/ucrt64

eboasson:
This takes the change for `PACKET_DESTINATION_INFO` from #2045 and bundles it with an improvement to the interface of `ddsrt_recvmsg`.

Closes #2027


dpotman (approved):
LGTM!

PR #2065, opened 2024-08-13
Add setting for controlling thread affinity

eboasson:
This adds a Thread/Scheduling/Affinity option, which is a list of unsigned 32-bit
integers indicating in some manner what CPUs to bind the thread to. Silently ignored on
platforms that do not support it (or where Cyclone doesn't provide support for it yet).

Currently only on Linux, where it is interpreted as a list of CPU ids.


noxpardalis (approved):
I wasn't super familiar with confining processes to processors but I had some fun going through the `cpuset` man pages and testing out `sched_setaffinity`. Based on my reading there this LGTM.

eboasson:
> I wasn't super familiar with confining processes to processors but I had some fun going through the `cpuset` man pages and testing out `sched_setaffinity`. Based on my reading there this LGTM.

Thank you for reviewing 🙏

PR #2071, opened 2024-08-19
add option to set realtime priority to iceorix notification thread (linux only)

MarcelJordense:
An option to set the realtime priority and the affinity to the psmx-iox notification thread is added. Currently this is implemented for linux only.


eboasson (approved):
Thank you @MarcelJordense !

PR #2078, opened 2024-08-22
fix compilation issue on qnx

MarcelJordense:
Fix compilation issue on QNX


eboasson (approved):
It works, it doesn't cause trouble with the CI, so I guess you've figured out that this helps on QNX. Thanks @MarcelJordense 🙂

PR #2083, opened 2024-09-02
Fix multicast check for loopback interface

eboasson:
This corrects the interface addresses used when checking whether multicast is supported on IPv4 even when the interface doesn't have the IFF_MULTICAST flag set as introduced in 3b66431d9894e06180701d91897609f95a005e77.

On Ubuntu it worked fine the way it was (at least for what I have seen), but not on Fedora
40. This appears to work fine on both.


dpotman (approved):
LGTM!

Issue #2068, opened 2024-08-15
Deserialization of sample failed due to reasons unknown, possibly related to a sequence


FirebarSim:
I am using the cyclonedds python bindings to test some proof of concept connectivity and processes to another machine. That machine has a production standard interface (that I have no control over) which sends a topic with mocked up IDL similar to below (unable to share the full IDL). 

```
struct ContributingData {
    string sensorName;
    short dataItemIndex;
};

typedef sequence<ContributingData, 20> ContributingDataType;

struct Data {
    short systemDataIndex;
    int dataValue;
    ContributingDataType contributingSensors;
};
```

On a machine running OSPL Test I can receive samples from the production VM, and if I create a sample with the same data on the Test VM I can also receive that at OSPL Test. However the packets from the Production VM on the Test VM produce the error "Unable to deserialize TopicName (reasons unknown)." I _think_ that packets from the Test VM are received at the Production Machine and return the OSPL Error "Malformed Packet", but I am not able to monitor this in real time with any great ease.

It appears that there is something wrong in the serialization, possibly in the IDLC generated python code (I note that at least one of the members is named format which is a Python built in function). However other errors along these lines appear to point to sequences, but I am not sure where to start with debugging.

Any help greatly appreciated! For additional context I have successfully communicated to this machine from cyclonedds using topics not containing populated sequences. But this is the first data type I am receiving that has a populated sequence.

eboasson:
> I note that at least one of the members is named format which is a Python built in function

I checked, that doesn't do any harm. I tried it with the above type with one of the fields renamed to `format`, and the CDR looks just fine to me.

If the type is no more complicated than this, I would double-check that the definition used in Cyclone is the same as the one used in OSPL, because it seems somewhat unlikely something so basic fails.

Then, I would verify that Cyclone (non-Python, just to be sure) receives the data correctly. Cyclone's `dynsub` example often comes in handy here. Just give it the topic name, then it should find the topic, retrieve & print the type definition and start printing the samples as they arrive.

If that all checks out, then I suppose the next check is to take one failing sample, and write a little test program in OSPL that publishes it. If you're adventurous, it is possible to run OpenSplice's IDL pre-processor to obtain the XML type definition that OpenSplice uses, then feed that into [pubsub](https://github.com/eboasson/opensplice-tools), like:
```
# cat x.idl
module X {
struct ContributingData {
    string sensorName;
    short dataItemIndex;
};
[...]

PR #2085, opened 2024-09-05
Improve performance of publishing complex types over PSMX only

eboasson:
It used to be that publishing "complex" types — anything requiring serialization, really — would always construct a serdata, then optionally copy the CDR into shared memory. We already know if we'll need both or will only need the shared memory copy, and so we can quite easily remove a copy. This gives rise to the new `dds_write_impl_serialize_into_loan`.

Then the question is one of implementation. The sertype already defined functions for this, but they could not handle key-only data and `sertype_default` implementation was horrible (serialize, look at the size, free the result). This has gave rise to `dds_stream_getsize_sample` and `dds_stream_getsize_key`.

That required testing, that made me find some cases where invalid data was not reject correctly in the write path, then I was fiddling with the serializer and so decided to also make the boolean handling a bit friendlier to inputs where `true` is encoded as a non-`0` byte, instead of only accepting `1`. (See #2068, this should fix that.)

This is quite difficult to test, but the paths do get hit in the PSMX tests and the serializer calls check against the length of the serialized data against the length computed by the `getsize` functions. The Python binding "fuzzer" test will be extended by a PR on the Python binding that also performs these checks on the types/data used there, which increases the coverage to a reasonable level. (Tested locally with some 1000 types.)

### DIFF
--- a/src/core/cdr/include/dds/cdr/dds_cdrstream.h
+++ b/src/core/cdr/include/dds/cdr/dds_cdrstream.h
@@ -205,19 +205,23 @@ DDS_EXPORT uint32_t dds_stream_countops (const uint32_t * __restrict ops, uint32
 size_t dds_stream_check_optimize (const struct dds_cdrstream_desc * __restrict desc, uint32_t xcdr_version);
 
 /** @component cdr_serializer */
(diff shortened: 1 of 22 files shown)


eboasson:
I see I have been a bit optimistic in creating a PR ... fortunately the trouble is caused by few stupid oversights. Looking into those, I discovered I also overlooked some more details on handling invalid inputs.

dpotman (approved):
Looks good, I've thoroughly checked the cdrstream `getsize` functions alongside the write functions and couldn't find any inconsistencies. 

(Nitpicking: the sentence "if get_serialized_size is 0" in the `dds_write_impl` comment confused me a bit, because initially I read it as: if the serialized_size is zero)

PR #2096, opened 2024-09-24
platform independet pylib determination

### LINKED ISSUE #2050: Python lib determination is not platform independent
It uses `python3` executable to determine the python libs, but on windows there is no `python3.exe`.
The windows python3 installer creates only `python.exe`.

See here:
https://github.com/eclipse-cyclonedds/cyclonedds/blob/d7db65c7f9def668af029890c351de6186977d58/src/tools/idlc/src/idlc/generator.c#L167

Of course a easy workaround is to duplicate the `python.exe` on windows and rename it to `python3.exe` ...

trittsv:
This PR closes #2050.

The python windows installer always creates `python.exe`, there is no `python3.exe`.
So we simply try again but using plain `python` to cover these cases.

@eboasson could you have a look? :)


eboasson (approved):
Why does Windows always have to be different, even when there's absolutely no reason to? Some things are simply to be suffered, and this seems a sensible way of reducing the suffering.

Issue #2022, opened 2024-06-05
Errors occur during high-load communication.


xygyo77:
Hello. I am validating an application using ros2 humble + cyclonedds.
I was doing a high load (3-5subscriptions each for 250-300publishers, interval 100ms) communication test, and in a few tens of seconds I get.
```
terminate called after throwing an instance of ‘rclcpp::exceptions::RCLError’
  what(): could not create service: failed to create topic [rr/subscriber_170_2/set_parameters_atomicallyReply] because the function was given invalid parameters, at . /src/rmw_node.cpp:1796, at . /src/rcl/service.c:124
[ros2run]: Aborted
```
occurred using SingleThreadedExecutor.
Please let me know what you find out.
It doesn't happen in the real environment, so there might be a problem with the usage, but I'll share it with you just in case.
```
    +-
    |       +-- sub
    |       |
    | pub --+-- sub
    |       |
    |       +-- sub
250 | (remove all pub/subs)
 |  |       +-- sub
300 |       |
pubs| pub --+-- sub
    |       |
    |       +-- sub
    | (remove all pub/subs)
    |  <...>
    |
    +-
```

---
When changed to MultiThreadedExecutor,
```
 . /src/core/ddsc/src/dds_handles.c:430: dds_handle_drop_childref_and_pin: Assertion `(cf & HDL_REFCOUNT_MASK) > 0' failed.
[ros2run]: Aborted
```
The following is the result. This one is mentioned at https://github.com/eclipse-cyclonedds/cyclonedds/issues/1679.

dyackzan:
Hey @xygyo77, I've been seeing this error intermittently occur when I run my ROS2 application continuously for a day or so. I've been trying to reproduce this based on the test case you've described here so I can debug whatever is going on. I created a test that loop creates & destroys a bunch of publishers and subscribers. For each publisher I create 10 subscribers and then publish 100 messages before destroying the publisher and subscribers and moving to the next iteration. But I'm not seeing this `Assertion `(cf & HDL_REFCOUNT_MASK) > 0' failed.` get hit.
Could you provide more info or code on the test you ran to reproduce this issue?

xygyo77:
@dyackzan Thank you for trying to reproduce it.
Please wait as I will try to find the records from that time.

xygyo77:
@dyackzan It took me a bit of time.
I tried to find the code and unfortunately could not reproduce it. Instead, the following exception occurred.
’
[...]

Issue #1679, opened 2023-05-02
Crash when initialising a big amount of Action CLients, Services and Publishers (ROS2 Humble)


pepeRossRobotics:
Hello, while testing an Action Server that dynamically generates a lot of Publishers, Action and Service clients, the loading of them increasingly becomes slower and eventually crashes. This is the error when it crashes
```
behavior_trees_action_server: ./src/core/ddsc/src/dds_handles.c:430: dds_handle_drop_childref_and_pin: Assertion `(cf & HDL_REFCOUNT_MASK) > 0' failed.
```

I manage to go around this issue by creating the ROS clients at a lower pace and only when needed

pepeRossRobotics:
Yes we were able to reproduce it back in the day.
We were using a Behavior Tree that created 1000's of action servers services and publishers. 

The issue happened at load time as it would create all of the instances crawl down to a stop then send the error.

We ended up going around the issue by creating singleton instances of the action servers, services and publishers.

eboasson:
I think #2119 will fix it (but beware of the `releases/0.10.x` vs `master` bit)

PR #2119, opened 2024-10-29
Allow more children per entity

eboasson:
Sometimes application overflow the children counters in the handle
flags. That leads to a bad crash, which is a bug, but it is also less
than ideal than one can't make a great many child entities.

With the way the bits are mapped, one can have 16k children of a single
entity and 4k concurrent calls on a single entity. This commit changes
that in two ways:

* On 32-bit platforms, it shifts the division of bits, allowing up to
  64k children and up to 1k concurrent calls on a single entity. It is
  clearly still possible to run into the same bug, and even slightly
  more likely to do so in progress with very many threads, but this
  seems like a minor improvement nonetheless.

* On 64-bit platforms, it changes the representation to an 64-bit
  integer (it simply uses an uintptr_t), which gives plenty of room in
  the field to substantially increase the width of the counters.

Just widening the counters is not an elegant solution, and 64-bits here
is a bit wasteful. Still, this should fix the problem at least in
practice.

I expect this will address the problems in #1679 and #2022 (although the latter doesn't quite seem to reach the required numbers, the symptom matches). As that all with ROS 2, it needs to be backported to the `releases/0.10.x` branch until Cyclone master/11.0-to-be and ROS 2 are friends again. (I expect it to be a trivial backport.)


dpotman (commented):
LGTM! 

Nit-pick: the (disabled) trace logging in `print_delete` in src/core/ddsc/src/dds_entity.c:378 still has a `ddsrt_atomic_ld32` on `cnt_flags`

eboasson:
> LGTM!
> 
> Nit-pick: the (disabled) trace logging in `print_delete` in src/core/ddsc/src/dds_entity.c:378 still has a `ddsrt_atomic_ld32` on `cnt_flags`

Great catch!

dpotman (approved):

PR #2120, opened 2024-10-29
Change default for PSMX-IOX keyed topics option

dpotman:
This changes the default of the `KEYED_TOPICS` option in the Iceoryx PSMX plugin to `true`, aligning it with the pre-PSMX behavior in Cyclone 0.10.x. Despite limitations in key support within the Iceoryx integration, enabling this option by default seems a reasonable choice.


eboasson (approved):
I agree. In fact, I've been bitten by that change in behaviour often enough myself and I am the one who changed it!

Thanks!

PR #2121, opened 2024-10-29
Support all debian architectue in cpack

AlexisPotvinARA:
Cpack enforces architecture debian architecture i386 or amd64  even when compiling for arm64.
Now the architecture is detected by dpkg as per the standard debian process:
https://github.com/Kitware/CMake/blob/f3f9ad9499526963a46c10733ab9e8b3c0ebfcb1/Modules/Internal/CPack/CPackDeb.cmake#L541


eboasson (approved):
I don't know anything about CPack, but fixing the architecture at x86/x64 is obviously wrong for non-x86/x64. This seems perfectly sensible to me.

I suppose the `redhat-release` ought to be changed in a similar way, but I guess that'll have to wait until someone with such a platform figures out how to do that.

PR #2124, opened 2024-11-05
Support for IDL wstring

eboasson:
I really think no-one should be using `wstring` but ... ROS 2 does have a wide string type, and so https://github.com/ros2/rmw_cyclonedds/pull/501 runs into a problem if Cyclone doesn't do wide strings.

It probably needs more testing, I haven't yet checked how to also do this for Python, and I am still debating whether it wouldn't be better to check that the wide strings received over the network use the surrogate pairs correctly ...


src/core/cdr/src/dds_cdrstream.c
```diff
@@ -3080,6 +3376,59 @@ static bool normalize_string (char * __restrict data, uint32_t * __restrict off,
...
+static bool normalize_wstring (char * __restrict data, uint32_t * __restrict off, uint32_t size, bool bswap, size_t maxsz)
+{
+  // maxsz = character count, includes terminating L'\0' that is in-memory but not on the wire (unlike string)
+  // taking the maximum allowed count on the wire as number uint16_t, not as number of code points (which may
+  // be lower if there are surrogate pairs in the input)
+  uint32_t sz;
+  assert (maxsz > 0);
+  if (!read_and_normalize_uint32 (&sz, data, off, size, bswap))
+    return false;
+  if ((size - *off) / 2u < sz || maxsz - 1 < sz)
```
dpotman, line 3404: Assumes size is in 2-byte units

src/core/cdr/src/dds_cdrstream.c
```diff
@@ -4122,6 +4533,13 @@ static void dds_stream_extract_key_from_key_prim_op (dds_istream_t * __restrict
       is->m_index += sz;
       break;
     }
+    case DDS_OP_VAL_WSTR: case DDS_OP_VAL_BWSTR: {
+      uint32_t sz = dds_is_get4 (is);
+      dds_os_put4 (os, allocator, sz);
+      dds_os_put_bytes (os, allocator, is->m_buffer + is->m_index, 2u * sz);
```
dpotman, line 4539: See comment on line 4653

src/core/cdr/src/dds_cdrstream.c
```diff
@@ -4439,6 +4865,7 @@ static void dds_stream_read_key_impl (dds_istream_t * __restrict is, char * __re
     case DDS_OP_VAL_2BY: *((uint16_t *) dst) = dds_is_get2 (is); break;
```
dpotman, line 4865: [lines 4723 and 4805] `WCHAR`, `WSTR` and `BWSTR` cases missing (cast `subtype` and `type` to enum so that the compiler shows a warning?)

src/core/cdr/src/dds_cdrstream.c
```diff
@@ -4754,7 +5228,7 @@ static bool prtf_simple_array (char * __restrict *buf, size_t * __restrict bufsi
       }
       break;
     case DDS_OP_VAL_BLN: case DDS_OP_VAL_2BY: case DDS_OP_VAL_4BY: case DDS_OP_VAL_8BY:
-    case DDS_OP_VAL_STR: case DDS_OP_VAL_BST:
+    case DDS_OP_VAL_STR: case DDS_OP_VAL_BST: case DDS_OP_VAL_WCHAR:
```
dpotman, line 5231: `WSTR` and `BWSTR` missing

dpotman (commented):
Everything looks mostly fine, though there are a few missing cases in the IDL compiler and the serializer, and (I believe) an incorrect unit for serialized length of a `wstring`.

eboasson:
@dpotman I believe the commits I just pushed address all the comments — and a few details extra that I discovered in the process. There are also now some tests for wstring support. Good thing I did that, because that's where [2fd2600](https://github.com/eclipse-cyclonedds/cyclonedds/pull/2124/commits/2fd2600799706c03aa3e383682085179893fca3b) comes from.

Now the problem with the tests is that I disabled them on Windows, or more precisely: when using Microsoft's compiler. I know the general issue, which is that you need an extra round of macro expansion when playing games with `__VA_ARGS__`. I tried liberally sprinkling `DDSRT_COUNT_PAIRS_MSVC_WORKAROUND` all over `DDSRT_FOREACH_PAIR...` and that mostly fixes it, but that leaves a handful of preprocessor errors (of the same kind) that I am seem to unable to fix without a bit of trial-and-error.

I would've done that trial-and-error if only my little old Windows PC would still let me do a remote desktop session ... That leaves the CI, but for this kind of trial-and-error going through the CI is practically infeasible. Maybe some kind soul can help me out here ...

dpotman:
> I would've done that trial-and-error if only my little old Windows PC would still let me do a remote desktop session ... That leaves the CI, but for this kind of trial-and-error going through the CI is practically infeasible. Maybe some kind soul can help me out here ...

With some trial-and-error, I think I've fixed it! https://github.com/eclipse-cyclonedds/cyclonedds/commit/229733d67fb59d7f3c3a11516a630dea13649e35

dpotman (approved):
Everything looks good now!

PR #2132, opened 2024-11-08
feature/implement-PID_PRODUCT_VERSION

trittsv:
This PR implements `PID_PRODUCT_VERSION` to be able to see the version on wireshark.

@eboasson could you have a look?

### DIFF
--- a/src/core/ddsi/include/dds/ddsi/ddsi_protocol.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_protocol.h
@@ -59,6 +59,13 @@ typedef struct {
   uint8_t id[2];
 } ddsi_vendorid_t;
 
+typedef struct {
+  uint8_t major;
+  uint8_t minor;
+  uint8_t release;
+  uint8_t revision;
+} ddsi_product_version_t;
+
 typedef struct ddsi_protocolid {
   uint8_t id[4];
 } ddsi_protocolid_t;
--- a/src/core/ddsi/src/ddsi__protocol.h
+++ b/src/core/ddsi/src/ddsi__protocol.h
@@ -264,6 +264,7 @@ typedef union ddsi_rtps_submessage {
 #define DDSI_PID_TRANSPORT_PRIORITY                  0x49u
 #define DDSI_PID_PROTOCOL_VERSION                    0x15u
 #define DDSI_PID_VENDORID                            0x16u
+#define DDSI_PID_PRODUCT_VERSION                     0x8000u
 #define DDSI_PID_UNICAST_LOCATOR                     0x2fu
 #define DDSI_PID_MULTICAST_LOCATOR                   0x30u
 #define DDSI_PID_MULTICAST_IPADDRESS                 0x11u
--- a/src/core/ddsi/src/ddsi_discovery_endpoint.c
+++ b/src/core/ddsi/src/ddsi_discovery_endpoint.c
@@ -160,10 +160,14 @@ static int sedp_write_endpoint_impl
   else
   {
     assert (xqos != NULL);
-    ps.present |= PP_PROTOCOL_VERSION | PP_VENDORID;
+    ps.present |= PP_PROTOCOL_VERSION | PP_VENDORID | PP_PRODUCT_VERSION;
     ps.protocol_version.major = DDSI_RTPS_MAJOR;
     ps.protocol_version.minor = DDSI_RTPS_MINOR;
     ps.vendorid = DDSI_VENDORID_ECLIPSE;
+    ps.product_version.major = DDS_VERSION_MAJOR;
+    ps.product_version.minor = DDS_VERSION_MINOR;
+    ps.product_version.release = DDS_VERSION_PATCH;
+    ps.product_version.revision = DDS_VERSION_TWEAK;
 
     assert (epcommon != NULL);
 
--- a/src/core/ddsi/src/ddsi_discovery_spdp.c
+++ b/src/core/ddsi/src/ddsi_discovery_spdp.c
@@ -81,13 +81,17 @@ void ddsi_get_participant_builtin_topic_data (const struct ddsi_participant *pp,
 
   ddsi_plist_init_empty (dst);
   dst->present |= PP_PARTICIPANT_GUID | PP_BUILTIN_ENDPOINT_SET |
-    PP_PROTOCOL_VERSION | PP_VENDORID | PP_DOMAIN_ID;
+    PP_PROTOCOL_VERSION | PP_VENDORID | PP_DOMAIN_ID | PP_PRODUCT_VERSION;
   dst->participant_guid = pp->e.guid;
   dst->builtin_endpoint_set = pp->bes;
   dst->protocol_version.major = DDSI_RTPS_MAJOR;
   dst->protocol_version.minor = DDSI_RTPS_MINOR;
   dst->vendorid = DDSI_VENDORID_ECLIPSE;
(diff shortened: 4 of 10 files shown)


src/core/ddsi/src/ddsi__protocol.h
```diff
@@ -264,6 +264,7 @@ typedef union ddsi_rtps_submessage {
 #define DDSI_PID_TRANSPORT_PRIORITY                  0x49u
 #define DDSI_PID_PROTOCOL_VERSION                    0x15u
 #define DDSI_PID_VENDORID                            0x16u
+#define DDSI_PID_PRODUCT_VERSION                     0x8000u
```
eboasson, line 267: The vendor-specific flags are defined below, e.g., line 349 where it defines `DDSI_PID_ADLINK_PARTICIPANT_VERSION_INFO`. 0x8000 is already defined in this file, inherited from OpenSplice but not used by Cyclone (see https://github.com/eclipse-cyclonedds/cyclonedds/blob/d8cef891b9a27cc7bc52477e1c7b2c3f8edf81ba/src/core/ddsi/src/ddsi_plist.c#L2110 and the table below it: it doesn't occur in either). There are a lot of parameters there that it doesn't grok, come to think of it.

I suspect `0x8000` has never been used by Cyclone, so I think it can be reclaimed.

src/core/ddsi/src/ddsi_discovery_endpoint.c
```diff
@@ -160,10 +160,14 @@ static int sedp_write_endpoint_impl
   else
   {
     assert (xqos != NULL);
-    ps.present |= PP_PROTOCOL_VERSION | PP_VENDORID;
+    ps.present |= PP_PROTOCOL_VERSION | PP_VENDORID | PP_PRODUCT_VERSION;
```
eboasson, line 163: I wouldn't add it to endpoint discovery: the version is constant for all endpoints in a participant, endpoint discovery only runs after participant discovery and participant discovery messages are periodic. So putting in participant discovery both ensures you eventually will find out, and avoids sending it for potentially thousands of endpoints.

src/core/ddsi/src/ddsi_discovery_spdp.c
```diff
@@ -81,13 +81,17 @@ void ddsi_get_participant_builtin_topic_data (const struct ddsi_participant *pp,
...
   dst->present |= PP_PARTICIPANT_GUID | PP_BUILTIN_ENDPOINT_SET |
-    PP_PROTOCOL_VERSION | PP_VENDORID | PP_DOMAIN_ID;
+    PP_PROTOCOL_VERSION | PP_VENDORID | PP_DOMAIN_ID | PP_PRODUCT_VERSION;
   dst->participant_guid = pp->e.guid;
   dst->builtin_endpoint_set = pp->bes;
   dst->protocol_version.major = DDSI_RTPS_MAJOR;
   dst->protocol_version.minor = DDSI_RTPS_MINOR;
   dst->vendorid = DDSI_VENDORID_ECLIPSE;
   dst->domain_id = gv->config.extDomainId.value;
+  dst->product_version.major = DDS_VERSION_MAJOR;
```
eboasson, line 91: If you look a bit further down in the function can also see that it already puts something called "adlink version info" in. I am not sure the format is ideal, nor that we should keep it at all, but the name is irrelevant for compatibility and we could clean it up a bit.

For example, we could also redefine `struct ddsi_adlink_participant_version_info` as
```C
typedef struct ddsi_cyclonedds_version
{
  uint8_t major;
  uint8_t minor;
  uint8_t release;
  uint8_t revision;
  uint32_t flags;
  uint32_t unused[3];
  char *internals;
} ddsi_cyclonedds_version_t;
```
It has proven useful to have a bunch of flags ready for use. The string can probably be replaced by something more useful, but if we really replace it by some other type (and not just change the contents) then we would need to change the parameter id to avoid upsetting "old" Cyclone versions.

OpenSplice doesn't interpret this if it comes from Cyclone, so we really have some freedom.

src/core/ddsi/src/ddsi_discovery_topic.c
```diff
@@ -41,10 +42,14 @@ static int ddsi_sedp_write_topic_impl (struct ddsi_writer *wr, int alive, const
   ps.topic_guid = *guid;
 
   assert (xqos != NULL);
-  ps.present |= PP_PROTOCOL_VERSION | PP_VENDORID;
```
eboasson, line 44: Like with the endpoints, I wouldn't add it for the topics either. I think we should just clean up/improve version information in SPDP.

src/core/ddsi/src/ddsi_plist.c
```diff
@@ -2049,6 +2049,7 @@ static const struct piddesc piddesc_omg[] = {
 #endif
   PP  (PROTOCOL_VERSION,                    protocol_version, Xox2),
   PP  (VENDORID,                            vendorid, Xox2),
+  PP  (PRODUCT_VERSION,                     product_version, Xox2),
```
eboasson, line 2052: Almost right! `Xox2` means `o` times 2, so two octets. You really want four, which means `Xox4` but that one doesn't exist. It could be added, but it would be equivalent to `Xox2, Xox2`, or even to `Xo, Xo, Xo, Xo`. I probably would go for the latter (out of laziness? who knows ...)

This would be a vendor-specific code, and so it would need to be moved to the vendor-specific code tables, though. And as I remarked before, I think I would first see if it wouldn't make sense to clean up the `adlink_participant_info` in a (sufficiently) backwards compatible manner.

src/security/core/src/dds_security_serialize.c
```diff
@@ -49,6 +49,7 @@
 #define PID_TRANSPORT_PRIORITY                  0x49u
 #define PID_PROTOCOL_VERSION                    0x15u
 #define PID_VENDORID                            0x16u
+#define PID_PRODUCT_VERSION                     0x8000u
```
eboasson, line 52: As the product version is only informational now, I don't think anything is to be gained by also burdening the serializer used for the DDS Security with it. The clear text SPDP message will be good enough for the purpose.

Note that we shouldn't even have a separate "DDS Security" serializer. It only exists because the in-memory representation in the spec'd plug-in interface is different, and the dice landed the wrong way when a choice had to be made between adding a second serializer or having only a single serializer and converting between in-memory representations. Given the security sensitivity of (in particular de-)serializers, we should get rid of the second one.

In short, I wouldn't add it here.

src/core/ddsi/include/dds/ddsi/ddsi_protocol.h
```diff
@@ -59,6 +59,13 @@ typedef struct {
   uint8_t id[2];
 } ddsi_vendorid_t;
 
+typedef struct {
+  uint8_t major;
+  uint8_t minor;
+  uint8_t release;
+  uint8_t revision;
```
eboasson, line 66: Right now it is major, minor, "something" ("patch"?) and I don't know that having four fields won't make life more complicated. What I would find really useful is to also have a commit hash in there (or perhaps just the first few bytes to keep the size in check).

eboasson (commented):
Thank you @trittsv ! I think it makes sense to put proper version information in the discovery data, but there are a few ways in which I think this can be improved upon.

I didn't dig deep enough in the sources of Cyclone (and of old versions) to be able to say exactly which parts of the old "junk" we need to keep, or whether it really would make more sense to keep a tweaked version of what is there today. A bit of both probably will be best.

I expect it needs another look for you and me before it is clear what we should do.

PR #2145, opened 2024-11-26
Remove `restrict` (for the most part)

### LINKED ISSUE #2140: Do not #define restrict
https://github.com/eclipse-cyclonedds/cyclonedds/blob/d8cef891b9a27cc7bc52477e1c7b2c3f8edf81ba/src/core/ddsc/include/dds/dds.h#L29

Since `restrict` is a valid C++ identifier, a macro should not make it disappear. It can lead to unnecessary errors in user code.

Example:

    #include <dds/dds.h>
    int restrict = 42;

Workaround is simple: Just rename the variable.

eboasson:
A lot of "restricts" were sprinkled throughout the code, for the most part completely unnecessary because it didn't even involve performance-sensitive code. C++ does not have the "restrict" keyword, and the work-around for that was #defining it away, which breaks C++ code using "restrict" as an identifier. In short, a bit of a mess.

This removes the vast majority, including in particular all in the API and thus eliminating any need for a C++-specific workaround. The only cases that remain are in the (de)serializers because:

* they are performance-sensitive
* they read/write through "char *" pointers that are allowed to alias anything
* a cursory inspection of the output of gcc suggests it does affects the code

This PR sits on top of #2124, mainly because of laziness. The two PRs are not entirely independent (because of the serializer), but this PR can easily be changed to apply to master if this one gets the green light but the other one does not (yet?). The changes exclusive to this PR are all in a single commit, so it is pretty easy to review the restrict part without having to wade through unrelated changes.

Fixes #2140

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "VxWorks")
 endif()
 
 if(${CMAKE_C_COMPILER_ID} STREQUAL "SunPro")
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m64 -xc99 -D__restrict=restrict -D__deprecated__=")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m64 -xc99 -D__deprecated__=")
   set(CMAKE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} -m64")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m64")
--- a/docs/manual/doxygen.conf.in
+++ b/docs/manual/doxygen.conf.in
@@ -21,7 +21,7 @@ GENERATE_HTML          = NO
 GENERATE_LATEX         = NO
 GENERATE_XML           = YES
 MACRO_EXPANSION        = YES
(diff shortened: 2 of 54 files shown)


dpotman (commented):
The changes look fine to me! I agree with removing most of the `restrict`s, as in many of these code blocks it isn't relevant because there are no loops doing memory access or not used in a path where performance is critical. 

Removing `restrict` for `const` parameters and only keep it for the non-const (e.g. in plist) seems to be fine, as it follows [this](https://www.gnu.org/software/c-intro-and-ref/manual/html_node/restrict-Pointer-Example.html) example. But on the other hand, e.g. the [`memcpy`](https://en.cppreference.com/w/c/string/byte/memcpy) functions in C do use it on both parameters, although that seems to be just for clarification and not strictly necessary. 

Not completely sure about some generic ddsrt functions like `ddsrt_strlcpy` and `ddsrt_strlcat`, that pass the parameters directly into `memcpy`. So `src` and `dst` need to be `restrict`, and wouldn't it be good to show that in the declaration of these functions?

In cdrstream, what the reason for removing `restrict` on the write functions? Some of them use `memcpy`, so I'd expect the `os` parameter (or the buffer in `os`) to be `restrict`?

eboasson:
I rebased the PR on current master, which needed a few touch-ups after the fixes in the wstring-support. Those are really trivial: `dds_stream_reuse_string_bound` and `wstring_from_utf16` gained a `restrict` (I overlooked it before).

dpotman (approved):

PR #2148, opened 2024-12-10
Add restrict qualifier to serializer output buffer

eboasson:
... and two unrelated and very minor details.

Thinking about the use of `restrict` while cleaning it up in #2145, it seemed sensible to also mark the output buffer of the serializer as `restrict`. Clearly things will go horribly wrong if it aliases any of the inputs of the serializer, it is one of those essentially untyped buffers and so clearly a case where qualifying it may avoid some reloads of things after storing data in it. I don't really think it is very likely, but ok.

Those restrictions don't apply to the caller's use of the buffer before or after. So `dds_ostream_t` shouldn't be changed ... This solves this by casting a pointer to a `dds_ostream_t` to a pointer of the (new type) `restrict_ostream_t` at the entry points of the serializer. All other casts between the `ostream` variants (and there were quite a few) have been eliminated.

Just the elimination of the multitude of casts was worth the effort. If objections are raised to the use of the `restrict` qualifier for this buffer, then it makes sense to drop just that single `restrict` keyword that this PR adds and keep all the other changes. (Perhaps with the `restrict_ostream_t` renamed, too 🙂)

### DIFF
--- a/src/core/cdr/src/dds_cdrstream.c
+++ b/src/core/cdr/src/dds_cdrstream.c
@@ -22,6 +22,35 @@
 #include "dds/cdr/dds_cdrstream.h"
 #include "dds/ddsc/dds_data_type_properties.h"
 
+typedef struct restrict_ostream_base {
+  unsigned char * restrict m_buffer;
+  uint32_t m_size;          /* Buffer size */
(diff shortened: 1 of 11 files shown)


dpotman (approved):
LGTM

PR #2154, opened 2025-01-09
Add check for nested optional members in key

### LINKED ISSUE #2151: IDL compiler allow optional key when nested in struct
The IDL compiler allows a value to be an optional key when the optional annotation is nested inside a struct. However, this behavior is not consistent with the specification in 7.2.2.4.4.4.8 Key Members, which states that "Key members shall never be optional, and they shall always have their 'must understand' bit set." To resolve this discrepancy, the IDL compiler and type object validation code should block this behavior.

Here is an example data structure that can be used to validate the error.

```
struct Plate {
    string number;
    @optional string state;
};
struct Car {
    @key Plate licensePlate;
    string color;
};
```

dpotman:
Optional members as part of a key in a data type are not allowed. A check for optional members in nested types was missing, and is introduced in this commit. The checks is added in the IDL compiler and in the runtime type meta-data validator.

Fixes #2151

### DIFF
--- a/src/core/ddsi/src/ddsi_typewrap.c
+++ b/src/core/ddsi/src/ddsi_typewrap.c
@@ -853,7 +853,7 @@ static dds_return_t xt_valid_type_flags (struct ddsi_domaingv *gv, uint16_t flag
 #define M DDS_XTypes_IS_MUST_UNDERSTAND
 #define K DDS_XTypes_IS_KEY
 #define D DDS_XTypes_IS_DEFAULT
-static dds_return_t xt_valid_member_flags (struct ddsi_domaingv *gv, uint16_t flags, uint8_t member_flag_kind)
+static dds_return_t xt_valid_member_flags (struct ddsi_domaingv *gv, uint16_t flags, uint8_t member_flag_kind, bool in_key)
 {
   dds_return_t ret = DDS_RETCODE_OK;
 
@@ -870,6 +870,10 @@ static dds_return_t xt_valid_member_flags (struct ddsi_domaingv *gv, uint16_t fl
     case MEMBER_FLAG_STRUCT_MEMBER:
       if (flags & ~(T1|T2|O|M|K|X))
         ret = DDS_RETCODE_BAD_PARAMETER;
+      if (in_key && (flags & O))
+        ret = DDS_RETCODE_BAD_PARAMETER;
+      if ((flags & O) && (flags & K))
+        ret = DDS_RETCODE_BAD_PARAMETER;
       break;
     case MEMBER_FLAG_UNION_MEMBER:
       if (flags & ~(T1|T2|D|X))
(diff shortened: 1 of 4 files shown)


src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -953,25 +959,27 @@ static dds_return_t xt_validate_impl (struct ddsi_domaingv *gv, const struct xt_
         return ret;
       for (uint32_t n = 0; n < t->_u.structure.members.length; n++)
       {
-        if ((ret = xt_valid_member_flags (gv, t->_u.structure.members.seq[n].flags, MEMBER_FLAG_STRUCT_MEMBER)))
+        DDS_XTypes_StructMemberFlag flags = t->_u.structure.members.seq[n].flags;
+        bool key = in_key || (flags & DDS_XTypes_IS_KEY);
```
eboasson, line 963: Would this not cause it to reject:
```
struct A { @key x; @optional y; };
struct B { @key A z; };
```
* validation starts with `B`, `in_key` = `false`
* field `z`: key flag is set, so validation of `A` starts with `in_key` = `true`
* when it gets to field `y`, `xt_valid_member_flags` is called with `in_key` = `true` and so rejects it at line 873 (`if (in_key && (flags & O))`)

In my understanding, because `A` contains some members marked `@key`, the members not marked `@key` (i.e., `y`) are not key fields. It `x` hadn't marked as key, then of course rejecting it would've been correct.

I think it needs a separate run over the structure members to check for the presence of some `@key` annotations.

dpotman: Good point, I missed that. In [88cfe84](https://github.com/eclipse-cyclonedds/cyclonedds/pull/2154/commits/88cfe84edfa1da5bb3b6d7b99ce4c8406294e44a) I've addressed this, and also take the members in the base type into account for this.

src/core/ddsi/src/ddsi_typewrap.c
```diff
@@ -957,10 +957,16 @@ static dds_return_t xt_validate_impl (struct ddsi_domaingv *gv, const struct xt_
...
+      bool has_key_members = false;
+      for (const struct xt_type *t1 = t; t1 && ddsi_xt_is_resolved (t1); t1 = t1->_u.structure.base_type ? &t1->_u.structure.base_type->xt : NULL)
+        for (uint32_t n = 0; !has_key_members && n < t1->_u.structure.members.length; n++)
+          has_key_members = (t1->_u.structure.members.seq[n].flags & DDS_XTypes_IS_KEY);
+
       for (uint32_t n = 0; n < t->_u.structure.members.length; n++)
       {
         DDS_XTypes_StructMemberFlag flags = t->_u.structure.members.seq[n].flags;
-        bool key = in_key || (flags & DDS_XTypes_IS_KEY);
+        bool key = (in_key && !has_key_members) || (flags & DDS_XTypes_IS_KEY);
```
eboasson, line 969: When I reviewed it the first time I completely forgot about base types 😱 Good thing you thought of it now 🙂

So ... on looking at this again ... in
```
struct A { x; };
struct B { @key A y; };
struct C { B z; };
```
`x` is not a key when considering `C` as the type of a topic, but it would be a key if considering `B` as the type of a topic. Should that make `C` fail verification if `x` were optional? What if no-one tries to make a topic out of `B`?

I can see arguments both ways: the one combination used has no problems, but then the IDL is nonsensical. So rejecting `B` and thus also `C` if `x` were made optional is reasonable.

I do think it would be wise to point out in a small comment that the `key` can be `true` here without the field actually acting as a key field in the topic type, and that we choose to reject cases like the above.

dpotman: Added the comment in 19a9679

eboasson (approved):
LGTM! Thank you @dpotman ☺️

PR #2158, opened 2025-01-14
Improve PSMX a bit: multiple plugins, interface improvements, documentation

eboasson:
... and some odds and ends ...

I'm happy to move the odds-and-ends into a separate PR if people want me to, it's just that I think they are some trivial things I came across while doing this.


src/core/ddsc/include/dds/ddsc/dds_psmx.h
```diff
@@ -32,365 +163,504 @@ extern "C" {
...
+    dds_data_type_properties_t data_type_props,
+    const struct ddsi_type *type_definition,
+    uint32_t sizeof_type);
+
+/**
+ * @brief Definition of the function for destructing and freeing a PSMX Topic
+ * @ingroup psmx
+ *
+ * Definition for a function that is called on deleting the topic in the DDS Domain.
+ * Called exactly once for each successful invocation of `dds_psmx_topic_new`, all
```
dpotman, line 277: `dds_psmx_topic_new` -> `create_topic` (or `create_topic_type`?)

And below: `dds_psmx_cleanup_topic_generic` -> `dds_psmx_topic_cleanup_generic`

(same in `dds__psmx.h` for `_int` variant)

src/core/ddsc/include/dds/ddsc/dds_psmx.h
```diff
@@ -32,365 +163,504 @@ extern "C" {
...
 
 /**
- * @brief Definition of function to write data on a PSMX endpoint
+ * @brief Definition of function for writing data via a PSMX Writer
  * @ingroup psmx
  *
- * @param[in] psmx_endpoint    The endpoint to publish the data on
- * @param[in] data    The data to publish
- * @returns a DDS return code
+ * Equivalent to `dds_psmx_endpoint_write_ih_fn` with `keysz` = 0 and `key` = NULL
```
dpotman, line 492: `dds_psmx_endpoint_write_ih_fn` -> `dds_psmx_endpoint_write_key_fn`

dpotman (commented):
Overall, it looks good! Below are a few comments regarding some details in the documentation and a few minor points in the code.

eboasson:
> Overall, it looks good! Below are a few comments regarding some details in the documentation and a few minor points in the code.

Thank you very much for those comments, they really are quite important 🙂

I believe I've addressed all of the comments, but one: the question whether it would not be better to provide the complete type instead of the minimal type to the plugin. I agree, I just need to figure out why the existing code retrieved the minimal type and what the consequence is of changing that.

eboasson:
@dpotman The commits I just pushed should take care of the remaining bit. I think it is worthy of a re-reviewing.

dpotman:
I had already started reviewing the commits for other fixes, and these last three look good as well! The PSMX config string handling is a bit tricky to understand, but the code comments and tests make it manageable😄. Just one small detail: the comment in `dds_pubsub_message_exchange_configstr` mentions _"We also add `SERVICE_NAME=config_name` if no `SERVICE_NAME` set"_ but this has now been renamed to `INSTANCE_NAME`. Other than that, the changes look good.

dpotman (approved):
That updated comment helps! I also noticed in CI that a C++ test (`invalid_qos`) fails, probably because of the PSMX instance name check.

eboasson:
> That updated comment helps!

Thanks 🙂

> I also noticed in CI that a C++ test (`invalid_qos`) fails, probably because of the PSMX instance name check.

That was also my conclusion. I'll do a PR for that and wait with merging this one until that one is ready to merge (or merged) before merging this one. I think that's the proper order.

PR #2163, opened 2025-01-20
Fix ddsrt init counter overflow

### LINKED ISSUE #2156: Reference counter overflow in ddsrt_init results in busy loop in ROS2 humble rclpy
Hello all,

We are experiencing a stall in of our ROS2 humble nodes in the field after continuous working for 3 weeks. 
The stall introduces itself as cpu throttling to 100% and the ROS2 callbacks to be not being serviced which points to a busy loop. I have traced the issue to a reference counter `init_status` overflowing to 0 after continuous increment over UINT_MAX32 in `ddsrt_init` function.

What is my theory?
When ROS2 calls for spin() function it is then eventually stepped into `Executor._wait_for_ready_callbacks` function which creates a list of subscriptions, timers, waitables.. and puts them into a `WaitSet` object [here](https://github.com/ros2/rclpy/blob/296425c4183135c048dff8fd22a4c0d44b39b0f4/rclpy/rclpy/executors.py#L742). This object creation path goes through `rcl_wait_set_init` -> `rmw_create_wait_set` -> `dds_create_waitset` -> `dds_init` -> `ddsrt_init` and increments the reference counter `init_status` there. But the inverse doesn't seem to hold true. Destruction of this object through `rcl_wait_set_fini` -> `rmw_destroy_wait_set` doesn't end the path on `ddsrt_fini` so the reference count never decreases which contradicts this [statement](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/modules.md#constructors-and-destructors).

What does this result in?
After (UINT_MAX32 - 0x80000001u / 1000HZ(my node's callback frequency) -> 24 days) `init_status` wraps to 0 and we hit retry loop [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/6befb5678745251060f18885483fc3560f6fa5ad/src/ddsrt/src/cdtors.c#L40C1-L40C6).
I would have expected this to happen other nodes as well but I think the prerequisite is single node in single container. If there was another node it'd would have incremented the counter to 1 so the busy wait status would have lasted intermittently(not 100% sure about this part).

Reproduction steps?
Here is my docker setup:

```
FROM ros:humble
SHELL ["/bin/bash", "-c"]
WORKDIR /workspace

ENV RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
RUN apt update && apt install -y python3-pip sqlite3 ros-humble-example-interfaces ssh gdb git ros-humble-rmw-cyclonedds-cpp ros-humble-cyclonedds-dbgsym ros-humble-rclcpp-dbgsym ros-humble-rmw-cyclonedds-cpp-dbgsym ros-$ROS_DISTRO-demo-nodes-py ros-$ROS_DISTRO-demo-nodes-cpp
COPY . .

# Set up SSH keys for GitHub access
RUN mkdir -p ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts

# Install Python dependencies from requirements.txt
RUN --mount=type=ssh python3 -m pip install -r requirements.txt

# Build the workspace
RUN source /opt/ros/$ROS_DISTRO/setup.bash && colcon build --cmake-args -DCMAKE_BUILD_TYPE=Debug
```

I create two instances of the container. I run
`ros2 run demo_nodes_py listener`
and
`ros2 run demo_nodes_py talker`
in each.

I inspect init_status using GDB + bfptrace and observe it incrementing without decrementing

```
sudo bpftrace -e '
u:/proc/149514/root/opt/ros/humble/lib/x86_64-linux-gnu/libddsc.so.0:ddsrt_init
{
    printf("Value of init_status: %u\n", * 0x745f3cfbf2e8);
}'

Value of init_status: 2147485586
Value of init_status: 2147485587
Value of init_status: 2147485588
Value of init_status: 2147485589
Value of init_status: 2147485590
Value of init_status: 2147485591
Value of init_status: 2147485592
Value of init_status: 2147485593
Value of init_status: 2147485594
Value of init_status: 2147485595
Value of init_status: 2147485596
Value of init_status: 2147485597
Value of init_status: 2147485598
Value of init_status: 2147485599
Value of init_status: 2147485600
Value of init_status: 2147485601
```

I use GDB to change the memory of `init_status` to a number close UINT_MAX32. After some time the issue surfaces:

```
Value of init_status: 4294967294
Value of init_status: 4294967295
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
Value of init_status: 0
```
CPU throttles and listening stops:

Screenshot from 2025-01-13 11-18-29

eboasson:
Fixes #2156

### DIFF
--- a/src/core/ddsc/src/dds_init.c
+++ b/src/core/ddsc/src/dds_init.c
@@ -98,9 +98,16 @@ dds_return_t dds_init (void)
     case CDDS_STATE_READY:
       assert (dds_global.m_entity.m_hdllink.hdl == DDS_CYCLONEDDS_HANDLE);
       ddsrt_mutex_unlock (init_mutex);
+      // Undo the ddsrt init counter increment: some other thread initialized
+      // the library and we continue with that pre-initialized state.  At some
+      // point dds_fini() will be called, and that will undo the ddsrt_fini()
+      // associated with the library initialization.
+      ddsrt_fini ();
       return DDS_RETCODE_OK;
     case CDDS_STATE_ZERO:
       ddsrt_atomic_st32 (&dds_state, CDDS_STATE_STARTING);
+      // Initializing Cyclone, this thread did the ddsrt_init() call that will
+      // be undone by dds_fini()
       break;
     default:
       ddsrt_mutex_unlock (init_mutex);
--- a/src/ddsrt/src/cdtors.c
+++ b/src/ddsrt/src/cdtors.c
@@ -19,49 +19,83 @@
 /* Sockets API initialization is only necessary on Microsoft Windows. The
    overly correct approach would be to abstract this a little further, but for
    now a pragmatic approach will suffice. */
-extern void ddsrt_winsock_init(void);
-extern void ddsrt_winsock_fini(void);
-extern void ddsrt_time_init(void);
-extern void ddsrt_time_fini(void);
+extern void ddsrt_winsock_init (void);
+extern void ddsrt_winsock_fini (void);
+extern void ddsrt_time_init (void);
+extern void ddsrt_time_fini (void);
 #endif
 
-#define INIT_STATUS_OK 0x80000000u
-static ddsrt_atomic_uint32_t init_status = DDSRT_ATOMIC_UINT32_INIT(0);
+#define INIT_STATUS_OK    0x80000000u
+#define INIT_COUNT_MASK   0x7fffffffu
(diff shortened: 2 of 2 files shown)


dpotman (commented):
LGTM

PR #2164, opened 2025-01-20
Use epoll instead of select on Linux

### LINKED ISSUE #2161: select error
The file descriptor exceeds 1024, resulting in a select exception and causing the CPU of the recv thread to surge. Why not use epoll.

eboasson:
Fixes #2161 — that's the plan anyway 🙂

### DIFF
--- a/src/core/ddsi/src/ddsi_sockwaitset.c
+++ b/src/core/ddsi/src/ddsi_sockwaitset.c
@@ -25,9 +25,12 @@
 #define MODE_KQUEUE 1
 #define MODE_SELECT 2
 #define MODE_WFMEVS 3
+#define MODE_EPOLL  4
 
 #if defined __APPLE__
 #define MODE_SEL MODE_KQUEUE
+#elif defined __linux
+#define MODE_SEL MODE_EPOLL
 #elif defined WINCE
 #define MODE_SEL MODE_WFMEVS
 #else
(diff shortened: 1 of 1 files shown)


src/core/ddsi/src/ddsi_sockwaitset.c
```diff
@@ -281,6 +284,252 @@ int ddsi_sock_waitset_next_event (struct ddsi_sock_waitset_ctx * ctx, struct dds
...
+  ddsrt_mutex_init (&ws->lock);
+  return ws;
+
+fail_fcntl:
+fail_add_trigger:
+  close (ws->pipe[0]);
+  close (ws->pipe[1]);
+fail_pipe:
+  close (ws->epfd);
+fail_kqueue:
```
dpotman, line 391:
```suggestion
fail_epoll_create:
```

src/core/ddsi/src/ddsi_sockwaitset.c
```diff
@@ -281,6 +284,252 @@ int ddsi_sock_waitset_next_event (struct ddsi_sock_waitset_ctx * ctx, struct dds
...
+  int ret;
+  ddsrt_mutex_lock (&ws->lock);
+  ret = add_entry_locked (ws, conn, ddsi_conn_handle (conn));
+  ddsrt_mutex_unlock (&ws->lock);
+  return ret;
+}
+
+void ddsi_sock_waitset_purge (struct ddsi_sock_waitset * ws, unsigned index)
+{
+  /* Sockets may have been closed by the Purge is called, but any closed sockets
```
dpotman, line 434: Some references to 'kqueue' in the comment (also on line 485), and a few words missing in the sentence '.. by the Purge is called' (also in the original in the kqueue implementation)

dpotman (commented):
Looks good and works fine on my Ubuntu 24! Only a few details in the code comments.

PR #2169, opened 2025-02-10
Bugfix/fix-android

trittsv:
It seems that Android builds on the master are broken since quite some time.

This PR fixes Android builds:
- `ushort` is not a valid datatype in C, replaced with `unsigned short`
- `pthread_attr_setaffinity_np` is not portable and android ndk does not seem to support it


eboasson (approved):
Thanks @trittsv !

I think at some point it would be better to error out on platforms where we don't support setting the thread affinity, but as that currently is the case for all platforms other than Linux, I don't see a reason why this PR would have to fix that.

PR #2170, opened 2025-02-10
bugfix/fix-vs2017-build

trittsv:
This PR fixes the build with older MSVC compilers like Visual Studio 2017.


eboasson (approved):
Thank you @trittsv !

PR #2172, opened 2025-02-11
Some fixes for edge cases

eboasson:
This PR contains fixes for the following problems:
* deleting a reader with outstanding PSMX loans
* waitset_wait not getting unblocked by set_status_mask
* a crash when trying to create the 33rd query condition on a reader
* silently ignoring thread scheduling affinity settings


dpotman (approved):
LGTM!

PR #2173, opened 2025-02-13
Bump Ubuntu versions on CI

eboasson:
Triggered by the failure of `precommit` on Ubuntu 20.04 after a VM update. It is already 2025, so we might as well switch from 20.04 / 22.04 to 22.04 / 24.04. The `precommit` problem goes away that way.

Unfortunately it also introduces some new problems ...


dpotman (commented):
Looks good, failing tests on CI are unrelated. Seems like `.azure/coverity-scan.yml:29` also needs a version bump, the Coverity build on CI also fails on the pre-commit step.

PR #2178, opened 2025-02-17
Fix segv on null param_seq in member annotations

eboasson:
The parameter sequence for a custom annotation is an optional field, and so may be a null pointer.


dpotman (approved):
LGTM!

PR #2180, opened 2025-02-20
ddsi_time_t.seconds should be uint32_t

binbowang1987:
ref DDSI-RTPS 9.3.2.1 IDL Definitions 
``` idl
struct Time_t {
  unsigned long seconds; // time in seconds
  unsigned long fraction; // time in sec/2^32
};
```

### DIFF
--- a/src/core/ddsi/src/ddsi__time.h
+++ b/src/core/ddsi/src/ddsi__time.h
@@ -21,11 +21,11 @@ extern "C" {
 #endif
 
 typedef struct {
-  int32_t seconds;
+  uint32_t seconds;
   uint32_t fraction;
 } ddsi_time_t;
-#define DDSI_TIME_INFINITE ((ddsi_time_t) { INT32_MAX, UINT32_MAX })
-#define DDSI_TIME_INVALID ((ddsi_time_t) { -1, UINT32_MAX })
+#define DDSI_TIME_INFINITE ((ddsi_time_t) { UINT32_MAX, UINT32_MAX - 1 })
+#define DDSI_TIME_INVALID ((ddsi_time_t) { UINT32_MAX, UINT32_MAX })
 
 typedef ddsi_time_t ddsi_duration_t;
 
--- a/src/core/ddsi/src/ddsi_plist.c
+++ b/src/core/ddsi/src/ddsi_plist.c
@@ -349,7 +349,7 @@ static dds_return_t deser_reliability (void * restrict dst, struct flagset *flag
     return DDS_RETCODE_BAD_PARAMETER;
   if (kind < 1 || kind > 2)
     return DDS_RETCODE_BAD_PARAMETER;
-  mbt.seconds = (int32_t) mbtsec;
+  mbt.seconds = mbtsec;
   mbt.fraction = mbtfrac;
   if (validate_external_duration (&mbt) < 0)
     return DDS_RETCODE_BAD_PARAMETER;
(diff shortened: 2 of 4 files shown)


binbowang1987:
when encounter this timestamp, it will be overflowed and cause cxx version to throw exception.
``` cpp
void dds::core::Time::sec(int64_t s)
{
    if(s < 0 && s != -1) {
        ISOCPP_THROW_EXCEPTION(ISOCPP_ERROR, "dds::core::Time::sec out of bounds");
    } else {
        /** @internal @bug OSPL-2308 RTF Time-ish coercion issue
        @see http://jira.prismtech.com:8080/browse/OSPL-2308 */
        sec_ =  s;
    }
}
```

CleanShot 2025-02-20 at 20 50 59@2x
CleanShot 2025-02-20 at 20 51 55@2x

eboasson:
Hi @binbowang1987, thank you for the PR ☺️ Don't you love the OMG specifications ... You're entirely correct that DDSI-RTPS 2.5 states this:

> ref DDSI-RTPS 9.3.2.1 IDL Definitions
> 
> ```idl
> struct Time_t {
>   unsigned long seconds; // time in seconds
>   unsigned long fraction; // time in sec/2^32
> };
> ```

but it also states:

> 9.4.2.9 Timestamp
> The PSM mapping for the `Timestamp` SubmessageElement defined in 8.3.5.8 is given by the following IDL
> ```
> typedef Time_t Timestamp;
> ```
> Following the CDR encoding, the wire representation of the `Timestamp` SubmessageElement is:
> ```
> 0...2...........8...............16.............24. ............. 31
> +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
> | long seconds                                                  |
> +---------------+---------------+---------------+---------------+
> | unsigned long fraction                                        |
> +---------------+---------------+---------------+---------------+
> ```

so ... great. I was pretty certain I copied it correctly from DDSI 2.1, and indeed, there it says in table 9.4:
> Mapping of the type
> ```
> struct {
>   long seconds; // time in seconds
>   unsigned long fraction; // time in sec/2^32
> } Time_t;
> ```

They just changed it! Yet another case of not caring about backwards compatibility. If you then look at the DDS DCPS 1.4 specification, which defines the semantics and API of DDS (the DDSI spec merely defines an interoperable networking protocol for this DCPS specification):
> ```
> struct Time_t {
>   long sec;
>   unsigned long nanosec;
> };
> ```
the place where the application can inject a timestamp that makes it onto the wire is the `write_w_timestamp` (and some similar operations), and those don't specify that the timestamp must be ≥0, so the application is allowed to set -1234.5678 as a timestamp.

Therefore the change in the DDSI that happened somewhere between 2.1 and 2.5 is actually not compatible with the top-level specification of DDS. And so, once again, the OMG working group for DDS failed to do at least a half-decent job.

That said, the C and C++ API's of Cyclone have always required the timestamp to be non-negative in `dds_write_ts` and friends (in contradiction to the DCPS 1.4 specification). That means the change in wire representation is pretty safe for Cyclone. The only case where it is a problem is in receiving data from (e.g.) a DDSI 2.1 peer with a negative timestamp, and that can be handled by choosing the correct transformation routine based on the DDSI version number in the protocol header. It will require bumping the protocol version Cyclone writes in the header, but I already have bumping that on my todo list.

Your PR seems to have failed all the checks in the CI. I need to look into that and also see what else needs to change given to operate correctly with DDSI 2.1 and for the C++ binding to handle valid negative time stamps on reception.

eboasson (commented):
It turns out life is more complicated than this: "duration" is unchanged, then there is the backwards compatibility with versions 2.1 and 2.2 ... Then there are some things in Cyclone that I discovered were broken already (insufficient validation of source timestamps in `dds_write_ts` and friends, what I believe to be incorrect validation of incoming source timestamps).

All-in-all it is quite a bit.

Instead of doing a "normal" review, I decided to take your PR and put some commits on top of it that would (I assume) pass my review. Perhaps you'd be willing to take a look at #2192?

binbowang1987:
Yes, my change does not considered version compatibility.  
Thanks for your work @eboasson , I will close this comment and follow https://github.com/eclipse-cyclonedds/cyclonedds/pull/2192.

PR #2181, opened 2025-02-24
Fix order of dimensions for typedef of array of bounded string.

mvandenhoek:
An array of bounded string was mapped incorrectly from IDL to C in the sense that it put the dimensions of the array and the string bound in the wrong order, i.e.

```idl
typedef string<16>  str_arr[3];
```
becomes
```c
typedef char Space_str_arr[17][3];
```
when it should be
```c
typedef char Space_str_arr[3][17];
```


eboasson (approved):
Thank you @mvandenhoek !

PR #2198, opened 2025-03-24
security add support for pkcs11

MarcelJordense:
For DDS security added support to have the private key's and the certificates stored in a hsm using the openssl pkcs11 provider

### DIFF
--- a/src/security/builtin_plugins/authentication/src/auth_utils.c
+++ b/src/security/builtin_plugins/authentication/src/auth_utils.c
@@ -127,67 +127,66 @@ DDS_Security_ValidationResult_t get_subject_name_DER_encoded(const X509 *cert, u
 
 static DDS_Security_ValidationResult_t check_key(EVP_PKEY *key, const char *key_type, bool isPrivate, DDS_Security_SecurityException *ex)
 {
-    DDSRT_UNUSED_ARG(key_type);
+  DDSRT_UNUSED_ARG(key_type);
 
-    if (EVP_PKEY_id(key) == EVP_PKEY_RSA)
+  if (EVP_PKEY_id(key) == EVP_PKEY_RSA)
+  {
+    if (isPrivate)
     {
-        if (isPrivate)
-        {
-            RSA *rsaKey = EVP_PKEY_get1_RSA(key);
-            const bool fail = (rsaKey && RSA_check_key(rsaKey) != 1);
-            RSA_free(rsaKey);
-            if (fail)
-            {
-                DDS_Security_Exception_set_with_openssl_error(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, DDS_SECURITY_VALIDATION_FAILED, "RSA key not correct : ");
-                return DDS_SECURITY_VALIDATION_FAILED;
-            }
-        }
+      const RSA *rsaKey = EVP_PKEY_get0_RSA(key);
+      const bool fail = (rsaKey && (RSA_check_key(rsaKey) != 1) && RSA_get0_p(rsaKey));
+      if (fail)
+      {
+        DDS_Security_Exception_set_with_openssl_error(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, DDS_SECURITY_VALIDATION_FAILED, "RSA key not correct : ");
+        return DDS_SECURITY_VALIDATION_FAILED;
+      }
     }
-    else
+  }
+  else
+  {
+    assert(EVP_PKEY_id(key) == EVP_PKEY_EC);
+    EC_KEY *ecKey = EVP_PKEY_get1_EC_KEY(key);
+    const bool fail = (ecKey && EC_KEY_check_key(ecKey) != 1);
+    EC_KEY_free(ecKey);
+    if (fail)
     {
-        assert(EVP_PKEY_id(key) == EVP_PKEY_EC);
-        EC_KEY *ecKey = EVP_PKEY_get1_EC_KEY(key);
-        const bool fail = (ecKey && EC_KEY_check_key(ecKey) != 1);
-        EC_KEY_free(ecKey);
-        if (fail)
-        {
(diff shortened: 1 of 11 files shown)


src/security/builtin_plugins/authentication/src/authentication.c
```diff
@@ -650,6 +655,45 @@ static DDS_Security_ValidationResult_t get_adjusted_participant_guid(X509 *cert,
...
+  dds_security_authentication_impl *auth = arg;
+  const char *pkcs11 = "pkcs11";
+
+  if (OSSL_PROVIDER_available(NULL, pkcs11))
+    return true;
+
+  ddsrt_mutex_lock(&auth->lock);
+  auth->pkcs11Provider = OSSL_PROVIDER_try_load(NULL, pkcs11, 1);
+  ddsrt_mutex_unlock(&auth->lock);
+  return (auth->pkcs11Provider != NULL);
```
eboasson, line 693: If `pkcs11Provider` is checked outside the lock, then why do we need the mutex here in the first place?

src/security/core/tests/common/cert_utils.c
```diff
@@ -134,3 +137,179 @@ char * generate_identity(const char * ca_cert_str, const char * ca_priv_key_str,
...
+    return true;
+#endif
+  return false;
+}
+
+char * generate_pkcs11_private_key(const char *token, const char *name, uint32_t id, const char *pin)
+{
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+  const char template[] = "pkcs11:token=%s;object=%s;id=%u?pin-value=%s";
+  size_t len = sizeof (template) + strlen (token) + strlen (name) + 8 + strlen (pin);
```
eboasson, line 163: Should that be `+ 10`?

src/security/core/tests/common/security_config_test_utils.c
```diff
@@ -395,3 +396,46 @@ char * get_permissions_config(char * grants[], size_t ngrants, bool add_prefix)
...
+    { "ALLOW_UNAUTH_PP", allow_unauth_pp ? "true" : "false", 1 },
+    { "ENABLE_JOIN_AC", enable_join_ac ? "true" : "false", 1 },
+    { "DISCOVERY_PROTECTION_KIND", pk_to_str (discovery_protection_kind), 1 },
+    { "LIVELINESS_PROTECTION_KIND", pk_to_str (liveliness_protection_kind), 1 },
+    { "RTPS_PROTECTION_KIND", pk_to_str (rtps_protection_kind), 1 },
+    { "TOPIC_RULES", topic_rules != NULL ? topic_rules : get_governance_topic_rule (NULL, false, false, false, false, PK_N, BPK_N), 1 },
+    { NULL, NULL, 0 }
+  };
+  char * config = ddsrt_expand_vars (governance_xml, &expand_lookup_vars, vars);
+  char * config_signed = signer (config, signer_arg);
```
eboasson, line 413: It looks me to that `get_governance_config(a,b,c,d,e,f,g)` is equivalent to `get_governance_config_ex(a,b,c,d,e,f,X,NULL,g)` with
```C
char *X(const char *config, void *arg) {
  (void)arg;
  return get_signed_data(config);
}'
```
And similarly for `get_permissions_config_ex`. With `get_signed_data` used only in this particular context, it would make sense to add that unused argumnet to `get_signed_data` to avoid the wrapper function.

Maybe I am overlooking a detail ...

src/security/builtin_plugins/authentication/src/auth_utils.c
```diff
@@ -127,67 +127,66 @@ DDS_Security_ValidationResult_t get_subject_name_DER_encoded(const X509 *cert, u
...
-            const bool fail = (rsaKey && RSA_check_key(rsaKey) != 1);
-            RSA_free(rsaKey);
-            if (fail)
-            {
-                DDS_Security_Exception_set_with_openssl_error(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, DDS_SECURITY_VALIDATION_FAILED, "RSA key not correct : ");
-                return DDS_SECURITY_VALIDATION_FAILED;
-            }
-        }
+      const RSA *rsaKey = EVP_PKEY_get0_RSA(key);
+      const bool fail = (rsaKey && (RSA_check_key(rsaKey) != 1) && RSA_get0_p(rsaKey));
```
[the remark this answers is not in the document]
MarcelJordense: The RSA_get0_p is used to detect if the key is stored on an HSM. When RSA_get0_p return 0 for the p parameter then it is assumed that the key is stored on a HSM and is not accessible. Thus when RSA_get0_p returns 0 then test on the valid the RSA private key can not be performed and the fail check will be disabled.

src/security/builtin_plugins/authentication/src/auth_utils.c
```diff
@@ -427,6 +426,262 @@ static DDS_Security_ValidationResult_t load_private_key_from_file(const char *fi
...
+
+  if (!(engine = ENGINE_by_id("pkcs11")))
+  {
+    DDS_Security_Exception_set(ex, DDS_AUTH_PLUGIN_CONTEXT, DDS_SECURITY_ERR_UNDEFINED_CODE, DDS_SECURITY_VALIDATION_FAILED, "Failed to find pkcs11 engine");
+    return DDS_SECURITY_VALIDATION_FAILED;
+  }
+
+  if (password)
+  {
+    const char *pin_str = "?pin_value=";
```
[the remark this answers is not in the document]
MarcelJordense: The pin value can be specified as part of the uri. Therefor we have decided to disable the adding of the pin value (password) to the uri for the moment. A comment has been added.

src/security/core/tests/common/cert_utils.c
MarcelJordense: Yes should be 10.

src/security/core/tests/common/security_config_test_utils.c
MarcelJordense: Yes, I agree.

eboasson (approved):
Thank you @MarcelJordense ☺️

PR #2202, opened 2025-04-08
Remove code for dependencies between participants

eboasson:
The code base inherited (from OpenSplice) some features to handle participants dependent on another participant, for example a participant for which all discovery and application data is handle by a gateway process.  This is a useful feature in principle, but it hasn't been used or tested in Cyclone in years and it now turns out to cause incorrect handling of the writer liveliness in combination with, of all implementations, OpenSplice.

This removes that (now broken) support because cleaning up, fixing the code and adding tests is too big a project at this time, and all the old code visible in the history.

### DIFF
--- a/src/core/ddsi/src/ddsi_proxy_participant.c
+++ b/src/core/ddsi/src/ddsi_proxy_participant.c
@@ -92,53 +92,6 @@ static void proxy_participant_replace_minl (struct ddsi_proxy_participant *proxy
   ddsrt_atomic_stvoidp (manbypp ? &proxypp->minl_man : &proxypp->minl_auto, lnew);
 }
 
-void ddsi_proxy_participant_reassign_lease (struct ddsi_proxy_participant *proxypp, struct ddsi_lease *newlease)
-{
-  ddsrt_mutex_lock (&proxypp->e.lock);
-  if (proxypp->owns_lease)
-  {
-    struct ddsi_lease *minl = ddsrt_fibheap_min (&ddsi_lease_fhdef_pp, &proxypp->leaseheap_auto);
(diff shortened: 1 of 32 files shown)


src/core/ddsi/src/ddsi_proxy_participant.c
```diff
@@ -282,15 +235,13 @@ void ddsi_proxy_participant_add_pwr_lease_locked (struct ddsi_proxy_participant
...
+  if (minl_prev != minl_new)
   {
     ddsrt_etime_t texp = ddsrt_etime_add_duration (ddsrt_time_elapsed (), minl_new->tdur);
     struct ddsi_lease *lnew = ddsi_lease_new (texp, minl_new->tdur, minl_new->entity);
     if (minl_prev == NULL)
     {
-      assert (manbypp);
-      assert (ddsrt_atomic_ldvoidp (&proxypp->minl_man) == NULL);
-      ddsrt_atomic_stvoidp (&proxypp->minl_man, lnew);
+      ddsrt_atomic_stvoidp (manbypp ? &proxypp->minl_man : &proxypp->minl_auto, lnew);
```
dpotman, line 244: `minl_auto` always has a lease set (in `ddsi_new_proxy_participant`), so I think the asserts that were here are still valid and this path should never store a lease in `proxypp->minl_auto`?

eboasson: I think you're right: it used to be that lease was present only if `owns_lease` was set, but this PR changes that. That I removed the asserts is not because of that change, but because I only later decided to remove all that code for `owns_lease` = `false` hence had to remove them, and then forgot to put them back.

I'll have another look just to be sure 😀

dpotman (approved):

PR #2205, opened 2025-04-10
Fix incorrect Type Identifier for multidimensional arrays in IDLC

dpotman:
This resolves an issue in IDLC where the discriminator value for a Type Identifier was incorrectly determined for multi-dimensional arrays. The maximum size of each individual array bound should be evaluated, not the total number of elements across all dimensions.


eboasson (approved):
Thank you! ☺️

PR #2209, opened 2025-04-16
Dynsub better json

eboasson:
Dynsub would print an extra comma before the first field in the output when printing an invalid sample of a type where the first key field was preceded by a non-key field.


dpotman (commented):
LGTM!

PR #2210, opened 2025-04-16
Fix use-after-free in `ddsi_local_reader_ary_insert`

eboasson:
When deleting a writer, it is first removed from the hash table, and only later
freed. When deleting a reader, it needs to be removed from the writer's array of reader
pointers, but the writer can no longer be found when it has been removed from said hash
table. This causes a dangling reader pointer to be left in the array.

If another reader is getting matched to the writer in parallel to this, it potentially
dereferences readers in this array to check the type pointer, to keep readers with the
same type together. This can result in a use-after-free.

For example:

reader 1 gets created, starts matching, locates writer W, thread is descheduled
writer W is deleted, gets removed from hash table, freeing is deferred
reader 2 is deleted, can't find W, leaves dangling pointer in reader ary
reader 1 matching continues, compares reader 1's type pointer with (what used to be)
  reader 2's type pointer found in the array

There is already a "valid" flag for the reader array, which gets reset once: when it is
removed from the hash table (all under the writer lock). Adding a reader to the array
needs to be skipped when the array is no longer valid.

### DIFF
--- a/src/core/ddsc/tests/CMakeLists.txt
+++ b/src/core/ddsc/tests/CMakeLists.txt
@@ -73,6 +73,7 @@ set(ddsc_test_sources
     "qos_set_match.c"
     "querycondition.c"
     "guardcondition.c"
+    "matchstress.c"
     "readcollect.c"
     "readcondition.c"
     "reader.c"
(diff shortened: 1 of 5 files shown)


dpotman:
For the proxy writer, clearing the 'valid' flag and removing the proxy writer entity from the entity index is done under `gv->lock`, and not the proxy writer's lock. However, in `ddsi_proxy_writer_add_connection` the proxy writer's lock `pwr->e.lock` is used. Maybe not really an issue, because the valid flag is cleared before removing the proxy writer entity from the index, but may be better to also use the pwr's lock instead in `ddsi_delete_proxy_writer`?

PR #2212, opened 2025-04-24
Fix link issue on musl platforms.

### LINKED ISSUE #2211: Unable to link against ddsc on musl platforms
Currently it is not possible to link against ddsc on musl platforms due to the use of `pthread_attr_setaffinity_np` which is not available on musl.

An example of the error is shown below:

```
    = note: /usr/local/bin/../lib/gcc/x86_64-linux-musl/9.2.0/../../../../x86_64-linux-musl/bin/ld: /tmp/rustcbv9cCJ/libcyclors-02cc33282643057d.rlib(threads.c.o): in function `ddsrt_thread_create':
            threads.c:(.text.ddsrt_thread_create+0x126): undefined reference to `pthread_attr_setaffinity_np'
```

gmartin82:
Changed guard condition to ensure glibc is present due to link issues on musl.

Fixes #2211.

### DIFF
--- a/src/ddsrt/src/threads/posix/threads.c
+++ b/src/ddsrt/src/threads/posix/threads.c
@@ -417,7 +417,7 @@ ddsrt_thread_create (
 
   if (tattr.schedAffinityN > 0)
   {
-#if defined (__linux) && !defined(__ANDROID__)
+#if defined (__linux) && && defined(__GLIBC__)
     cpu_set_t cpuset;
     CPU_ZERO (&cpuset);
     for (uint32_t i = 0; i < tattr.schedAffinityN; i++)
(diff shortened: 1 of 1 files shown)


src/ddsrt/src/threads/posix/threads.c
```diff
@@ -417,7 +417,7 @@ ddsrt_thread_create (
 
   if (tattr.schedAffinityN > 0)
   {
-#if defined (__linux) && !defined(__ANDROID__)
+#if defined (__linux) && && defined(__GLIBC__)
```
noxpardalis, line 420: Small typo:
```suggestion
#if defined (__linux) && defined(__GLIBC__)
```

noxpardalis (changes requested):

src/ddsrt/src/threads/posix/threads.c
gmartin82: Fixed.

noxpardalis (approved):
Looks good to me. Bionic also doesn't define `__GLIBC__` so this should correctly exclude Android but perhaps someone else in the team should also take a look to confirm.

eboasson (approved):

PR #2213, opened 2025-04-29
Fix IDLC opcode generation for nested collection types

dpotman:
This fixes an issue in the opcode generation in IDLC for struct members that have the `@key` annotation and are of a type that is a nested collection. In the opcodes, the key flag was not added for the additional `ADR` opcodes for the nested types, e.g. for the following IDL:

```
struct n { @key sequence<short> n1[2]; };
struct t { n t1; };
```


eboasson (approved):
Thanks!

PR #2214, opened 2025-04-29
Type info in built-in topics for local entities

eboasson:
The data used for generating the built-in topic sample for a local entity did not include the type information, and as a consequence a process looking at its own entities would see "no type information available" in this data.


src/core/ddsi/src/ddsi_topic.c
```diff
@@ -104,7 +104,7 @@ dds_return_t ddsi_new_topic (struct ddsi_topic **tp_out, struct ddsi_guid *tpgui
   tp_qos->present |= DDSI_QP_TYPE_INFORMATION;
   tp_qos->type_information = ddsi_sertype_typeinfo (sertype);
```
dpotman, line 105: These lines above can be removed

dpotman (commented):
Looks good, just one small detail.

dpotman (approved):

PR #2216, opened 2025-05-02
Fix some issues with inheritance

eboasson:
Assignability check failing with multi-level inheritance

Key offsets not computed correctly by "type builder" for appendable types i.c.w. inheritance


dpotman (approved):
LGTM

PR #2226, opened 2025-05-19
Graceful shutdown debmon socket

eboasson:


trittsv:
Works great and solves all the instability issues with the debug-monitor.

PR #2230, opened 2025-05-28
Support for appendable types and optional members in XCDR1 encoding

dpotman:
This PR adds support for XCDR1 encoding for data types with struct members annotated with `@optional`. When optional members are present, the preferred encoding for writers is set to XCDR2, but this can be overridden either by using the `@data_representation` annotation in IDL or by explicitly setting the data-representation to XCDR1 in the QoS. 

To support the XCDR1 parameter header, the serializer requires the member ID for optional members. Therefore a table containing the member IDs for these members is appended to the CDR stream serializer opcodes. 

Note: this PR is currently a draft, as there are some open issues, e.g. missing test cases for normalizing and reading the short parameter header.


eboasson (commented):
I had looked at it pretty carefully some time ago and the changes relative to that are thankfully very minor 🙂 . It looks good to me except for one naming detail and one place where the alignment offset for XCDR1 optionals when the optional field is absent.

Once that's dealt with I'll have one last good look, but I don't expect anything further

dpotman:
Thanks for the re-review. I've addressed the findings in [1254f90](https://github.com/eclipse-cyclonedds/cyclonedds/pull/2230/commits/1254f90b5fe7330f71290690c26651bace72d98b). I also pushed some other fixes, and a rewrite of the MSVC warning suppression commit (so that required a force-push). The list of open issues is getting smaller: the Python fuzzer fails on a type builder check, and testing of non-present optionals in XCDR1 combined with specific alignment offset is missing. I'll address that next.

src/core/ddsc/tests/cdrstream.c
```diff
@@ -1900,49 +1953,74 @@ CU_Test (ddsc_cdrstream, check_wstring_valid)
...
-    dds_free (data);
+static bool eq_CdrStreamXcdr1Optional_t1 (const void *va, const void *vb)
+{
+  const CdrStreamXcdr1Optional_t1 *a = va;
+  const CdrStreamXcdr1Optional_t1 *b = vb;
+  if ((a->f1 == NULL || b->f1 == NULL) && a->f1 != b->f1)
+    return false;
+  if (a->f1 == NULL && b->f1 == NULL)
+    return true;
+  return *(a->f1) == *(b->f1) && a->k == b->k;
```
eboasson, line 1977: I'm not convinced by this: if `f1` is a null pointer in both `a` and `b` but `k` differs, it still returns `true`. I also admit to misreading line 1973 at first because my brain was somehow expecting a dereference there — obviously then I'd write that check differently myself 🙂  

Perhaps something like:
```C
  if (a->k != b->k)
    return false;
  if ((a->f1 == NULL) != (b->f1 == NULL))
    return false;
  if (a->f1 && *a->f1 != *b->f1)
    return false;
  return true;
```
but I don't really mind anything except for the check on `k`.

src/tools/idlc/src/libidlc/libidlc__descriptor.c
```diff
@@ -2587,6 +2587,11 @@ static idl_retcode_t get_ctype_memberids(const idl_pstate_t *pstate, struct desc
     }
   }
 
+  /* TO-DO: XCDR1 for mutable types is currently not supported, so
+     don't include the MID table for mutable types */
+  if (idl_is_extensible(ctype->node, IDL_MUTABLE))
```
eboasson, line 2592: For mutable we never need it, right? That would make the code is correct also when we do support mutable XCDR1 and the comment would be better phrased as "Mutable encoding doesn't need a MID table".

dpotman: You're right, fixed in  45c2257

src/core/ddsc/tests/cdrstream.c
dpotman: Thanks for catching this, fixed in 45c2257 (also for `t2` and `t3`)

src/core/cdr/src/dds_cdrstream.c
```diff
@@ -1018,22 +1096,27 @@ static void dds_stream_get_ops_info1 (const uint32_t *ops, uint32_t nestc, struc
             info->data_types |= DDS_DATA_TYPE_CONTAINS_BITMASK;
             break;
           case DDS_OP_VAL_SEQ:
-            ops = dds_stream_get_ops_info_seq (ops, insn, nestc, info);
             info->data_types |= DDS_DATA_TYPE_CONTAINS_SEQUENCE;
+            info->in_xcdr1_delimited_scope = false;
```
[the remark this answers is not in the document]
dpotman: Yes, good point, I've fixed this and added some test-cases

eboasson (approved):
Thanks! 🥳

PR #2232, opened 2025-06-05
Fix permissions example XML in the DDS Security documentation

dpotman:
Fixes the `subject_name` in the permissions example XML in the DDS Security documentation, so that it matches the subject name used for the identity. Also updates the configuration XML used in the example.

### DIFF
--- a/docs/manual/_static/example_permissions.xml
+++ b/docs/manual/_static/example_permissions.xml
@@ -3,7 +3,7 @@
      xsi:noNamespaceSchemaLocation="https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd">
   <permissions>
     <grant name="default_permissions">
-      <subject_name>emailAddress=alice@cycloneddssecurity.adlinktech.com,CN=Alice Example,O=Example Organization,OU=Organizational Unit Name,L=Locality Name,ST=OV,C=NL</subject_name>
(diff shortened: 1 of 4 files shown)


docs/manual/_static/example_permissions.xml
```diff
@@ -3,7 +3,7 @@
      xsi:noNamespaceSchemaLocation="https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd">
   <permissions>
     <grant name="default_permissions">
-      <subject_name>emailAddress=alice@cycloneddssecurity.adlinktech.com,CN=Alice Example,O=Example Organization,OU=Organizational Unit Name,L=Locality Name,ST=OV,C=NL</subject_name>
+      <subject_name>emailAddress=alice@cycloneddssecurity.zettascale.com,CN=Alice Example,O=Example Organization,OU=Organizational Unit Name,L=Locality Name,ST=OV,C=NL</subject_name>
```
eboasson, line 6: I think it would be better to use something like [`example.com`](https://en.wikipedia.org/wiki/Example.com)

```suggestion
      <subject_name>emailAddress=alice@example.com,CN=Alice Example,O=Example Organization,OU=Organizational Unit Name,L=Locality Name,ST=OV,C=NL</subject_name>
```

eboasson (approved):

PR #2250, opened 2025-07-23
Documentation improvements

poetinger:
> Fixes a few minor typos during documentation review.

This now includes more review comment improvements from various sources, all bundled into one PR.


poetinger:
> "behaviour" and "serialisation" aren't so much typo's as English words, whereas "behavior" is mere dialect 🤣:
> 
> * "serialize | ˈsɪərɪəlʌɪz | (British English also serialise)"
> * "behaviour | bɪˈheɪvjə | (behavior US English)"
> 
> I do agree there is an undesirable inconsistency in their use (the worst for "serialize"/"serialise"). If Cyclone today straddles the mid-Atlantic ridge and we want to prevent it from being torn apart by the continental drift, we need to choose a side. I would instinctively choose the old world, that goes without saying ...
> 
> (Also don't forget the pre-commit check is failing on the CI)

I will open an issue for this, more thorough, standardization for a future PR. We can pick which side of the Atlantic for this :)

I will update the hashes as well.

docs/dev/data path - receive.svg
poetinger, line 1: Is this actually correct to delete?

PR #2255, opened 2025-07-29
Gracefully handle time jumps

### LINKED ISSUE #1995: Event thread is freezed when time is changed to past
When the time is changed to past (we changed time to 2 years ago), the event thread is freezed on `xevent_thread` - `ddsrt_cond_waitfor`. Because of this, the discovery is disconnected due to the participant lease duration after changing time to past.

Below is my temporary solution to fix it.

The `ddsrt_cond_waitfor` function is not implemented as monotonic on src/ddsrt/src/sync/posix/sync.c, it just calculates `dds_time() + reltime` and calls `ddsrt_cond_waituntil`.

The monotonic `ddsrt_cond_waitfor` can be implemented like below. (However, this is only implemented using pthread, and if you want to be compatible with other code, you should use other code. Like adding `ddsrt_cond_init` with is_monotonic parameter or `ddsrt_cond_init_w_monotonic` and so on.)

```
diff --git a/src/core/ddsi/src/q_xevent.c b/src/core/ddsi/src/q_xevent.c
index 9f8a79a2..fc3d6646 100644
--- a/src/core/ddsi/src/q_xevent.c
+++ b/src/core/ddsi/src/q_xevent.c
@@ -11,6 +11,7 @@
  */
 #include <math.h>
 #include <stdlib.h>
+#include <pthread.h>
 
 #include "dds/ddsrt/atomics.h"
 #include "dds/ddsrt/heap.h"
@@ -522,7 +523,13 @@ struct xeventq * xeventq_new (struct ddsi_domaingv *gv, size_t max_queued_rexmit
   evq->queued_rexmit_msgs = 0;
   evq->gv = gv;
   ddsrt_mutex_init (&evq->lock);
-  ddsrt_cond_init (&evq->cond);
+  // ddsrt_cond_init (&evq->cond);
+
+  pthread_condattr_t attr;
+  pthread_condattr_init(&attr);
+  pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
+  pthread_cond_init(&evq->cond.cond, &attr);
+  evq->cond.is_monotonic = true;
 
   evq->cum_rexmit_bytes = 0;
   return evq;
diff --git a/src/ddsrt/include/dds/ddsrt/sync/posix.h b/src/ddsrt/include/dds/ddsrt/sync/posix.h
index c13d8e7d..033b4b78 100644
--- a/src/ddsrt/include/dds/ddsrt/sync/posix.h
+++ b/src/ddsrt/include/dds/ddsrt/sync/posix.h
@@ -24,6 +24,7 @@ extern "C" {
 
 typedef struct {
     pthread_cond_t cond;
+    bool is_monotonic;
 } ddsrt_cond_t;
 
 typedef struct {
diff --git a/src/ddsrt/src/sync/posix/sync.c b/src/ddsrt/src/sync/posix/sync.c
index 89dce5d9..3ffbebe8 100644
--- a/src/ddsrt/src/sync/posix/sync.c
+++ b/src/ddsrt/src/sync/posix/sync.c
@@ -131,8 +131,16 @@ ddsrt_cond_waitfor(
   assert(cond != NULL);
   assert(mutex != NULL);
 
+  int64_t base_time;
+  if (cond->is_monotonic) {
+      ddsrt_mtime_t time =  ddsrt_time_monotonic();
+      base_time = time.v;
+  } else {
+      base_time = dds_time();
+  }
+
   return ddsrt_cond_waituntil(
-    cond, mutex, ddsrt_time_add_duration(dds_time(), reltime));
+    cond, mutex, ddsrt_time_add_duration(base_time, reltime));
 }
 
 void
```

eboasson:
The main commit in this PR allows blocking with timeouts based on a monotonic clock rather than the system clock. This should address the various problems with time jumps people have been facing. I haven't actually tested it with time jumps yet because of some practicalities, but perhaps some kind souls are willing to give it a go as it is.

Edit: I accidentally pushed some more onto this branch, but it fits well enough with the subject: some platforms have low-resolution clocks that interact with the scheduler but can provide a high-resolution clock that is unrelated to scheduling. Using the latter for benchmarking seems sensible.

The other commits are just housekeeping and fixing some trivialities.

Fixes #1995 https://github.com/ros2/rmw_cyclonedds/issues/276

### DIFF
--- a/src/core/ddsc/src/dds__types.h
+++ b/src/core/ddsc/src/dds__types.h
@@ -474,6 +474,8 @@ typedef struct dds_waitset {
      acquired while holding an ancestor's lock, but a waitset must be capable of triggering on
      events on its parent */
   ddsrt_mutex_t wait_lock;
+
+  /* etime: dds_waitset_wait timeout */
   ddsrt_cond_etime_t wait_cond;
   size_t nentities;         /* [wait_lock] */
   size_t ntriggered;        /* [wait_lock] */
--- a/src/core/ddsi/include/dds/ddsi/ddsi_domaingv.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_domaingv.h
@@ -309,7 +309,7 @@ struct ddsi_domaingv {
   ddsrt_avl_tree_t typelib;
   ddsrt_avl_tree_t typedeps;
   ddsrt_avl_tree_t typedeps_reverse;
-  ddsrt_cond_etime_t typelib_resolved_cond;
+  ddsrt_cond_etime_t typelib_resolved_cond; // etime: create_topic_descriptor timeout
 #endif
 #ifdef DDS_HAS_TOPIC_DISCOVERY
   ddsrt_mutex_t topic_defs_lock;
@@ -317,7 +317,7 @@ struct ddsi_domaingv {
 #endif
 
   ddsrt_mutex_t new_topic_lock;
-  ddsrt_cond_etime_t new_topic_cond;
+  ddsrt_cond_etime_t new_topic_cond; // etime: find_topic timeout
   uint32_t new_topic_version;
 
   /* security globals */
--- a/src/core/ddsi/include/dds/ddsi/ddsi_endpoint.h
+++ b/src/core/ddsi/include/dds/ddsi/ddsi_endpoint.h
@@ -72,7 +72,7 @@ struct ddsi_writer
   struct ddsi_endpoint_common c;
   ddsi_status_cb_t status_cb;
   void * status_cb_entity;
-  ddsrt_cond_mtime_t throttle_cond; /* used to trigger a transmit thread blocked in throttle_writer() or wait_for_acks() */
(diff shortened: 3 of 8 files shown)


dpotman (commented):
Looks good! A few minor things:
- the test code for the C++ API is using `ddsrt_cond_waitfor`, so that needs an update (or a wrapper for this function in DDSRT?)
- `sync.c` for FreeRTOS still has `ddsrt_cond_waitfor`
 
And at some places (e.g. waitset, xeventq, writer throttling) it may be helpful to add a comment with the rationale for using `mtime_t` or `etime_t` (although technically it makes no difference in many cases)

noxpardalis (approved):
The new changes look good to me and address @dpotman’s points. All tests also run fine locally.

I also took a gander at the associated PR for the C++ repo: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/546 and it looks fine. I can do a full review once this is merged and the noise from the CI failures on the C++ repo disappear on a re-run.

PR #2260, opened 2025-07-31
Add support for mutable types in XCDR1

dpotman:
This adds support for XCDR1 (de)serialization of mutable types in the CDR stream serializer and also includes a few related changes:

- The serialization order for the CDR used to calculate the keyhash for an XCDR1 endpoint now always uses the member IDs (if present)
- The extract_key_from_data functions for the optimized path are simplified, because a type with keys in appendable or mutable nested types will be handled in the non-optimized path
- The IDLC xtests are updated so that these also do a run with XCDR1
- Update cdrstream `get_ops_info` so that appendable or mutable aggregated types used in recursion are correctly detected

Some of the CI builds fail on the `idlc_test_bitmask` test with a "load of misaligned address for type 'uint64_t'", which may be a result of the XCDR1 alignment rules for parameter payload? (but I don't really understand why this error is not triggered for XCDR2, which also has max 4-byte alignment..)

**Note that this PR includes the commits of #2259**


eboasson (approved):
Thank you @dpotman ! It looks good to me, there were a few details that I decided to fix in a few small commits instead of putting comments on the PR etc.

PR #2266, opened 2025-08-13
feature/reader-received-bytes-statistic

trittsv:
This PR implement the statistic `received_bytes` for readers.

@eboasson could you have a look? :)

### DIFF
--- a/src/core/ddsi/tests/receive_packet.c
+++ b/src/core/ddsi/tests/receive_packet.c
@@ -97,6 +97,11 @@ static void stop_and_teardown (void)
   teardown ();
 }
 
+struct serdata {
+  struct ddsi_serdata c;
+  uint32_t size;
+};
+
 static void serdata_free (struct ddsi_serdata *sd) {
   ddsrt_free (sd);
 }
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Thanks @trittsv, that was bound to come after the previous one. The change looks good to me and it is counting incoming network traffic just like the previous one was counting outgoing network traffic.

There is a problem though: the CI turned red. This change obviously seemingly cannot possibly cause that, so I did take a look. It turns a test I added some time ago uses a very sparingly populated `serdata` implementation where the function for retrieving the size isn't set.

The patch below fixes that. I considered fixing it by returning a constant, but decided against it. If you could add this to the PR then I can merge it.
 
```
diff --git i/src/core/ddsi/tests/receive_packet.c w/src/core/ddsi/tests/receive_packet.c
index 7e542df94..118cc22f0 100644
--- i/src/core/ddsi/tests/receive_packet.c
+++ w/src/core/ddsi/tests/receive_packet.c
@@ -97,6 +97,11 @@ static void stop_and_teardown (void)
   teardown ();
 }
 
+struct serdata {
+  struct ddsi_serdata c;
+  uint32_t size;
+};
+
 static void serdata_free (struct ddsi_serdata *sd) {
   ddsrt_free (sd);
 }
@@ -105,10 +110,11 @@ static struct ddsi_serdata *serdata_from_ser (const struct ddsi_sertype *st, enu
 {
   (void) fragchain;
   printf ("Yay! %zu bytes\n", size);
-  struct ddsi_serdata *sd = ddsrt_malloc (sizeof (*sd));
-  ddsi_serdata_init (sd, st, kind);
-  sd->hash = st->serdata_basehash;
-  return sd;
+  struct serdata *sd = ddsrt_malloc (sizeof (*sd));
+  ddsi_serdata_init (&sd->c, st, kind);
+  sd->size = (uint32_t) size;
+  sd->c.hash = st->serdata_basehash;
+  return &sd->c;
 }
 
 static struct ddsi_serdata *serdata_to_untyped (const struct ddsi_serdata *d)
@@ -123,11 +129,18 @@ static bool serdata_eqkey (const struct ddsi_serdata *a, const struct ddsi_serda
   return true;
 }
 
+static uint32_t serdata_get_size (const struct ddsi_serdata *sd0)
+{
+  struct serdata *sd = (struct serdata *) sd0;
+  return sd->size;
+};
+
 static const struct ddsi_serdata_ops serdata_ops = {
   .from_ser = serdata_from_ser,
   .free = serdata_free,
   .to_untyped = serdata_to_untyped,
-  .eqkey = serdata_eqkey
+  .eqkey = serdata_eqkey,
+  .get_size = serdata_get_size
 };
 
 static void sertype_free (struct ddsi_sertype *st)
```

eboasson (approved):

PR #2270, opened 2025-08-26
Improve asymmetric deny_rule handling for topics

### LINKED ISSUE #2267: Publish-only or subscribe-only security permissons fail
Hi,

We are trying to implement restrictions on what messages different participants can subscribe to and publish, but we are only able to allow both read and write access, but not read-only or write-only. An example:

We can have an authentication service and multiple authentication clients. We want to enable the clients to send the authentication request, but we do not want other clients to be able to read/subscribe to the request message. We also want to restrict the sending of authentication-responses to only the server. 

The reason for the failure is that the topic creation fails if either subscription or publishing is denied. We don't get to trying to create the writer or reader.

Attached is an example project with authentication server and clients. The auth_client application should, according to our understanding of the standard, be able to to start, but auth_client2 should fail. With current master (a couple of months old) also auth_client fails to start.

[dds_cxx_deny_rules.tar.gz](https://github.com/user-attachments/files/21935529/dds_cxx_deny_rules.tar.gz)

We have debugged it a bit and have created a patch to allow the topic creation if either subscription or publishing is allowed:
[CycloneDDS-fix-asymetric-deny_rule.tar.gz](https://github.com/user-attachments/files/21935648/CycloneDDS-fix-asymetric-deny_rule.tar.gz)

We tried to follow the specification, in that the first rule that matches, is the one that is used.

Best regards,
  Kåre

kasars:
Allow the creation of topics if either subscribe or publish is allowed

Fixes #2267

### DIFF
--- a/src/security/builtin_plugins/access_control/src/access_control.c
+++ b/src/security/builtin_plugins/access_control/src/access_control.c
@@ -2014,13 +2014,39 @@ static bool is_topic_allowed_by_permissions (const struct permissions_parser *pe
   const struct allow_deny_rule *rule;
   if (!rule_iter_init (&it, permissions, domain_id, identity_subject_name, ex))
     return false;
+
+  bool subscribe_rule_found = false;
+  bool publish_rule_found = false;
+
+  bool subscribe_allowed = true;
+  bool publish_allowed = true;
+
   while ((rule = rule_iter_next (&it)) != NULL)
   {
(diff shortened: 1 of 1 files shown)


kasars:
Waiting for the eclipsefdn/eca :)

eboasson:
The ECA remained happy, so I've followed your suggestion with #2272

PR #2276, opened 2025-09-05
Work around RTI bugs in XCDR1 mutable encoding

eboasson:
This works around two issues with RTI's handling of mutable structs in XCDR1:

* RTI expects/requires/produces 4-byte padding in the length fields of the parameters (forbidden by XTypes 1.3, required by XTypes 1.1 — go figure);
* RTI can't handle "must understand" on a key field even though the member ids are as expected (we believe "must understand" is required, but what can you do ...)


dpotman (commented):
Looks good, only a question regarding the param length for optionals.

PR #2282, opened 2025-09-09
Fix dynamic type API for adding primitive struct members

dpotman:
Also add some additonal tests and a missing check in the code for setting the default flag when completing an enum type.


eboasson (approved):
Thank you @dpotman !

PR #2287, opened 2025-09-16
sysdef. related improvements 

Splinter1984:
- Allow empty `user/topic/group_data` to be set from `sysdef`.
- `entity_factory` enabled, but no way to disable created entities.

Additionally extend `qos_provider` tests with file-read test.


dpotman (approved):
LGTM

PR #2296, opened 2025-09-30
Sertype default compare typeinfo

eboasson:
This is #1698 updated to the current version and with a `memcmp` added.


dpotman (approved):
LGTM!

PR #2298, opened 2025-10-01
feat: improved library build system when building as third party

### LINKED ISSUE #2297: feat: Support for building as a vendored library
I have the need to include CycloneDDS as part of my main project, without building it in advance and using it as an imported CMake target, e.g. `add_subdirectory(CycloneDDS)`.
I found out this very challenging because apparently the majority of CMake was structured to not support this feature, e.g. usage of `CMAKE_PROJECT_NAME` instead of `PROJECT_NAME`.

Davide-1998:
Update: rebased to CyloneDDS/master branch and added minimal features for usage as third party

Fix #2297 
This PR adds the following:  
  - Added aliases for targets avoiding multiple definitions with foreign code
  - `cdr` target becomes part of DDSC without having its own project
  - Rename `tls_callback_func` in `idl` and `ddsrt` targets to avoid multiple definition errors
  - Enforced usage of folders locals to the CycloneDDS project rather than the global CMAKE_PROJECT
  - Added a new option `IDL_LIBRARY_TYPE` when generating the compiled IDL target to allow user to chose the preferred library type, by defaults falls back to *INTERFACE* for compatibility
  - Added documentation on building CycloneDDS as a third party library on a generic project


eboasson (commented):
Thank you very much @Davide-1998 ! I am not much of a CMake user and I have never gotten around to trying to solve some vague reports about difficulties building it as a sub-project. So it is really nice to see a PR for addressing those issues.

It looks good to me, even if I there are some things I am not sure of yet. I'd appreciate it if you could help me with them.

Davide-1998:
@eboasson friendly ping.
These last 13 commits are the minimal changes to use CycloneDDS as a third party, without altering too much the logic of the CMake itself.
I have rebased the fork on the previous master commit, thus all changes I had before do not exist anymore.

Davide-1998:
Hi, given the amount of changes that this PR carries, and how the comment history is now broken due to some force pushes on my side, I have decided to close this PR in favor of some smaller ones, hoping to ease the review process.  
Following is the list of the new PRs covering the same functionalities here proposed:  

- #2438 : fixes `tls_callback_func` multiple definitions
- #2439: fixes usage of CycloneDDS as vendored library

PR #2309, opened 2025-10-27
feat: add compile commands for cmake

nomanfound:
This PR add CMAKE_EXPORT_COMPILE_COMMANDS for cmake so that it will generate `compile_commands.json`.
It will help vscode extension or language server to parse code and headers correctly.

Hope it helps!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@
 cmake_minimum_required(VERSION 3.16)
 # C++ is only for Iceoryx plugin
 project(CycloneDDS VERSION 0.11.0 LANGUAGES C CXX)
+set(CMAKE_EXPORT_COMPILE_COMMANDS true)
 
 # Set a default build type if none was specified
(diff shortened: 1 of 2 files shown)


poetinger:
In my personal opinion, I don't think adding this to the default `cmake` config generation is the right way to do this. While it doesn't hurt to create a file if somebody isn't using the `compile_commands.json`, it is potentially going to get in the way of somebody if they do. Consider the following:

- If the user is manually adding some paths in their compile commands, theses will be overwritten anytime somebody regenerates the config.
- This assumes they use a folder called `build`, but some (including myself), use a multiple config, single install method (e.g. `build-debug`, `build-release`, `build-security`) and a single `compile-commands.json`.

I think it might be good to let each person determine whether they wish to generate the file, and its simple enough with the flag `-DCMAKE_EXPORT_COMPILE_COMMANDS=On` once and then move it to the base.

PR #2321, opened 2025-11-15
implement property-qos propagate getter and setters

trittsv:
This PR implements getter and setter for property propagate on discovery.

Previously it was not possible to propagate the Property-Qos on the network for the c++ and python wrappers.

@eboasson could you have a look?

### DIFF
--- a/src/security/core/tests/config.c
+++ b/src/security/core/tests/config.c
@@ -22,6 +22,11 @@
 #include "common/config_env.h"
 #include "common/test_identity.h"
 
+#define BUILTIN_PROPS \
+  "1:\"__ProcessName\":\"*\"," \
+  "1:\"__Pid\":\"*\"," \
+  "1:\"__Hostname\":\"*\"" \
+
 #define PROPLIST(init_auth, fin_auth, init_crypto, fin_crypto, init_ac, fin_ac, perm_ca, gov, perm, pre_str, post_str, binprops)         \
   "property_list={" pre_str                                             \
   "0:\"" DDS_SEC_PROP_AUTH_LIBRARY_PATH "\":\""WRAPPERLIB_PATH("dds_security_authentication_wrapper")"\","                         \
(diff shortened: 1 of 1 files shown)


src/core/ddsc/include/dds/ddsc/dds_public_qos.h
```diff
@@ -482,6 +482,23 @@ dds_qset_prop (
...
+ * @brief Set value if the property shall be included in discovery or not.
+ *
+ * @param[in,out] qos - Pointer to a dds_qos_t structure that contains the property
+ * @param[in] name - Pointer to name of the property
+ * @param[in] propagate - Bool value to propagate or not
+ * 
+ * @returns - false if any of the arguments is invalid, the qos is not present in the qos object or there was no property found with the provided name
+ */
+DDS_EXPORT bool
+dds_qset_prop_propagate (
```
eboasson: I would instead add a function `dds_qset_prop_propagate(dds_qos_t *qos, const char *name, const char *value, bool propagate)` and define `dds_qset_prop` using that function: `dds_qset_prop(qos, name, value) { dds_qset_prop_propagate(qos, name, value, false); }`. I think that's a lot easier than having to set the properties by name and then set the propagation.

src/core/ddsc/include/dds/ddsc/dds_public_qos.h
```diff
@@ -1010,6 +1027,23 @@ dds_qget_prop (
   const char * name,
   char ** value);
 
+/**
```
eboasson: Here I would follow the same pattern as what I suggested for the setter: a `dds_qget_prop_propagate` that *also* returns the setting of the "propagate" flag (if a non-null pointer is passed in), then implement `dds_qget_prop` as:
```
bool dds_qget_prop (const dds_qos_t *qos, const char *name, char **value)
{
  return dds_qget_prop_propagate (qos, name, value, NULL);
}
```

(And again the same for the binary property one.)

trittsv:
@eboasson thanks a lot for the review, i updated the functions as you noted and added also the binary properties propagate functions. Furthermore it includes your patch so that the default participant qos are always set.

May you can have a look again? :)

PR #2343, opened 2026-02-03
Fix of enum warnings for MSVC 2022 ROS2 windows rolling build

knmcguire:
Hi! 

This PR is fixing the enum comparison warnings for MSVC 2022 builds for ROS2 on windows (warning[ C5286 C5287](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c5200-through-c5399)). Cyclone-dds 0.10.x was resulting in over 1500 warnings (See this on the ros2 ci: https://ci.ros2.org/job/test_ci_windows/687/)

With this patch in place, the warnings were resolved (see this test build on [ros2 ci](https://ci.ros2.org/job/test_ci_windows/693/)).

I'll leave this in **draft** so that you and the ROS PMC can discuss if this change is desired. An alternative would be to fix this all by all local cmake suppressions of these warnings, but that was a wack-a-mole that required that suppression in a lot of cmake. This was a patch that required the least amount of code change but perhaps not something to be desired. 

Related to https://github.com/ros2/ci/issues/832#issuecomment-3786789226

Just let me know what you think.

### DIFF
--- a/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
+++ b/src/core/ddsc/include/dds/ddsc/dds_opcodes.h
@@ -384,25 +384,24 @@ enum dds_stream_typecode {
 /**
  * @ingroup serialization
  * @brief primary type code for DDS_OP_ADR, DDS_OP_JEQ
- * Convinience pre-bitshifted values.
+ * Convenience pre-bitshifted values, defined as macros to allow
+ * bitwise OR with dds_stream_opcode without enum type mixing warnings.
  */
-enum dds_stream_typecode_primary {
-  DDS_OP_TYPE_1BY = DDS_OP_VAL_1BY << 16, /**< one byte simple type (char, octet) */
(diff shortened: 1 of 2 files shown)


knmcguire:
By the way, I did a check with the suppressions you have suggested at that same location. that seems to have worked! Thanks for this!

See this run with suppressions [Build Status](https://ci.ros2.org/job/ci_windows/27107/) (only 1 msbuild warning)
Compared to this run without suppressions: [Build Status](https://ci.ros2.org/job/ci_windows/27021/) (1800+ msbuild warnings)

I had a faint memory that I did see these suppressions in the cmakelist before in the cyclone dds master (but I can not find it now) and therefore tried something similar as well, but this was a local docker test so I can not find a run of this. At least, I was too quickly to brush it away when the build failed.

Let me try to clear out all warnings and make a new minimal PR with those suppressions.

PR #2349, opened 2026-02-18
RTI Compatibily in XCDR2

ErwanGa:
RTI can't handle "must understand" on key fields when using XCDR2 The same issue has already been patched for XCDR1 format

Setting must_understand if FLAG_MU, rather than if (FLAG_MU | FLAG_KEY)


ErwanGa:
I've signed it just 2 minutes ago :-)
Erwan

PR #2367, opened 2026-03-12
Github Actions Support

### LINKED ISSUE #2366: Github Actions support
While working on the build system of CycloneDDS I found myself challenged in ensuring a fully compatible solution.
The main limitation is the Azure CI/CD pipeline which requires an Azure Developer account to be run and, since it comes under subscription, not many open-source developers have it.

For this reason it would be optimal to also add support for Github Actions, thus allowing any developer on this platform to provide their contribution without requiring a PR to run the CI.

I link to the following issue a possible implementation of CI/CD using Github Actions that follows, as close as possible, the one written for Azure.

Davide-1998:
Fixes #2366

Note:
This PR adds an extra CI validation layer and ensures the workflow is usable by developers without an Azure Developer account.

Changes:
- Corrects errors on Azure pipelines
- Introduce CI pipeline using Github Actions for Windows, Ubuntu and MacOS
- Updates Build system to support the new CI
- Updates the `symbol_export` test code to directly search the symbols of interest inside the built ddsc shared library
- Enforced parallel building to reduce the building time during CI/CD
- Allowed maximum thread usage for CTest
- Fix building with Iceoryx2 support due to renaming of strategies introduced by latest iceoryx2 commit eclipse-iceoryx/iceoryx2@8d20af8dc164c11e129092530bdd43a068c2d0d2


Davide-1998:
@eboasson Friendly ping

.azure/templates/build-test.yml
```diff
@@ -108,7 +108,7 @@ steps:
             -DCMAKE_BUILD_SHARED_LIBS=${SHAREDLIBS:-on} \
             -DCMAKE_INSTALL_PREFIX=install \
             ${GENERATOR:+-G} "${GENERATOR}" -A "${PLATFORM}" -T "${TOOLSET}" ../iceoryx_meta
-      cmake --build . --config ${BUILD_TYPE} --target install -- ${BUILD_TOOL_OPTIONS}
+      cmake --build . --parallel --config ${BUILD_TYPE} --target install -- ${BUILD_TOOL_OPTIONS}
```
eboasson, line 111: The `--parallel` isn't going to improve build times a lot, I think, because it already sets `-maxcpucount:4` and `-j 4`. I am pretty sure I tried it myself at one point but decided against it. I just wonder why ... The only thing I can think of is that it increased the memory usage too much. That is plausible for `gcc -fanalyze` with the early versions of the static analyzer. The CI run for this PR succeeded, so I guess it's fine.

.azure/templates/build-test.yml
```diff
@@ -160,7 +160,7 @@ steps:
             -DENABLE_SECURITY=${SECURITY:-on} \
             -DENABLE_LIFESPAN=${LIFESPAN:-on} \
             -DENABLE_DEADLINE_MISSED=${DEADLINE_MISSED:-on} \
-            -DSKIP_DEADLINE_UPDATE={DEADLINE_UPDATE_SKIP:-off} \
+            -DSKIP_DEADLINE_UPDATE=${DEADLINE_UPDATE_SKIP:-off} \
```
eboasson, line 163: Good catch! Thank you! I tried, it looks like it must have always skipped them in the CI.

.azure/templates/build-test.yml
```diff
@@ -173,21 +173,23 @@ steps:
...
   - bash: |
       set -e -x
       cd build
       UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 \
-        ctest -j 4 --output-on-failure -T test -E '^CUnit_ddsrt_random_default_random$' -C ${BUILD_TYPE}
+        ctest -j --output-on-failure -T test -E '^CUnit_ddsrt_random_default_random$' -C ${BUILD_TYPE}
     condition: ne(variables['testing'], 'off')
     name: test
     displayName: Test
+    retryCountOnTaskFailure: 3 # Run CTest 3 times to counteract Timeout and Out of Memory events
```
eboasson: I have many times considered retrying tests a few times. The problem I have with that is that I have caught a few very nearly unreproducible bugs thanks to a single failure/crash. But I do hate that it isn't green almost always. Perhaps it is worth trying this.

I would prefer to do it more fine-grained. CTest also has options for this:
```
  --repeat until-fail:<n>, --repeat-until-fail <n>
                               = Require each test to run <n> times without
                                 failing in order to pass
  --repeat until-pass:<n>      = Allow each test to run up to <n> times in
                                 order to pass
  --repeat after-timeout:<n>   = Allow each test to run up to <n> times if it
                                 times out
```

.azure/templates/build-test.yml
```diff
@@ -173,21 +173,23 @@ steps:
...
   - bash: |
       set -e -x
       cd build
       ${SHELL} ../src/tools/ddsperf/sanity.bash
+    condition: eq(variables['sharedlibs'], 'on')
```
eboasson: Why this? I thought it worked fine also with static libs: https://dev.azure.com/eclipse-cyclonedds/cyclonedds/_build/results?buildId=8283&view=logs&jobId=5b78434c-c39a-531f-e7af-0422f4799d71&j=5b78434c-c39a-531f-e7af-0422f4799d71&t=e1299c93-df53-5d1b-e2e9-466bc0c610a4

.github/workflows/misra-regressions-check.yml
```diff
@@ -40,22 +40,22 @@ jobs:
         with:
```
eboasson: I have stopped looking at MISRA: after having dealt with MISRA for real, I am convinced its primary purpose is making life miserable. The issues that it tries to protect against are real, the approach it advocates in most cases is pretty sensible, but the strict application of it makes code unreadable. Combined with the low quality of the checkers, it becomes worse than useless.

So *I* would remove the misra check workflow altogether.

src/core/ddsc/CMakeLists.txt
```diff
@@ -149,6 +149,6 @@ install(
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
   COMPONENT dev)
 
-if((BUILD_TESTING) AND ((NOT DEFINED MSVC_VERSION) OR (MSVC_VERSION GREATER "1800")))
+if((BUILD_TESTING AND BUILD_IDLC) AND ((NOT DEFINED MSVC_VERSION) OR (MSVC_VERSION GREATER "1800")))
```
eboasson: For cross-builds you do want testing even though `BUILD_IDLC` is false. I think it makes more sense to define a `HAVE_IDLC` variable with a manual override, and which is set to true if `BUILD_IDLC` or if cross-compiling and IDLC is found in the path.

src/core/xtests/symbol_export/CMakeLists.txt
```diff
@@ -10,20 +10,35 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
```
eboasson, line 10: I do like verifying the presence of symbols by using the run-time linker. However, the "old" test verified not only that all expected symbols were available, but also that no other symbols were present. That was deliberate: I think if one makes an effort not to export everything, one also wants to make it a bit harder to accidentally add symbols.

I don't quite know how to verify the latter without running `nm` or equivalent on the generated library.

But other than that, I think this is a significant improvement to the test implementation, even if there are a few things I would do differently:

* I would use `ddsrt_dlopen` &c. to load the library and probe the symbols. I think that alone would eliminate the platform dependent code.
* Instead of calling `resolve_symbol` for each expected symbol, I think it would make sense to have a text file with one symbol per line, let the test executable read that file.

Perhaps you can have another look at the "symbol_export" bits? (Or even take it out of this PR and do it at some later date, because so far, the other comments are all trivial.)

src/psmx_iox/src/psmx_iox2_impl.c
```diff
@@ -893,7 +893,7 @@ static bool psmx_iox2_init_writer (psmx_iox2_endpoint_t *ep, psmx_iox2_topic_t *
   
   dds_reliability_kind_t rel;
   (void) dds_qget_reliability (qos, &rel, NULL);
-  iox2_port_factory_publisher_builder_unable_to_deliver_strategy (&publisher_builder_handle, rel == DDS_RELIABILITY_BEST_EFFORT ? iox2_unable_to_deliver_strategy_e_DISCARD_SAMPLE : iox2_unable_to_deliver_strategy_e_BLOCK);
```
eboasson: I ran into this one as well and fixed it already ...

src/security/core/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ target_include_directories(security_core
         "$<BUILD_INTERFACE:$<TARGET_PROPERTY:ddsc,INTERFACE_INCLUDE_DIRECTORIES>>"
 )
 
-if(BUILD_TESTING)
+if(BUILD_TESTING AND BUILD_IDLC)
```
eboasson: See earlier comment on `BUILD_IDLC`

src/tools/CMakeLists.txt
```diff
@@ -16,5 +16,5 @@ add_subdirectory(_confgen)
 if(BUILD_IDLC)
   add_subdirectory(idlpp)
   add_subdirectory(idlc)
+  add_subdirectory(ddsperf)
```
eboasson, line 22: See earlier comment on `BUILD_IDLC`

CMakeLists.txt
```diff
@@ -417,11 +417,11 @@ set(generated_defconfig_src  "${CMAKE_CURRENT_SOURCE_DIR}/src/core/ddsi/defconfi
 
 add_subdirectory(compat)
 add_subdirectory(src)
-if(BUILD_EXAMPLES)
+if(BUILD_EXAMPLES AND BUILD_IDLC)
```
eboasson: See earlier comment on `BUILD_IDLC`

eboasson (commented):
Hi @Davide-1998 ! Thank you so much for this effort (and for pinging me).

I didn't look at the actual workflow definitions in detail: given the care with which you went over the Azure ones, I'm pretty sure they are fine. Obviously I won't merge it without having a good look at them, but as there are a few minor details I might as well postpone that bit of drudgery.

The CI scripting fortunately rarely changes, so keeping the Azure one and the GH one in sync shouldn't be too hard, but it would be good to ensure (somehow) that they are in sync. They seem quite similar, but I doubt it is possible to really combine the important bits. For now, I'll probably stick to the Azure one (it gives 10 parallel jobs for free for open-source projects, so it is not too bad), but I've looked at the GH actions a few times and wondered whether I shouldn't migrate.

.azure/templates/build-test.yml
Davide-1998: I agree that adding `--parallel` on azure pipeline does not yield a great difference in execution times, nonetheless on GH actions it allowed to drop the overall building+test time more than half.
Thus I decided to add the `--parallel` flag to try fasten it up and also for coherence with the commands on the GH pipeline.

src/psmx_iox/src/psmx_iox2_impl.c
Davide-1998: Good to know, when I first wrote the fix there were no new commits on the master branch.
I will revert the change then, thus to avoid conflicts with the master branch

.github/workflows/misra-regressions-check.yml
Davide-1998: Thanks for sharing your feedback about MISRA, I will remove it immediately.

.azure/templates/build-test.yml
Davide-1998: I added the `--repeat after-timeout:3` to allow only tests that timeout to be retried. I believe it is the best course of action to ensure "single crash bugs" to still be detectable, however errors like `live memory grew too much` will still cause a failure of the tests.

src/core/xtests/symbol_export/CMakeLists.txt
Davide-1998: I think it is best to follow your suggestion of removing the changes to symbol_export_test from this PR, thus keeping the overall work relevant to the CI.

eboasson (approved):
Thanks @Davide-1998 !

PR #2398, opened 2026-05-22
Add support for IntervalZero's RTX64 platform.

MI-IntervalZero:
In the CMake scripts, "RTX64" is used to identify the RTX64 platform.

In the source code, the combination of "WIN32" and "UNDER_RTSS" macros is used to identify code specific to the RTX64 platform.

See also: ports/rtx64/README.md


eboasson (commented):
Thanks @MI-IntervalZero !

Most of the CI is failing:
```
+ pre-commit run --all-files
[INFO] Initializing environment for local.
[INFO] Installing environment for local.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
bison-parser-hashcheck...................................................Passed
confgen-hashcheck........................................................Failed
- hook id: must-match-hash
- exit code: 1
```
Which is because you did not include the generated configuration files (`.xsd`, `.rnc`, `.md`, etc.) in the commit. That can easily be fixed: it simply means these files:

* docs/manual/config/config_file_reference.rst
* docs/manual/options.md
* etc/cyclonedds.rnc
* etc/cyclonedds.xsd
* src/core/ddsi/defconfig.c

need to be regenerated from the new sources (running `cmake . && make` in the build directly usually suffices) and included in the commit.

However ... There is a problem which I only now realize: `src/core/ddsi/defconfig.c` contains the code to default-initialize a configuration struct for those who want to use `dds_create_domain_with_rawconfig` (I think the RTX64 users might be among them). With the different default under RTX64 than under other platforms, and the code generation tool only seeing the preprocessor output for `ddsi__cfgelems.h`, it'll be next-to-impossible to get that generated source file to have the desired values in all cases.

(We could do something in the configuration string interpretation, allowing one to add overrides for specific platforms/devices `"14720B (RTX64=1500B; raweth=1500B)"` or something. For numeric values that'd be doable, but I think outside the scope of this PR!)

So, it probably is better to leave the defaults as they are, removing all changes to `ddsi__cfgelems.h`, and simply override the value in `ddsi_udp_init`. I don't like it, but the alternatives are all rather more complicated ...

The Python/C++ checks are fine, so I expect that undoing the changes in `ddsi__cfgelems.h` will make it pass CI.

eboasson (approved):
Thank you!

PR #2399, opened 2026-05-25
Fix take_next_instance for invalid samples

disganaitis:
Take next_nonempty_instance_by_id was ignoring invalid samples when checking if an instance is empty


eboasson (approved):
LGTM, thanks @disganaitis !

PR #2401, opened 2026-05-26
Reduce stack size usage in ddsperf.

MI-IntervalZero:
The goal is to avoid using 64+KB of thread's stack in ddsperf.

Use static storage instead of stack storage for the "union data data" variables defined in the maybe_send_ping() and pubthread() functions.

There is no risk because maybe_send_ping() is called from the main() function only, and pubthread() is the entry point of the "pub" thread which is spawned no more than once in the ddsperf application.


eboasson (approved):
That's a useful change. Thank you @MI-IntervalZero

PR #2406, opened 2026-06-01
Replace call to strdup by call to ddsrt_strdup.

MI-IntervalZero:
strdup is part of POSIX but hasn't been part of the standard C lib until C23. A simple way to improve code portability of the dynsub example  is to replace stdup by ddsrt_strdup.


eboasson (approved):
Thanks @MI-IntervalZero !

PR #2409, opened 2026-06-09
Add support for building under `aarch64` for Windows

noxpardalis:
I was setting up the builds with a vendored Cyclone C lib for the Rust API and noticed that Cyclone DDS doesn't currently build under the `aarch64-msvc` target. I didn't add a new OS in the matrix for the `azure-pipelines.yml` but I guess it would be pretty easy (just specifying the correct image from the set of images supported on Azure).


eboasson (approved):
Thanks @noxpardalis !

PR #2410, opened 2026-06-11
Fix thread-local storage section name under Windows

noxpardalis:
I ran into some issues with the `TLS_OUT_OF_INDEXES` when vendoring Cyclone for the Rust API. This seems to happen when doing a static build of Cyclone for Windows in `debug`.

According to https://web.archive.org/web/20251116170835/http://www.nynaeve.net/?p=183 (which the original TLS comments that we have in Cyclone reference):

> The C runtime uses this quirk of the compiler to create an array of
> null terminated function pointers to TLS callbacks (with the pointer
> stored in the ".CRT$XLZ" section being the null terminator).

I'm not 100% sure whether the XLZ section is just the null terminator or is also executed, but correcting the section name fixes this assert on static Windows builds (under debug):

```c
dds_return_t ddsrt_thread_cleanup_push(void (*routine)(void *), void *arg)
{
...
  assert(cleanup != TLS_OUT_OF_INDEXES);
```

The `cleanup` var was not being initialized by `ddsrt_thread_init` without this fix.

### DIFF
--- a/src/ddsrt/src/cdtors.c
+++ b/src/ddsrt/src/cdtors.c
@@ -195,17 +195,17 @@ DDSRT_WARNING_CLANG_ON(missing-prototypes)
 // logic isn't exactly as trivial as for example determining the endianness of
 // a platform, so keeping this close to the implementation is probably wise.
 #if __MINGW32__
-  PIMAGE_TLS_CALLBACK __crt_xl_tls_callback__ __attribute__ ((section(".CRT$XLZ"))) = ddsrt_cdtor;
+  PIMAGE_TLS_CALLBACK __crt_xl_tls_callback__ __attribute__ ((section(".CRT$XLY"))) = ddsrt_cdtor;
 #elif _WIN64
   #pragma comment (linker, "/INCLUDE:_tls_used")
   #pragma comment (linker, "/INCLUDE:tls_callback_func")
-  #pragma const_seg(".CRT$XLZ")
+  #pragma const_seg(".CRT$XLY")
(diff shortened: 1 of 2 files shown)


src/ddsrt/src/cdtors.c
```diff
@@ -195,17 +195,17 @@ DDSRT_WARNING_CLANG_ON(missing-prototypes)
 // logic isn't exactly as trivial as for example determining the endianness of
 // a platform, so keeping this close to the implementation is probably wise.
```
eboasson, line 213: I would also add a reference to https://learn.microsoft.com/en-us/cpp/c-runtime-library/crt-initialization?view=msvc-170 : that takes some of the magic out.

eboasson (commented):
Thanks @noxpardalis !

I think it would be better to update the comment to explain at least somewhat what is happening here. It turns out that there is actually a decent reference in the Microsoft documentation.

Perhaps you can have a look at this and come up with an improved version of the comment?

noxpardalis:
@eboasson

> Thanks @noxpardalis !
> 
> I think it would be better to update the comment to explain at least somewhat what is happening here. It turns out that there is actually a decent reference in the Microsoft documentation.
> 
> Perhaps you can have a look at this and come up with an improved version of the comment?

While trying to find information on this I had read somewhere that `XLB` was also reserved and so while I understood that XLA and XLZ where probably just the sentinels and non-configurable I wasn't sure what other reserved sections there were.

Thanks for linking the docs because reading through them makes it clearer that B-Y is just fully available. I did a pass to improve the documentation in 8e7f5ac.

eboasson (approved):
Thanks!

Issue #2413, opened 2026-06-13
QoS provider failes to load xml with tags


trittsv:
The QoS Provider fails when trying to load the Qos-Provider XML file.
It would be cool to have these xml-attributes.

Issues:
1. xsi:noNamespaceSchemaLocation:
`1781339165.009964 []   16196523: Error parsing system definition XML: Unknown attribute 'xsi:noNamespaceSchemaLocation' (error code -1, line 19)`
2. is_default_qos:
`1781340210.309587 []   16283810: Error parsing system definition XML: Unknown attribute 'is_default_qos' (error code -1, line 20)`
3. base_name:
`1781340251.449959 []   16283810: Error parsing system definition XML: Invalid value for attribute 'base_name' (error code -1, line 34)`

The XML:
````xml
<dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:noNamespaceSchemaLocation="../schema/dds_qos_profiles.xsd">
    <qos_library name="Shapes_Lib">
        <qos_profile name="Default_Profile" is_default_qos="true">
            <datawriter_qos>
                <writer_data_lifecycle>
                    <autodispose_unregistered_instances>false</autodispose_unregistered_instances>
                </writer_data_lifecycle>
            </datawriter_qos>
            <datareader_qos>
                <history>
                    <kind>KEEP_LAST_HISTORY_QOS</kind>
                    <depth>6</depth>
                </history>
            </datareader_qos>
        </qos_profile>

        <qos_profile name="Batching" base_name="Default_Profile">            
            <datawriter_qos>
                <batch>
                    <enable>true</enable>
                    <max_flush_delay>
                        <sec>1</sec>
                        <nanosec>0</nanosec>
                    </max_flush_delay>
                    <max_data_bytes>LENGTH_UNLIMITED</max_data_bytes>
                    <max_samples>10</max_samples>
                </batch>
                <history>
                    <kind>KEEP_LAST_HISTORY_QOS</kind>
                    <depth>10</depth>
                </history>
           </datawriter_qos>
        </qos_profile>
    </qos_library>
</dds>
```

Splinter1984:
Hi, @trittsv.
[...]

PR #2414, opened 2026-06-15
resolve qos_lib scope profiles reference

Splinter1984:
This changes allow to resolve not only `<qos_lib>::<qos_profile>` ref. within `base_name`, but `<qos_profile>`. Thus `base_name` ref. on `<qos_profile>` will be resolved with profile current `qos_lib`. Such kind of reference allowed only with use inside of `qos_lib` and `qos_profile` scope. All other cases treat as parse error.

> [!NOTE]
> this pull request may resolve topics highlighted by the issue <https://github.com/eclipse-cyclonedds/cyclonedds/issues/2413>

### DIFF
--- a/src/core/ddsc/src/dds_sysdef_parser.c
+++ b/src/core/ddsc/src/dds_sysdef_parser.c
@@ -879,11 +879,43 @@ static int split_ref (const char *ref, char **lib, char **local_name)
     return *type != NULL ? SD_PARSE_RESULT_OK : SD_PARSE_RESULT_INVALID_REF; \
   }
 
-RESOLVE_REF_FNDEF(qos_profile, qos, qos_profiles)
 RESOLVE_REF_FNDEF(domain, domain, domains)
 RESOLVE_REF_FNDEF(participant, participant, participants)
 RESOLVE_REF_FNDEF(node, node, nodes)
 RESOLVE_REF_FNDEF(application, application, applications)
+
+static int proc_attr_resolve_qos_profile_ref (struct parse_sysdef_state * const pstate, const char *type_ref, struct dds_sysdef_qos_profile **qos_profile)
+{
+  char *lib_name, *local_name;
+  if (*qos_profile != NULL)
+    return SD_PARSE_RESULT_DUPLICATE;
+
+  struct dds_sysdef_qos_lib *lib = NULL;
+  if (split_ref (type_ref, &lib_name, &local_name) != SD_PARSE_RESULT_OK)
+  {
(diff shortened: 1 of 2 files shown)


src/core/ddsc/src/dds_sysdef_parser.c
```diff
@@ -879,12 +879,44 @@ static int split_ref (const char *ref, char **lib, char **local_name)
...
+static int proc_attr_resolve_qos_profile_ref (struct parse_sysdef_state * const pstate, const char *type_ref, struct dds_sysdef_qos_profile **qos_profile)
+{
+  char *lib_name, *local_name;
+  if (*qos_profile != NULL)
+    return SD_PARSE_RESULT_DUPLICATE;
+
+  struct dds_sysdef_qos_lib *lib = NULL;
+  if (split_ref (type_ref, &lib_name, &local_name) != SD_PARSE_RESULT_OK)
+  {
+    if (pstate->current->parent != NULL &&
```
eboasson: Wouldn't it be clearer to walk the parents looking for a `ELEMENT_KIND_QOS_LIB`? Like (untested):
```
tmp = pstate->current->parent;
while(tmp && tmp->kind != ELEMENT_KIND_QOS_LIB)
  tmp = tmp->parent;
if (tmp == NULL)
  return SD_PARSE_RESULT_ERR;
else
  lib = tmp;
```
Or something along those lines, there are many possible variations. I think it makes a perfectly clear that in the `lib` is a library, without having to check the structure of the element tree.

src/core/ddsc/tests/sysdef_parser.c
```diff
@@ -18,7 +18,7 @@
```
eboasson, line 18: Perhaps we can also add a test for the name resolution change?

eboasson (commented):
Thanks @Splinter1984 !

src/core/ddsc/src/dds_sysdef_parser.c
Splinter1984: suggestion looks perfect. 338d72d3

src/core/ddsc/tests/sysdef_parser.c
Splinter1984: sure added in 338d72d3

eboasson (commented):
Hi @Splinter1984 the changes look good in principle, but there's a test that fails now. Can you have a look, please?

eboasson (approved):
Thanks @Splinter1984 !

PR #2416, opened 2026-06-16
resolve qos_profile base ref. for qos-provider.

Splinter1984:
resolve chain of inheritance for qos_profile against other profiles. Update documentation with `base_name` ref. examples and python api code-block. Extend testing of the qos_provider also.

> [!NOTE]
> Additional changes required for resolution of the topic <https://github.com/eclipse-cyclonedds/cyclonedds/issues/2413>


src/core/ddsc/src/dds_qos_provider.c
```diff
@@ -283,6 +342,8 @@ dds_return_t dds_create_qos_provider_scope (const char *path, dds_qos_provider_t
 {
   dds_return_t ret = DDS_RETCODE_OK;
   struct dds_sysdef_system *sysdef;
+  if (path == NULL)
```
eboasson, line 346: I don't mind making `NULL` a supported value here, but I think we should then also update the documentation.

eboasson (commented):
Thanks @Splinter1984 !

src/core/ddsc/src/dds_qos_provider.c
Splinter1984: resolved in 5ce9a4c :ok_hand:

eboasson (approved):
Thanks @Splinter1984 !

PR #2417, opened 2026-06-18
OSS-Fuzz: Fix build for CXX17 requirements

arthurscchan:
This PR fixes the build configuration for the fuzzers to accommodate the CXX17 requirements.


eboasson (approved):
Thanks @arthurscchan ! Strange, I can't find any email warning me of a build failure (other than old ones that I fixed), but https://introspector.oss-fuzz.com says it failed, and I find it entirely plausible that bumping the C++ version could fix it.

Let's see what happens.

PR #2433, opened 2026-07-27
Fix build when ENABLE_QOS_PROVIDER=OFF

banburybill:
Currently building with -DENABLE_QOS_PROVIDER=NO fails:

```
[271/291] Linking C executable bin/symbol_export_test
FAILED: [code=1] bin/symbol_export_test 
: && /usr/bin/cc -O2 -g -DNDEBUG -Wl,--dependency-file=src/core/xtests/symbol_export/CMakeFiles/symbol_export_test.dir/link.d src/core/xtests/symbol_export/CMakeFiles/symbol_export_test.dir/symbol_export.c.o -o bin/symbol_export_test  -Wl,-rpath,/home/ultra/build/cyclonedds/lib  lib/libddsc.so.11.0.1 && :
/usr/bin/x86_64-linux-gnu-ld.bfd: src/core/xtests/symbol_export/CMakeFiles/symbol_export_test.dir/symbol_export.c.o: in function `main':
/home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1250:(.text.startup+0x34e3): undefined reference to `dds_sysdef_init_sysdef'
/usr/bin/x86_64-linux-gnu-ld.bfd: /home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1251:(.text.startup+0x34f4): undefined reference to `dds_sysdef_init_sysdef_str'
/usr/bin/x86_64-linux-gnu-ld.bfd: /home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1252:(.text.startup+0x34fe): undefined reference to `dds_sysdef_fini_sysdef'
/usr/bin/x86_64-linux-gnu-ld.bfd: /home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1253:(.text.startup+0x350d): undefined reference to `dds_sysdef_init_data_types'
/usr/bin/x86_64-linux-gnu-ld.bfd: /home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1254:(.text.startup+0x351c): undefined reference to `dds_sysdef_init_data_types_str'
/usr/bin/x86_64-linux-gnu-ld.bfd: /home/ultra/src/cyclonedds/src/core/xtests/symbol_export/symbol_export.c:1255:(.text.startup+0x3526): undefined reference to `dds_sysdef_fini_data_types'
collect2: error: ld returned 1 exit status
[...]

PR #2434, opened 2026-07-27
detach sysdef_parser priv. from qos_provider pub.

Splinter1984:
original issue by (@banburybill): <https://github.com/eclipse-cyclonedds/cyclonedds/pull/2433>

### DIFF
--- a/src/core/ddsc/CMakeLists.txt
+++ b/src/core/ddsc/CMakeLists.txt
@@ -48,6 +48,7 @@ prepend(srcs_ddsc "${CMAKE_CURRENT_LIST_DIR}/src/"
   dds_psmx.c
   dds_guid.c
   dds_err_check.c
+  dds_sysdef_parser.c
 )
 
 if(ENABLE_TYPELIB)
@@ -55,7 +56,6 @@ if(ENABLE_TYPELIB)
 endif()
 if(ENABLE_QOS_PROVIDER)
   list(APPEND srcs_ddsc
-    "${CMAKE_CURRENT_LIST_DIR}/src/dds_sysdef_parser.c"
     "${CMAKE_CURRENT_LIST_DIR}/src/dds_sysdef_validation.c"
     "${CMAKE_CURRENT_LIST_DIR}/src/dds_qos_provider.c")
 endif()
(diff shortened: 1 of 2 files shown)


src/core/ddsc/CMakeLists.txt
```diff
@@ -48,14 +48,14 @@ prepend(srcs_ddsc "${CMAKE_CURRENT_LIST_DIR}/src/"
   dds_psmx.c
   dds_guid.c
   dds_err_check.c
+  dds_sysdef_parser.c
```
eboasson, line 51: The idea behind the `ENABLE_QOS_PROVIDER` switch is that the entire QoS provider can be left out if it isn't needed. So this file should remain conditional.

src/core/ddsc/tests/CMakeLists.txt
```diff
@@ -230,6 +230,7 @@ set_tests_properties(
 		REQUIRED_FILES ${CUnit_ddsc_config_simple_udp_file}
 		ENVIRONMENT "${CUnit_ddsc_config_simple_udp_env}")
 
+if(ENABLE_QOS_PROVIDER)
```
eboasson, line 233: A nitpick: I'd indent the conditional part

src/core/ddsc/tests/CMakeLists.txt
```diff
@@ -143,6 +143,7 @@ if(ENABLE_TYPELIB)
     "typebuilder.c"
     "dynamic_type.c"
     "typewrap.c"
+    "sysdef_parser.c"
```
eboasson, line 146: I want to be able to build without including anything from the QoS provider, so the tests can't be present either if `ENABLE_QOS_PROVIDER` is false.

What you need for that is exactly what @banburybill did in https://github.com/eclipse-cyclonedds/cyclonedds/pull/2433/changes/e9f612740d6a8fe33260cfdcc24b20c1087dfa1d : make it conditional in `symbol_export.c`.

eboasson (commented):
Thanks @Splinter1984 for looking into #2433, but the solution proposed by @banburybill better fits the intention behind the `ENABLE_QOS_PROVIDER` flag. So I'll close this PR and reopen (and merge) that one.

Splinter1984:
Sure! Thank you for your remarks.

PR #2438, opened 2026-08-05
fix: tls_callback_func already defined

Davide-1998:
This PR fixes #2422 by renaming `tls_callback_func` in `ddsrt` and `idl` libraries avoid duplication when statically linked:  

- *src/ddsrt/src/cdtors.c* : `tls_callback_func_ddsrt`
- *src/idl/src/string.c* : `tls_callback_func_idl`

PR #2439, opened 2026-08-05
fix: CMake for third party usage

### LINKED ISSUE #2297: feat: Support for building as a vendored library

Davide-1998:
This PR fixes #2297 but depends on #2438.


fuzz/fuzz_handshake/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ endif()
 # dependencies are not yet tracked by every cmake distribution. This makes it
 # necessary to source the cmake configuration files from the build directories
 # of protobuf (using the CONFIG argument to find_package)...
-set(PROTOBUF_ROOT "${CMAKE_BINARY_DIR}/../LPM/external.protobuf")
+set(PROTOBUF_ROOT "${CMAKE_CURRENT_BINARY_DIR}/../LPM/external.protobuf")
```
eboasson: Doesn't this change the directory where protobuf is to be found?

I am not sure if that breaks OSS-Fuzz builds or not because running the OSS-Fuzz infrastructure helpers is too hard for me locally these days. I do know that my hack to work locally (`fuzz/local.sh`) puts LPM in a fixed location and that this would break that build. I'm fine with `LPM` being in `fuzz`, I'd just like to keep everything working!

Perhaps it is also not worth changing these. Why would you want to build the fuzzers in a subproject?

src/core/cdr/CMakeLists.txt
```diff
@@ -23,48 +18,11 @@ set(srcs_cdr
...
 
-if(${CMAKE_PROJECT_NAME} STREQUAL "CycloneDDS")
-  target_sources(ddsc PRIVATE ${srcs_cdr} ${hdrs_private_cdr})
-  target_include_directories(ddsc PRIVATE "${CMAKE_CURRENT_LIST_DIR}/include")
-
-  install(
-    DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/include/"
-    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
-    COMPONENT dev)
-else()
```
eboasson, line 34: Doesn't removing this break using the CDR support code as a stand-alone library? I don't care for that use too much and I don't know that many people depend on it, but I would like to keep that possibility alive.

README.md
```diff
@@ -188,6 +188,52 @@ At this point you are ready to use Eclipse Cyclone DDS in your own projects.
 
 Note that the default build type is a release build with debug information included (RelWithDebInfo), which is generally the most convenient type of build to use from applications because of a good mix between performance and still being able to debug things.  If you'd rather have a Debug or pure Release build, set `CMAKE_BUILD_TYPE` accordingly.
 
+### Third party library
+
+It is also possible to build CycloneDDS as part of your project by treating it as a third party library.  
+Following is an example *CMakeLists.txt* using [FetchContent_Declare](https://cmake.org/cmake/help/latest/module/FetchContent.html) to put CycloneDDS into scope.
+
+```cmake
+cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
```
eboasson: `TARGET_RUNTIME_DLLS` isn't available in CMake 3.16, I guess the easiest way to deal with that is to change the minimum required in this example to 3.21.
```suggestion
cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
```

src/core/CMakeLists.txt
```diff
@@ -58,9 +58,9 @@ if(ENABLE_SECURITY)
     $<BUILD_INTERFACE:$<TARGET_PROPERTY:security_api,INTERFACE_INCLUDE_DIRECTORIES>>)
 endif()
 
-include(cdr/CMakeLists.txt)
-include(ddsi/CMakeLists.txt)
-include(ddsc/CMakeLists.txt)
+add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/cdr")
```
eboasson: I think the main consequence of replacing the `include` by `add_subdirectory` would be the scope of any variable definitions. I don't think these three `CMakeLists.txt` do anything like that, so I suppose it is safe to change it, and it does look like good hygiene. Besides, the CI is green, so it can't be too bad.

But what I don't understand is why this needs the `${CMAKE_CURRENT_SOURCE_DIR}`?

eboasson (commented):
Thanks for helping me to get the CMake stuff done well, @Davide-1998 !

I have a few straightforward comments, but apart from those this looks good to me.

fuzz/fuzz_handshake/CMakeLists.txt
Davide-1998: Hi, I didn't notice that the LPM folder would have been populated by the `local.sh` script and assumed it was another folder in the project. Given this new information, I will update the PR, having the folder structure as it was before.  
I don't know how much sense may have to build the fuzzers when CycloneDDS is used as a third party, but being these included in the project when `LIB_FUZZING_ENGINE` is set, seemed reasonable to also update their CMake.  
If this is not the intended behavior or the fuzzers must be restricted to only when CycloneDDS is built as a standalone package, I can add few lines in the CMake to avoid adding them when used as third party

src/core/cdr/CMakeLists.txt
Davide-1998: Yes, the current version of CMakeLists.txt for the cdr library will break any other CMake project that tries to link to `cdr::cdr` target.  
This was necessary due to the `cdr` project shadowing the `ddsc` target and thus forcing any third party user to address the ddsc library as `cdr::ddsc`.  
I agree that my solution is quite disruptive by enforcing `cdr` to be fully part of the main `ddsc` library and removing the homonymous target.  
If you want, I can delve further into re-enabling the `cdr` library to allow both `CycloneDDS::cdr` and `CycloneDDS::ddsc` targets to be valid.

README.md
Davide-1998: Yes, you are totally right!  
Thank you for catching it 💪🏻

src/core/CMakeLists.txt
Davide-1998: Hi, as you pointed out, there is no need to specify `CMAKE_CURRENT_SOURCE_DIR` given that `add_subdirectory` already resolves the input argument with respect to the current CMakeLists.txt file if it is a relative path.  
This has been an oversight caused by how I am used to write CMake files.  
I will restore the paths to just have the relevant folders

Davide-1998:
Hi, I have applied the suggestions:  
- Re-enable `cdr` library: now is addressable as `CycloneDDS::cdr` when building as a third party
- Restored `CMAKE_BINARY_DIR` path for the `fuzz` subfolder
  - also added a further check in the main CMakeLists.txt to build the fuzz folder only when CycloneDDS is the uppermost project
- Updated CMake version to `3.21` in the example
- Restored the relative paths in `core/CMakeLists.txt`

I also rebased this PR to the latest commit on master and solved the merge conflicts inside `src/tools/_confgen/CMakeLists.txt` where new instructions where added using the `CMAKE_SOURCE_DIR` variable

eboasson (approved):
Thanks @Davide-1998 !

PR #2450, opened 2026-08-18
Replace `bison` parser with a recursive-descent parser

eboasson:
This PR switches IDLC to a recursive-descent IDL parser (not quite handwritten thanks to ChatGPT). The existing Bison-generated parser remains available for now as a compile-time option. The main goal is to remove the normal build’s dependence on generated parser output for IDL parsing, making this part of the compiler easier to review. The new parser uses the existing scanner and AST construction machinery, so the intended behavior is parity with the previous parser rather than a user-visible language change.

**What changed**

- A new IDL parser covering the constructs currently handled by IDLC, including modules, structs, typedefs, arrays, template types, constants, annotations, enums, bitmasks, and unions.
- The hand-written parser is used by default with `-DENABLE_IDL_HAND_PARSER=ON`, the Bison parser is used when `-DENABLE_IDL_HAND_PARSER=OFF`.
- Add parser-output checking script (`scripts/compare-idlc-outputs.sh`) for easy checking that the generated output is unchanged.


0x5v3n:
@eboasson I'm not a big fan of the `ENABLE_IDL_HAND_PARSER` name. It feels a bit like a temporary variable name (like `myvariable`). Since this option replaces the Bison parser, maybe we could name it something like `ENABLE_IDL_CYCLONE_PARSER` instead?

Alternatively, could we just remove the Bison parser entirely so we don't have to maintain two different parsers in the project?

eboasson:
> @eboasson I'm not a big fan of the `ENABLE_IDL_HAND_PARSER` name. It feels a bit like a temporary variable name (like `myvariable`). Since this option replaces the Bison parser, maybe we could name it something like `ENABLE_IDL_CYCLONE_PARSER` instead?
> 
> Alternatively, could we just remove the Bison parser entirely so we don't have to maintain two different parsers in the project?

Hi @0x5v3n, the intent is to delete the Bison parser entirely. I left it in for now just in case there's a problem with the new one. Perhaps that is being overly cautious, it is not like one can't simply report a bug and go back to the old one temporarily.

eboasson:
@0x5v3n I decided to follow your suggestion and remove the old one altogether. The force push comes from rebasing on current master, it really is just two added commits.

PR #2454, opened 2026-08-29
Fix out-of-bounds base64 decode reads

### LINKED ISSUE #2441: b64_decode() reads one byte past the heap buffer for `<value>` payloads of length ≡ 2 (mod 4)
`b64_decode()` (`src/core/ddsc/src/dds_sysdef_parser.c:1890`) bounds its loop on the input and output
indices, but the loop body reads four bytes per iteration unconditionally. For an unpadded payload
whose length is ≡ 2 (mod 4), the final iteration reads one byte past the end of the heap allocation
holding the XML element text.

Reachable through the public `dds_create_qos_provider()` API via the `<user_data>`, `<topic_data>`
and `<group_data>` QoS policies.

Affected: 11.0.0, 11.0.1 and `master`. `b64_decode` was introduced in `ff640ac5` (2024-04-26) for
11.0.0 and does not exist on any 0.x branch, so the 0.10.x line used by ROS 2 is unaffected.

On reporting channel: I read `SECURITY.md` and the Eclipse process it links, and I'm filing publicly
on the same reasoning as the companion issue — the input is a configuration document the application
chose to load, and comparable parser findings (#1054, #1555) were handled on the public tracker.
Reclassify it if you see it differently.

## Reproducer

```xml
<dds><qos_library name="L"><qos_profile name="P"><domain_participant_qos><user_data><value>AB</value></user_data></domain_participant_qos></qos_profile></qos_library></dds>
```

Passed to `dds_create_qos_provider()`, built with `-DSANITIZER=address`. Verbatim ASan output,
trimmed to the relevant frames plus the allocation site:

```
==42676==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000002d3 at pc 0x000100eb1df8 bp 0x00016f3df090 sp 0x00016f3df088
READ of size 1 at 0x6020000002d3 thread T0
    #0 0x000100eb1df4 in b64_decode dds_sysdef_parser.c:1903
    #1 0x000100eb2260 in set_USERDATA_VALUE dds_sysdef_parser.c:2138
    #2 0x000100e97fc4 in proc_elem_data dds_sysdef_parser.c:2432
    #3 0x000100effe68 in parse_element xmlparser.c:719
    ...
   #16 0x000100eca7e0 in dds_create_qos_provider dds_qos_provider.c:265

0x6020000002d3 is located 0 bytes after 3-byte region [0x6020000002d0,0x6020000002d3)
allocated by thread T0 here:
    ...
    #3 0x0001049c96c8 in save_payload xmlparser.c:450
```

## Cause

```c
  for (size_t i = 0, j = 0; i < sz && j < buff_len; i+=4, j+=3)
  {
    ...
    uint32_t tmp = (uint32_t)(base64_dtable[text[i]] << 0x06U) | base64_dtable[text[i+1]];
    ...
    tmp = (safe << 0x0CU) | ((uint32_t)(base64_dtable[text[i+2]] << 0x06U) | base64_dtable[text[i+3]]);
```

The loop condition constrains `i` and `j`; the body reads `text[i+1]`, `text[i+2]` and `text[i+3]`
regardless. The allocation is exact-fit — `save_payload` (`src/ddsrt/src/xmlparser.c:450`) does
`ddsrt_malloc (last - first + 2)` and NUL-terminates, so the block is exactly `strlen + 1` bytes.

Only residue 2 over-reads. With no `=` present, `buff_len = sz * 3 / 4`:

| `sz % 4` | example | outcome |
|---|---|---|
| 0 | `QUJD` (sz=4, buff_len=3) | quantum complete, in bounds |
| 1 | `QUJDR` (sz=5, buff_len=3) | final iteration fails `j < buff_len`; loop exits first |
| 2 | `AB` (sz=2, buff_len=1) | `i=0`, `j=0` both hold; reads `text[3]` — **1 byte past a 3-byte block** |
| 3 | `QUJDRA` +`=` (sz=7 → padd at 6, buff_len=4) | `=` shortens `buff_len`; in bounds |

Generally for `sz = 4k + 2`: `buff_len = 3k + 1`, the last iteration has `i = 4k` and `j = 3k`, both
conditions hold, and it reads `text[4k+3]` while the allocation ends at index `4k + 2`. The over-read
is always exactly one byte — `j < buff_len` stops the loop, so it never walks. A `=` never makes it
worse: `padd - text <= sz` always, so `buff_len` only shrinks.

## Impact

Small, and I'd rather say so than dress it up. A one-byte heap over-read in a configuration parser,
not on any network path — no file under `src/core/ddsi/` references sysdef, and the only non-test
callers of `dds_sysdef_init_sysdef*` are in `dds_qos_provider.c`.

**The over-read byte does not reach decoded output.** In the over-reading iteration
`cp_sz = min(buff_len - j, 3) = 1` (`:1906`), so only `chunk[0]` is copied out, and `chunk[0]` is
derived solely from `text[i]` and `text[i+1]` — both in bounds. `text[i+3]` feeds `chunk[1]` and
`chunk[2]`, which are computed and then discarded. So there is no information disclosure; the risk is
the read itself — a SEGV if the allocation abuts an unmapped page, an abort under a hardening
allocator or ASan, and UB regardless.

## Why it has survived

The unpadded path *is* exercised — `tests/sysdef_parser.c` feeds `QUJD` (`:40`), `REVG` (`:56`),
`R0hJ` (`:58`) and `SlNM` (`:59`) through `all_constructs` (`:165`) and `all_constructs_values`
(`:170`), none of them padded. But every one is length 4. No test feeds a base64 `<value>` whose
length is ≡ 2 (mod 4), which is the only residue that triggers it.

(For completeness, since it looks like a counterexample: `:47-48` do contain two-character values —
`<ownership_strength><value>19</value>` and `<transport_priority><value>20</value>` — but those are
numeric policies that never reach `b64_decode`.)

## Suggested fix

Read each sextet only if it is in range, treating out-of-range as zero:

```c
    unsigned char c1 = (i + 1 < sz) ? text[i+1] : (unsigned char) 0;
    unsigned char c2 = (i + 2 < sz) ? text[i+2] : (unsigned char) 0;
    unsigned char c3 = (i + 3 < sz) ? text[i+3] : (unsigned char) 0;
    uint32_t tmp = (uint32_t)(base64_dtable[text[i]] << 0x06U) | base64_dtable[c1];
    ...
    tmp = (safe << 0x0CU) | ((uint32_t)(base64_dtable[c2] << 0x06U) | base64_dtable[c3]);
```

This is output-identical to the current code on every input. For the residue-2 case that reason is
the `cp_sz == 1` clamp described above — `c2` and `c3` only ever feed bytes that are discarded. For
every other residue the substituted reads are in bounds and unchanged. I extracted the function and
diffed both implementations:

| input | current | with fix |
|---|---|---|
| `QUJD` | `41 42 43` | `41 42 43` |
| `QUJDRA==` | `41 42 43 44` | `41 42 43 44` |
| `QUJDRA` | `41 42 43 44` | `41 42 43 44` |
| `Zm9vYmE` | `66 6f 6f 62 61` | `66 6f 6f 62 61` |
| `AB` | `00` | `00` |
| `QQ==` | `41` | `41` |

Applied to `394da46f` it removes the ASan report, and `ddsc_sysdef_parser` and `ddsc_qos_provider`
produce assert counts identical to baseline (199 and 627, zero failures) on clang/macOS and GCC 13.3
on Linux, warning-free.

Worth flagging what **not** to do, because it's the obvious one-liner and it's wrong: bounding the
loop with `i + 3 < sz` silently corrupts output, because `buff_len` is still returned at the larger
value while the final partial quantum is skipped. Measured: `QUJDRA` → `41 42 43 00`, and `Zm9vYmE`
→ `66 6f 6f 00 00` — the latter entirely in bounds today.

Separately, and at your discretion: `b64_decode` accepts any length and silently mis-decodes
non-multiple-of-4 input rather than rejecting it. Rejecting `sz % 4 != 0` would close the safety and
correctness gaps together, but it's a behaviour change for existing callers so I've kept it out of
the minimal fix.

## Related

A NULL-dereference crash in the same parser is filed as #2440. There is currently no fuzz coverage of
`dds_sysdef_parser.c`; I have a `fuzz_qos_provider` target that `build_fuzzers` and `check_build`
both accept, and would be glad to send it as a PR.

Found with automated tooling; the analysis is mine.

timothyanderson096-ocdealcheck:
## Summary

Fixes the out-of-bounds look-ahead reads in `b64_decode()` when processing an incomplete Base64 input quantum.

The decoder now reads each optional sextet only when its index remains within the input buffer. This preserves the existing decoded output for both complete and partial inputs while removing the one-byte heap-buffer-overflow reported in #2441.

A regression test exercises the two-character payload through the public `dds_create_qos_provider()` API.

Fixes #2441.

## Validation

* Unpatched AddressSanitizer build reproduced the one-byte heap-buffer-overflow in `b64_decode()`.
* Patched ASan `ddsc_qos_provider`: 8 tests passed, 629 assertions passed.
* Patched ASan `ddsc_sysdef_parser`: 3 tests passed, 199 assertions passed.
* `git diff --check`: passed.

## AI assistance

OpenAI Codex assisted with implementation and test execution. GDN was used to reproduce the baseline fault and verify the patched behaviour. I reviewed the resulting change and evidence and take responsibility for this contribution.


eboasson (approved):
Thank you @timothyanderson096-ocdealcheck !

PR #2455, opened 2026-09-01
Stricter bounds checking on input sequences

eboasson:
For some element types, the in-memory representation may be larger than the CDR representation. These cases need to be checked against the 32-bit allocation limit.


noxpardalis (approved):
LGTM!

I haven't come across using `setjmp` and `longjmp` like this before (not that I've used them much to begin with). It's a fun alternative to mocking in the tests here.

PR #2456, opened 2026-09-03
OSS-Fuzz: Fix fuzzer build path error

arthurscchan:
In PR #2449, the inclusion directory paths of some fuzzers has been changed in their corresponding CMakeLists.txt. But there is one missing changes for fuzz_handshake and causes the OSS-Fuzz, specifically fuzz_handshake, build failed since then. This is gone undetected because that is not included in any CI nor cyclonedds build. This PR fixes that and re-enable the correct build for OSS-Fuzz fuzzers.


eboasson (approved):
Thanks!

eboasson:
I used to get emails from OSS-Fuzz when the build broke, but that apparently is no longer the case ... thank you for discovering it, and also for fixing it.

arthurscchan:
Yes, it does happens. And just FYI, I was being lucky because I found that out when I am starting to develop more fuzzers for the cyclonedds project. You may expect to see some additional fuzzers creation PR in the coming days. Thanks.
@eboasson

PR #2459, opened 2026-09-08
Reject system definitions without a root element

### LINKED ISSUE #2440: dds_create_qos_provider() segfaults on an empty or element-free QoS profile file
`dds_create_qos_provider()` dereferences a NULL pointer and crashes when given a system-definition
document that parses without error but opens no `<dds>` root element. **A zero-byte file is
sufficient.**

Affected: 11.0.0, 11.0.1 and `master`. The QoS provider does not exist on any 0.x branch, so the
0.10.x line used by ROS 2 is unaffected.

On reporting channel: I read `SECURITY.md` and the Eclipse Vulnerability Reporting process it links.
I'm filing publicly because this needs an application to feed its own configuration parser a document
it chose to load — no trust boundary is crossed in a default deployment — and because #1054 (a
fuzz-found NULL-deref in the config parser) and #1555 (a heap overflow reachable from a malformed
RTPS submessage) were both handled on the public tracker. If you classify it differently, say so and
I'll follow whatever process you prefer.

## Reproducer

```c
#include "dds/dds.h"
#include "dds/ddsc/dds_public_qos_provider.h"

int main (int argc, char **argv)
{
  dds_qos_provider_t *p = NULL;
  dds_create_qos_provider (argv[1], &p);   /* SEGV on an empty file */
  return 0;
}
```

```
$ : > empty.xml && ./repro empty.xml
```

Built against `394da46f` with `-DSANITIZER=address -DENABLE_QOS_PROVIDER=ON`. Verbatim ASan output,
trimmed after frame `#4`:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==42573==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000030 (pc 0x000102ee0048 bp 0x00016d3ce710 sp 0x00016d3ce180 T0)
==42573==The signal is caused by a READ memory access.
==42573==Hint: address points to the zero page.
    #0 0x000102ee0048 in dds_validate_qos_lib dds_sysdef_validation.c:105
    #1 0x000102ee31b8 in read_validate_sysdef dds_qos_provider.c:99
    #2 0x000102ee2770 in dds_create_qos_provider_scope dds_qos_provider.c:348
    #3 0x000102ee2530 in dds_create_qos_provider dds_qos_provider.c:265
    #4 0x000102a30ae4 in main repro.c:7
```

`0x30` is `offsetof(struct dds_sysdef_system, qos_libs)`, confirmed with a local `offsetof` probe.
Reproduces 10/10, with and without sanitizers, and in the OSS-Fuzz x86_64 container.

Any document consumed without opening an element does it. Tested on `394da46f`:

| input | result |
|---|---|
| empty file (0 bytes) | SEGV |
| `<!-- only a comment -->` | SEGV |
| `<?` | SEGV |
| `<` | no crash — syntax error, `DDS_RETCODE_ERROR` |
| `<x/>` | no crash — `CHECK_PARENT_NULL` (`dds_sysdef_parser.c:2607`, macro at `:83-87`) rejects it |
| `<dds/>` | no crash — root opens, `pstate->sysdef` allocated |

## Cause

**1. `sysdef_parse()` reports success while leaving the out-parameter NULL.**
`src/core/ddsc/src/dds_sysdef_parser.c:3134`:

```c
static dds_return_t sysdef_parse(struct ddsrt_xmlp_state *xmlps, struct parse_sysdef_state *pstate, struct dds_sysdef_system **sysdef)
{
  dds_return_t ret = DDS_RETCODE_OK;
  if ((ret = ddsrt_xmlp_parse (xmlps)) != SD_PARSE_RESULT_OK)
  {
    SYSDEF_ERROR ("Error parsing system definition XML: %s (error code %d, line %d)\n", pstate->err_msg, ret, pstate->err_line);
    ret = DDS_RETCODE_ERROR;
    if (pstate->sysdef != NULL)
      dds_sysdef_fini_sysdef (pstate->sysdef);
  }
  else
  {
    *sysdef = pstate->sysdef;
  }

  return ret;
}
```

`pstate->sysdef` is assigned in exactly one place, `dds_sysdef_parser.c:2601`, inside the
`ddsrt_strcasecmp (name, "dds") == 0` branch of `proc_elem_open`. It is initialised to NULL at
`:3169`. An empty document produces `TOK_EOF` on the first token, `ddsrt_xmlp_parse` returns `0`,
`SD_PARSE_RESULT_OK == 0` (`dds__sysdef_parser.h:47`), and so the function returns
`DDS_RETCODE_OK` with `*sysdef == NULL`.

**2. `read_validate_sysdef()` validates that NULL.** `src/core/ddsc/src/dds_qos_provider.c:90`:

```c
static dds_return_t read_validate_sysdef(const char *path, struct dds_sysdef_system **sysdef)
{
  dds_return_t ret = DDS_RETCODE_OK;
  struct dds_sysdef_system *def;
  if ((ret = read_sysdef (path, &def)) != DDS_RETCODE_OK)
  {
    QOSPROV_ERROR("Failed during read sysdef: %s\n", path);
    goto err_read;
  }
  if ((ret = dds_validate_qos_lib (def, PROVIDER_ALLOWED_QOS_MASK)) != DDS_RETCODE_OK)
  {
    QOSPROV_ERROR("Failed during validate sysdef: %s\n", path);
    goto err_validate;
  }
  *sysdef = def;
```

`def` is uninitialised, `read_sysdef` returns OK having written NULL into it, and
`dds_validate_qos_lib` dereferences unconditionally at `dds_sysdef_validation.c:105`:

```c
  for (const struct dds_sysdef_qos_lib *qos_lib = sysdef->qos_libs; qos_lib != NULL; qos_lib = (struct dds_sysdef_qos_lib *) qos_lib->xmlnode.next)
```

## Why the empty-file case matters

`read_sysdef` (`dds_qos_provider.c:28-52`) treats an argument starting with `<` as an inline
document and anything else as a filesystem path, including a `file://` URI, which it `fopen`s at
`:40`. So an application that loads a QoS profile from a configured path takes a hard process crash
if that file is empty or truncated — no return code to check, nothing to catch. Zero-length files
are a normal outcome of a failed write, an interrupted deploy, or an unwritten config.

This also reaches the language bindings: `cyclonedds-python`'s `QosProvider` and `cyclonedds-cxx`'s
`QosProviderDelegate` both wrap this call, so a Python application pointed at a bad file loses the
interpreter.

## The same pattern exists twice

`dds_sysdef_init_data_types_impl` (`dds_sysdef_parser.c:3425`) has the identical shape at
`:3447-3450`:

```c
  else
  {
    *type_meta_data = pstate.type_meta_data;
  }
```

I have not built a reproducer for that path, but whatever you do for `sysdef_parse` probably wants
doing there too.

## Suggested fix

In `sysdef_parse`, treat "no root element" as a parse failure:

```c
  else if (pstate->sysdef == NULL)
  {
    SYSDEF_ERROR ("No system definition root element found\n");
    ret = DDS_RETCODE_ERROR;
  }
  else
  {
    *sysdef = pstate->sysdef;
  }
```

`DDS_RETCODE_ERROR` to match the parse-failure branch above it. In that branch `pstate->sysdef` is
NULL by construction, so nothing leaks — an element arriving while `pstate->current == NULL` already
returns `SD_PARSE_RESULT_SYNTAX_ERR` from `CHECK_PARENT_NULL` and takes the existing error path.

Separately, `read_validate_sysdef` would be more robust initialising `def = NULL` and rejecting it
before validating, so the public entry point doesn't depend on the parser's contract.

**What I actually tested:** only the `sysdef_parse` guard above. Applied to `394da46f`, it removes
the crash for all six inputs in the table, and `ddsc_sysdef_parser` and `ddsc_qos_provider` produce
assert counts identical to baseline (199 and 627, zero failures) on both clang/macOS and GCC 13.3 on
Linux, warning-free. I did **not** implement or test the `read_validate_sysdef` half, nor the
`data_types` twin.

Existing tests already encode the invariant: `tests/sysdef_parser.c`'s `assert_parse_result`
(`:116-126`) asserts `sysdef != NULL` only when `ret == OK`, and no existing test feeds a root-less
document.

## Related

There is currently no fuzz coverage of `dds_sysdef_parser.c` or `dds_qos_provider.c` — none of the
seven targets under `fuzz/` reach `dds_sysdef_init_sysdef_str()`. I have a `fuzz_qos_provider` target
following the existing `fuzz/fuzz_*` layout, guarded on `ENABLE_QOS_PROVIDER`, which
`infra/helper.py build_fuzzers` and `check_build` both accept. Happy to open that as a PR if it'd be
useful.

Found with automated tooling; the analysis is mine.

vpfkfl753:
Fixes #2440 using the parser guard suggested there.

Reject system-definition documents that contain no `<dds>` element. The parser currently returns success with a null system definition, which crashes the QoS provider during validation.

Adds regression tests for empty strings/files, comments and XML declarations, plus a check that `<dds/>` remains valid. The sysdef parser and QoS provider suites pass under AddressSanitizer (13 tests); the new cases fail on the original code.


eboasson (commented):
Thank you @vpfkfl753 ! It looks good to me and I'd like to merge it, but I can't unless you sign the [ECA](https://www.eclipse.org/legal/eca/). (It doesn't automatically revalidate, so please let me know when you signed it, or if you decide not to sign it.)

vpfkfl753:
Thanks, I've signed the ECA. The check is passing now.

eboasson (approved):
Thank you!

PR #2461, opened 2026-09-11
ports/zephyr: update the Zephyr port to Zephyr v4.3+

axrbl:
## Summary

Bring the Zephyr port of CycloneDDS up to date with Zephyr v4.3+. The port
carried a number of Zephyr-only shims and "ignored socket option" paths that are
no longer needed (and no longer work) against current Zephyr, which now provides
a POSIX layer for sockets, threads and the environment.

The change is split into 18 focused commits (ddsrt, ddsi, ports/zephyr); each one
builds on its own.

## ddsrt

* build the POSIX `environ` implementation on Zephyr instead of the Zephyr-only
  copy (`src/environ/zephyr/environ.c` is removed);
* `ifaddrs`: adapt to the v4.3 `struct net_if_addr_ipv4` layout
  (`unicast[].ipv4.is_used`, `...ipv4.address.in_addr`, per-address `netmask`)
  and drop the `ifaddrs-v3.7pre.c` stop-gap;
* `process`: use `CONFIG_KERNEL_BIN_NAME` as the process name;
* `threads`: follow the rename of `CONFIG_MAX_PTHREAD_COUNT` to
  `CONFIG_POSIX_THREAD_THREADS_MAX`, drop the unused `_open()` stub, and honour
  `PTHREAD_EXPLICIT_SCHED` (the POSIX layer implements it now);
* `sockets`: let the Zephyr POSIX layer handle `IP_ADD/DROP_MEMBERSHIP` in
  `getsockopt` and all multicast options in `setsockopt` except
  `IP_MULTICAST_IF` (see Notes), delete the obsolete `IP_MULTICAST_*` fallback
  defines, and use Zephyr's own `recvmsg` instead of the local shim;
* `strtod`: pass the `float` to `snprintf` as `double`.

## ddsi

* do not request packet destination info on Zephyr (`IP_PKTINFO`/`IPV6_PKTINFO`
  are not implemented; asking for it leaves every received packet without a
  usable destination address);
* pass the multicast TTL / loopback flag as `unsigned int`, which is what the
  Zephyr socket layer expects.

## ports/zephyr

* do not build the host-only tools (`dyntypelib`/`xmltype`) when cross
  compiling -- they would be linked with the host compiler against the
  Zephyr-targeted `libddsc.a`;
* enable IGMP in the example configuration (CycloneDDS discovers peers over IPv4
  multicast; without IGMP membership the participant never receives it);
* fix an extra parenthesis in the Hello World IPv4 peer check;
* drive the Hello World example through `config.xml` + `CYCLONEDDS_URI` like the
  other examples, instead of building a raw `ddsi_config` from internal headers;
* finish the Hello World restructuring that the other examples already had:
  the entry point is `helloworld_main.c` and the pub/sub translation units come
  from `copy_examples.sh`.

## Upstream defects fixed along the way

These are the reason the port's own examples were never built:

1. `ports/zephyr/src/main.c` had an unmatched `)` in the IPv4 peer check and
   could not compile when `CONFIG_NET_CONFIG_PEER_IPV4_ADDR` was set.
2. `GENERATE_CDDS_CONF("config.xml" ...)` feeds the XML file to the `xxd`
   subprocess through its `INPUT_FILE`; relative paths are resolved against the
   build directory, so the generated header came out empty and syntactically
   invalid (`{ ..., , 0 }`). The new call site passes an absolute path.
3. The Hello World example was left half-way through the sample restructuring:
   the CMakeLists referred to `helloworld_pub.c`/`helloworld_sub.c` (produced by
   `copy_examples.sh`) while the entry point was still called `main.c`.

## Notes

* `IP_MULTICAST_IF` keeps being ignored on Zephyr: Zephyr rejects the `ip_mreqn`
  form, and a failed/partial `setsockopt` leaves the socket unable to receive
  multicast. The transmit interface is taken from the routing table instead.
* No application-side `environ` definition is needed anymore: the configuration
  is published with `setenv("CYCLONEDDS_URI", ...)` and the POSIX layer owns
  `environ`. The previously proposed "mark environ weak in Zephyr" change is
  therefore not required.

## Known limitations

* `sched_get_priority_min()` / `sched_get_priority_max()` are still not
  implemented by Zephyr's POSIX layer while `ddsrt`'s POSIX thread layer calls
  them, so an application-side stub is needed until either side is fixed.
* RAM footprint is still high: a minimal configuration needs on the order of
  1 MB, so MCU targets with small per-core RAM need a custom memory layout.

## Tested on

* `native_sim` (Zephyr v4.3+): builds and runs; publish/subscribe verified
  end-to-end against a host-side CycloneDDS subscriber.
* POSIX/Linux build of the same tree plus a `ddsperf` loopback self-test:
  0 errors, `lost 0`.
* Cross-compile for Infineon TriCore (`kit_tc4x9_std_trb/tc499pp/cpu0`): library
  and application compile; linking needs the RAM notes above.
* The `ports/zephyr` example itself now builds with `-DBUILD_HELLOWORLD_PUB=1`
  (it did not before this series).


fj-blanco:
Hi, I've been testing this branch while working with Cyclone DDS on Zephyr, including `native_sim` and ESP32-S3 builds.

I found two small additions on top of the current PR head that were needed in my tests:

* [`[77921a7](https://github.com/fj-blanco/cyclonedds/commit/77921a7544bbee41a9771f9b8445d7377e1e1053)`](https://github.com/fj-blanco/cyclonedds/commit/77921a7544bbee41a9771f9b8445d7377e1e1053) initializes the `recvmsg()` control buffer before calling the Zephyr implementation. I hit this because Zephyr may inspect the existing control headers while filling ancillary data.
* [`[530f5d4](https://github.com/fj-blanco/cyclonedds/commit/530f5d42baa23d418e70cbc0bae65cd076a8a6bb)`](https://github.com/fj-blanco/cyclonedds/commit/530f5d42baa23d418e70cbc0bae65cd076a8a6bb) enables `CONFIG_POSIX_PRIORITY_SCHEDULING`, which provides the `sched_get_priority_min/max()` functions mentioned in the known limitations above.

Both commits are signed off and are on my `contrib/pr-2461-review-fixes` branch.

Would you prefer to cherry-pick them, or should I open a small PR against your branch?

